PR #7, opened 2022-04-05
feat(datatable): add styles and docs for loadingindicator

Janobob:
* feature(datatable): add styles and docs for loadingindicator

Feature has been requested by an employee of swiss post.


gfellerph:
@Janobob, the datatable styles should match the other tables with a white seperator line between the rows:
image

PR #95, opened 2022-05-06
chore: Create dependency-review.yml

gfellerph:


gfellerph:
Not needed at the moment -> use dependabot.

Issue #143, opened 2022-06-20
Hoist styles folder structure to get rid of publishConfig dist directory


gfellerph:
publishConfig is not supported by pnpm and creates serious drawbacks trying to get it to play nicely. That's why lerna bootstrap is in the init script. If it's possible to get rid of all the use cases for this setting (see web-components), the monorepo complexity can be reduced greatly.

gfellerph:
Might get obsolete when https://github.com/pnpm/pnpm/issues/3901#event-7060649429 is published.

PR #152, opened 2022-06-29
fix(web-styles): publish the `dist` subdirectory

driimus:
Fixes the release workflow, which currently fails for the `web-styles` package because the publish command targets the `dist` folder, and `dist/package.json` points to `/dist/dist`, a directory that doesn't exist. Haven't looked into why looking for a `package.json` file in `dist/dist` doesn't produce an error, but transforming `dist/package.json` should fix the issue.

I have tried it out locally against `verdaccio` and `changeset publish` worked as expected. Feel free to close the PR if you have different plans for fixing the issue, I'm more interested in letting you know what went wrong.

### Why only `web-styles`?
Not 100% sure as to why `swisspost-intranet-header` doesn't have the same issue, but my guess is:
https://github.com/swisspost/common-web-frontend/blob/9203419db831f4eca3fca92b3f6b55194e0701e4/packages/angular-components/projects/swisspost-intranet-header/package.json#L9
When the `publish` command runs in `dist/swisspost-intranet-header`, it will end up pointing to the current directory, meaning that the relative path ends up being correct.

### Context
I had a similar issue as the one described in changesets/action#191 (in my case, the failure was TypeScript compilation errors). 

The difference is that mine would be fixable by changesets/changesets#844, because `pnpm` does redirect compilation errors to `stdout` by default (can be overridden by `use-stderr` in `.npmrc`).

### DIFF
--- a/packages/web-styles/gulpfile.js
+++ b/packages/web-styles/gulpfile.js
@@ -1,10 +1,11 @@
+const fs = require('fs');
+const path = require('path');
 const gulp = require('gulp');
 const sass = require('sass');
 const gulpSass = require('gulp-sass')(sass);
 const gulpPostCss = require('gulp-postcss');
 const autoprefixer = require('autoprefixer');
 const options = require('./package.json').sass;
-
 /*
  * Copy task
  */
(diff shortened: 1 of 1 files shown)


gfellerph (approved):

gfellerph:
@driimus absolutely amazing, thank you so much for taking the time to investigate and fix the issue. Also, you are our first external contributor. Please accept my humble offering of 💯 fake internet points.

We're planning to improve the web-styles situation with #143, but until then, I'm glad to try your fix.

driimus:
Whoops, forgot to commit the imports. Should be good now.

PR #159, opened 2022-07-12
fix(web-styles): adds styles for form-range in forced-color/high-contrast mode

oliverschuerch:


.changeset/great-parrots-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/web-styles': patch
+---
+
+Moves /.stylelintrc.js to /web-styles/.stylelintrc, so it works correctly with the IDE
```
gfellerph, line 5: A change like this is irrelevant to consumers of our packages, therefore I'd not add it to the changelog. Can you delete this changeset?

.changeset/chilled-seahorses-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/web-styles': patch
+---
+
+add styles for form-range in fourced-color/high-contrast mode
```
gfellerph, line 5:
```suggestion
Add styles for form-range in fourced-color/high-contrast mode
```

packages/web-styles/src/components/form-range.scss
```diff
@@ -43,3 +44,46 @@
...
   }
 }
+
+@include utilities.high-contrast-mode() {
+  .form-range {
+    // linear-gradient is forced to the value of "none" in forces-color mode
+    // so, the "forced-color-adjust" property is necessary for "linear-gradient" to continue to work
+    forced-color-adjust: none;
+
+    &:not(:disabled):not(.disabled) {
```
gfellerph, line 54:
```suggestion
    &:not(:disabled, .disabled) {
```
The :not pseudo selector accepts lists now 🎉 https://caniuse.com/css-not-sel-list

packages/web-styles/src/components/form-range.scss
```diff
@@ -43,3 +44,46 @@
...
+
+      &::-moz-range-progress {
+        background: ButtonBorder;
+      }
+
+      &::-moz-range-thumb {
+        background-color: ButtonFace;
+        border-color: ButtonBorder;
+      }
+    }
```
gfellerph, line 74: Can you add :focus-visible styles for HCM? At the moment I can't see my keyboard focus on the slider. Either an outline on the whole slider or just the thumb, but more pronounced than the shadow in non-HCM.

gfellerph (changes requested):

packages/web-styles/src/components/form-range.scss
oliverschuerch: I just copied the existing selector ;)
So I will change both of them.

gfellerph (approved):
Good stuff, thank you!

gfellerph (approved):

PR #162, opened 2022-07-14
fix(topic-teaser): refactors component topic-teaser, to make it work as expected by design

oliverschuerch:
Corrects a render-issue (in breakpoint 780px) and fixes the design lacks.


packages/web-styles/src/components/topic-teaser.scss
```diff
@@ -13,72 +18,96 @@
...
+      > .row {
+        justify-content: flex-end;
+      }
     }
 
-    .topic-teaser-content {
-      @include media-breakpoint-up(xxl) {
-        padding-left: 4.5rem;
+    .topic-teaser-image-container {
+      @each $key in topic-teaser.$grid-keys-lower {
```
gfellerph, line 51: Similarly, @include media-breakpoint-down($grid-keys-lower[last item]) instead of the loop.

gfellerph (changes requested):

packages/web-styles/src/components/topic-teaser.scss
```diff
@@ -13,72 +18,96 @@
...
+      padding-right: map.get(grid.$grid-container-padding, xxl) * 0.5;
+    }
+  }
+
   &:not(.topic-teaser-reverse) {
-    > .container {
-      @include media-breakpoint-up(lg) {
-        padding-right: 0;
+    @each $key in topic-teaser.$grid-keys-upper {
+      @include media-breakpoint-only($key) {
```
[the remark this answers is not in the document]
oliverschuerch: In this case your right. I will change it.

packages/web-styles/src/components/topic-teaser.scss
oliverschuerch: In this case you are wrong. In each `$grid-keys-lower` breakpoint a `magin-left` or `margin-right` is set on the `.topic-teaser-image-container`, depending on the `grid.$grid-container-padding` variable.
So it is absolutely necessary to iterate over each of these breakpoints.

gfellerph (approved):

PR #169, opened 2022-07-15
chore(form-select/custom-select): Fixes floating-label visibility, when rendered without option-tags or without selected option

oliverschuerch:
Includes example for form-select floating-label (with and without options).
Includes examples for custom-select floating-label (with and without options/selected option).
Unifies the static texts of the demo examples.


packages/web-demo/src/app/bootstrap/components/form-select/form-select-demo/form-select-demo.component.html
```diff
@@ -1,10 +1,17 @@
...
     <option value="3">Three</option>
   </select>
-  <label for="floatingSelect">.form-select-lg</label>
+  <label for="floatingSelect">Floatinglabel form-select</label>
+</div>
+
+<div class="form-floating">
+  <select aria-label="Floating label select example" class="form-select" id="floatingSelectEmpty">
+    <!-- no options -->
+  </select>
```
gfellerph, line 13: I think the example with no selected option for the custom select is enough to show the behaviour. I'd rather not put bad examples like selects without options in the docs. They look broken. If you agree, can you remove the two examples without options?

gfellerph (approved):

PR #187, opened 2022-07-22
chore(tabs): Refactors tabs styles, to get rid of the class text-auto

oliverschuerch:


packages/web-styles/src/components/tabs.scss
```diff
@@ -48,69 +49,75 @@
...
+    padding: nav.$nav-link-padding;
+    transition: background-color 100ms;
+    border-right: 1px solid transparent;
+    border-left: 1px solid transparent;
+    outline-color: currentColor;
+    opacity: 0.7;
+    color: var(--text-color);
+
+    // Tabs with backgrounds
+    @include color-mx.has-background() {
```
gfellerph, line 65: @oliverschuerch, I'd like to discuss this mixin with you again.

gfellerph (changes requested):

packages/web-styles/src/components/tabs.scss
```diff
@@ -33,13 +34,20 @@
 }
 
 .tabs {
+  --tabs-title-bg-opacity: 1;
```
gfellerph, line 37: Do we want to declare default variables for components within the :root { } selector (but inside this file)? This would enable the whole page to make use of that variable if tabs are included (might be necessary for the page to use this var somewhere else, in this case I don't see it as very useful but I would like to set a precedent).

packages/web-styles/src/variables/_color.scss
```diff
@@ -117,10 +117,15 @@ $colors: map.merge(
 );
 $colors: map.merge($background-colors, $colors);
 
-// Compile a list of dark backgrounds, used in the :is selector mixin at mixins/color
+// Compile lists of backgrounds, used in the :is selector mixins at mixins/color
+$all-backgrounds: () !default;
```
gfellerph, line 121: Do we still need the `$all-backgrounds` variable?

oliverschuerch: Of course not! *shameonme*

packages/web-styles/src/components/tabs.scss
oliverschuerch: I don't see the point, why should anyone use this variable outside the tabs-component?
I would rather discuss why this is not a scss-variable.
Otherwise I would always prefer to declare css-variables like this in the component itself and not pollute the :root selector with it.
But maybe your arguments will make sense to me if you explain them.

gfellerph (approved):

PR #188, opened 2022-07-22
chore(option-variables): Remove unused scss options

oliverschuerch:
Removes the following variables from the project:
- $isolate-components / --isolate-components
- $variable-prefix


packages/web-styles/src/themes/bootstrap/_overrides-variables.scss
```diff
@@ -1,6 +1,7 @@
 @forward '../../variables/animation';
 @forward '../../variables/breakpoints';
 @forward '../../variables/color';
+@forward '../../variables/features';
```
gfellerph, line 4: Can you move this file to the themes/bootstrap folder as they are bootstrap only features?

gfellerph (changes requested):

packages/web-styles/src/themes/bootstrap/_overrides-variables.scss
oliverschuerch: The file is now located under `themes/bootstrap/_overrides-features.scss`.

gfellerph (approved):

PR #217, opened 2022-07-27
fix(web-styles-migration): typo in scope of package name

gfellerph:


packages/web-styles-migrations/package.json
```diff
@@ -1,28 +1,28 @@
 {
-  "name": "@swispost/web-styles-migrations",
+  "name": "@swisspost/web-styles-migrations",
```
alizedebray, line 2: It's better like that!

alizedebray (approved):

PR #225, opened 2022-07-27
build: with pnpm 7.7.0, lerna becomes obsolete

gfellerph:
When pnpm 7.7 gets released, we can remove lerna from the build because pnpm starts to support the publishConfig.directory setting.

Currently, 7.7.0-0 is a pre-release (https://github.com/pnpm/pnpm/releases/tag/v7.7.0-0). Once it's been released properly, this change can be merged.


gfellerph:
Current issue: dist folders do not exist when `pnpm install` is run, therefore the workspaces are not linked correctly and when building, the packages don't exist in the node_modules of the web-demo.

gfellerph:
Needs to be re-investigated later.

PR #250, opened 2022-08-03
feat(cards): add card story

LarissaMS:


packages/web-styles/src/components/card.stories.js
```diff
@@ -0,0 +1,116 @@
+import './card.scss';
+
+export default {
+    title: 'Components/Card',
```
gfellerph, line 4: This file uses 4 spaces for indentation, but only 2 should be used. Press <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>P</kbd> -> Format document in Visual Studio Code to format it correctly.

gfellerph (changes requested):

gfellerph:
- [ ] Also add a changeset

gfellerph (approved):

PR #257, opened 2022-08-04
feat(blockquote): add blockquote story

LarissaMS:


packages/web-styles/src/components/blockquotes.stories.js
```diff
@@ -0,0 +1,113 @@
+import './type.scss';
+
+export default {
+    title: 'Components/Blockquotes',
```
gfellerph, line 4: Use two spaces for indentation.

gfellerph (changes requested):

gfellerph (approved):

PR #285, opened 2022-08-09
fix(color): remove remaining secondary color classes

gfellerph:


oliverschuerch (commented):
I just searched `secondary` over the hole project and found 104 results in 40 files.
If we do not reintegrate secondary variable, we have to fix them all.
The only reason we do not fix them all could be, that we don't want to update the web-demo package anymore.

image

oliverschuerch (approved):

PR #286, opened 2022-08-09
fix(button-group): let white space break for button-group

gfellerph:


oliverschuerch (commented):
Shouldn't buttons have the statement `text-align: center;` in this case?

oliverschuerch (approved):

PR #288, opened 2022-08-09
fix(alerts): button size on small screens

gfellerph:


oliverschuerch (commented):
The made changes are ok, but I would prefer to let the buttons show right-aligned and side-by-side as long as possible. And it should be possibly done with a few simple changes.
@gfellerph what do you think about this approach?

| How it looks now | How it could look |
| - | - |
| image | image |
|  image | image |
| image | image |
| image | image |
| image | image |

packages/web-styles/src/components/alert.scss
```diff
@@ -179,14 +161,17 @@
   }
 
   .alert-buttons {
-    flex: 0 0 auto;
-    margin-left: alert.$alert-padding-x;
+    display: flex;
+    flex-wrap: wrap;
+    gap: spacing.$size-regular;
+    align-self: flex-end;
+    margin-top: spacing.$size-regular;
```
oliverschuerch, line 168:
```suggestion
```

oliverschuerch (commented):
If an alert is positioned fixed, set the property `flex-direction` in the selector `.alert-action` to `column`, so the content and the action-buttons are displayed below each other.

oliverschuerch:
@gfellerph I thing you missed my general comment regarding `flex-direction`!?

gfellerph:
@oliverschuerch, I updated the layout so buttons can break to a newline on smaller screens and don't squish the content too much but are still center aligned if there is little content.

oliverschuerch (approved):

PR #289, opened 2022-08-09
fix(button-group): consistently use secondary buttons for the btn-group

gfellerph:


oliverschuerch (commented):
Do we need to update the documentation?
At least a note that the class of the button-variant must not be changed would be useful in my opinion. Because every other button-variant class (like `btn-primary`) may result in misrepresentation.
image

oliverschuerch (approved):

PR #383, opened 2022-08-22
chore(deps): update dependencies

gfellerph:


oliverschuerch:
Check https://github.com/nrwl/nx/issues/10776, maybe this is the solution for the broken linting problem...

oliverschuerch:
@gfellerph if you run `pnpm update`, the error ("no such file or directory nx/src/utils/app-root") disappears.
But if you run `pnpm bootstrap` again afterwards, it is broken again.

PR #390, opened 2022-08-22
fix(textarea): hide placeholder inside of floating-label to avoid display error

oliverschuerch:
The placeholder attribute must stay, as it is needed for the floating-label to work correctly.
And it must at least contain one char (e.g. "&nbsp;").


gfellerph:
@oliverschuerch you can use `opacity: 0;` to hide it. I think it's safer than the documentation approach.

gfellerph (changes requested):
@oliverschuerch you can use opacity: 0; to hide it. I think it's safer than the documentation approach.

gfellerph (approved):

PR #403, opened 2022-08-23
fix(subnavigation): add hcm styles

oliverschuerch:


gfellerph (changes requested):
@oliverschuerch, the border bottom is not yet displayed in HCM. I think it could be helpful in most situations to discern the content from the navigation.

gfellerph (approved):
Top

PR #432, opened 2022-08-29
chore(dependecies): monday morning updates

gfellerph:


oliverschuerch (approved):
If you've not allready done, do not forget to test the packages with breaking changes such as `core-js` in the demo package.

PR #433, opened 2022-08-29
fix(input-file): Fix display errors on safari

oliverschuerch:
I was able to eliminate almost all of the graphical errors.
Unfortunately this leads to the fact that you have to set the button translations via content CSS-property and these are dependent on the document-language.
The integrated languages are: en (default), de, fr, it


gfellerph (approved):
I think it's a good workaround until we have the web component and I think it's ok to expect every document to have a html lang attribute.

PR #447, opened 2022-08-30
feat(intranet-header): correcting and completing the property documentation

oliverschuerch:


packages/demo/src/styles.scss
```diff
@@ -83,15 +83,19 @@ section {
...
-  background: rgb(50, 50, 50);
-  color: white;
-  white-space: pre;
-  max-width: 100%;
-  overflow: auto;
-  margin-bottom: 0.8em;
-
-  max-height: 450px;
+code {
+  color: darkviolet;
```
gfellerph, line 87: For regular inline code elements, this would work better. Maybe we need to reset some of the styles when used in tables or for the block variants.

```suggestion
  padding: 0.1em 0.4em;
  margin: 0;
  font-size: 85%;
  background-color: rgba(175,184,193,0.3);
  border-radius: 6px;
```

gfellerph (changes requested):

gfellerph (approved):

PR #449, opened 2022-08-31
feat(topic-teaser): add topic-teaser story

LarissaMS:


gfellerph (approved):

gfellerph:
@oliverschuerch, this [import of another comonent](https://github.com/swisspost/design-system/pull/449/files#diff-07f4a4e3dfb240dc03c7e579ca9a178ebf2420de00d74e0520186263387b7187R14) is necessary as the topic teaser is dependent on the background styles. Do you think it's the right approach?

oliverschuerch:
> @oliverschuerch, this [import of another comonent](https://github.com/swisspost/design-system/pull/449/files#diff-07f4a4e3dfb240dc03c7e579ca9a178ebf2420de00d74e0520186263387b7187R14) is necessary as the topic teaser is dependent on the background styles. Do you think it's the right approach?

I don't get the point? Let's talk in person what the problem is.

gfellerph (approved):

PR #461, opened 2022-09-01
chore!: rename and move features.scss

oliverschuerch:


.changeset/khaki-balloons-look.md
```diff
@@ -0,0 +1,5 @@
+---
```
gfellerph, line 1: This file is unneeded because it has no impact on consumers of the library.

gfellerph (changes requested):

gfellerph (approved):

PR #484, opened 2022-09-26
feat: add percy snapshot functionality

oliverschuerch:
Add the script which can generate storybook-snapshots with the power of percy and cypress.
Add also the first cypress test which generates snapshots from the button-overview story and sends them to percy.io.


swisspost-devs:

gfellerph (approved):

PR #529, opened 2022-10-03
feat(workflow): add initial comment as swisspost-dev user

oliverschuerch:


swisspost-devs:
Preview environment links will be posted here, as soon as they are ready!

PR #537, opened 2022-10-03
chore(components): set publishConfig access to public

gfellerph:


packages/components/package.json
```diff
@@ -19,6 +19,9 @@
     "dist/",
     "loader/"
   ],
+  "publishConfig": {
+    "access": "public"
```
oliverschuerch, line 23: https://docs.npmjs.com/cli/v8/using-npm/config#access
Says: "... Unscoped packages always have an access level of public. ..."
So the definition is not nessesairy, but should also not be a problem.

oliverschuerch (approved):

packages/components/package.json
gfellerph: @oliverschuerch this is, however, a scoped project (@swisspost). Therefore the default is private.

PR #542, opened 2022-10-03
build(root): remove dependency lerna

gfellerph:
Since 7.9, pnpm supports the publishConfig.directory setting, which enables us to remove lerna from the bootstrap script and as a dependency. Local workspace packages that configure a publishConfig.directory should now also set linkDirectory to true.


oliverschuerch (approved):

swisspost-devs:

PR #589, opened 2022-10-13
feat(components): add component button

oliverschuerch:


oliverschuerch:
We close this PR, because of reasons.

PR #607, opened 2022-10-19
feat(migration): add migrate binary to migrations package

oliverschuerch:


oliverschuerch:
This PR is not necessary because we desided to integrate the actual migration package into the styles package.

PR #628, opened 2022-10-24
feat(styles-migrations): add migration ng-button-classes

oliverschuerch:


packages/styles/schematics/migrations/buttons copy/index.ts
```diff
@@ -0,0 +1,22 @@
...
+class OutlinedButtonClassesUpdate extends CssClassesUpdate {
+    searchValue = 'btn-outline-' + oneOf(themeColors);
+    replaceValue = (color: string) => 'btn-' + color;
+}
+
+// The .btn-inverted class no longer exists
+class InvertedButtonClassesUpdate extends CssClassesUpdate {
+    searchValue = 'btn-inverted';
+    replaceValue = '';
+}
```
gfellerph, line 22: Is there a space in the folder name?

gfellerph (approved):

packages/styles/schematics/migrations/buttons copy/index.ts
oliverschuerch: This file has been commited mistakenly. It's a copy of the buttons migration and there is no entry in the migrations.json for that. I removed it.

PR #641, opened 2022-10-25
fix(migration-utils): add regex element selectors

oliverschuerch:


gfellerph (approved):

packages/styles/schematics/migrations.json
```diff
@@ -27,10 +27,10 @@
       "description": "Replaces ng-button classes.",
       "factory": "./migrations/ng-buttons/index"
     },
-    "migration-forms": {
+    "migration-form-controls": {
       "version": "5.0.0",
-      "description": "https://getbootstrap.com/docs/5.0/migration/#forms",
-      "factory": "./migrations/forms/index"
+      "description": "Applies automatic migrations on form-control elements.",
+      "factory": "./migrations/form-controls/index"
```
alizedebray, line 33: Is this change necessary? "form-controls" only refers to inputs, but migration also changes other types of form elements

packages/styles/schematics/utils/css/css-classes-update.ts
```diff
@@ -7,6 +7,18 @@ export abstract class CssClassesUpdate {
     abstract searchValue: string;
     abstract replaceValue: string | Replacer;
 
+    get tag(): RegExp | null {
+        return this.tagSelector ? new RegExp(`^${this.tagSelector}$`) : null;
+    }
+
+    get class(): RegExp | null {
+        return this.classSelector ? new RegExp(`^${this.classSelector}$`) : null;
```
alizedebray, line 15: In the majority of cases the class selector matches the search value, therefore it might be good for this getter to return `this.searcher` instead of `null` if no class selector is defined. What do you think?

packages/styles/schematics/utils/css/css-migration.ts
```diff
@@ -15,16 +15,16 @@ export class CssMigration {
...
+            .filter(u => u.class === null || element?.attributes.some(a => a.name === 'class' && a.value.split(' ').some(v => u.class !== null && u.class.test(v))))
+            .filter(u => u.attribute === null || element?.attributes.some(a => u.attribute !== null && u.attribute.test(a.name)));
     }
 
     currentTemplate (element: TmplAstTemplate | null) {
         this.filteredUpdates = this.updates
-            .filter(u => !u.tagSelector || u.tagSelector === element?.tagName)
-            .filter(u => !u.classSelector || element?.attributes.filter(a => a.name === 'class' && a.value.split(' ').includes(u.classSelector)))
-            .filter(u => !u.attributeSelector || element?.attributes.some(a => a.name === u.attributeSelector));
+            .filter(u => u.tag === null || (u.tag !== null && u.tag.test(element?.tagName ?? '')))
```
alizedebray, line 25: This could be simplified: 
`u.tag === null || u.tag.test(element?.tagName ?? '')`

Same for the two other filters

packages/styles/schematics/utils/css/css-classes-update.ts
oliverschuerch: The `class` property only act as a filter for the existing updates and we're not aware of what is defined in the `searcher` at this point.
So if the `classSelector` is not defined, the `class` filter should be `null`, to not filtering the nodes by something else (like the  `searcher` property), and therefore avoid side-effects.

packages/styles/schematics/utils/css/css-migration.ts
oliverschuerch: Can you provide an simplified example of this line?

packages/styles/schematics/migrations.json
oliverschuerch: For me form-controls refers to all form-controls (input, select, textarea, checkbox, radio, etc.) but I can see your point and will rename it.

packages/styles/schematics/utils/css/css-migration.ts
alizedebray:
```suggestion
            .filter(u => u.tag === null || u.tag.test(element?.tagName ?? ''))
```
Either the tag selector is null or the test should be true

packages/styles/schematics/utils/css/css-classes-update.ts
alizedebray: The searcher aims to find a class that needs to be updated, in most cases if an item does not contain that class it can be filtered out.

Looking at the form migrations, the `classSelector` almost always matches the `searchValue` (custom select menu, custom checkbox, custom radio, custom control input, custom control label, custom inline control, switch , toggle switch)

In special cases such as `.form-text`, the classSelector can always be overridden

alizedebray: But if you think setting a classSelector for each migration is clearer, that's fine with me too

oliverschuerch: I will as it is for the moment (never touch a running system).

PR #644, opened 2022-10-26
docs(styles-migration): add automatic migration flag for multiselect class update

oliverschuerch:


swisspost-devs:

gfellerph (changes requested):
Wait until other PRs are merged and update this branch. It's not clear where the error with the test demo-build comes from, but I'd not merge the change if it does not run through.

gfellerph (approved):

oliverschuerch:
@gfellerph If we have a change in the styles package as well, it is working fine. I don't know why, but it must has something to do with the dependency building `...[origin/main]...`.
The demo package has one dependency to the styles package and one to the intranet-header.
But the intranet-header has also a dependency to the styles package.
If these two packages are build at the same time, the intranet-header package can not get the styles dependency, because it is not built yet.

Just a guess...

gfellerph:
pnpm should actually resolve the dependency tree and figure out that the styles package has to be built before the intranet header, but it's a good hint, will test locally.

PR #653, opened 2022-10-28
build(cypress): move binary into node_modules for caching

gfellerph:


.github/workflows/build.yaml
```diff
@@ -13,13 +13,7 @@
...
 name: Build demo app
 on:
-  pull_request:
-    paths:
-      - 'packages/components-angular/**'
-      - 'packages/demo/**'
-      - 'packages/styles/src/**'
-      - 'packages/styles/*'
-      - '!packages/styles/jest.*'
+  pull_request
```
oliverschuerch, line 16: Why do you change this lines?

oliverschuerch (changes requested):

oliverschuerch (approved):

swisspost-devs:

oliverschuerch (commented):
Have a look into the test.yaml workflow. There is code included to cache the cypress binary which does probably not work together with your .npmrc change.

PR #656, opened 2022-10-28
fix: remove cypress binary postinstallation from package files

oliverschuerch:


swisspost-devs:

oliverschuerch:
Manually run the postintallation script on the cypress node_module did not work in github workflow.
I inserted an `use_cache` flag into the pnpm-setup action instead, to disable the cache in the test workflow only.

gfellerph (approved):

PR #669, opened 2022-10-31
fix(styles-migration): fix ng-button label class migration

oliverschuerch:


swisspost-devs:

gfellerph (approved):

PR #676, opened 2022-11-01
chore: reset changelogs before initial release

gfellerph:
Resetting changelogs before the first major release creates a summary of changes in just one update instead of all alpha and beta increments which is more readable. All changesets created are still listed in the newly generated changelog and the message format follows the new standard.


swisspost-devs:

Issue #155, opened 2022-07-11
Components: Collapsible


gfellerph:
Create a collapsible web component.

**Features:**
- [x] Name: post-collapsible
- [x] Animates height, unless prefers-reduced-motion: reduced is set
- [x] Does **not** use `<details>/<summary>` internally (https://adrianroselli.com/2019/04/details-summary-are-not-insert-control-here.html)
- [x] Click on header toggles the collapsible, if no header is present, an external toggler can be used to toggle via JS API
- [x] Header is focusable, focus is visible
- [x] Collapsible elements can be nested
- [x] Plays nicely together with #200 
- [x] Complies with https://www.w3.org/WAI/ARIA/apg/patterns/accordion/ and https://www.accessibility-developer-guide.com/examples/widgets/accordion/

**Options:**
- [x] Can be open or closed initially

**Methods:**
- [x] JS API for toggling with optional force mode (can pass toggle(collapse: true) to always close the collapsible)

**Tests**
- [x] Toggle API can be called to toggle and with force parameter
- [ ] ARIA-Attributes are correct depending on state
- [ ] Collapsibles can be nested and have correct ARIA attributes
- [x] Height of an open collapsible body is the same as the height of the content (ensure no content is hidden)

**Default:**
```html
<post-collapsible>
  <h2 slot="header">Collapsible header</h2>
  <p slot="body">Collapsible body. slot="body" could be omitted.</p>
</post-collapsible>
```

**Without header:**
```html
<post-collapsible>
  <p>The default slot is the body slot.</p>
</post-collapsible>
```

**Reference implementation:**
https://telekom.github.io/scale/?path=/docs/components-accordion--standard
https://developers.arcgis.com/calcite-design-system/components/accordion/ (especially for accessibility)

PR #683, opened 2022-11-01
feat(components): #155 collapsible component

alizedebray:


swisspost-devs:

gfellerph:
Had an idea about how to manage the slot issue. It's not ideal, but leaves the definition of the header level and layout entirely in the hands of the project while also retaining accessibility standards:

```html
<post-collapsible>
  <h2 slot="header">
    <post-collapsible-button>
      <post-icon name="island" />
      Some heading content here
    </post-collapsible-button>
  </h2>
  <div>I'm the body of the collapsible in the default slot</div>
</post-collapsible>
```

Another approach would include less flexibility but more stability:

```html
<post-collapsible>
  <post-collapsible-header level="2" slot="header">
    <post-icon name="island" />
    Some header text
  </post-collapsible-header>
  <div>Some collapsible content</div>
</post-collapsible>

<!-- which renders as -->
<post-collapsible>
  <post-collapsible-header>
    <h2>
      <button aria-expanded="...">
        <post-icon ... />
        Some header text
      </button>
    </h2>
  </post-collapsible-header>
  <div class="collapsible-content-container">
    <div>Some collapsible content</div>
  </div>
</post-collapsible>

```

oliverschuerch:
My solution to manage the slot issue is - of course - inspired by the way how it's done in vuejs.

```tsx
import { Component, Host, Element, Prop, h } from '@stencil/core';

@Component({
  tag: 'post-collapse',
  // styling the header, header-button and the body can be overwritten by the user with the css :part() selectors
  styleUrl: 'collapse.scss',
  shadow: true
})
export class PostButton {
  @Element() host: HTMLElement;

  // use some guid generator to get a strong and falid guid (do not use one, only using Math.random)
  collapseId: guid = '2f9fc627-39f7-4eed-905f-987365523c05';
  // define headerTag default as 'h2' but give the possibility to change it to 'h1', 'h3', 'h4', 'h5' or 'h6'
  @Prop() headerTag: string = 'h2';

  @Prop({ mutable: true }) collapsed = false;

  // all the code alize allready defined to manage the collapse to work...

  render () {
    // this gives us the possibility to add the header only if a header-slot is defined
    let headerDom = [];

    if (this.host.querySelector('[slot="header"]')) {
      // We can add custom attributes to header by creating propertis like 'headerClass' or 'headerButtonRole', if there is really a need for that.
      // Otherwise I would prefer to leave the elements as clean as possible, because they need to be like so, because of semantic reasons!
      headerDom.push(
        <this.headerTag
          id={`${this.collapseId}--header`}
          part="header"
        >
          <button
            aria-expanded={`${!this.collapsed}`}
            aria-controls={`${this.collapseId}--body`}
            part="header-button"
          >
            <slot name="header">Provide some fallback content (if possible/necessary)</slot>
          </button>
        </this.headerTag>
      );
    }

    return (
      <Host class="post-collapse">
        { headerDom }

        <div
          id={`${this.collapseId}--body`}
          aria-labelledby={`${this.collapseId}--header`}
          part="body"
        >
          {/* default slot, every thing you define directly inside of the post-collapse component without an explicit slot wrapper goes here */}
          <slot>Provide some fallback content (if possible/necessary)</slot>
        </div>
      </Host>
    );
  }
}
```

gfellerph:
**Discussion summary**:
- It's easier to have one web-component only to reduce complexity, separation of concerns is less an issue because it is a small component
- It's better to reduce some flexibility concerning the heading level but increase stability of the component. The user might not need control over the heading element at all. Optionally, a custom ID can be passed via props and custom styles can be applied via the parts API.
- The collapsible body content should live in the default slot
- It's better to have stable default IDs as opposed to random GUIDs to improve URL sharing capabilities and possible other browser functionality related to IDs.

alizedebray:
The collapsible component now matches the last discussion's outcomes. 

Props are currently not managed with constants as this introduce side effects in the auto-generated documentation (see the "Default" column in the [readme](https://github.com/swisspost/design-system/blob/feat/580-component-buttons/packages/components/src/components/button/readme.md), also appeared in storybook "Docs" tab)

packages/components/src/utils/get-prop-attributes.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function getPropAttributes(props: Record<string, unknown> | undefined): string {
+  if (!props) return '';
+
+  return Object.keys(props).reduce((attributes, propName) => {
+    const propValue = props[propName];
+    if (typeof propValue !== 'undefined') attributes += ` ${toKebabCase(propName)}="${propValue}"`;
+
+    return attributes;
+  }, ' ');
```
oliverschuerch, line 15: It's just a different way to write your lines, please take over or leave out whatever you like.
But notice, that in your code you would write `propName="propValue"` even if the propValue is an empty string or null.

```suggestion
  return Object.entries(props)
    .reduce((attributes, [ propName, propValue ]) => attributes.concat(propValue ? ` ${toKebabCase(propName)}="${propValue}"` : '', ' ');
```

packages/components/src/utils/get-element-height.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright 2022 by Swiss Post, Information Technology
+ */
+
+export function getElementHeight(el: HTMLElement): string;
+export function getElementHeight(el: HTMLElement, classWhenShown: string): string;
+export function getElementHeight(el: HTMLElement, classesWhenShown: string[]): string;
+export function getElementHeight(el: HTMLElement, classesWhenShown: any = []): string {
+  if (!Array.isArray(classesWhenShown)) classesWhenShown = [classesWhenShown];
+
+  const classesToAdd = classesWhenShown.filter(klass => !el.classList.contains(klass));
```
oliverschuerch, line 11: Please do not use german property names.
With for example "cssClass" you are more specific and closer to the way it would be named in backend code like C#.
But you are free to use other names as well...
when it is as clear as here what you are iterating about, I would also absolutely agree with an abbreviation like "c".

```suggestion
  const classesToAdd = classesWhenShown.filter(c => !el.classList.contains(c));
```

packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  @Element() host: HTMLElement;
+
+  hasHeader: boolean;
+  headingTag: string | undefined;
+  collapsibleId: string;
+  collapsibleElement: HTMLElement;
+
+  componentWillLoad() {
+    this.hasHeader = !!this.host.querySelector('[slot="header"]');
```
oliverschuerch, line 33: Users could use slot-binding on multiple elements.

```suggestion
    this.hasHeader = this.host.querySelectorAll('[slot="header"]').length > 0;
```

packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -0,0 +1,115 @@
...
+    if (!this.hasHeader) {
+      console.warn('Be sure to bind the post-collapsible to its control using aria-controls and aria-expanded attributes. More information here: https://getbootstrap.com/docs/5.2/components/collapse/#accessibility');
+    }
+
+    this.headingTag = checkValidHTMLTag(`h${this.headingLevel}`, 'The post-collapsible element requires a heading level between 1 and 6.');
```
oliverschuerch, line 38: This check will produce a wrong positiv value, if someone enters "-1", "-2", etc. as heading-level property.
Because "h-1" is basically a correct html-tag.

packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    this.collapsibleId = this.host.id || `post-collapsible-${nextId++}`;
+    this.collapseClasses = this.collapsed ? 'collapse' : 'collapse show';
```
oliverschuerch, line 41: I would prefer to have a private function to set the collapse classes, because you need to this multiple times (line 41 and 60).
If we ever need to update the classes, we can do it only at one place.

packages/components/src/utils/get-element-height.ts
```diff
@@ -0,0 +1,19 @@
...
+  if (classesToAdd.length) el.classList.add(...classesToAdd);
+
+  const scrollHeight = el.scrollHeight;
+
+  if (classesToAdd.length) el.classList.remove(...classesToAdd);
+
+  return `${scrollHeight}px`;
```
oliverschuerch, line 18: If you need the return value for further calculations, it is possible like this.

```suggestion
  return scrollHeight;
```

packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  startTransition() {
+    const expandedHeight = getElementHeight(this.collapsibleElement, 'show');
+
+    this.collapseHeight = this.collapsed ? expandedHeight : '0px';
+    this.collapseClasses = 'collapsing';
+
+    setTimeout(() => {
+      this.collapseHeight = this.collapsed ? '0px' : expandedHeight;
+    });
```
oliverschuerch, line 73: If the util "getElementHeight" returns a number (without 'px'):

```suggestion
    const expandedHeight = getElementHeight(this.collapsibleElement, 'show');

    this.collapseHeight = `${this.collapsed ? expandedHeight : 0}px`;
    this.collapseClasses = 'collapsing';

    setTimeout(() => {
      this.collapseHeight = `${this.collapsed ? 0 : expandedHeight}px`;
    });
```

packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  componentWillLoad() {
+    this.hasHeader = !!this.host.querySelector('[slot="header"]');
+    if (!this.hasHeader) {
+      console.warn('Be sure to bind the post-collapsible to its control using aria-controls and aria-expanded attributes. More information here: https://getbootstrap.com/docs/5.2/components/collapse/#accessibility');
+    }
+
+    this.headingTag = checkValidHTMLTag(`h${this.headingLevel}`, 'The post-collapsible element requires a heading level between 1 and 6.');
+
+    this.collapsibleId = this.host.id || `post-collapsible-${nextId++}`;
```
oliverschuerch, line 40: I don't know why, but this produces `undefined--header` and `undefined-collapse` classes in the dom:
image

oliverschuerch (changes requested):

packages/components/src/utils/get-element-height.ts
alizedebray: 👍

packages/components/src/utils/get-prop-attributes.ts
alizedebray: I'm still used to avoid `Object.entries` because of IE but you're right, that time is over! I just updated this.

Regarding the solution you proposed, if the propValue is "0" or "false", it will not be rendered.
With Stencil components, we should only have props with a default value (different from null or an empty string) or no value at all (undefined). For me, the only cases where a prop may be set to null or an empty string is at runtime when the user is playing with the component. In this case, s.he should be able to write anything and incorrect attributes would be caught with errors.

packages/components/src/utils/get-element-height.ts
alizedebray: The term "klass" is quite conventional, it is at least used by [Angular ](https://github.com/angular/angular/search?q=klass)and [VueJS](https://github.com/vuejs/core/search?q=klass). But I guess we can agree on a name to line up in the future.

packages/components/src/components/post-collapsible/post-collapsible.tsx
alizedebray: I could not reproduce this issue. Did you set anything special as an prop?

image

packages/components/src/components/post-collapsible/post-collapsible.tsx
alizedebray: It is then probably overkill to just check if a string is contained in a list. What do you think of [this](https://github.com/swisspost/design-system/pull/683/commits/73d6d1fa387b5f376cb108c5b04641f8c6728e89) instead?

packages/components/src/components/post-collapsible/post-collapsible.tsx
alizedebray: I just refactored with a private method (see [this commit](https://github.com/swisspost/design-system/pull/683/commits/fd299a3944e4284ab8dfbf72b29cff1a9c07a04b)) but it seems less readable, don't you think?

packages/components/src/utils/get-element-height.ts
oliverschuerch: I never saw this in VueJS in the past 5 years. But if you both like to use it I'm fine with that.

packages/components/src/components/post-collapsible/post-collapsible.tsx
oliverschuerch: I just realized, that I criticized the wrong line of code.
What we really should do is to validate the Property before using it!
I found how to validate props in stencil here: https://stenciljs.com/docs/properties#prop-validation, so...

```javascript
export class PostCollapsible {
  @Prop({ mutable: true }) headingLevel?: number = 2;
  
  ...

  this.headingTag = `h${this.headingLevel}`;

  ...
  
  // if we could abstract this in a utility-function, that would be great
  @Watch('headingLevel')
  headingLevelValidation(newValue: string, _oldValue: string) {
    if (![1, 2, 3, 4, 5, 6].includes(newValue)) {
      throw new Error('The post-collapsible element requires a heading level between 1 and 6.');
    }
  }
};
```

For the given case it should not be necessary to use a utility like "checkValidHTMLTag" anymore after this.

packages/components/src/components/post-collapsible/post-collapsible.tsx
oliverschuerch: Your right! But if you use the class only to return the correct value and asign this to the variable it should become much clearer:

```javascript
this.collapseClasses = this.getCollapseClasses();

// more lines of code

private getCollapseClasses() {
  return this.collapsed ? 'collapse' : 'collapse show';
}
```

packages/components/src/components/post-collapsible/post-collapsible.tsx
oliverschuerch: I'm happy that I could not reproduce it anymore either. So you can ignore it...

packages/components/src/components/post-collapsible/post-collapsible.tsx
alizedebray: I had seen this page, however in the storybook the component is reloaded each time you modify an argument and since the @Watch decorator is not called on load, the error is never displayed.

Please check [my last commit](https://github.com/swisspost/design-system/pull/683/commits/a6949b0421069f8c27b0fe045ad3d8e17161d3a0) where I mixed heading level check with @Watch and with componentWillLoad hook.

oliverschuerch (changes requested):

oliverschuerch (approved):

packages/documentation/src/stories/components/collapsible/collapsible.docs.mdx
```diff
@@ -0,0 +1,70 @@
+import { Meta, Canvas, Story, Source, ArgsTable } from '@storybook/addon-docs';
+
+<Meta title="Components/Collapsible" />
+
+# Collapsible
+
+**Toggle the visibility of content across your project.**<br/>
+The `<post-collapsible>` component is used to show and hide content.
+Collapsing an element will animate the height from its current value to 0.
```
oliverschuerch, line 9:
```suggestion
<p className="lead">Toggle the visibility of content across your project.<br/>
The `<post-collapsible>` component is used to show and hide content.
Collapsing an element will animate the height from its current value to 0.</p>
```

packages/documentation/src/stories/components/collapsible/collapsible.docs.mdx
```diff
@@ -0,0 +1,70 @@
...
+### Custom Trigger
+
+The `<post-collapsible>` component exposes a `.toggle()` method that allows to trigger the collapse programmatically.
+This method is asynchronous and returns a promise that resolves with the current open state.
+It optionally takes a boolean parameter that forces open when `true` or close when `false`.
+
+```typescript dark
+const collapsible = document.querySelector('#collapsibleId') as HTMLPostCollapsibleElement;
+collapsible.toggle(/* open: boolean */).then((/* isOpen: boolean */) => {});
+```
```
oliverschuerch, line 58: Maybe we should also provide a non-typescript example.
What do you think?

packages/documentation/src/stories/components/collapsible/collapsible.docs.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+A header is then not mandatory and can be replaced by an external control.
+
+In this case, to ensure good accessibility, identify the collapsible with an `id`,
+then add an `aria-controls` attribute to your control element referencing this id.
```
oliverschuerch, line 63:
```suggestion
then add an `aria-controls` attribute to your control element referencing this `id`.
```

packages/documentation/src/utils/component-properties.ts
```diff
@@ -0,0 +1,5 @@
+export function definedProperties(properties: Record<string, unknown>) {
+  return Object.fromEntries(
+    Object.entries(properties).filter(([_, value]) => typeof value !== 'undefined')
+  );
```
oliverschuerch, line 4:
```suggestion
  return Object.values(properties).filter(value => typeof value !== 'undefined')
  );
```

I'm not 100% sure if it's doing the same thing as your logic, but it seems so to me.
Please check and ignore if this is not the case.

packages/documentation/src/stories/components/collapsible/collapsible.stories.tsx
```diff
@@ -0,0 +1,143 @@
...
+      name: 'Content',
+      control: {
+        type: 'text',
+      },
+      table: {
+        category: 'Content',
+      },
+    },
+  },
+};
```
oliverschuerch, line 36: Please visit the docs https://storybook.js.org/docs/react/writing-stories/introduction to ensure the correct setup here.
 The default export of a stories file, should be of type `Meta` or `ComponentMeta<typeof component>`.
 
 But as you only use the `displayName` for the `component` property, I'm not sure how it must be written at this point.

packages/documentation/package.json
```diff
@@ -18,14 +18,20 @@
...
   },
   "dependencies": {
     "@swisspost/design-system-components": "workspace:1.0.4",
+    "@swisspost/design-system-components-react": "workspace:0.0.1",
     "@swisspost/design-system-styles": "workspace:5.1.3",
+    "html-react-parser": "3.0.4",
+    "js-beautify": "1.14.7",
     "react": "18.2.0",
     "react-dom": "18.2.0",
-    "html-react-parser": "3.0.4"
```
oliverschuerch, line 28: The `html-react-parse` package is used within the `alert.stories.tsx`.
Try to replace it there as well.

packages/documentation/src/stories/components/collapsible/collapsible.stories.tsx
```diff
@@ -0,0 +1,143 @@
...
+  argTypes: {
+    content: {
+      name: 'Content',
+      control: {
+        type: 'text',
+      },
+      table: {
+        category: 'Content',
+      },
+    },
```
oliverschuerch, line 34: Define also controls for the other props. e.g. `collapsed`, `headingLevel`, ...

oliverschuerch (changes requested):

packages/documentation/src/utils/component-properties.ts
alizedebray: I need `definedProperties` to return an object with only properties that have a defined value.
The solution you proposed returns an array of set values ​​but the keys are lost.

packages/components-react/src/components/index.ts
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import * as Components from "./stencil-generated/index";
+
+Object.entries(Components).forEach(([name, component]: [string, React.NamedExoticComponent]) => {
+  component.displayName = name.replace(/\B([A-Z])/g, '-$1').toLowerCase();
+});
+
+export * from "./stencil-generated/index";
```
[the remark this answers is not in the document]
alizedebray: I moved the logic to `preview.js` as it is only used there.

packages/components/src/utils/tests/property-validators.spec.ts
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2022 by Swiss Post, Information Technology
+ */
+
+import { booleanValidator, oneOfValidator } from '../property-validators';
+
+describe('property-validators', () => {
+  let errorMessage: string;
+  let validatorParameters: unknown[];
+  let validator: (...params: unknown[]) => void;
```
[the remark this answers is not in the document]
alizedebray: I renamed `booleanValidator` for `checkBoolean` and `oneOfValidator` for `checkOneOf`

packages/documentation/src/stories/components/collapsible/collapsible.docs.mdx
alizedebray: Do you mean having both TS and JS or having JS only?

The JS version is very similar to the TS except it gives extra information with `as HTMLPostCollapsibleElement`:
```suggestion
```typescript dark
const collapsible = document.querySelector('#collapsibleId');
collapsible.toggle(/* open: boolean */).then((/* isOpen: boolean */) => {});
```
```

File: packages/documentation/package.json (hunk 14 above)
Comment (alizedebray) [reply packages/documentation/package.json:28]: The package is still there, it was just moved because of the automatic alphabetization.

**Comment by swisspost-devs** [comment]

File: packages/components/src/components.d.ts [hunk 18]
```diff
@@ -6,65 +6,52 @@
  */
```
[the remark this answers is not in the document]
alizedebray: The generated stencil component readme.md file are now ignored as well as all stencil-generated files in the react component package. 
However ignoring components.d.ts [is not recommended](https://stenciljs.com/docs/typed-components) and causes build issues thus I let this file in the version control.

packages/documentation/.storybook/preview.js
```diff
@@ -1,12 +1,20 @@
+import { extractArgTypes, extractComponentDescription, setStencilDocJson } from '@pxtrn/storybook-addon-docs-stencil';
+import docJson from '@swisspost/design-system-components/dist/docs.json';
+import * as Components from '@swisspost/design-system-components-react';
+import beautify from 'js-beautify';
+import React from 'react';
+import { renderToString } from 'react-dom/server';
```
oliverschuerch, line 6:
```suggestion
import { renderToStaticMarkup } from 'react-dom/server';
```

This function is similar to the `renderToString` function but does not create some react specific attributes.
See: https://reactjs.org/docs/react-dom-server.html#rendertostaticmarkup

packages/documentation/.storybook/preview.js
```diff
@@ -51,28 +61,17 @@ export const parameters = {
...
-          // repalce noRefCheck functions
-          .replace(/function noRefCheck\(\)\s?{}/g, '() => {}')
-
-          // remove brackets from "{value}" attribute-values
-          .replace(/([a-zA-Z][a-zA-Z0-9-_:.]+)={([^}]*}?)}/g, (_m, g1, g2) => `${g1}="${g2}"`)
-
-          // replace "className" attributes with "class"
-          .replace(/className/g, 'class')
+      const reactElement = <JsxParser jsx={snippet} renderInWrapper={false}/>;
+      const htmlString = renderToString(reactElement);
```
oliverschuerch, line 65:
```suggestion
      const htmlString = renderToStaticMarkup(reactElement);
```

packages/documentation/.storybook/preview.js
```diff
@@ -51,28 +61,17 @@ export const parameters = {
...
 
-          // replace "htmlFor" attributes with "for"
-          .replace(/htmlFor/g, 'for')
+      return beautify.html(
```
oliverschuerch, line 67: Can we use pretter here? We use it allready in some other files and it would be nice to not have multiple packages installed, that literally do the same thing.
If the output is totally different, I would off course accept the beautify package ;)

oliverschuerch (changes requested):

packages/documentation/.storybook/preview.js
alizedebray: I used your implementation except for the `.replace()` methods which no longer seem useful since `renderToStaticMarkup` already handles these changes.

gfellerph (approved):
Fix the code smell and you're good to go. There are a lot of changes here so I think it makes sense to merge it and go from there. Otherwise this PR will stay open forever.

oliverschuerch (approved):

PR #689, opened 2022-11-02
chore(demo): switched to styles version number

gfellerph:
Switched to displaying the styles package version number instead of the demo app version number. The styles version is more important to developers than thenow disconnected demo version number


swisspost-devs:

alizedebray (approved):

PR #697, opened 2022-11-03
build(actions): fix deprecated set_output

gfellerph:
According to this post https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/, the set_output can be replaced by the newer env file approach.


alizedebray (approved):

swisspost-devs:

PR #698, opened 2022-11-04
fix(demo): update target Angular version to 14

gfellerph:
Updated the target Angular version to 14 in the migration instructions. The Intranet Header is not compatible with older versions.


oliverschuerch (approved):

swisspost-devs:

PR #712, opened 2022-11-04
fix: add version to package updated action

gfellerph:


.github/workflows/release.yaml
```diff
@@ -48,7 +48,7 @@ jobs:
       # Check if demo version changed
       - name: Demo version check
         id: demoCheck
-        uses: MontyD/package-json-updated-action
+        uses: MontyD/package-json-updated-action@1.0.1
```
alizedebray, line 51: Shouldn't this be "@v1.0.1" with a "v"?

gfellerph: The GitHub Action marketplace defines it like that, but for other Actions we used the "v". I think it should work both ways. 
image
https://github.com/marketplace/actions/package-version-updated

alizedebray: Ok! I had only seen the usage example in which there is no version

alizedebray (approved):

PR #687, opened 2022-11-02
feat(styles-migration): create migrations with the new cheerio util

oliverschuerch:

PR #713, opened 2022-11-04
feat(styles-migration): add automatic migration for bootstrap form-control, form-select and textarea components

oliverschuerch:


swisspost-devs:

oliverschuerch:
Review #687 first. This will make it easier to review this PR.

gfellerph:
@oliverschuerch can you target that PR at #687? This way we can merge the whole feature when it's complete and it's easier to review this PR.

.changeset/spicy-boats-judge.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-demo': patch
+---
+
+Upated migration documentation for bootstrap `form-control`, `form-select` and `textarea` components.
+Removed floating-label-wrapper and form-control-lg-class changes from `textarea` and `form-file` components, because they never existed.
```
gfellerph, line 6: Is this something that changed from v4 to v5 or just in v5 itself? I think it should be elaborated a little to make sense to end-users.

gfellerph (approved):

.changeset/spicy-boats-judge.md
oliverschuerch: In v4 `textarea` and `form-file` components did not work with floating-labels or at least, there are no examples shown in the docs. Therefore these two components must have gotten a `floating-label` visuality only in v5.
We should just remove the second sentence, so we're fine.

PR #724, opened 2022-11-07
feat(styles-migration): add automatic migration for form-select component

oliverschuerch:


swisspost-devs:

gfellerph (approved):

PR #735, opened 2022-11-08
feat(styles-migration): add automatic migration for bootstrap blockquote component

oliverschuerch:


swisspost-devs:

packages/styles/schematics/utils/dom/migration-rule.ts
```diff
@@ -24,7 +24,12 @@ const CHEERIO_OPTIONS: cheerio.CheerioOptions = {
 const PRETTIER_OPTIONS: prettier.Options = {
   parser: 'html',
   plugins: [htmlParser],
-  htmlWhitespaceSensitivity: 'strict'
+  printWidth: 1000,
+  bracketSpacing: true,
+  bracketSameLine: true,
+  htmlWhitespaceSensitivity: 'ignore',
```
gfellerph, line 30: Are you sure about that? I'm not sure we can assume this will not break any interface.

oliverschuerch: Which option property gives you a headache?

Keep in mind the following:
1. Only manipulated elements will be processed with prettier.format.
2. These elements are given to prettier.format as a single-line-string, because they are preprocessed by cheerio.

gfellerph (approved):

PR #784, opened 2022-11-18
fix(styles): set tooltip opacity to 1

gfellerph:


swisspost-devs:

oliverschuerch (approved):

PR #786, opened 2022-11-18
fix(intranet-header): include default user image

gfellerph:
The default user image was broken when viewed from outside the postnet because internal user images cannot be accessed from the internet. Added the png as base64 encoded string to make it available to anybody.


swisspost-devs:

oliverschuerch (approved):

PR #793, opened 2022-11-21
fix(datepicker): alignment of days when outside="hidden"

gfellerph:
When outside days get hidden, they collapse in their flex context and change the alignment of the first days of the month. Adding flex-grow 1 ensures the hidden days still keep their space so the day labels on top match correctly with the calendar days.


swisspost-devs:

alizedebray (approved):

PR #812, opened 2022-11-25
feat(documentation): add import instruction to intranet header documentation

riscie:
Adds import instructions for the intranet-header component because I stumbled upon it myself earlier.
image

### DIFF
--- a/.changeset/slow-tables-kick.md
+++ b/.changeset/slow-tables-kick.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-demo': patch
+---
+
+Adds import instructions to intranet-header documentation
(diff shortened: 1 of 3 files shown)


swisspost-devs:

.changeset/slow-tables-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-demo': patch
+---
+
+Adds import instructions to intranet-header documentation
```
gfellerph, line 5:
```suggestion
Added import instructions to the intranet-header documentation
```

gfellerph (approved):
@riscie, thank you so much for contributing. I suggest just a slightly different wording on the changeset to align it with the rest, but otherwise the change is an improvement to the docs.

PR #813, opened 2022-11-25
feat(documentation): add alert stories

oliverschuerch:


swisspost-devs:

packages/documentation/stories/components/alert/alert.stories.tsx
```diff
@@ -0,0 +1,238 @@
...
+        '2035',
+        '2101'
+      ],
+      table: { category: 'Icon' }
+    },
+    iconVariant: {
+      name: 'Icon Color',
+      control: {
+        type: 'select',
+        labels: {
```
gfellerph, line 102: Icon color on the alert should be one of the contrast colors, devs should not color it differently in order to keep the color contrast as high as possible. There's functionality in place to automatically handle this in high contrast mode as well using `filter: invert(1)`. I'd not provide the option to color icons here.

packages/documentation/stories/components/alert/alert.stories.tsx
```diff
@@ -0,0 +1,238 @@
...
+          'alert-warning': 'Warning',
+          'alert-info': 'Info'
+        }
+      },
+      options: [
+        'alert-primary',
+        'alert-notification',
+        'alert-success',
+        'alert-danger',
+        'alert-error',
```
gfellerph, line 69: Should we deprecate one of the error/danger classes as well?  Bootstrap does not provide error, only danger.

packages/documentation/stories/components/alert/alert.stories.tsx
```diff
@@ -0,0 +1,238 @@
...
+        'alert-warning',
+        'alert-info'
```
gfellerph, line 71: Bootstrap provides a `secondary` alert class. At the moment, there is one type of alert we do not cover here with a light grey background (https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/Components?node-id=2412%3A68325&t=3Z7AnhTO6FV4qHgT-4). We could either use the secondary variant to represent the light grey alert or create another variant while defining another default for secondary alerts. This can be handled by a separate ticket.

packages/documentation/.storybook/preview.js
```diff
@@ -30,5 +30,10 @@ export const parameters = {
       color: /(background|color)$/i,
       date: /Date$/
     }
+  },
+  docs: {
+    source: {
+      excludeDecorators: true
```
gfellerph, line 36: What effect does that have?

oliverschuerch: If we decorate (wrap) the output code with some html elements, these elements will show up in the code panel. If we exclude decorators in source component, they will not not.
This has nothing to do with angular decorators ;)

packages/documentation/stories/components/alert/alert.stories.tsx
```diff
@@ -0,0 +1,238 @@
...
+          'warning': 'warning',
+          'danger': 'danger',
+        }
+      },
+      options: [
+        '',
+        'white',
+        'primary',
+        'success',
+        'info',
```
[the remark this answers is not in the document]
oliverschuerch: In my opinion, fixing stuff on-the-fly should not be part of the workflow "converting the given demo-app documentation to storybook".
If we need to remove the info color for alerts (for example) this should happen first in the styles and afterwards we can update the demo-app and storybook documentations and if necessary write a migration, to reflect this.

packages/documentation/stories/components/alert/alert.stories.tsx
```diff
@@ -0,0 +1,238 @@
...
+          'alert-success': 'Success',
+          'alert-danger': 'Danger',
+          'alert-error': 'Error',
+          'alert-warning': 'Warning',
+          'alert-info': 'Info'
+        }
+      },
+      options: [
+        'alert-primary',
+        'alert-notification',
```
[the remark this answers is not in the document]
oliverschuerch: I agree.

packages/documentation/stories/components/alert/alert.stories.tsx
oliverschuerch: This is control is only there if you integrate a custom icon (which is by default dark). Don't you think it is important to show the possibility to change the icon color in this case?

packages/documentation/stories/components/alert/alert.stories.tsx
oliverschuerch: In my opinion we could remove the `alert-error` variant.

packages/documentation/stories/components/alert/alert.stories.tsx
oliverschuerch: 1. If there are new styles, we should first update the alert styles (which is not part of this task).
2. We do not have a style for alert-secondary. And we should not start mixing bootstrap and the Design-System together. Not everything that is possible in bootstrap is possible in the Design System and vice versa.

packages/documentation/stories/components/alert/alert.stories.tsx
gfellerph: This would be part of the icon documentation

packages/documentation/stories/components/alert/alert.stories.tsx
gfellerph: #825

packages/documentation/stories/components/alert/alert.stories.tsx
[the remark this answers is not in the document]
gfellerph: #825

packages/documentation/stories/components/alert/alert.stories.tsx
oliverschuerch: Created a new Ticket #827

packages/documentation/src/stories/components/alert.stories.tsx
```diff
@@ -0,0 +1,392 @@
...
+
+  // every time you want to parse one or more elements to a JSX.Element[], concat all the elements to a string and parse them all
+  // even if it would be possible to write some of them directly
+  // the reason for this is, that react elements in a list need a "key" property, single elements don't
+  // and the parser only creates this property, when it receives multiple elements to parse
+  const content: JSX.Element[] | JSX.Element | string = parse(`<h4 class="alert-heading" key="title">${args.title}</h4>${args.content}`);
+
+  return <div className={ classes } role="alert">
+    {/* Dismissible Button */}
+    { (args.dismissible || args.fixed) && !args.action ? <button className="btn-close" data-dismiss="alert" aria-label="Close" onClick={ (e: React.MouseEvent) => onShowToggle(e, args, updateArgs) }></button> : null }
```
alizedebray, line 201: In the snippet, this function is replaced by `function noRefCheck(){}` which makes the code difficult to understand.

image

packages/documentation/src/stories/components/alert.stories.tsx
```diff
@@ -0,0 +1,392 @@
...
+export const Variant: Story = Template.bind({});
+Variant.parameters = {
+  controls: {
+    exclude: [
+      'Icon',
+      'Icon Color',
+      'No Icon',
+      'Dismissible',
+      'Fixed',
+      'Action',
```
alizedebray, line 242:
```suggestion
      'Action Buttons',
```

Most stories have this error, it needs to be replaced in each one.

packages/documentation/src/stories/components/alert.stories.tsx
```diff
@@ -0,0 +1,392 @@
...
+        'alert-warning',
+        'alert-info'
+      ],
+      table: {
+        category: 'Variant',
+        type: {
+          summary: 'cssClass'
+        },
+        defaultValue: {
+          summary: 'alert-{variant}'
```
alizedebray, line 72: As it stands, the "Default" column in the documentation table doesn't make much sense. It would be nice to add the "." in front of the summary to specify that these are CSS classes and even to change the title of the column if it is possible with storybook.

image

packages/documentation/src/stories/components/alert.stories.tsx
```diff
@@ -0,0 +1,392 @@
...
+      }
+    },
+    icon: {
+      name: 'Icon',
+      description: 'Defines a custom icon.',
+      control: {
+        type: 'select',
+        labels: {
+          'null': 'Default',
+          '1001': 'Envelope (1001)',
```
alizedebray, line 83: Icons do not display correctly when used on dark backgrounds.

image

alizedebray (changes requested):

packages/documentation/src/stories/components/alert.stories.tsx
```diff
@@ -0,0 +1,392 @@
...
+    { (args.dismissible || args.fixed) && !args.action ? <button className="btn-close" data-dismiss="alert" aria-label="Close" onClick={ (e: React.MouseEvent) => onShowToggle(e, args, updateArgs) }></button> : null }
+
+    {/* Alert Content */}
+    { args.action ? <div className="alert-content">{ content }</div> : content }
+
+    {/* Alert Action Buttons */}
+    {
+      args.action ?
+      <div className="alert-buttons">
+        <button className="btn btn-primary btn-animated" onClick={ (e: React.MouseEvent) => onShowToggle(e, args, updateArgs) }><span>Akcepti</span></button>
```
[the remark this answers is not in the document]
oliverschuerch: I already fixed this by adding a `transformSource` function in the preview.js.

Can you explain how you would work with the "h" function at this point?

packages/documentation/src/stories/components/alert.stories.tsx
oliverschuerch: I have already updated the table.

packages/documentation/src/stories/components/alert.stories.tsx
oliverschuerch: This is a known issue, we will fix with the new icons.

alizedebray:
I just noticed a display issue with the fixed alert and the alert with action buttons: 
image

oliverschuerch:
@alizedebray Works on my machine.
Does the problem persist?
If "yes": Can you give more context (browser, controls states, ...)?

oliverschuerch:
@alizedebray
I was able to detect the render issue in firefox.
However, this is more of a firefox bug than a issue in our component.
Firefox incorrectly places "fixed" positioned containers, which have a parent with a "transform" statement.
I'll try to find a solution for this, but am not sure if there is one...

gfellerph (approved):
If the styles issue for the icons is fixed, I think we're good to go.

oliverschuerch:
@alizedebray Can you please check if the error you reported is now fixed?
Thanks ;)

packages/documentation/src/stories/components/alert/alert.stories.tsx
```diff
@@ -0,0 +1,305 @@
...
+        '2035',
+        '2101'
+      ],
+      table: {
+        category: 'General'
+      }
+    },
+    dismissible: {
+      name: 'Dismissible',
+      description: 'Adds the dismissible styles.<br/><div className="alert alert-info" style="padding: 0.2rem 0.5rem 0.2rem 2rem; min-height: 0; background-position: 0.5rem center; background-size: 1rem; font-size: inherit;">Do not forget to add the structural adjustments!</div>',
```
alizedebray, line 108: Descriptions are HTML strings, so they should use `class`, not `className`.

Also, they are displayed within spans in the args table, so it's best to only use inline elements and not divs.

```suggestion
      description: 'Adds the dismissible styles.<br/><span class="bg-info rounded light lh-1 d-flex align-items-center gap-mini py-micro px-mini mt-mini"><em class="pi pi-2106 pi-small"></em>Do not forget to add the structural adjustments!</span>',
```

alizedebray (changes requested):
Thank you Olivier for all these changes! 

Last thing, I noticed some unwanted padding around the alerts header:
image

packages/documentation/src/stories/components/alert/alert.stories.tsx
oliverschuerch: Hei @alizedebray 
Thanks for your latest review!

- The "unwanted padding around the alerts header" is coming from the file `@use './layouts/portal';` inside of the `@swisspost/design-system-styles/index.scss`. I don't know why this styles are not present in the generated `@swisspost/design-system-styles/index.css`, but it has nothing to the with a wrong setup of the alert component.
- Descriptions are html strings, that get rendered by the storybook react app and therefore, they need to have the attribute `className` instead of `class`. In the output they are renamed to `class="..."` for sure.
- Your right! But if we do so, we could break some of the expected behaviour on our own components (for example, that the alert should be display: block; per default, etc.).
Because of this reason, I would see this as a downside of storybook, which we can not change, but leave our code as it shoud be.
Is this ok for you?

alizedebray:
> Thank you Olivier for all these changes!
> 
> Last thing, I noticed some unwanted padding around the alerts header: image

The problem appears on several components when using portal styles. I created a separate issue #879.

alizedebray (approved):
Well done! 👍

PR #878, opened 2022-12-13
feat(documentation): add toast stories

oliverschuerch:


swisspost-devs:

packages/documentation/src/stories/components/toast/toast.docs.mdx
```diff
@@ -0,0 +1,56 @@
...
+```
+
+## Auto close
+
+<Canvas>
+  <Story id="components-toast--auto-close" />
+</Canvas>
+
+Toasts should be closed automatically after a while, because they usually overlay the regular content.
+Every change on the screen should be left for at least 6 seconds so that the user can perceive it.
```
alizedebray, line 44:
```suggestion
Each change on the screen should remain for at least 6 seconds for the user to perceive it.
```

packages/documentation/src/stories/components/toast/toast.docs.mdx
```diff
@@ -0,0 +1,56 @@
...
+We therefore recommend a display duration of at least 10-12 seconds, since the message still has to be
```
alizedebray, line 45:
```suggestion
We therefore recommend a display time of at least 10-12 seconds, since the message still has to be
```

packages/documentation/src/stories/components/toast/toast.docs.mdx
```diff
@@ -0,0 +1,56 @@
...
+read.
+
+In addition, if the user moves the mouse or the focus to the `.toast` element, the automatic closing should be disabled. If the mouse/focus leaves the element again, the timeout for the automatic closing can be restarted.
```
alizedebray, line 48:
```suggestion
In addition, if the user moves the mouse or the focus to the `.toast` element, the automatic closing should be disabled. If the mouse/focus leaves the element again, the automatic closing delay should be restarted.
```

packages/documentation/src/stories/components/toast/toast.docs.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+## Accessibility
+
+Toasts are intended to be small interruptions to your visitors or users, so to help those with screen readers and similar assistive technologies, you should wrap your toasts in an `aria-live` region. Changes to live regions (such as injecting/updating a toast component) are automatically announced by screen readers without needing to move the user’s focus or otherwise interrupt the user. Additionally, include `aria-atomic="true"` to ensure that the entire toast is always announced as a single (atomic) unit, rather than just announcing what was changed (which could lead to problems if you only update part of the toast’s content, or if displaying the same toast content at a later point in time). If the information needed is important for the process, e.g. for a list of errors in a form, then use the [alert component](/?path=/docs/components-alert--default) instead of toast.
```
alizedebray, line 52:
```suggestion
Toasts are intended to be small interruptions to your visitors or users, so to help those with screen readers and similar assistive technologies, you should wrap your toasts in an `aria-live` region. Changes to live regions (such as injecting/updating a toast component) are automatically announced by screen readers without the need to move the user’s focus or interrupt in any way. Additionally, include `aria-atomic="true"` to ensure that the entire toast is always announced as a single (atomic) unit, rather than just announcing what changed (which could lead to problems if you only update part of the toast’s content, or if you show the same toast content later in time). If the information is important in the process, e.g. for a list of errors in a form, then use the [alert component](/?path=/docs/components-alert--default) instead of the toast.
```

alizedebray (changes requested):

packages/documentation/src/stories/components/toast/toast.docs.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+Note that the live region needs to be present in the markup before the toast is generated or updated. If you dynamically generate both at the same time and inject them into the page, they will generally not be announced by assistive technologies.
+
+You also need to adapt the `role` and `aria-live` level depending on the content. If it’s an important message like an error, use `role="alert" aria-live="assertive"`, otherwise use `role="status" aria-live="polite"` attributes.
```
gfellerph, line 56: Nice write up. Please note that ng-bootstrap already provides aria-live attributes on the container, so this is only relevant for users of the plain bootstrap toasts. For them, we could also link to the relevant bootstrap docs section (https://getbootstrap.com/docs/5.1/components/toasts/#stacking).

packages/documentation/src/stories/components/toast/toast.stories.tsx
```diff
@@ -0,0 +1,319 @@
...
+      { args.show && component }
+    </div>;
+  } else {
+    return component;
+  }
+};
+
+export const Default: Story = Template.bind({});
+Default.decorators = [
+  (Story: Story, { args }) => {
```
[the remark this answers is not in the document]
oliverschuerch: Sounds legit

packages/documentation/src/stories/components/toast/toast.docs.mdx
oliverschuerch: Haha, you mean nice copy-paste 🤣

It was my intention, to use this Page only to document the bootstrap/toast component (at least for the moment).
Hope this is ok?

Would you prefer to just link over to bootstrap or to implement an stacking example by our own?

packages/documentation/src/stories/components/toast/toast.stories.tsx
```diff
@@ -0,0 +1,394 @@
...
+      table: {
+        category: 'General'
+      }
+    },
+    position: {
+      name: 'Position',
+      description: 'Defines the component\'s positionning.',
+      control: {
+        type: 'radio',
+        lables: {
```
alizedebray, line 132:
```suggestion
        labels: {
```

alizedebray (changes requested):

packages/documentation/src/stories/components/toast/toast.docs.mdx
gfellerph: image

Yes, documenting plain bootstrap only is ok. A stacking example would be nice though (stop disabling that button and let me hammer out as many toasts as I can).

alizedebray (approved):

PR #889, opened 2022-12-16
fix(styles): unwrap complex :is selector

gfellerph:
Sass seems not entirely ready for :is. When integrating in another project, this mixin causes sass to silently fail and generate an empty output. The bug is not recreatable in a different environment and therefore we could not raise an issue with sass (yet).


swisspost-devs:

alizedebray (approved):

PR #929, opened 2023-01-03
feat(styles): add font only css export

gfellerph:
Added a font-face definition CSS file for use with SAP cloud solutions as they need an externally hosted CSS file.


swisspost-devs:

PR #945, opened 2023-01-09
chore(documentation): update button stories

oliverschuerch:


swisspost-devs:

packages/documentation/src/utils/react.ts
```diff
@@ -0,0 +1,7 @@
+export function forEach (iterable: any[] | object, callback: Function) {
+  const input = Object.assign(iterable);
+
+  return Object
+    .entries(input)
+    .map(([key, value]) => callback({ key, value }));
```
alizedebray, line 6: Why not just use `.map((value, index) => {...});` on the arrays?

oliverschuerch: Because we're able to iterate over objects as well as arrays with the given solution.

PR #951, opened 2023-01-13
chore(documentation): update card stories

oliverschuerch:


swisspost-devs:

swisspost-devs:

packages/documentation/src/stories/components/card/card.docs.mdx
```diff
@@ -0,0 +1,80 @@
...
+### Image overlay
+
+<Canvas>
+  <Story id="components-card--image-overlay" />
+</Canvas>
+
+<div className="alert alert-danger mb-3">
+  Don't do this in production. It's very difficult for content editors to always ensure that text is
+  legible on an image for every browser size and all different languages.
+</div>
```
gfellerph, line 80: image

Text without background on an image is generally not allowed. There will be an example when we have a hero component, but for cards I don't think we need an image with overlayed text. 

Hero: https://www.experience-hub.ch/document/2803#/organisms-module/hero

gfellerph (changes requested):

packages/documentation/src/stories/components/card/card.docs.mdx
oliverschuerch: It's only here because it is possible to do this by bootstrap and it is documented in the old docs.
What do you think i should do with it?

gfellerph: I'd remove it from the docs

oliverschuerch: As discussed face-to-face I will remove only the code view and update the "Don't do this..." text.
Additionaly I created a ticket for @moniesan to document such stuff on the experience hub too:
https://github.com/swisspost/design-system-planning/issues/157

gfellerph (approved):

PR #953, opened 2023-01-13
chore(documentation): update heading stories

oliverschuerch:


swisspost-devs:

packages/documentation/src/stories/components/button/button.docs.mdx
```diff
@@ -1,6 +1,6 @@
 import { Meta, Canvas, Story, ArgsTable } from '@storybook/addon-docs';
 
-<Meta title="Design System/Button" />
+<Meta title="Components/Button" />
```
gfellerph, line 3: I think this change should be in another PR, but it's a detail.

gfellerph (approved):

packages/documentation/src/stories/components/button/button.docs.mdx
oliverschuerch: It's not only a detail, it is actually doing nothing, because it is in a docs.mdx file.

PR #988, opened 2023-01-23
feat(documentation): add page structures of stories we still need to write

oliverschuerch:
- Added TODOs on the pages.
- Changesets should be created as soon as one or more of the stories gets written. And therefore are not part of this PR.


swisspost-devs:

packages/documentation/src/stories/foundation/zindex/zindex.stories.mdx
```diff
@@ -0,0 +1,30 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta
+  title="Foundations/Z-Index"
```
gfellerph, line 4: I'm not happy with that title, but we can discuss that when the documentation is written.

gfellerph (approved):

alizedebray (approved):

packages/documentation/src/stories/foundation/zindex/zindex.stories.mdx
oliverschuerch: I chaneged it to "elevation". Hope you can agree with that?

PR #1003, opened 2023-01-23
chore(icons): create workflow to regularly update icon SVGs

alizedebray:
Missing steps in the current workflow:
- [x] generation of a body for the PR (should ideally list the name of all added, deleted and modified icons or at least their count)
- [x] generation of a changeset file
- [x] triggering of the documentation deployment workflow

The workflow should also be tested on real GitHub environment


swisspost-devs:

PR #1011, opened 2023-01-27
fix(internet-header): patched relative URLs in search recommendations

gfellerph:
Patched relative URLs for the "most searched services" feature of the search box. Relative URLs are always resolved to the base "https://post.ch"


swisspost-devs:

oliverschuerch (approved):

PR #1038, opened 2023-01-30
chore(styles): remove obsolete dependencies

gfellerph:


swisspost-devs:

oliverschuerch (approved):
I did not check if the package is still running without these packages.
Only reviewed the code it self.
If I should check the packages ability to run, leave me a note.

PR #1040, opened 2023-01-30
chore(icons): create workflow to regularly update icon SVGs

alizedebray:


.github/workflows/fetch-icons.yaml
```diff
@@ -0,0 +1,112 @@
...
+#
+# https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+#
+###
+
+name: Fetch Icons
+
+on:
+  schedule:
+    - cron: '* * * * 0'
```
gfellerph, line 18: Can you add a comment explaining what that means?

.github/workflows/fetch-icons.yaml
```diff
@@ -0,0 +1,112 @@
...
+      - name: Exit if no changes
+        if: steps.changed-files.outputs.any_modified == 'false'
+        run: |
+          git checkout main
+          git branch --delete ${{ steps.current-branch.outputs.branch }}
+          exit 1
+
+      - name: Create Changeset
+        run: |
+          echo ---                                      > $CHANGESET
```
gfellerph, line 63: Not sure if we're missing a `>` here.

```suggestion
          echo ---                                      >> $CHANGESET
```

.github/workflows/fetch-icons.yaml
```diff
@@ -0,0 +1,112 @@
...
+          labels+=( "Deleted icons" );  changes["Deleted icons"]=$ADDED_ICONS
+
+          for label in "${labels[@]}"; do
+            if [[ ${changes[$label]} != '' ]]; then
+              echo "## $label"                                                                         >> $PR_BODY_FILE
+              for file in ${changes[$label]}; do
+                echo "- $(echo $file | sed 's/^packages\/icons\/public\/svg\///' | sed 's/\.svg$//')"  >> $PR_BODY_FILE
+              done
+            fi
+          done
```
gfellerph, line 100: I think this needs some explanation as well. Is this just writing the PR body file or setting git labels as well?

.github/workflows/fetch-icons.yaml
```diff
@@ -0,0 +1,112 @@
+###
+#
+# Create a pull request with up-to-date icon SVGs
+#
+# This workflow does not have secrets access
+# when run from a fork. Artifacts are uploaded
+# and used in a subsequent workflow with more
+# privileges which does not run unsafe commands.
+#
+# https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
```
gfellerph, line 10: I don't think this comment is up to date and I also think that this workflow should not run on forks because it uses a github token and the fetchSVG command needs API secrets to be set in the env. We could prevent the workflow from running on a fork with something described here https://github.com/orgs/community/discussions/26409#discussioncomment-3251822 and I think the owner would be 'swisspost' in this case. What do you think?

gfellerph (changes requested):
Looks really nice

.github/workflows/fetch-icons.yaml
```diff
@@ -0,0 +1,112 @@
...
+          BRANCH_NAME: ${{ steps.current-date.outputs.date }}-update-icon-svgs
+
+      - name: Fetch Icons
+        run: pnpm --filter design-system-icons fetchSVGs
+
+      - name: Get Changes
+        id: changed-files
+        uses: tj-actions/changed-files@v35
+        with:
+          files: ./packages/icons/public/svg/**
```
oliverschuerch, line 52: This path will change to: `./packages/icons/public/post-icons/**`

oliverschuerch (changes requested):

.github/workflows/fetch-icons.yaml
alizedebray: This is currently only for the PR body. Do I also need to add tags?

.github/workflows/fetch-icons.yaml
alizedebray: It is a good idea, I just updates this.

.github/workflows/fetch-icons.yaml
gfellerph: No, git tags would be unnecessary. In this case, everything is good.

.github/workflows/fetch-icons.yaml
```diff
@@ -0,0 +1,125 @@
...
+
+on:
+  schedule:
+#            ┌───────────── minute (0 - 59)
+#            │ ┌───────────── hour (0 - 23)
+#            │ │ ┌───────────── day of the month (1 - 31)
+#            │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+#            │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+#            │ │ │ │ │
+    - cron: '0 0 * * 0'  # the job will run every Sunday at 0:00
```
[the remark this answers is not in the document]
alizedebray: Thanks ! But not to lie, it comes directly from the doc 😁

PR #1045, opened 2023-02-01
fix: deploy to next.design-system.post.ch

gfellerph:
Switched from styles.design-system to next.design-system.post.ch for better naming of the beta docs.


swisspost-devs:

oliverschuerch (approved):

PR #1058, opened 2023-02-02
fix(header): missing margin for lg viewport

gfellerph:


swisspost-devs:

packages/internet-header/src/index.html
```diff
@@ -36,6 +36,7 @@
         header.addEventListener('headerLoaded', e => console.log('header loaded'));
         await customElements.whenDefined(header.tagName.toLocaleLowerCase());
         console.log('header defined');
+        header.activeRoute = 'https://int.post.ch/de/kundencenter/kundencenter';
```
oliverschuerch, line 39: This has probably nothing to do with a margin.
But I think you know what you're doing ;)

oliverschuerch (approved):

packages/internet-header/src/index.html
gfellerph: It's just for testing the margin

PR #1062, opened 2023-02-02
fix(styles): default table borders

gfellerph:


swisspost-devs:

packages/styles/src/variables/components/_tables.scss
```diff
@@ -23,9 +21,10 @@ $table-hover-bg: rgba(color.$black, 0.075) !default;
 $table-active-bg: $table-hover-bg !default;
 
 $table-border-width: commons.$border-width !default;
-$table-border-color: rgba(var(--post-contrast-color-inverted-rgb), 0.4) !default;
+$table-border-color: rgba(var(--post-contrast-color-rgb), 0.4) !default;
+$table-group-separator-color: rgba(var(--post-contrast-color-rgb), 0.4) !default;
 
-$table-head-bg: rgba(var(--post-contrast-color-inverted-rgb), 0.6) !default;
+$table-head-bg: rgba(var(--post-contrast-color-rgb), 0.6) !default;
```
alizedebray, line 27: It is not a Bootstrap variable and it is not used in tables.scss thus it can probably be removed.

gfellerph: I opened a new issue for that -> #1078

PR #1064, opened 2023-02-02
test: disable demo percy snapshot

gfellerph:
This test creates a lot of unnecessary snapshots of a page that has a random sort order of contributors which always leads to regression tests failing for all kinds of PRs. Renaming disables the test but keeps it for reference.


swisspost-devs:

gfellerph:
Unlinked the repo in the percy settings.

PR #1067, opened 2023-02-03
fix(docs): style headings more specifically

gfellerph:
Headings for the docs have been styled in a way that affected headings inside components. To prevent this, a more specific selector is being used now.


swisspost-devs:

alizedebray (approved):

PR #1068, opened 2023-02-03
fix(intranet-header): update currentUserID typing

gfellerph:
Updated the type for prop currentUserId to string


swisspost-devs:

alizedebray (approved):

PR #1082, opened 2023-02-07
fix(header): convert env to lowercase for comparison

gfellerph:
Fixed environment string comparison when upper case characters are used to define the environment property


swisspost-devs:

alizedebray (approved):

PR #1120, opened 2023-02-15
fix: remove slugify

gfellerph:
It's no longer necessary to create a slug for places.post.ch detail pages.


swisspost-devs:

oliverschuerch (approved):

PR #1154, opened 2023-02-22
fix(styles): update contrast function

gfellerph:
Using a new algorithm to calculate color contrast as the old one did not satisfy WCAG 2.1 AA in all cases (success button). The new check is based on this blogpost https://codetuts.dev/text-contrast-sass/ with adaptions for modern sass syntax and our needs.


swisspost-devs:

alizedebray (approved):

PR #1155, opened 2023-02-22
fix(styles): floating labels should be truncated

gfellerph:
Fixed an issue with floating labels on input fields that prevented the label from being truncated correctly if it was overflowing the text box.


swisspost-devs:

alizedebray (approved):

PR #1164, opened 2023-02-23
chore(components): add proper linting to the components package

alizedebray:


.changeset/small-ads-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Added linting configuration to ensure all Stencil components conform to the [Stencil style guide](https://stenciljs.com/docs/style-guide).
```
gfellerph, line 5: For internal changes, it's not necessary to write a changeset, because we're not changing anything on the components and therefore it's less relevant for devs (https://github.com/swisspost/design-system/blob/main/CONTRIBUTING.md#versioning)

packages/components/.eslintrc.json
```diff
@@ -0,0 +1,76 @@
...
+    "@stencil-community/own-methods-must-be-private": "error",
+    "@stencil-community/own-props-must-be-private": "error",
+    "@stencil-community/prefer-vdom-listener": "error",
+    "@stencil-community/props-must-be-public": "error",
+    "@stencil-community/props-must-be-readonly": "error",
+    "@stencil-community/render-returns-host": "error",
+    "@stencil-community/required-jsdoc": "error",
+    "@stencil-community/reserved-member-names": "error",
+    "@stencil-community/single-export": "error",
+    "@stencil-community/strict-mutable": "error"
```
gfellerph, line 74: Why did you override these values, are they not defined by @stencil-community/recommended?

gfellerph (changes requested):

packages/components/.eslintrc.json
alizedebray: According to @stencil-community/stencil-eslint README, these are all [recommended rules](https://github.com/stencil-community/stencil-eslint#recommended-rules). However, using @stencil-community/recommended alone, I noticed that most of these errors were actually not shown and I had to add them explicitly to get them to work.

.changeset/small-ads-leave.md
alizedebray: I just removed the changeset.

packages/components/.eslintrc.json
gfellerph: Ah, I see. The standard rules could be an installable package as well... I think it's fine like this for now.

gfellerph (approved):

PR #1169, opened 2023-02-23
chore(components): refactor property checkers

alizedebray:


swisspost-devs:

.changeset/thirty-carrots-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Refactored property checking utility functions to allow more focused unit testing.
```
gfellerph, line 5: This is not a relevant change for end users, so I'd not create a changeset (https://github.com/swisspost/design-system/blob/main/CONTRIBUTING.md#versioning)

packages/components/src/utils/property-checkers/check-one-of.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright 2023 by Swiss Post, Information Technology
+ */
```
gfellerph, line 3: I'm not sure if this is required, since we have a license at every package root containing a copyright notice. Do you have more infos about this?

gfellerph (changes requested):

packages/components/src/utils/property-checkers/check-one-of.ts
alizedebray: Good question, this is added automatically as soon as I create a new file with intelliJ but I don't know if it is necessary in our case. I will delete that for now.

gfellerph (approved):

PR #1181, opened 2023-02-24
feat(components): add post-tabs component

alizedebray:


swisspost-devs:

packages/components/src/components/post-tab-panel/post-tab-panel.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Hides the tab panel programmatically.
+   */
+  @Method()
+  async hide() {
+    this.isShown = false;
+    await onTransitionEnd(this.tabPanelElement).then(() => {
+      this.host.remove();
+    });
```
gfellerph, line 63: Will `transitionend` be fired if the user prefers reduced motion? We could use the [web animation API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API/Using_the_Web_Animations_API) to do the transitions. This will return a promise that resolves anyways.

packages/components/src/utils/property-checkers/check-exists.ts
```diff
@@ -0,0 +1,7 @@
+export function checkExists<T = unknown>(value: T, error: string) {
+  const valueIsUndefined = typeof value === 'undefined';
```
gfellerph, line 2: Why is this a string?

```suggestion
  const valueIsUndefined = typeof value === undefined;
```

packages/styles/src/components/tabs/_tab-title.scss
```diff
@@ -0,0 +1,150 @@
...
+  padding: nav.$nav-link-padding;
+  transition: background-color 100ms;
+  border-right: 1px solid transparent;
+  border-left: 1px solid transparent;
+  outline-color: currentColor;
+  opacity: 0.7;
+  color: var(--post-contrast-color);
+
+  &:focus {
+    opacity: 0.7;
```
gfellerph, line 21: Is already defined in the standard rule.

```suggestion
```

packages/components/src/components/post-tab-panel/post-tab-panel.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Shows the tab panel programmatically.
+   */
+  @Method()
+  async show() {
+    this.associatedHeader.insertAdjacentElement('afterend', this.host);
+    setTimeout(() => {
+      this.isShown = true;
+    });
```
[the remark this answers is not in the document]
alizedebray: A timeout is necessary here for the transition to be done correctly: a first rendering adds the panel to the DOM without the `.show` class, then a second rendering adds the `.show` class causing a change of styles and therefore a transition.
I added a comment.

packages/components/src/utils/property-checkers/check-exists.ts
alizedebray: Because `typeof` returns a string (I used `typeof` as a [best practice](https://thisthat.dev/variable-undefined-vs-typeof-variable-undefined/))

packages/components/src/components/post-tab-panel/post-tab-panel.tsx
alizedebray: I wrote the [onTransitionEnd](https://github.com/swisspost/design-system/blob/main/packages/components/src/utils/on-transition-end.ts) function to return a promise that directly resolves if the user prefers reduced motion but waits for the transition to end otherwise.

The Web Animations API is also a good alternative but it involves rewriting Bootstrap animations.

packages/components/src/components/post-tab-header/post-tab-header.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  /**
+   * Activates the tab programmatically.
+   */
+  @Method()
+  async activate() {
+    this.isActive = true;
+    if (this.isLoaded) {
+      this.activated.emit();
+    }
```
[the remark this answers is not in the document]
alizedebray: This applies to interactions from outside of the component, but here the method is also called from inside by the `validateActive` method which is first executed with the `componentWillLoad` hook.

packages/components/src/components/post-tab-panel/post-tab-panel.tsx
gfellerph: I see. There is one more benefit I used in the header. If the animation should be cancelled because the user cancelled the action (clicking everywhere like a madman), the animation can be reversed instantly by calling `animation.reverse()`. This could lead to snappier interactions because orchestration is easier and predictable. We could discuss this next Thursday.

gfellerph:
- A lot of slot checks are happening. Can this be simplified?

oliverschuerch:
@alizedebray sorry to contaminate your PR, but I had to put it somewhere.

I just did some tests to see if a totally different design of the tab components might work.
As a starting point I used the design of the tabs comoponents from the bootstrap-vue project: https://bootstrap-vue.org/docs/components/tabs.

This has resulted in a two-component solution, but it has two major problems:

1. Nested slots (add elements to a slot in the parent-component from the child-component) are not realizable with web-components in the same way as with JS-frameworks. Only top-level children can be slotted.
As a result, JS-logic was needed to make the components work "correctly".

2. The slotchange event has been used to correctly include updates in one specific slot into the post-tabs component.
Unfortunately the slotchange event is not able to recognize text-node changes. Instead it only detects if a slotted-element is added or removed. A mutationObserver is also not an alternative, since this also does not recognize changes to pure text-nodes.

Conclusion:
Maybe someone find the example helpful.
But as long as it is not possible to include non top-level children into a slot of a web-component it's not really a good solution.

Nevertheless, I would like to share the output:

```tsx
// post-tab.tsx component (stenciljs)
import { Component, h, Host } from '@stencil/core';

@Component({
  tag: 'post-tab',
  shadow: true,
})
export class PostTab {
  render() {
    return (
      <Host>
        {/* the "tab" slot is defined in a template-tag, because it's not where it should be at the end of the render processs */}
        <template>
          <li class="tab-item" slot="tabs">
            <a href="#">
              <slot name="tab">tab</slot>
            </a>
          </li>
        </template>

        <div class="tab-content">
          <slot>content</slot>
        </div>
      </Host>
    );
  }
}
```

```tsx
// post-tabs.tsx wrapper component (stenciljs)
import { Component, Element, h, Host } from '@stencil/core';

@Component({
  tag: 'post-tabs',
  shadow: true,
})
export class PostTabs {
  @Element() el!: HTMLPostTabsElement;

  render() {
    return (
      <Host>
        <ul class="tabs">
          <slot name="tabs"></slot>
        </ul>

        <div class="tabs-content">
          {/*
            all the post-tab components and there content goes here
            we need to extract the post-tab template content (which should be placed in the post-tabs "tabs" slot) by js
          */}
          <slot></slot>
        </div>
      </Host>
    );
  }

  componentDidRender() {
    // after every rendering of the component, we need to update the top-level children of post-tabs
    this.assignHeadSlot();
  }

  componentDidLoad() {
    this.el.querySelectorAll('post-tab').forEach(tab => {
      tab.shadowRoot.querySelector('slot[name="tab"]').addEventListener('slotchange', () => {
        // if the post-tab "tab" slot content changes, we need to update the top-level children of post-tabs
        this.assignHeadSlot();
      });
    });
  }

  assignHeadSlot() {
    // we need to remove previously added post-tab template contents
    this.el.querySelectorAll(':scope > [slot="tabs"]').forEach(tab => tab.remove());

    // then need to clone all post-tab template contents as a top-level child into the post-tabs component
    this.el.querySelectorAll('post-tab').forEach(tab => {
      // the post-tab template content
      const tabContent = tab.shadowRoot.querySelector('template [slot="tabs"]');
      // the post-tab "tab" slot element
      const tabSlot = tabContent.querySelector('[name="tab"]');
      // the post-tab "tab" slotted elements
      const tabSlotContent = tab.querySelectorAll('[slot="tab"]');

      if (tabSlotContent.length > 0) {
        // remove the post-tab "tab" slot default content
        tabSlot.innerHTML = '';
        // append assignedElements to the slot itself and make them its default/fallback value
        // we need to do this, because after cloning the post-tab template content to the post-tabs component, the slot will not work as expected anymore
        tabSlotContent.forEach(el => {
          tabSlot.append(el.innerHTML);
        });
      }

      // append a clone of the updated template content element to the post-tabs component (as top-level children)
      // this will make the cloned element a slotted elemente of the post-tabs "tabs" slot
      this.el.append(tabContent.cloneNode(true));
    });
  }
}
```

```html
// index.html
<post-tabs>

  <!-- post-tab element will be slotted into the "default" slot of the post-tabs component -->
  <post-tab>
    <!-- span[slot="tab" will be slotted into the "tab" slot of the post-tab conponent -->
    <span slot="tab">tab 1</span>
    <!-- everything else will be slotted into the "default" slot of the post-tab component -->
    <div>
      Lorem ipsum, dolor sit amet consectetur adipisicing elit. Consectetur doloribus eius
      veritatis omnis, corrupti rem aliquid! At facere minus impedit deleniti incidunt.
      Voluptas officiis laudantium reprehenderit magnam quia ducimus ullam.
    </div>
  </post-tab>

  <post-tab>
    <span slot="tab">tab 2</span>
    <div>
      Lorem ipsum, dolor sit amet consectetur adipisicing elit. Consectetur doloribus eius
      veritatis omnis, corrupti rem aliquid! At facere minus impedit deleniti incidunt.
      Voluptas officiis laudantium reprehenderit magnam quia ducimus ullam.
    </div>
  </post-tab>
</post-tabs>
```

gfellerph:
- Use the 3 component approach with custom props for connecting header and panel.
- Don't do checks if child components are in the right place
- Put the tab-panel-header into the correct slot if it's not already assigned
- Fall back to index based connection between header and tabs if nothing is specified

packages/components/src/components/post-tab-panel/post-tab-panel.tsx
```diff
@@ -0,0 +1,28 @@
...
+   * The name of the panel, used to associate it with a tab header.
+   */
+  @Prop() readonly name: string;
+
+  render() {
+    return (
+      <div
+        aria-labelledby={`${this.name}--tab`}
+        class="tab-pane fade active show"
+        id={`${this.name}--panel`}
```
alizedebray, line 21: This id is not secure enough to be sure that there will never be duplicates.

@gfellerph @oliverschuerch @imagoiq 
Should we go for the id/for solution where devs will be responsible for providing unique ids or should we stick with the name/panel approach and create uniq ids ourselves?

gfellerph: According to the resolution, keep the custom prop names for connecting header and panel and generate uinique ids for internal usage with `crypto.randomUUID` (but this will only work on https and localhost, not on http pages) or use a custom function for generating more or less random strings (always start with a letter in order to be compliant with id requirements).

alizedebray: Ok, thanks for the clarification

packages/documentation/src/stories/components/tab/tab.stories.tsx
```diff
@@ -21,18 +21,17 @@ export default {
 
 const Template = () => (
   <PostTabs>
-    <PostTabHeader slot="tab">First</PostTabHeader>
-    <PostTabPanel slot="panel">
+    <PostTabHeader slot="tabs" panel="first">First</PostTabHeader>
+    <PostTabHeader slot="tabs" panel="second">Second</PostTabHeader>
+    <PostTabHeader slot="tabs" panel="third">Third</PostTabHeader>
```
imagoiq, line 26: Can you explain why you are using tabs in plural, please?

imagoiq (approved):

packages/components/src/components.d.ts
```diff
@@ -119,9 +166,34 @@ declare namespace LocalJSX {
...
+    }
+    interface PostTabs {
+        /**
+          * The name of the panel that is initially shown.
+         */
+        "activePanel"?: string;
+        /**
+          * An event emitted whenever a new tab becomes active. The payload is the name of the associated panel.
+         */
+        "onTabChange"?: (event: PostTabsCustomEvent<string>) => void;
```
imagoiq, line 189:
```suggestion
        "onTabChange"?: (event: PostTabsCustomEvent<PostTabPanel["name"]>) => void;
```
Nitpick. I'm not sure that it will work with the optional character.

packages/documentation/src/stories/components/tab/tab.stories.tsx
alizedebray: Because it is the slot where all the tabs are: 

image

packages/components/src/components.d.ts
alizedebray: This file is generated.
To update the declaration, we should update the component:
```
/**
 * An event emitted whenever a new tab becomes active.
 * The payload is the name of the associated panel.
 */
@Event() tabChange: EventEmitter<HTMLPostTabPanelElement['name']>;
```

Is that what you had in mind?
If so, would you change it everywhere a `string` is used to type a panel name variable (i.g. in the `show` method or for the `panel` prop of the tab header)?

imagoiq: I would say if possible. And just to clarify, it's only a proposal, I know that it can be annoying in some situations, but it helps to read the code.

alizedebray: I used `HTMLPostTabPanelElement['name']` in most place

packages/documentation/src/stories/components/tabs/tabs.snapshot.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+type Story = StoryObj<HTMLPostTabsElement>;
+
+export const Tabs: Story = {
+  render: (_args: HTMLPostTabsElement, context: StoryContext<HTMLPostTabsElement>) => {
+    return html`
+      <div>
+        ${['bg-white', 'bg-dark'].map(bg => html`
+          <div
+            class="${bg} mt-3"
+            style="padding: 1rem; display: flex; flex-direction: column; gap: 1rem;"
```
imagoiq, line 20: I think you can use bootstrap here as well

packages/styles/src/components/tabs/_tabs-wrapper.scss
```diff
@@ -0,0 +1,69 @@
...
+  @include utilities-mx.reset-list;
+  display: flex;
+  flex-wrap: nowrap;
+  overflow-x: auto;
+
+  /* prevent scroll chaining on x scroll */
+  overscroll-behavior-x: contain;
+  white-space: nowrap;
+
+  // Scrolling fix, make the tabs scroll a little more so the right most tab is not
```
imagoiq, line 47: I think it's missing a word after 'a little more'

imagoiq (changes requested):
On Firefox (at least), after multiple clicks, the content of the tab disappears:

https://github.com/swisspost/design-system/assets/12294151/b79a03c1-fdde-4220-8ae7-1b064575898b

packages/documentation/src/stories/components/tabs/tabs.snapshot.stories.tsx
alizedebray: We had styles in most snapshot stories, I created PR #1844 to fix this.

alizedebray:
> On Firefox (at least), after multiple clicks, the content of the tab disappears:

Good point, I prevented tab switch when clicking an active tab.

packages/styles/src/components/tabs/_tabs-wrapper.scss
imagoiq: Sorry, I've misread it I think. Maybe this help to read it:

```suggestion
  // Scrolling fix to make the tabs scroll a little more so the right most tab is not
```
or

```suggestion
  // Scrolling fix: make the tabs scroll a little more so the right most tab is not
```

or

```suggestion
  // Make the tabs scroll a little more so the right most tab is not
```

packages/components/src/components/post-tab-panel/post-tab-panel.tsx
```diff
@@ -0,0 +1,38 @@
...
+  componentWillLoad() {
+    // get the id set on the host element or use a random id by default
+    this.panelId = this.host.id || crypto.randomUUID();
+  }
+
+  render() {
+    return (
+      <Host data-version={version} id={this.panelId}>
+        <div
+          aria-labelledby={`${this.panelId}--tab`}
```
gfellerph, line 28: This won't work since different random id's will be generated for every component:

image

It's important that only one component generates the ID and distributes it to the others. We could let the post-tab-header generate the id and write it from there on the connected post-tab-panel in the componentDidLoad.

packages/components/src/components/post-tabs/post-tabs.tsx
```diff
@@ -0,0 +1,161 @@
...
+    this.showing.onfinish = () => {
+      this.showing = null;
+    };
+  }
+
+  render() {
+    return (
+      <Host data-version={version}>
+        <div class="tabs-wrapper">
+          <ul class="tabs nav" role="tablist">
```
gfellerph, line 151: The semantics don't work with voice over. Voice over announces "[tab text], tab, 1 of 1". It does not recognise the other tabs as belonging to the same group. Might be due to the fact that the web component is the child of the ul and not the li. Not sure how to handle this though.

packages/documentation/src/stories/components/tabs/tabs.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+const meta: Meta<HTMLPostTabsElement> =  {
+  title: 'Components/Tabs',
+  component: 'post-tabs',
+  render: renderTabs,
+  parameters: {
+    badge: [BADGE.BETA, BADGE.NEEDS_REVISION],
+  },
+  argTypes: {
+    activePanel: {
+      name: 'active-panel',
```
gfellerph, line 15:
```suggestion
      name: 'Initial panel',
      description: 'The initially active panel. Changing this value after initialization won\'t have any effect.'
```

packages/styles/src/components/tabs/_tab-title.scss
```diff
@@ -0,0 +1,153 @@
...
+  transition: background-color 100ms;
+  border-right: 1px solid transparent;
+  border-left: 1px solid transparent;
+  outline-color: currentColor;
+  opacity: 0.7;
+  color: var(--post-contrast-color);
+  text-decoration: none;
+
+  &:focus {
+    opacity: 0.7;
```
gfellerph, line 23: Opacity is defined twice.

```suggestion
  opacity: 0.7;
  color: var(--post-contrast-color);
  text-decoration: none;

  &:focus {
```

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
gfellerph, line 1: Apparently, later v8 versions write 6.0 again, after some people got really angry https://github.com/pnpm/pnpm/issues/6648.

```suggestion
lockfileVersion: '6.0'
```

gfellerph (changes requested):
We can also open tickets for fixes that we want to deliver later.

packages/components/src/components/post-tab-panel/post-tab-panel.tsx
```diff
@@ -0,0 +1,38 @@
...
+  @State() panelId: string;
+
+  /**
+   * The name of the panel, used to associate it with a tab header.
+   */
+  @Prop() readonly name: string;
+
+  componentWillLoad() {
+    // get the id set on the host element or use a random id by default
+    this.panelId = this.host.id || crypto.randomUUID();
```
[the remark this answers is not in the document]
gfellerph: This should be removed because it's not working, see https://github.com/swisspost/design-system/pull/1181#discussion_r1302861651

packages/documentation/src/stories/components/tabs/tabs.stories.tsx
gfellerph: You're right, I wanted to highlight the fact that the component does not update and that that's intended behavior.

packages/components/src/components/post-tabs/post-tabs.tsx
alizedebray: I cannot reproduce this issue, did you use NVDA?

packages/documentation/src/stories/components/tabs/tabs.stories.tsx
alizedebray: I updated the description.

packages/components/src/components/post-tabs/post-tabs.tsx
alizedebray: The semantic works fine with NVDA, I created the ticket #1869 to find a solution with VoiceOver.

packages/components/src/components/post-tab-panel/post-tab-panel.tsx
alizedebray: I updated the behavior: now the header and panel both generate their ID's separately then, once all headers and panels are fully loaded the post-tabs gets the ID's and uses them to set aria attributes.

gfellerph (approved):
Can't believe it, but it finally happens.

imagoiq (approved):

PR #1192, opened 2023-02-27
chore(icons): improve how the "Fetch Icons" writes PR body

alizedebray:
The "Fetch Icons" PR is now created using a personal access token so that all necessary checks and deployments are triggered.

In addition, all modified icons are now taken into account when writing the PR body but only the name of the first 15 is listed:

image


gfellerph (approved):
Awesome!

PR #1214, opened 2023-03-03
chore: deprecate alert-error and alert-notification

gfellerph:
Deprecated the `alert-error` and the `alert-notification` variants of the alert component. These variants will be removed in the next major version.


swisspost-devs:

oliverschuerch (approved):

PR #1228, opened 2023-03-07
fix: add sort order plugin

gfellerph:
Sort attributes on SVG icons so builds are deterministic. This ensures that the automatic icon update PR only updates icons that have changed in content.


swisspost-devs:

oliverschuerch (approved):

PR #1230, opened 2023-03-07
feat(styles): decouple fixed position and actions for alerts

gfellerph:
Decoupled the fixed and the action button styles for the alert component. Alerts can now be fixed to bottom without having action buttons and can have action buttons without being fixed to the bottom of the page.


oliverschuerch (approved):
Nice work!

Issue #1239, opened 2023-03-08
How to keep old documentation alive


gfellerph:
Is there a way on netlify to keep old documentation pages alive?

## First findings

- Branch deploys (previews) are being deleted after 90 days (unless it's a GitHub branch deploy, which is not linked in our use case)
- Production deploys are kept forever and have a unique URL (e.g. https://6401b5c167e11f27f2f741f8--swisspost-design-system-next.netlify.app). At the moment it's not possible to automatically link a unique production deploy to a specific Design System version, but a list could be established manually, at least for major and minor releases
- If a public json containing a version : url mapping is published on https://design-system.post.ch, even old applications would always fetch the latest json and could link to newer versions

## Second findings
Automating the generation of the versions file is pretty hard. Steps would look like:
1. Deploy the production site in order to get the correct URL
2. Read URL from success message
3. Generate versions.json
4. Re-deploy the whole site with the new versions.json
5. Commit new versions.json to main

++ Always up to date
-- Two deploys necessary
-- Two commits to main on publish

Alternative approach, always lagging behind one version but would be ok because the current version is always on the same URL:
1. When generating the release preview PR, query netlify for the latest deployed target URL (is that possible?)
2. Generate versions.json
3. Commit and open PR with updated versions.json (will include currently deployed versions, but not the soon to be deployed next version)
4. The one-behind versions.json will get automatically deployed with the regular workflow

++ Can use existing workflows, no duplicate deploys or commits, is possible because link to latest deploy does not change
++ Can preview the versions.json before committing to main
+- versions.json is always one deploy behind
+- not sure if possible

Example script to update `versions.json`:
```js
const fs = require('fs');
const path = require('path');
const packageJSON = require('../package.json');
const versionsFile = require('../src/assets/versions.json');
const minimist = require('minimist');
const args = minimist(process.argv.slice(2));

const interestingPackages = [
  "@angular/core",
  "@swisspost/design-system-styles",
  "@swisspost/design-system-intranet-header",
  "bootstrap",
  "@ng-bootstrap/ng-bootstrap",
];

const getInterestingPackages = () => {
  return Object.entries({ ...packageJSON.dependencies, ...packageJSON.devDependencies}).reduce((acc, [key, value]) => {
    if (interestingPackages.includes(key)) {
      acc[key] = value;
    }

    return acc;
  }, {})
}

const writeVersions = () => {
  console.log(args);
  const newVersions = [{
    title: `v${packageJSON.version}`,
[...]

PR #1241, opened 2023-03-10
feat(docs): add version switch

gfellerph:
This is the most simple implementation of a version switch and will need some automation. Question is how. More details can be found in #1239.


swisspost-devs:

packages/demo/src/app/app.component.html
```diff
@@ -1,135 +1,441 @@
 <div class="layout-container header-big">
-  <sp-intranet-header siteTitle="Swiss Post Design System" languages="en" logoUrl="/">
+  <ng-template #versionDropdown ngbDropdown>
+      <label class="visually-hidden" for="versionDropdownButton" id="versionDropdown">Version Selection for older documentation versions</label>
+      <button class="btn btn-link dropdown-button" id="versionDropdownButton" ngbDropdownToggle>
+        v{{ version }}
```
alizedebray, line 5: Showing the demo version here is confusing because it does not match any of the versions actually installed in the projects. I'd rather show the version of the `@post/design-system-styles` package since, for now, this is the base design system package. What do you think?

packages/demo/src/app/app.component.html
```diff
@@ -1,135 +1,441 @@
...
+      <div ngbDropdownMenu aria-labelledby="versionDropdown">
+        <a
+          *ngFor="let v of versions"
+          [href]="v.version === version ? 'https://design-system.post.ch' : v.url"
+          [class.active]="version === v.version"
+          ngbDropdownItem
+        >
+          <span class="bold">{{ version === v.version ? v.title + ' (latest)' : v.title }}</span>
+          <p class="small mt-2">
+            Angular {{v.dependencies['@angular/core'].split('.')[0]}}
```
alizedebray, line 17: It's good to show the versions of the dependencies! 

However the number of versions being destined to grow, we would benefit from a more condensed display. We could for example show icons: 
image

packages/demo/src/assets/versions.json
```diff
@@ -0,0 +1,28 @@
...
+    "title": "v5.3.1",
+    "version": "5.3.1",
+    "description": "Pattern documentation, code snippets and implementation guidelines for the Design System Styles.",
+    "url": "https://6401b58ebf0f4e298ee2dbe7--swisspost-web-frontend.netlify.app/",
+    "dependencies": {
+      "@angular/core": "14.2.12",
+      "@ng-bootstrap/ng-bootstrap": "12.1.2",
+      "@swisspost/design-system-intranet-header": "3.1.1",
+      "@swisspost/design-system-styles": "5.4.1",
+      "bootstrap": "5.1.3"
```
alizedebray, line 12: The versions are [miss matched](https://ng-bootstrap.github.io/#/getting-started#dependencies). Do you think we could do something about that?

packages/demo/src/app/app.component.html
gfellerph: This would work for now because I wrote the versions.json manually. As soon as we're looking to automate this, we'd have to follow the version of the demo/documentation package.json because deployment is following this. What we could do, is declare these two packages as [fixed](https://github.com/changesets/changesets/blob/main/docs/fixed-packages.md) or [linked](https://github.com/changesets/changesets/blob/main/docs/linked-packages.md).

packages/demo/src/assets/versions.json
gfellerph: Yes, we just re-write the versions.json.

packages/demo/src/app/app.component.html
gfellerph: I simplified the displayed versions (just showing major updates). We need to discuss how we automate this exactly as it's not easy, but I think I have a better understanding of how we could do it in the future.

packages/demo/src/app/app.component.html
gfellerph: Done.

packages/demo/src/styles.scss
```diff
@@ -153,3 +153,10 @@ code {
     }
   }
 }
+
+// Take this, you inflexible header custom content slots
+sp-intranet-header .settings:has(.version-switch) {
+  > div > .p-3 {
+    padding: 0 !important;
```
alizedebray, line 160:
```suggestion
    .version-switch { margin: -1rem; }
```
`:has()` is not yet supported by Firefox

alizedebray (changes requested):

packages/demo/src/assets/versions.json
alizedebray: ng-bootstrap 12 is theoretically not compatible with Angular > 14.1
Support was added as part of [13.0.0](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CHANGELOG.md#1300-2022-07-28) (the same version that removed ngb buttons)

We should probably downgrade to Angular 14.0.0 or do you think that's not a problem?

packages/demo/src/styles.scss
gfellerph: Done, removed the global styles.

packages/demo/src/assets/versions.json
gfellerph: As discussed, we'll keep it like this because our tests with this version combo were successful and because a newer Version with compatible Angular/ng-bootstrap versions is available.

gfellerph:
The URL to the live system will be patched by #1260, I left it on the current deploy here for testing reasons (version switch would not be displayed because there is no versions.json on the current production deploy).

alizedebray (approved):

PR #1251, opened 2023-03-13
fix: add deprecation notice for sizing variables

gfellerph:


swisspost-devs:

alizedebray (approved):

PR #1253, opened 2023-03-13
chore(styles): refactor DOM migration schematics

alizedebray:
- removed the RegEx util which is no longer used
- removed the `DomMigration` class which was an unnecessary middleman as it contained no logic
- removed the "I" in the interface name `IDomUpdate` to match [Angular naming conventions](https://github.com/angular/angular/blob/main/docs/NAMING.md)


swisspost-devs:

gfellerph (approved):

PR #1276, opened 2023-03-21
fix(docs): use btn-md instead of empty string

gfellerph:
Empty key values break storybook control rendering and lead to navigation hiccups. Using a default here is better because it's the correct class. The field is initially empty, so the default state without explicit class is still documented, only when size: md is explicitly chosen, the class will show up in the code block.


swisspost-devs:

alizedebray (approved):

PR #1278, opened 2023-03-21
fix(header): shipment info in search suggestions not reset

gfellerph:
Fixed a bug where a shipment information in the search suggestions was not reset after the search query changed.


swisspost-devs:

alizedebray (approved):

PR #1279, opened 2023-03-21
test: move snapshot tests to docs

gfellerph:
Percy works better with just one snapshot run per repo. When using per package snapshot tests, multiple percy projects would need to be set up and configured.


swisspost-devs:

PR #1282, opened 2023-03-21
chore(github): update the fetch icons workflow

alizedebray:
The PR created by the "Fetch icons" workflow now looks like [this](https://github.com/swisspost/design-system/pull/1281): the PR body is more compact and the changset more complete.


gfellerph (approved):
Sweet!

PR #1288, opened 2023-03-22
fix(demo): add auto as default placement setting

gfellerph:
Updated the documentation for tooltips regarding the "auto" placement option which can be used to place the tooltip where it fits if there is not enough space for the initial placement.


swisspost-devs:

alizedebray (approved):

PR #1289, opened 2023-03-22
fix: move non-production dependencies to devDependencies

gfellerph:
Fixed package dependencies. Moved dependencies that are not being used in production to devDependencies in order to simplify and speed up the package install process.


swisspost-devs:

alizedebray (approved):

Issue #1294, opened 2023-03-23
Update heading size


gfellerph:
Headings in current implementation do not match the design. 

Figma: https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/Components?node-id=5962%3A32961&t=vhqkjNAQJRJLeOQe-1
![image](https://user-images.githubusercontent.com/1659006/227138337-3448c214-87a8-4455-9f3b-1d7e7ebfdf00.png)

Demo: https://design-system.post.ch/#/bootstrap-samples/typography
![image](https://user-images.githubusercontent.com/1659006/227138381-f31b7dfc-65cc-4fa6-ae50-6d7926753a55.png)

```[tasklist]
### Tasks
- [x] Update storybook foundation/typography documentation
```

Issue #1194, opened 2023-02-27
Update line heights


gfellerph:
According to the [WCAG 2.1 on text spacing](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html), line-heights should be 1.5 for body text.

Design specifications: https://www.figma.com/file/ojCcgC5Zd12eUSzq6V5m24/Foundations?node-id=3%3A950&t=uemycRw9g7CNz7E7-0

```[tasklist]
### Tasks
- [x] Default line-height should be 1.5
- [x] Header line-height should be 1.2
- [x] Deprecate old line-height variables
- [x] Update storybook foundation/typography documentation
```

PR #1298, opened 2023-03-23
feat: line height and font size update

gfellerph:
This PR fixes two tickets:

- https://github.com/swisspost/design-system/issues/1294
- https://github.com/swisspost/design-system/issues/1194

Todo
- [x] Update storybook foundation/typography documentation


swisspost-devs:

packages/styles/src/variables/_type.scss
```diff
@@ -135,6 +167,7 @@ $font-curve-big: (
   'xxl': $font-size-small-huge,
 );
 
+// Deprecated
```
oliverschuerch, line 170: Why would we deprecate font-curves? Or do you want to create new ones with an updated naming?

oliverschuerch (approved):

PR #1317, opened 2023-03-29
fix(stepper): misaligned text of last item on mobile

gfellerph:


packages/demo/src/app/post-sample/components/stepper/stepper-demo/stepper-demo.component.html
```diff
@@ -13,11 +13,11 @@ <h2 class="visually-hidden">Order progress, step 3 of 4</h2>
       [attr.aria-current]="isCurrent(step) ? 'step' : undefined"
       class="stepper-item"
     >
-      <a *ngIf="i < currentIndex" [routerLink]="getPathTo(step)" class="stepper-link">
+      <a *ngIf="i < currentIndex" class="stepper-link" (click)="currentIndex = i">
```
alizedebray, line 16: This might be misleading for some users, maybe it can be done from the demo-page so that it does not show in the snippet. What do you think?

alizedebray (approved):

oliverschuerch (approved):

PR #1318, opened 2023-03-30
test: add alert snapshot test

gfellerph:


swisspost-devs:

packages/documentation/src/stories/components/alert/alert-snapshot.stories.mdx
```diff
@@ -0,0 +1,59 @@
...
+        noIcon: [true, false],
+        icon: ['null', '1001'],
+        dismissible: [true, false],
+        action: [true, false],
+      })
+        .filter(args => !(args.noIcon && args.icon === 'null'))
+        .map(args => {
+          console.log(args);
+          return { ...args, show: true };
+        })
```
oliverschuerch, line 22:
```suggestion
        .map(args => ({ ...args, show: true }))
```

oliverschuerch (changes requested):

packages/documentation/cypress/snapshots/styles/alert.snapshot.ts
```diff
@@ -0,0 +1,6 @@
+describe('Alert', () => {
+  it('default', () => {
+    cy.visit('/iframe.html?id=hidden-alert-snapshot-test--page');
+    cy.percySnapshot('Alerts', { widths: [320, 600, 1440] });
```
alizedebray, line 4: As this is going to be used for all snapshot tests, creating a [custom Cypress command](https://docs.cypress.io/api/cypress-api/custom-commands) might be worth it, don't you think?

It could take two parameters: the component name and a list of breakpoints, then the URL and screen width would be set from it.

gfellerph: Screen widths could be different from component to component. Some of them just need to be tested on one screen width, saving a lot of snapshots. It's just two lines and abstracting these into a function that needs all the params anyways does not provide enough helpful abstraction to warrant a helper in my opinion.

alizedebray (approved):

PR #1320, opened 2023-03-31
test: add snapshot tests for headings

gfellerph:


swisspost-devs:

oliverschuerch (commented):
It could be helpful to prefix every heading with it's corresponding html element:
"h1 - Lorem ipsum dolor sit"

What do you think?

oliverschuerch (approved):

PR #1323, opened 2023-03-31
fix(form-file): file-selector-button visualization in firefox

oliverschuerch:


swisspost-devs:

alizedebray (changes requested):
Vertical alignment is not the same as the basic controls:
image

alizedebray (approved):

PR #1324, opened 2023-03-31
fix(styles): fix collapsible component styles

alizedebray:
I took advantage of the accordion display issue fix to remove the Bootstrap styles.


swisspost-devs:

packages/styles/src/components/accordion.scss
```diff
@@ -1,94 +1,124 @@
...
+  --post-accordion-button-padding: #{accordion.$accordion-button-padding};
+  --post-accordion-icon-size: #{accordion.$accordion-icon-size};
+  --post-accordion-button-cursor: #{accordion.$accordion-button-cursor};
+  --post-accordion-button-disabled-opacity: #{accordion.$accordion-button-disabled-opacity};
+  --post-accordion-button-focus-box-shadow: #{accordion.$accordion-button-focus-box-shadow};
+  --post-accordion-icon-transition: #{accordion.$accordion-icon-transition};
+  --post-accordion-icon-transform: #{accordion.$accordion-icon-transform};
+  --post-accordion-body-padding: #{accordion.$accordion-body-padding};
+  --post-accordion-heading-margin: #{accordion.$accordion-heading-margin};
+  --post-accordion-heading-font-weight: #{accordion.$accordion-heading-font-weight};
```
oliverschuerch, line 25: Did you define CSS variables because you want to give the users easy access to all of those styles properties?
- If the answer is "no": Why don't you use the SCSS variables directly instead of defining some CSS variables out of it?
- If the answer is "yes": Why don't you define the variables on the host?

packages/styles/src/components/accordion.scss
```diff
@@ -1,94 +1,124 @@
...
+    opacity: var(--post-accordion-button-disabled-opacity);
+  }
+
+  &:focus {
+    outline: none;
+    box-shadow: var(--post-accordion-button-focus-box-shadow);
+  }
+
+  &::after {
+    @include icon-mixins.pi(accordion.$accordion-icon-name);
```
oliverschuerch, line 61: It would be nice if you could use the new icon system :)

oliverschuerch (changes requested):

packages/styles/src/components/accordion.scss
alizedebray: The answer is "yes", I appreciate the flexibility offered by CSS variables (but I'm open to discussing it further!).

If by "host" you mean the `.accordion`, I didn't declare the variables there because there is no `.accordion` in a collapsible: the `.accordion` is just a wrapper for multiple `.accordion-item`, i.e. several collapsible.

packages/styles/src/components/accordion.scss
alizedebray: Is there a way to use the new icon system in a Sass file?
The new system works with HTML markups, doesn't it?

alizedebray: I opened [this issue](https://github.com/swisspost/design-system/issues/1361) to find a way to use the new icon system within the styles package.

packages/styles/src/components/accordion.scss
alizedebray: Discussion moved to #1380

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
imagoiq, line 1: It came several time in PRs, I think we should leave it as it. I guess the tool is right

packages/styles/src/components/accordion.scss
```diff
@@ -1,94 +1,109 @@
...
-      background-image: $accordion-hcm-icon;
-    }
-
-    &:not(.collapsed)::after {
-      background-image: $accordion-hcm-icon;
+.accordion-body {
+  padding: accordion.$accordion-body-padding;
+  border-block-start: accordion.$accordion-border-width solid accordion.$accordion-border-color;
+
+  // TODO: find a way to style headings within web component
```
imagoiq, line 89: Is it tracked somewhere? I mean, those TODO are too often forgotten. Or perhaps you intended to work on it after this PR? Or is there already an established process for those?

imagoiq (approved):

packages/styles/src/components/accordion.scss
gfellerph: The `:part()` could be good here and there should be a ticket if it's not in the scope of this PR.

alizedebray: I fixed the 3 TODOs that were in this file: 
- "find a way to style icon within web component on dark background": I reintroduced the mask-image for the accordion button so that the icon color is always inherited from the button color
- "find a way to style WHCM within web component": this was not a problem, media queries are taken into account by the shadow DOM
- "find a way to style headings within web component": we decided not to style specifi slotted element from web components

oliverschuerch (approved):

gfellerph (changes requested):
Just noticed some issues, will update

packages/styles/src/components/accordion.scss
```diff
@@ -1,94 +1,103 @@
...
+}
 
-  &:last-of-type {
-    border-bottom: 0;
-  }
+.accordion-header {
+  @include type-mixins.font-curve(accordion.$accordion-header-font-curve);
+  font-weight: accordion.$accordion-header-font-weight;
+  line-height: accordion.$accordion-header-line-height;
+  margin: 0;
```
gfellerph, line 19: CSS is awesome

image

For the text-truncate class to work, the span has to be `display: inline-block` (leaves that invisible margin-bottom or `display: block`. I think the use-case here, because it's documented, warrants a dedicated workaround where we specifically style set the span with class text-truncate to display: block. What do you think?

alizedebray: I added a `display: flex;` on the accordion button so that we don't have to style unknown children

gfellerph: I don't think that worked, looks like this for me:

image

Now the fancy title is wrapping and the overflow is still there.

alizedebray: My bad, the truncated title work in Firefox but I did not see the "Fancy title"... 
I just updated the PR to give a `display: block` to `.text-truncate` elements.

gfellerph (approved):
Thanks for the update @alizedebray! Nice to see one less bootstrap-dependent component 💪

> As an aside / personal opinion: exclusive accordions should die in a hellish fire; they deserve nothing more after all the frustration they’ve caused me and countless others—please let me see content when and how I want it…
- https://github.com/openui/open-ui/issues/725#issuecomment-1664173112

packages/styles/src/components/accordion.scss
imagoiq: FYI For those headings, I didn't know, but it seems that you can override style by using !important: https://www.matuzo.at/blog/2023/pros-and-cons-of-shadow-dom/#con-styling-slotted-content

PR #1350, opened 2023-04-19
352 clean up alerttoast styles

alizedebray:


swisspost-devs:

packages/styles/src/variables/components/_alert.scss
```diff
@@ -1,51 +1,37 @@
...
-$alert-font-size: type.$font-size-bigger-regular; // Design System only
-$alert-font-size-small: type.$font-size-small; // Design System only
-
-$alert-toast-width: 25rem; // Design System only
-$alert-toast-width-small-breakpoint: 21rem; // Design System only
-
-$alert-toast-box-shadow-hover: 0 0 1rem 0 rgba(var(--post-contrast-color-rgb), 0.4); // Design System only
-$alert-close-hover-opacity: 0.4; // Design System only
+  'info' color.$info 2106,
+  'cookies' color.$gray-background 2201
```
oliverschuerch, line 36: I would like to have a more generic name for this new color variable.
Something like `gray` or `light`...

Because it can be used for other topics in the future and then it will feel wrong to use the given `alert-cookies` class.

packages/styles/src/variables/components/_toasts.scss
```diff
@@ -1,18 +1,19 @@
...
+$toast-margin: spacing.$spacer;
+$toast-padding-x: spacing.$size-regular;
+$toast-padding-y: spacing.$size-small-regular;
+$toast-font-size: type.$font-size-tiny;
+$toast-heading-font-size: type.$font-size-small;
+$toast-hr-margin-top: spacing.$size-micro;
+$toast-hr-margin-bottom: spacing.$size-small-regular;
+$toast-icon-size: spacing.$size-big;
+
+$toast-variants: alert.$alert-variants;
```
oliverschuerch, line 19: This means, we should move the deprication message for the toast/notification variants up on the demo app.
Because we can not deprecate colors for the toast but not for the alert.

image

The deprecation warning should be included before the "Toast / Small notification bar" heading.

oliverschuerch (changes requested):

packages/styles/src/components/alert.scss
```diff
@@ -5,206 +5,128 @@
...
-      align-self: flex-end;
-    }
-  }
-
-  @include media-breakpoint-down(sm) {
-    align-items: center;
-  }
+  z-index: commons.$zindex-fixed;
 
   @include media-breakpoint-up(xxl) {
```
alizedebray, line 97: @oliverschuerch 
With my last two commits, I removed CSS variables and used mixins instead.
It works well for the majority of styles, except here where I had to replace many properties with calculations.
What do you think?

oliverschuerch (commented):
The default story in the storybook alert component should have the new `alert-gray` variant as well.

packages/styles/src/components/alert.scss
gfellerph, line 1: Not all alerts/notifications are visible as such in high contrast mode:

image

A solution could be to draw a border around the items in HCM. We can fix the icons in the icons PR.

gfellerph (changes requested):

packages/styles/src/components/alert.scss
oliverschuerch, line 1: - The dismiss button is vertically aligned center when showing the alert with action-buttons but it's aligned top in all the other alerts.
- The Icon alignes to the far left, on big screens (fixed), while the dismiss button stays within the container (not sure if this is wanted).

image

.changeset/clever-kangaroos-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Updated close button styles.
```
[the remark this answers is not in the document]
alizedebray: The close button changed for all components: alerts, toasts, badges, modals.

packages/styles/src/components/alert.scss
alizedebray: I just updated the HCM styles. 
There is still an issue with the custom icons:

alizedebray: - The vertical alignment of the dismiss button follows  [design guidelines](https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?type=design&node-id=156-232&mode=design&t=EEzivsyJn73xohnN-0). I suggest we leave it as is for this PR and discuss this point with Ruven or Dimitri later.
image
image

- I cannot reproduce your second point. Have you tested it recently?
image

oliverschuerch: I don't know why, but it's gone now! 😀

oliverschuerch (approved):

packages/styles/src/components/alert.scss
gfellerph: This would be fixed by the icons PR, no? Because then it will be possible to use `color` to color the icons, or am I wrong? If this is the case, we can merge already.

.changeset/clever-kangaroos-switch.md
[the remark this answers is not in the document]
gfellerph: Excellent changeset description ^^

gfellerph (approved):

packages/styles/src/components/alert.scss
alizedebray: Yes, it will work properly when using the <post-icon/> to define a custom alert icons.

PR #1368, opened 2023-04-27
feat(internet-header): add aria attributes to mobile menu toggle

alizedebray:
Added attributes recommended here: https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/

### DIFF
--- a/packages/internet-header/src/components/post-main-navigation/post-main-navigation.tsx
+++ b/packages/internet-header/src/components/post-main-navigation/post-main-navigation.tsx
@@ -250,6 +250,7 @@ export class PostMainNavigation implements HasDropdown, IsFocusable {
         <nav
           id="post-internet-header-main-navigation"
           class={{ 'main-navigation': true, 'open': this.mobileMenuOpen }}
+          role="menu"
           ref={el => (this.flyoutElement = el)}
         >
           <h1 class="visually-hidden">{headerConfig.translations.navMainAriaLabel}</h1>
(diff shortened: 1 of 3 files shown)


swisspost-devs:

packages/internet-header/src/components/post-main-navigation/post-main-navigation.tsx
```diff
@@ -250,6 +250,7 @@ export class PostMainNavigation implements HasDropdown, IsFocusable {
         <nav
           id="post-internet-header-main-navigation"
           class={{ 'main-navigation': true, 'open': this.mobileMenuOpen }}
+          role="menu"
```
oliverschuerch, line 253: Do you think the main navigation should have the role menu also on desktop or only on mobile devices?

oliverschuerch (commented):
onFocus `button.menu-button` nvda reads the following:
IS: "Menü Menü Schaltfläche reduziert Untermenü."
SHOULD: "Hauptnavigation Menü Schaltfläche reduziert Untermenü."

packages/internet-header/src/components/post-main-navigation/post-main-navigation.tsx
alizedebray: From what I understood of the [ARIA: menu role documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role) it seems to be appropriate for the desktop version as well. 

> A menu can be a permanently visible list of controls or a widget that can be made to open and close.

PR #1400, opened 2023-05-11
fix(internet-header): fix the scroll error in the 2nd-level navigation overlay on mobile devices

oliverschuerch:
The same behavior can be done with the CSS property `overscroll-behavior: contain;`. But it only works on scrollable containers, which is the reason why I keep the existing logic for the `post-internet-breadcrumbs` and `post-search components`.

image


swisspost-devs:

alizedebray (approved):

PR #1402, opened 2023-05-12
fix(internet-header): add aria-label on skiplinks ul element

oliverschuerch:


swisspost-devs:

alizedebray:
I checked the current Post Portal and the title "Navigieren auf post.ch" is used for the whole header: 
image

While in you solution it is used for skip links only:
image

alizedebray (changes requested):
I checked the current Post Portal and the title "Navigieren auf post.ch" is used for the whole header: 
image

While in you solution it is used for skip links only:
image

alizedebray (approved):

PR #1413, opened 2023-05-24
chore: update editorconfig of styles package

oliverschuerch:
Removed all the redundant config properties.

### DIFF
--- a/packages/styles/.editorconfig
+++ b/packages/styles/.editorconfig
@@ -1,4 +1,0 @@
-# http://editorconfig.org
-
-[*.md]
-max_line_length = 0
--- a/packages/styles/.editorconfig
+++ b/packages/styles/.editorconfig
@@ -1,15 +1,4 @@
 # http://editorconfig.org
-root = true
-
-[*]
-charset = utf-8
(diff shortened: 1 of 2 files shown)


swisspost-devs:

packages/styles/.editorconfig
```diff
@@ -1,15 +1,4 @@
...
-charset = utf-8
-indent_style = space
-indent_size = 2
-insert_final_newline = true
-trim_trailing_whitespace = true
-end_of_line = lf
-
 
 [*.md]
 max_line_length = 0
```
gfellerph, line 4: Do we even need this? If not, the file can be deleted entirely.

oliverschuerch: I did not check if the properties are really needed, but only compared the editorconfig files.
Actually this property doesn't seem to do anything in our code.
But to not break anything, I moved the property to the main editorconfig and set the value to "off" as described in the documentation:
https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#max_line_length

gfellerph (approved):

PR #1416, opened 2023-05-25
fix(internet-header): abort parallel header search requests per endpoint

oliverschuerch:


swisspost-devs:

.changeset/silent-terms-deny.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/internet-header': patch
+---
+
+Added abort controller for all existing post-search search endpoints, to avoid running parallel calls for the same endpoint and therefore get wrong answers shown in the search suggestion list.
```
gfellerph, line 5:
```suggestion
Added abort controller for all existing post-search search endpoints, to avoid running parallel calls for the same endpoint and therefore get outdated suggestions in the search suggestion list.
```

packages/internet-header/src/services/search/coveo.service.ts
```diff
@@ -15,25 +17,36 @@ export const getCoveoSuggestions = async (query: string): Promise<CoveoCompletio
...
+          coveoJSON.completions.map(completion => ({
+            ...completion,
+            redirectUrl: getSearchPageUrl(completion.expression),
+          })),
+        );
+      })
+      .catch(error => {
+        if (error.name === 'AbortError') {
+          console.warn('Coveo suggestions request aborted.');
+        } else {
```
gfellerph, line 44: I'd not log abort errors as they are expected to happen. This will just spam the console.

```suggestion
        if (error.name !== 'AbortError') {
```

packages/internet-header/src/services/search/parcel.service.ts
```diff
@@ -43,13 +45,26 @@ export const getParcelInfo = async (
...
+    return new Promise((resolve, reject) => {
+      fetch(url, { signal: parcelController.signal })
+        .then(response => response.json())
+        .then((trackAndTraceResult: TrackAndTraceInfo) => {
+          resolve({ ...trackAndTraceResult, ok: trackAndTraceResult.ok });
+        })
+        .catch(error => {
+          if (error.name === 'AbortError') {
+            console.warn('Track and trace request aborted.');
+          } else {
```
gfellerph, line 62: Same same.
```suggestion
          if (error.name !== 'AbortError') {
```

packages/internet-header/src/services/search/coveo.service.ts
```diff
@@ -15,25 +17,36 @@ export const getCoveoSuggestions = async (query: string): Promise<CoveoCompletio
...
+          console.error(
+            `Fetching coveo suggestions failed, ${error}\nDid you add "*.coveo.com" to your connect-src content security policy and tried turning off your adblocker?`,
+          );
+          reject(error);
+        }
```
gfellerph, line 49: Would it be a good idea to reject anyway, no matter what error happened?

```suggestion
        }
        
        reject(error);
```

packages/internet-header/src/services/search/parcel.service.ts
```diff
@@ -43,13 +45,26 @@ export const getParcelInfo = async (
...
+            console.error(`Could not check track and trace API due to error: ${error.message}`);
+            reject(error);
+          }
```
gfellerph, line 65:
```suggestion
          }
          reject(error);
```

packages/internet-header/src/services/search/places.service.ts
```diff
@@ -20,21 +22,29 @@ export const queryPlaces = async (query: string): Promise<GeocodeLocation[]> =>
...
+    fetch(url, { signal: placesController.signal })
+      .then(response => response.json())
+      .then((geocodeJSON: GeocodeResponse) => {
+        if (!geocodeJSON.ok) throw new Error(geocodeJSON.info);
+        resolve(geocodeJSON.locations);
+      })
+      .catch(error => {
+        if (error.name === 'AbortError') {
+          console.warn('Geocoder request aborted.');
+        } else {
```
gfellerph, line 40:
```suggestion
        if (error.name !== 'AbortError') {
```

packages/internet-header/src/services/search/places.service.ts
```diff
@@ -20,21 +22,29 @@ export const queryPlaces = async (query: string): Promise<GeocodeLocation[]> =>
...
+          console.error(
+            `Fetching places failed, ${error}\nDid you add "places.post.ch" to your connect-src content security policy?`,
+          );
+          reject(error);
+        }
```
gfellerph, line 45:
```suggestion
        }
        reject(error);
```

gfellerph (changes requested):

packages/internet-header/src/services/search/coveo.service.ts
oliverschuerch: Failing to resolve or reject a promise just fails to ever change the state from "pending" to anything else. This doesn't cause any fundamental problem in Javascript because a promise is just a regular Javascript object. The promise will still get garbage collected (even if still pending) if no code keeps a reference to the promise.

The real consequence here is what does that mean to the consumer of the promise if its state is never changed? Any `.then()` or `.catch()` listeners for resolve or reject transitions will never get called. Most code that uses promises expects them to resolve or reject at some point in the future. If they don't, then that code generally never gets to finish its work.

Or if code is using `await` on the promise instead of `.then()`, then that function will just remain suspended forever on that await. The rest of the function will be dead code and will never execute.

PR #1418, opened 2023-05-26
feat(styles): use @swisspost/design-system-icons

alizedebray:
I added a Gulp task which generates a SASS map from all icons available in the @swisspost/design-system-icons package.

Some icons do not exist in the icon package: 1041, 2062-f, 8003, and all 9900+. We still have to decide what to do with this.


swisspost-devs:

gfellerph (changes requested):
Generally, we're trying to get rid of icons defined as CSS backgrounds. One big factor is CSS output size:

image

This PR would double the gzipped size. The plan is however to migrate everybody to the new icon web component. This would allow us to remove the deprecated CSS icons, reducing gzipped CSS size to a third instead. Let's discuss this change and it's implications in some more detail.

gfellerph (approved):
Amazing work, thank you so much 💪

PR #1434, opened 2023-05-31
chore: exclude bootstrap from minor patch glob

gfellerph:
Bootstrap keeps introducing breaking changes to their minor versions, this time a bunch of additional variables that are not defined when trying to build.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -15,7 +15,7 @@
     },
     {
       "matchPackagePatterns": ["*"],
-      "excludePackageNames": ["typescript"],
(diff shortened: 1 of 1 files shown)


oliverschuerch:
Are you 100% sure about that?
Because last time it was not really a breaking change, but only for custom bootstrap css builds:
https://getbootstrap.com/docs/5.3/migration/#new-_mapsscss

gfellerph:
Not needed anymore, fixed the bug with bootstrap in the minor/patch PR.

PR #1438, opened 2023-05-31
feat(documentation): add background utilities stories

oliverschuerch:


swisspost-devs:

packages/documentation/src/stories/utilities/background/background.stories.mdx
```diff
@@ -15,7 +17,78 @@ import { Meta } from '@storybook/addon-docs';
...
+    </div>
+  </Story>
+</Canvas>
+
+## Opacity
+
+Use the css variable `--post-bg-opacity` on a background container, to modify the opacity of its `background-color` property.
+
+<p className="text-secondary fst-italic">
+  Node: We do not recommend low opacity, because you have to ensure a minimal contrast ratio of
```
alizedebray, line 73:
```suggestion
  Note: We do not recommend low opacity because you have to ensure a minimal contrast ratio of
```

packages/documentation/src/stories/utilities/background/background.stories.mdx
```diff
@@ -15,7 +17,78 @@ import { Meta } from '@storybook/addon-docs';
...
+  4.5:1, between background and foreground. Read more about this topic on or{' '}
```
alizedebray, line 74:
```suggestion
  4.5:1 between background and foreground. Read more about this topic on our{' '}
```

packages/documentation/src/stories/utilities/background/background.stories.mdx
```diff
@@ -15,7 +17,78 @@ import { Meta } from '@storybook/addon-docs';
...
+### Available backgrounds
+
+<BackgroundContainers></BackgroundContainers>
+
+## Nested
+
+It's possible to nest background containers.
+
+<p className="text-secondary fst-italic">
+  Note: Every container defines it's own text color and inverted text color css variables, to always
```
alizedebray, line 50:
```suggestion
  Note: Every container defines its own text color and inverted text color css variables to always
```

packages/styles/src/components/utilities.scss
```diff
@@ -67,9 +67,6 @@ span.spacer {
 @each $name, $color in color.$background-colors {
   .bg-#{$name} {
     --post-bg-rgb: #{color-fn.rgb-values($color)};
-    $shadow-color: sass-color.adjust($color, $lightness: -15%);
```
alizedebray, line 70: This seems like a breaking change, although it's very likely that no one is using these variables...

alizedebray (changes requested):

gfellerph (commented):
@oliverschuerch, the foundations/color page and this have some overlap. Can we reduce info in one or the other?

packages/styles/src/components/utilities.scss
oliverschuerch: I discussed it with @gfellerph.
This was mistakenly added in the elevation PR. And needs to be removed again.

oliverschuerch:
> @oliverschuerch, the foundations/color page and this have some overlap. Can we reduce info in one or the other?

I'm not sure if we should remove anything from one of the documentation pages, because available colors must not necessarly match the available background colors.
I would only add the missing primary, secondary, tertiary colors in the color documantation page and maybe change the code-snippets from `background-color` to `color`.
@gfellerph what do you think?

gfellerph:
@oliverschuerch, I think we can go forward with that and update it once there is clearer guidance.

.changeset/spotty-papayas-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Removed unused background utility shadow variables.
```
gfellerph, line 5:
```suggestion
Removed unused variable `$shadow-color`. It was added by mistake.
```

gfellerph (approved):

gfellerph (approved):

gfellerph (approved):

PR #1461, opened 2023-06-05
fix(styles): Safari display issue with floating labels

gfellerph:
A rendering issue in Safari clipped text input in a floating label input field where the floating label intersected with the current value of the input. Even having a transparent background, the input was invisible below the floating label. Setting the height of the floating label correctly fixes the rendering issue because now the label does no longer cover the value.


swisspost-devs:

alizedebray (approved):

PR #1462, opened 2023-06-05
fix(header): focus behavior on mouseleave

gfellerph:
Fixed the focus behavior when navigating the flyout with a mouse. Hovering and leaving navigation items that trigger a flyout will no longer focus the main navigation entry after the flyout has been closed.


swisspost-devs:

alizedebray (approved):

PR #1474, opened 2023-06-07
feat(docs): use the post-icon component in the documentation

alizedebray:


gfellerph (changes requested):
Thanks for this change!

gfellerph (approved):

PR #1475, opened 2023-06-07
feat(intranet-header): use post-icon component in the header

alizedebray:


swisspost-devs:

packages/components-angular/projects/intranet-header-showcase/src/main.ts
```diff
@@ -3,6 +3,9 @@ import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 
 import { AppModule } from './app/app.module';
 import { environment } from './environments/environment';
+import { defineCustomElements } from '@swisspost/design-system-components/loader';
+
+defineCustomElements();
```
oliverschuerch, line 8: Do we really want to define all the custom Elements or only the post-icon component?

packages/components-angular/projects/intranet-header/src/lib/swisspost-intranet-header.component.ts
```diff
@@ -319,10 +340,22 @@ export class SwissPostIntranetHeaderComponent implements OnInit, OnChanges, Afte
...
+  public get showNavigationLabel() {
+    return this.localization['showNavigationLabel'][this.lang.toLowerCase()];
+  }
+
+  public get hideNavigationLabel() {
+    return this.localization['hideNavigationLabel'][this.lang.toLowerCase()];
+  }
+
   public isLanguageActive(lang: string) {
     return this.lang.toLowerCase() === lang.toLowerCase();
```
oliverschuerch, line 360: It could be a good idea, to create a `this.langLow = this.lang.toLowerCase()` variable in the constructor, to reduce function calls every single time we need the lang lowercased.
What do you think?

oliverschuerch (changes requested):

packages/components-angular/projects/intranet-header-showcase/src/main.ts
gfellerph: Components are lazy loaded by default so there's no harm done here. Also, this is just the showcase for local testing.

oliverschuerch (approved):

packages/components-angular/projects/intranet-header/src/lib/swisspost-intranet-header.component.ts
oliverschuerch: I don't see the changes I requested, but anyway...
As long as everything is working I'm fine with it.

oliverschuerch (approved):

oliverschuerch:
@alizedebray Is there a reason, why this PR is still not merged?
Should I remove my approval?
Thanks for clearification.

alizedebray:
Outdated.

PR #1495, opened 2023-06-12
fix(stlyes): alert action buttons for alerts

gfellerph:
Error only appeared on the storybook visual regression testing page.


swisspost-devs:

PR #1497, opened 2023-06-12
fix(demo): add href to make link focusable

gfellerph:
Stepper links make no sense without being a link back to this step. A href or [routerLink] is needed anyways. Regression was added because of https://github.com/swisspost/design-system/pull/1317/files#r1153405992. A href attribute might be less missleading than an internal [routerLink].


swisspost-devs:

PR #1487, opened 2023-06-11
chore(icons): :point_up: update icons

swisspost-devs:
# Design system icons are now up to date!
## Added icons
2524, 2528, 2529, 2530, 2531, 2532, 2533, 2534 and 2535

PR #1500, opened 2023-06-12
feat(styles): update svg-icon-map

alizedebray:
The svg-icon-map was merged after the icon PR #1487, so the new icons were missing. It should update automatically from now on.


swisspost-devs:

alizedebray:
Solved with #1499.

PR #1505, opened 2023-06-12
feat(styles): update tooltips position

alizedebray:


swisspost-devs:

PR #1512, opened 2023-06-14
fix: adjust default legend styles sizes

geekrumper:
Using `$font-size-18` for default legend styles


swisspost-devs:

PR #1513, opened 2023-06-14
fix: add italic defintion to fonts.scss

yannikstuker:


swisspost-devs:

oliverschuerch (commented):
Why did you removed the `options.$font-base-path` variable?

We should consider adding the `font-base-path` again to all `@font-face` definitions or remove the option variable itself (would be a breaking change).

yannikstuker:
It was not intentional to delete the variable, I readded them

gfellerph (approved):

gfellerph:
@yannikstuker thank you for contributing!

PR #1516, opened 2023-06-14
feat(textarea): update resizer to match figma for webkit

RedlineTriad:
fire and forget

### DIFF
--- a/packages/styles/src/components/floating-label.scss
+++ b/packages/styles/src/components/floating-label.scss
@@ -174,7 +174,7 @@
         }
       }
     }
-    
+
     :-webkit-resizer {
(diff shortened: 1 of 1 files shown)


swisspost-devs:

oliverschuerch (commented):
This will not work in Firefox!

> "This feature is non-standard and should not be used without careful consideration."

Source: [caniuse.com](https://caniuse.com/mdn-css_selectors_-webkit-resizer)

RedlineTriad:
Thats why I mentioned it in the title that it will only work for webkit, but @gfellerph said that it would still be an improvement.

gfellerph:
> Thats why I mentioned it in the title that it will only work for webkit, but @gfellerph said that it would still be an improvement.

I'd call it progressive enhancement. It's non-critical for the correct functioning of the text-area, just looks nice for the browsers that support it.

oliverschuerch:
Duplicate of #1765.

PR #1537, opened 2023-06-19
fix(docs): update paths in the icons getting started page

alizedebray:
Since the snippet to set the meta for icons now uses the path `/post-icons` instead or `/public/post-icons` the postinstall script must be updated.


swisspost-devs:

gfellerph (approved):

PR #1546, opened 2023-06-20
chore(docs): update paths in the icons getting started

alizedebray:


packages/documentation/src/stories/icons/getting-started.stories.mdx
```diff
@@ -135,6 +135,11 @@ You can do this with two different solutions:
 <post-icon name="1000" base="/base-path/to/your/own/icon-folder"></post-icon>
 ```
 
+### Check the Content Security Policy
+
+If your project includes the <a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy" target="_blank">Content-Security-Policy</a> response header (usually in `index.html`),
+make sure that `default-src` is correctly set to `'self'`.
```
gfellerph, line 141: Do we want to change the blueprint to `self` or do we want to describe here what policies are actually needed for loading the icons (I think `img-src: self` or `img-src: unpkg.com` will do).

gfellerph, line 141:
```suggestion
make sure that `default-src` is set to `'self'`.
```

gfellerph (approved):

gfellerph (approved):

PR #1612, opened 2023-07-06
chore: update CONTRIBUTING.md

gfellerph:
Added information about usage of custom properties

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -141,7 +141,7 @@ When a new package is added to the repo, a few things need to be taken care of.
 ## Authoring web-components
 
 ### CSS Custom Properties
-As per resolution of the [discussion about sass variables vs. CSS custom properties](https://github.com/swisspost/design-system/discussions/1380), CSS custom properties should be used if they provide meaningful ways to interact with the component. Per default, not every possible value should be  a custom property.
+As per resolution of the [discussion about sass variables vs. CSS custom properties](https://github.com/swisspost/design-system/discussions/1380), CSS custom properties should be used if they provide meaningful ways to interact with the component. Per default, not every possible value should be a custom property.
 
 #### Do
 - Use custom properties to define themable colors, for example `color` or `border-color` for dark-mode support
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -137,6 +137,21 @@ When a new package is added to the repo, a few things need to be taken care of.
 
   > ⚠ On publish, the `package.json` gets copied into the `./dist` folder. This leads to an incorrect publish path because npm now tries to publish from `./dist/dist`. You'll need a pre-publish script that removes the `directory` key from the `publishConfig` (see the [styles package pre-publish workflow](./packages/styles/gulpfile.js) for an example).
 
+
+## Authoring web-components
+
+### CSS Custom Properties
+As per resolution of the [discussion about sass variables vs. CSS custom properties](https://github.com/swisspost/design-system/discussions/1380), CSS custom properties should be used if they provide meaningful ways to interact with the component. Per default, not every possible value should be  a custom property.
```
imagoiq, line 144:
```suggestion
As per resolution of the [discussion about sass variables vs. CSS custom properties](https://github.com/swisspost/design-system/discussions/1380), CSS custom properties should be used if they provide meaningful ways to interact with the component. Per default, not every possible value should be a custom property.
```
Nitpick

CONTRIBUTING.md
```diff
@@ -137,6 +137,21 @@ When a new package is added to the repo, a few things need to be taken care of.
...
+
+#### Do
+- Use custom properties to define themable colors, for example `color` or `border-color` for dark-mode support
+- Use custom properties when updating the value with JavaScript is necessary
+- Use custom properties when a simple option is available for tweaking the component from the outside
```
imagoiq, line 149: I'm not sure about what is the meaning of this sentence. What does this imply?

imagoiq (changes requested):

CONTRIBUTING.md
gfellerph: The last point is referring to options that could be tweaked from the light dom or host. If we had a Design System where the brand color could be changed, the brand color could be a custom property that each sub-brand could set to their specific needs. Other use cases might include a component that can change it's size based on one custom property and devs can easily choose between variants by setting one property. I can't think of any real use-cases right now, but there might be more in the future. This would be an alternative to setting props on the component where it's useful to set an option globally for all instances of the component on the page. With CSS custom properties, the option can be set once (for the whole page or even a specific scope), with props it would need to be set on each instance of the component.

imagoiq (approved):

Issue #1297, opened 2023-03-23
File inputs are broken in firefox


gfellerph:
![image](https://user-images.githubusercontent.com/1659006/227217819-160c7643-af94-47a5-8f35-9b4c880bc55d.png)

https://design-system.post.ch/#/bootstrap-samples/form-file

They should look the same as on chrome:
![image](https://user-images.githubusercontent.com/1659006/227218001-4a536a2e-893a-47a6-a156-f6597a7f17f9.png)

PR #1614, opened 2023-07-06
fix(form-file): file-selector-button cutoff in Safari

imagoiq:
# What

- Use old workaround removed in https://github.com/swisspost/design-system/pull/1323, but exclusively for Safari.
- Support plural when the input as the `multiple` attribute.
- Fix file selector button padding when the input is in `valid` or `invalid` state.

# UI

## Webkit
File_input_Webkit

##  Firefox
File_input_Firefox

##  Chrome
File_input_Chrome


packages/styles/src/components/forms.scss
```diff
@@ -51,25 +51,90 @@ select.form-control-rg:not([size]):not([multiple]) {
...
+      &::after {
+        display: flex;
+        align-items: center;
+        content: 'Choose File';
+        position: absolute;
+        top: 0;
+        bottom: 0;
+        right: 0;
+        z-index: 2;
+        padding-inline: forms.$input-padding-y;
```
imagoiq, line 74: This is needed for `valid` and `invalid` state. Previously, we inherited padding value from Bootstrap. It works for all cases except those valid and invalid state where the padding right and left aren't the same. Issue was already there in https://github.com/swisspost/design-system/issues/1297

packages/styles/src/components/forms.scss
```diff
@@ -51,25 +51,90 @@ select.form-control-rg:not([size]):not([multiple]) {
       border-left: inherit;
     }
 
+    // Fix Safari cutoff when change the initial position of the selector button.
+    @supports (background: -webkit-named-image(i)) {
```
imagoiq, line 55: This come from: https://www.bram.us/2021/06/23/css-at-supports-rules-to-target-only-firefox-safari-chromium/#safari

packages/styles/src/components/forms.scss
```diff
@@ -51,25 +51,90 @@ select.form-control-rg:not([size]):not([multiple]) {
...
+      &[multiple] {
+        &::after {
+          content: 'Choose Files';
+
+          [lang='de'] & {
+            content: 'Dateien auswählen';
+          }
+
+          [lang='fr'] & {
+            content: 'Sélect. fichiers';
```
gfellerph, line 101: Not sure if we should shorten texts. Would it be "Sélectionner des fichiers" or "Choisir des fichiers"? I get that it's quite long. This article recommends to not use abbreviations https://www.experience-hub.ch/document/2777#/schreiben-digitaler-inhalte/einfache-sprache. What do you think?

packages/styles/src/components/forms.scss
```diff
@@ -51,25 +51,90 @@ select.form-control-rg:not([size]):not([multiple]) {
...
+          }
+        }
+      }
+    }
```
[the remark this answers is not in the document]
gfellerph: This has already been the source of confusion for some people. We'll provide a much improved file upload web-component that will fix this issue. For the native one, we're just providing on a best effort base. I suspect that use cases where these languages differ are quite rare and if they happen, the user most likely understands both languages.

packages/styles/src/components/forms.scss
imagoiq: Good that we have a best practice. I think you are totally right, both look good to me, so I'll take the short version.

PR #1623, opened 2023-07-10
chore(deps): update dependency prettier to v3 + stylelint

imagoiq:
### What
* Update together
  * prettier from `2.8.8` to `3.0.0`
  * stylelint-prettier from `2.0.0` to `4.0.0`
  * stylelint-config-sass-guidelines from `9.0.1` to `10.0.0`
* Replaced Prettier with Prettier standalone version for `format-code.pipe.ts` as it's the version to be used in browser without node dependencies. 

stylelint-config-sass-guidelines was not waiting because it has a dependency on stylelint-order (see Philipp's comment: 
https://github.com/bjankord/stylelint-config-sass-guidelines/issues/214)

close: https://github.com/swisspost/design-system/pull/1619
close: https://github.com/swisspost/design-system/pull/1620


packages/styles/.stylelintrc
```diff
@@ -5,12 +5,11 @@
...
   "rules": {
-    "order/properties-alphabetical-order": null,
-    "order/order": null,
     "max-nesting-depth": 5,
     "selector-max-compound-selectors": 5,
     "selector-max-id": 1,
     "selector-no-qualifying-type": [true, { "ignore": ["attribute", "class"] }],
-    "function-url-quotes": null
+    "function-url-quotes": null,
+    "scss/dollar-variable-colon-space-after": "always-single-line"
```
imagoiq, line 13: This rule is conflicting with the new prettier line break: https://prettier.io/blog/2023/07/05/3.0.0.html#fix-line-break-in-css-declaration-with-comma-14208httpsgithubcomprettierprettierpull14208-by-mvorisekhttpsgithubcommvorisek

packages/styles/schematics/utils/dom-migration/get-dom-migration-rule.ts
```diff
@@ -117,9 +117,14 @@ export function getDomMigrationRule(...updates: DomUpdate[]): Rule {
               // remove old "element" out of tree file
               treeUpdateRecorder.remove(source.start, source.end - source.start);
               // write new "element" into the tree file
+
+
+              const content = await prettier.format(distElement, PRETTIER_OPTIONS);
+              const contentWithoutBreak = content.replace(/(\n|\r\n)$/, '');
```
imagoiq, line 123: https://github.com/prettier/prettier/issues/4459#issuecomment-982661922

imagoiq:
There seems to be some issue using Prettier standalone to transform embedded code. In addition, the typing was not updated so it's harder to work with.

https://www.npmjs.com/package/@types/prettier
https://github.com/prettier/prettier/issues/15078

packages/styles/src/components/button-group.scss
```diff
@@ -17,7 +17,9 @@
 
     // added a transition delay on the "z-index" so the button only moves back after all other transitions
     z-index: 0;
-    transition: z-index 0s ease (0.5 * button.$btn-transition-duration), button.$btn-transition;
+    transition:
+      z-index 0s ease (0.5 * button.$btn-transition-duration),
+      button.$btn-transition;
```
imagoiq, line 22: Line break rules has been fixed: https://prettier.io/blog/2023/07/05/3.0.0.html#fix-line-break-in-css-declaration-with-comma-14208httpsgithubcomprettierprettierpull14208-by-mvorisekhttpsgithubcommvorisek

imagoiq:
Preview doesn't work with v7 currently.

Here is the error: 
image

[Format method from prettier is now async](https://prettier.io/blog/2023/07/05/3.0.0.html#change-public-apis-to-asynchronous-12574httpsgithubcomprettierprettierpull12574-12788httpsgithubcomprettierprettierpull12788-12790httpsgithubcomprettierprettierpull12790-13265httpsgithubcomprettierprettierpull13265-by-fiskerhttpsgithubcomfisker) and storybook `transform` parameter doesn't seem to support it (it's hard to find some doc about it). Prettier recommends to use `@prettier/sync` if we still need a synchronous method, but it's not the standalone version and therefore the issue above with the `path` from Node. Adding the following code to `package.json` doesn't have an impact either:

```
  "browser": {
    "path": false
  }
```

PR #1625, opened 2023-07-11
fix(demo): tsconfig defined path in eslint should be from the root

imagoiq:
Code_lAs5ZGhr4X

### DIFF
--- a/packages/demo/.eslintrc.js
+++ b/packages/demo/.eslintrc.js
@@ -1,4 +1,4 @@
-{
+module.exports = {
   "root": true,
   "ignorePatterns": [
     "projects/**/*"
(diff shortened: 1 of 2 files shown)


packages/demo/.eslintrc.js
```diff
@@ -12,6 +12,7 @@
         "project": [
           "tsconfig.json"
         ],
+        "tsconfigRootDir": __dirname,
```
imagoiq, line 15: It was working in CI without this, but not on local. This is the recommended way to do it (and probably the only way): https://typescript-eslint.io/linting/typed-linting/monorepos#one-tsconfigjson-per-package-and-an-optional-one-in-the-root

alizedebray (approved):

PR #1627, opened 2023-07-12
chore(documentation): migrate changelog page to v7

imagoiq:
* Replace react-markdown with ["native" markdown from Storybook.](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#importing-plain-markdown-files-with-transcludemarkdown-has-changed)

What's missing from the version 6 is the style "dark" (dark background color) on code elements. But this should be handled through style so that the color adapt to the current theme. Currently, the background stays dark when we switch the theme.

```mdx
<ReactMarkdown
  rehypePlugins={[rehypeRaw]}
  components={{
    code({ node, inline, className, children, ...props }) {
      const match = /language-(\w+)/.exec(className || '');
      return !inline && match ? (
        <Source language={match[1]} dark code={String(children).replace(/\n$/, '')} />
      ) : (
        <code className={className} {...props}>
          {children}
        </code>
      );
    },
  }}
  children={changelog}
/>
```

### DIFF
--- a/packages/documentation-v7/src/stories/misc/changelog.stories.mdx
+++ b/packages/documentation-v7/src/stories/misc/changelog.stories.mdx
@@ -1,0 +1,16 @@
+import { Markdown, Meta, Source } from '@storybook/blocks';
+import changelog from '../../../../styles/CHANGELOG.md?raw';
+
+<Meta
+  title="Misc/ChangeLog"
+  parameters={{
+    previewTabs: {
+      canvas: {
+        hidden: true,
+      },
+    },
+    viewMode: 'docs',
+  }}
+/>
+
+<Markdown>{changelog}</Markdown>
--- a/packages/documentation-v7/src/stories/misc/changelog.stories.mdx
+++ b/packages/documentation-v7/src/stories/misc/changelog.stories.mdx
@@ -3,14 +3,6 @@ import changelog from '../../../../styles/CHANGELOG.md?raw';
 
 <Meta
   title="Misc/ChangeLog"
-  parameters={{
-    previewTabs: {
-      canvas: {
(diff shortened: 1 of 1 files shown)


packages/documentation-v7/src/stories/misc/changelog.stories.mdx
```diff
@@ -0,0 +1,16 @@
+import { Markdown, Meta, Source } from '@storybook/blocks';
```
oliverschuerch, line 1: Do we really need the `Source` import?

oliverschuerch (changes requested):

packages/documentation-v7/src/stories/misc/changelog.stories.mdx
imagoiq: No you right, I'm too used to automatic removal of unused imports. Thanks!

oliverschuerch (approved):

packages/documentation-v7/src/stories/misc/changelog.stories.mdx
```diff
@@ -0,0 +1,16 @@
+import { Markdown, Meta } from '@storybook/blocks';
+import changelog from '../../../../styles/CHANGELOG.md?raw';
+
+<Meta
+  title="Misc/ChangeLog"
+  parameters={{
```
alizedebray, line 6: These parameters can most likely be removed: there is no canvas in the page and the `viewMode` is no longer needed in storybook 7 since docs have been separated from stories.

imagoiq: Thanks! Yes, I don't see any changes.

PR #1630, opened 2023-07-13
feat(documentation): POC for lit component inside mdx

imagoiq:
Here is a proof-of-concept to use lit component inside mdx with the homepage of the design system.

It's rather raw and not optimized.

### DIFF
--- a/packages/documentation-v7/src/stories/home-tile.components.scss
+++ b/packages/documentation-v7/src/stories/home-tile.components.scss
@@ -1,0 +1,82 @@
+@use '@swisspost/design-system-styles/core.scss' as post;
+
+$tile-size: 120px;
+$tile-rg-size: 100px;
+
+.tile {
+    display: flex;
+    flex-direction: column;
+    width: $tile-size;
+    height: $tile-size;
+    background-color: post.$white;
+    border-radius: post.$size-regular;
+    box-shadow: 0 2px 5px 0 rgba(post.$aubergine-dark, 0.3);
+    text-decoration: none;
+    color:inherit;
+
+    .tile--title {
+      flex: 0 0 auto;
+      padding: post.$size-mini post.$size-micro;
+      background-color: post.$yellow;
+      border-top-left-radius: inherit;
+      border-top-right-radius: inherit;
+      font-size: $tile-size * 0.1;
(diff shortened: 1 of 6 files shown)


packages/documentation-v7/src/stories/home-tile.components.scss
```diff
@@ -0,0 +1,82 @@
...
+.tile {
+    display: flex;
+    flex-direction: column;
+    width: $tile-size;
+    height: $tile-size;
+    background-color: post.$white;
+    border-radius: post.$size-regular;
+    box-shadow: 0 2px 5px 0 rgba(post.$aubergine-dark, 0.3);
+    text-decoration: none;
+    color:inherit;
```
imagoiq, line 15: This line is a new addition from the transformation.

packages/documentation-v7/src/stories/home-tile.components.scss
```diff
@@ -0,0 +1,82 @@
...
+      padding: post.$size-mini post.$size-micro;
+      background-color: post.$yellow;
+      border-top-left-radius: inherit;
+      border-top-right-radius: inherit;
+      font-size: $tile-size * 0.1;
+      font-weight: post.$font-weight-bold;
+      box-shadow: 0 2px 0 1px rgba(post.$yellow, 0.5);
+
+      ~ .tile--icon {
+        ::slotted(img) {
```
imagoiq, line 28: `::slotted` needed to be added with the transformation.

packages/documentation-v7/src/stories/home-tile.components.ts
```diff
@@ -0,0 +1,30 @@
...
+              <slot></slot>
+              </span>
+          </a>`
+        : html`
+          <div class="tile">
+              ${ this.title ? html`<span class="tile--title">${ this.title }</span>` : null }
+              <span class="tile--icon">
+              <slot></slot>
+              </span>
+          </div>`
```
imagoiq, line 27: [Static expressions](https://lit.dev/docs/templates/expressions/#static-expressions) could be potentially used to avoid duplication here, but I ran into the following issue while trying to set an attribute (e.g. `href`) as an expression: "String contains an invalid character".

PR #1635, opened 2023-07-17
feat(styles): update buttons

imagoiq:


packages/styles/src/components/button.scss
```diff
@@ -42,7 +42,7 @@
 
 // Primary
 .btn-primary {
-  border-color: transparent;
+  border-color: rgba(var(--post-contrast-color-rgb), 0.8);
```
imagoiq, line 45: L34 and L45 are to address "Change all colors from transparent colors to full-tone colors (background, color, border)" task. On this one, I'm not sure if this is the goal of this task.

.changeset/long-geckos-clap.md
```diff
@@ -0,0 +1,8 @@
+---
+'@swisspost/design-system-styles': minor
+---
+
+- Update padding of tertiary buttons from 24px to 4px
+- Update font-size progression to 12/14/16/18
```
imagoiq, line 6: The two first ones are important, I think. Is "minor" a good fit? Or perhaps "major"?

packages/styles/src/components/button.scss
```diff
@@ -127,6 +123,20 @@
   }
 }
 
+.btn-default,
+.btn-secondary,
+.btn-link,
+.btn-tertiary {
+  // Invert icon on dark backgrounds
+  @include color-mx.on-dark-background() {
+    color: var(--post-contrast-color-inverted-rgb);
```
alizedebray, line 132:
```suggestion
    color: var(--post-contrast-color-inverted);
```

.changeset/long-geckos-clap.md
alizedebray: To me, none of the changes here are breaking or add functionality, so I would consider them all patches.

In the future, if you come across a case where you introduced both a minor and a major change, you can add multiple changesets to your PR.

packages/styles/src/components/button.scss
```diff
@@ -127,6 +123,20 @@
   }
 }
 
+.btn-default,
+.btn-secondary,
+.btn-link,
+.btn-tertiary {
```
alizedebray, line 129: Here you could create a `transparent-background-button` mixin that you will use in secondary and tertiary button code blocks. This would help keep the code separated and therefore easy to read.

packages/styles/src/mixins/_button.scss
```diff
@@ -20,10 +20,13 @@
 
 @mixin button-size($size: md) {
   min-height: map.get(button.$btn-height-map, $size);
-  padding: 0 map.get(button.$btn-padding-x-map, $size);
   font-size: map.get(button.$btn-font-size-map, $size);
   gap: map.get(button.$btn-padding-x-map, $size) * 0.5;
 
+  &:not(.btn-link, .btn-tertiary) {
```
alizedebray, line 26: I'm not sure this extra selector is necessary because it increases CSS output and makes padding harder to override if ever needed. What do you think?

alizedebray (changes requested):

packages/styles/src/components/button.scss
imagoiq: Ahhh ok thanks! I didn't know that --post-gray-# is actually dynamic according to the theme.

packages/styles/src/mixins/_button.scss
imagoiq: So, I'm not sure what's the best here. I need to increase specificity at some point to override the specific padding added in the sizes. The specificity is the same for both classes, but the size classes arrived later in the cascade.

I could as well:
* Set the important flag to the padding on `.btn-link, .btn-tertiary`.
* Increase specificity of the padding by targeting `.btn.btn-link, .btn.btn-tertiary`
* …

packages/styles/src/components/button.scss
imagoiq: Good idea, and thanks for naming. I used a placeholder to avoid duplicating output.

packages/styles/src/placeholders/_button.scss
```diff
@@ -3,7 +3,7 @@
 @use './../variables/components/button';
 
 %btn-default {
-  border: button.$btn-border-width solid rgba(var(--post-contrast-color-rgb), 0.2);
+  border: button.$btn-border-width solid rgba(var(--post-contrast-color-rgb), 0.6);
```
alizedebray, line 6: This border still has an alpha.

.changeset/long-geckos-clap.md
alizedebray, line 1: Verbs are still in the present tense, they must be in the past tense (https://github.com/swisspost/design-system/pull/1635#discussion_r1266353508).

packages/styles/src/placeholders/_button.scss
```diff
@@ -23,3 +23,12 @@
     color: rgba(var(--post-contrast-color-rgb), 0.6);
   }
 }
+
+%btn-transparent-background {
```
alizedebray, line 27: The text color an :hover text color could probably be part of this placeholder as well.

alizedebray (changes requested):

packages/styles/src/mixins/_button.scss
alizedebray: `:where()` [takes a list of selectors as an argument](https://developer.mozilla.org/en-US/docs/Web/CSS/:where). It is not the case here, right?

Tertiary buttons already have overridden padding in the [current Design System](https://design-system.post.ch/#/bootstrap-samples/buttons). Is there something that has changed that we need to increase the specificity or is this a preventative measure?

imagoiq: > :where() [takes a list of selectors as an argument](https://developer.mozilla.org/en-US/docs/Web/CSS/:where). It is not the case here, right?

I'm not sure what do you mean. [Pseudo-classes are selectors](https://www.w3.org/TR/selectors-4/#pseudo-classes), no? 
It seems to work, at least with the right displayed specificity, on WebKit:
zV0ZoWBzQ4

> Tertiary buttons already have overridden padding in the [current Design System](https://design-system.post.ch/#/bootstrap-samples/buttons). Is there something that has changed that we need to increase the specificity or is this a preventative measure?

It is true, but only with the version without an additional class to specify the size. In this case, the padding comes from the additional class, and it's the same for primary, secondary and tertiary, which is not the case for the latest variant according to Figma.

alizedebray (approved):

packages/styles/src/placeholders/_button.scss
```diff
@@ -3,11 +3,10 @@
 @use './../variables/components/button';
 
 %btn-default {
```
alizedebray, line 5: There are still several rgba colors here

alizedebray (changes requested):

alizedebray (approved):

PR #1638, opened 2023-07-18
feat(documentation): migrate post-icon to storybook v7 + lit

imagoiq:
Some changes:
* Title is auto-generated now, so I removed it
* There is a double line-break between components in the source view. We can remove it, but I thought that visual it's not bad to make a larger separation.

|  Before |  After |
|---|---|
| image |  image |

* I removed the following parameters as it didn't seem to have an impact on anything:
```
  parameters: {
    controls: {
      exclude: ['Base', 'Flip Horizontally', 'Flip Vertically', 'Scale', 'Rotate', 'Animation'],
    },
  },
 ```

PR #1639, opened 2023-07-19
chore(documentation-v7): add utility pages

alizedebray:
- Added following pages: 
image

- Configured prettier for code snipets to have expected indentation


oliverschuerch (approved):

packages/documentation-v7/.storybook/preview.ts
```diff
@@ -47,10 +45,9 @@ const preview: Preview = {
       container: DocsLayout,
       source: {
         excludeDecorators: true,
+        transform: (snippet: string) => format(snippet, prettierOptions),
```
imagoiq, line 48: There was this comment "ensure the string is not empty ('') because the Source component breaks if it is" in the last version. Isn't it still necessary? I left it in my PR: https://github.com/swisspost/design-system/pull/1638/files but I was not sure.

packages/documentation-v7/src/stories/utilities/background/background.docs.mdx
```diff
@@ -0,0 +1,82 @@
...
+export const backgroundColors = parse(scss);
+
+<Meta
+  title="Utilities/Background"
+  parameters={{ badges: [BADGE.TODO] }}
+/>
+
+# Background
+
+<div className="lead">
```
imagoiq, line 20: Nitpick: we can use class now instead of className

imagoiq (approved):

packages/documentation-v7/src/stories/utilities/background/background.docs.mdx
imagoiq: Ah I got confused because I tried in the icons.docs.mdx, and it gets rendered (but still display a console error)

packages/documentation-v7/.storybook/preview.ts
alizedebray: I checked and this is no longer the case: returning an empty string makes the Source component to display the original snippet.

PR #1646, opened 2023-07-20
feat(styles): update stepper

imagoiq:
Here are the fixes from https://github.com/swisspost/design-system/issues/1636


packages/demo/src/app/post-sample/components/stepper/stepper-demo/stepper-demo.component.html
```diff
@@ -11,15 +11,12 @@ <h2 class="visually-hidden">Order progress, step 3 of 4</h2>
     <li
       *ngFor="let step of steps; index as i"
       [attr.aria-current]="isCurrent(step) ? 'step' : undefined"
-      class="stepper-item"
+      class="stepper-item stepper-interactive"
     >
-      <a *ngIf="i < currentIndex" class="stepper-link" (click)="currentIndex = i" href="#">
-        <span class="visually-hidden">Complete:</span>
+      <a class="stepper-link" (click)="currentIndex = i" href="#">
+        <span class="visually-hidden" *ngIf="i < currentIndex">Complete:</span>
```
gfellerph, line 17: From an accessibility perspective, it makes no sense that the active step is a link, because clicking it would not have any effect. Should we use a `<span>` for the active item?

gfellerph (changes requested):

.changeset/rare-kids-drum.md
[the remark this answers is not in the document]
alizedebray: It was me who asked to separate the changeset because it seems strange to me to have "clarified how to use the stepper" in the styles changelog... But it is true that the second changelog does not specify which component is affected by the changes.

packages/styles/src/components/stepper.scss
```diff
@@ -104,35 +98,38 @@
   }
 }
 
+.stepper-interactive {
```
alizedebray, line 101: This new class seems like a breaking change as projects need to add it to their steppers to not lose interactivity on completed steps...

Can't we just say `a.stepper-link` are interactive and `span.stepper-link` aren't?
This probably implies a larger refactoring to attached the `::before` to the `stepper-link` instead of the `stepper-item`.

alizedebray (changes requested):

.changeset/rare-kids-drum.md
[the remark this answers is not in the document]
imagoiq: Is there a convention for the context, or is it fine like this?

gfellerph (approved):

alizedebray (changes requested):
It seems that this criteria is no longer met when using a `a` tag for the 4th step: "Set background color black to interactive next step badge on hover (currently only the label is changed)"

imagoiq:
> It seems that this criteria is no longer met when using a `a` tag for the 4th step: "Set background color black to interactive next step badge on hover (currently only the label is changed)"

Yes, sorry. We discussed it at some meeting and decided to enable interactivity on the steps after the current step. So this is not anymore valid. I've updated the the tasklist.

.changeset/rare-kids-drum.md
```diff
@@ -0,0 +1,11 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Stepper:
+
+- Normalize font styles and weights.
+- Position the step label under the the badge on small viewport.
```
alizedebray, line 8:
```suggestion
- Position the step label under the badge on small viewport.
```

.changeset/rare-kids-drum.md
```diff
@@ -0,0 +1,11 @@
...
+- Center the step label when this text is multiline.
```
alizedebray, line 9:
```suggestion
- Center the step label when its is multiline.
```

alizedebray (approved):

PR #1662, opened 2023-07-25
fix(styles): replace jest by scss alone and make tests work again

imagoiq:
Currently, the command `jest --OnlyChanged` using `jest-scss-transformer.js` is not working. If you try to run `pnpm styles:test` (and remove the `--onlyChanged` parameter in the styles/packages.json), you'll end up with:
Code_fz2jLc4ceM
The newest version of sass requires node packages like `fs` to read the input file, but it doesn't have access through jest. Discussed with @gfellerph, here jest is an unnecessary overhead. The only benefit is the colored syntax in the terminal for errors.


packages/styles/gulpfile.js
```diff
@@ -88,33 +87,47 @@ gulp.task('transform-package-json', (done) => {
...
+ */
+gulp.task('sass:tests', () => {
+  return gulp.src('./tests/**/*.scss').pipe(
+    gulpSass.sync({
+      includePaths: options.includePaths,
+      quietDeps: true,
+    }),
+  );
+});
+
```
imagoiq, line 130: Only relevant change. Other changes are the output of prettier.

alizedebray (approved):

packages/styles/tests/TESTING.md
gfellerph, line 1: Thanks for the update!

gfellerph (approved):

PR #1703, opened 2023-07-27
chore(documentation-v7): migrate intranet header documentation

alizedebray:


gfellerph (approved):

gfellerph:
@alizedebray, is the empty story only used for the props table?

alizedebray:
> @alizedebray, is the empty story only used for the props table?

Yes, removing the story results in an error.
I also tried declaring the meta with a different file extension than `.stories.ts` but that did not work either.

PR #1704, opened 2023-07-27
chore: update contribution guidelines

gfellerph:


imagoiq (changes requested):
The images are too heavy (especially the radar). I don't think it's necessary to have it in 5k (+ there is the figma title)

Here are some optimized versions: 
88a8d899-6d3f-4f08-bf32-2f1328252d67 _1000

cb114c7a-d14a-4cb8-842f-eaa8f503607c

gfellerph:
Updated the file links, thx @imagoiq.

imagoiq (approved):
Good!

By the way, should we add Typescript to the radar as well?

gfellerph:
> Good!
> 
> By the way, should we add Typescript to the radar as well?

It's almost a given at this point, but why not.

PR #1755, opened 2023-08-08
chore(documentation-v7): add range stories

imagoiq:


packages/documentation-v7/src/stories/components/range/range.stories.ts
```diff
@@ -0,0 +1,289 @@
...
+      : null,
+  ];
+
+  const control: TemplateResult = html`
+    <input
+      id="${id}"
+      class="${classes}"
+      type="range"
+      value="${args.value}"
+      .value="${args.value}"
```
imagoiq, line 198: We need the two. One is to have the attribute. The other with the dot is to consider it as a prop and update internal state.

packages/documentation-v7/src/stories/components/range/range.stories.ts
```diff
@@ -0,0 +1,288 @@
...
+      type="range"
+      value="${args.value}"
+      .valueAsNumber="${args.value}"
+      min="${args.useBoundaries ? args.min : nothing}"
+      max="${args.useBoundaries ? args.max : nothing}"
+      step="${args.useBoundaries ? args.step : nothing}"
+      ?disabled="${args.disabled}"
+      aria-label="${useAriaLabel ? args.label : nothing}"
+      ?aria-invalid="${VALIDATION_STATE_MAP[args.validation]}"
+      @change="${(e: MouseEvent) => updateArgs({ value: (e.target as HTMLInputElement).value })}"
```
gfellerph, line 204: Provides more responsive feedback.

```suggestion
      @input="${(e: MouseEvent) => updateArgs({ value: (e.target as HTMLInputElement).value })}"
```

packages/documentation-v7/src/stories/components/range/range.stories.ts
```diff
@@ -0,0 +1,288 @@
...
+      html`
+        <input
+          id="${inputId}"
+          class="form-control mw-giant"
+          type="text"
+          inputmode="decimal"
+          value="${args.value}"
+          .valueAsNumber="${args.value}"
+          ?disabled="${args.disabled}"
+          @change="${(e: Event) => updateArgs({ value: (e.target as HTMLInputElement).value })}"
```
gfellerph, line 240:
```suggestion
          @input="${(e: Event) => updateArgs({ value: (e.target as HTMLInputElement).value })}"
```

gfellerph (changes requested):
Don't know where it's coming from, but choosing `input` leads to an error on https://preview-1755--swisspost-design-system-next-v7.netlify.app/?path=/story/components-range--default&args=showValue:input.

packages/documentation-v7/src/stories/components/range/range.stories.ts
```diff
@@ -0,0 +1,288 @@
...
+      html`
+        <input
+          id="${inputId}"
+          class="form-control mw-giant"
+          type="text"
+          inputmode="decimal"
+          value="${args.value}"
+          .value="${parseFloat(args.value)}"
+          ?disabled="${args.disabled}"
+          @input="${(e: Event) => updateArgs({ value: (e.target as HTMLInputElement).value })}"
```
oliverschuerch, line 240: The `@input` binding leads to a `NaN` output in the input if a user wants to enter a negative number.
We should probably change this to an `@change` binding.

Reproduction:
- Set `Boundaries` control to `true`
- Set `Minium Value` to a negative number
- Set `Show Value` to `input`
- Try to enter a negative number in the input

oliverschuerch (changes requested):

packages/documentation-v7/src/stories/components/range/range.stories.ts
imagoiq: The parseFloat was the issue actually, but thanks for the catch!

packages/documentation-v7/src/stories/components/range/range.stories.ts
oliverschuerch: @imagoiq Since we do not use `parseFloat` anymore to set the value, could we switch back to `@input`?

imagoiq: Woops missing a push here. It's back

oliverschuerch (approved):

alizedebray (approved):

gfellerph (approved):

PR #1757, opened 2023-08-08
chore(*): Improve Intellij IDE support

imagoiq:
Improve language recognition support and reformatting for Intellij IDE users.

At the moment, using web-types seems to be the only way to support lit (at least basically) language. See:
  - https://youtrack.jetbrains.com/issue/WEB-32640/lit-HTML-support
  - https://plugins.jetbrains.com/docs/intellij/websymbols-web-types.html
  - https://www.npmjs.com/package/@web-types/lit

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,3 +14,6 @@ trim_trailing_whitespace = true
 max_line_length = off
 insert_final_newline = false
 trim_trailing_whitespace = false
+
+[*.html]
+ij_html_quote_style = double
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
   },
   "devDependencies": {
     "@changesets/cli": "2.26.2",
+    "@web-types/lit": "2.0.0-3",
     "start-server-and-test": "2.0.0"
   },
   "engines": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -11,6 +11,9 @@ importers:
       '@changesets/cli':
         specifier: 2.26.2
         version: 2.26.2
+      '@web-types/lit':
+        specifier: 2.0.0-3
+        version: 2.0.0-3
       start-server-and-test:
         specifier: 2.0.0
         version: 2.0.0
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -61,6 +61,7 @@
   },
   "devDependencies": {
     "@changesets/cli": "2.26.2",
+    "@web-types/lit": "2.0.0-3",
```
imagoiq, line 64: Unfortunately, there is no better way than adding a new package. This package is not often updated and is mainly made of JSON files. Providing that Intellij has quite some usage at the Post, I think it's reasonable to add this package.

.editorconfig
```diff
@@ -14,3 +14,6 @@ trim_trailing_whitespace = true
 max_line_length = off
 insert_final_newline = false
 trim_trailing_whitespace = false
+
+[*.html]
+ij_html_quote_style = double
```
imagoiq, line 19: Intellij reads Prettier fairly. Our prettier config says that we want to use single quote (everywhere). But in HTML, it looks awkward. Unfortunately, there is no standard way to indicate to use double quote for HTML only with Editorconfig (https://github.com/editorconfig/editorconfig/issues/410) and Prettier.

gfellerph: Does it make sense to have the editorconfig at all if we have prettier?

package.json
imagoiq: Yes good idea

.editorconfig
imagoiq: When I format with Prettier through Intellij, I got single quote on HTML (with a default, install). Prettier doesn't allow specifying double quote for HTML. I'm not sure why vscode let us use double quote. I think it's built-in. I'm able to specify the quote type it on my Intellij Config, but as you say, it would make sense to be ready for any developer who would like to do a PR here.

gfellerph: We could also just use the default (https://prettier.io/docs/en/options#quotes) which is based on the number of escapes necessary. I don't have a personal preference other than reduce complexity.

oliverschuerch: @imagoiq would an override for html files only work?
https://prettier.io/docs/en/configuration.html#configuration-overrides

oliverschuerch (approved):

PR #1764, opened 2023-08-08
chore(releases): add workflow for releasing old documentation

gfellerph:
This workflow takes care of a few things. Old releases live as their own site on netlify. On the current documentation, we redirect e.g. /v4 to the old documentation site. The branch name is important and should always follow the format `release/v[major release]` -> `release/v3`.

1. Creating a site on netlify for the old release. If the site already exists, this job will fail but the workflow will continue to run
2. Opening a PR to update the redirects file on the current documentation
3. Releasing the new old documentation to its netlify site

**Disclaimer**: In order to test this workflow, it needs to be on main.


.github/workflows/release-old-demo.yaml
```diff
@@ -0,0 +1,140 @@
...
+  update-redirect:
+    name: Update the _redirects file
+    runs-on: ubuntu-latest
+    needs: [create-site, get-version]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Find _redirects file
```
oliverschuerch, line 59: Is this file comming from netlify or what?
Could be nice to write some line about it here.

oliverschuerch (changes requested):

.github/workflows/release-old-demo.yaml
gfellerph: It's a settings file for netlify.

gfellerph:
Closing as not relevant anymore, will be superseded by a better workflow.

PR #1768, opened 2023-08-09
feat(documentation-v7): migrate Select component stories

b1aserlu:


packages/documentation-v7/src/stories/components/select/select.docs.mdx
```diff
@@ -0,0 +1,74 @@
...
+<Canvas of={SelectStories.Default} />
+<div className="hide-col-default">
+  <Controls of={SelectStories.Default} />
+</div>
+
+## SCSS Imports
+
+To import all styles from the design-system-styles package:
+
+```scss dark
```
oliverschuerch, line 24:
```suggestion
```scss
```

packages/documentation-v7/src/stories/components/select/select.docs.mdx
```diff
@@ -0,0 +1,74 @@
...
+@use '@swisspost/design-system-styles/index.scss';
+```
+
+To import only the styles required for this component:
+
+```scss dark
```
oliverschuerch, line 30:
```suggestion
```scss
```

packages/documentation-v7/src/stories/components/select/select.stories.ts
```diff
@@ -0,0 +1,328 @@
+import type { Args, Meta, StoryContext, StoryObj } from '@storybook/web-components';
+import { html, unsafeStatic } from 'lit/static-html.js';
+import { BADGE } from '../../../../.storybook/constants';
+import { LitElement } from 'lit';
+import { useArgs } from '@storybook/preview-api';
```
oliverschuerch, line 5: Import node modules first (group them if they come from the same module) and custom imports second.

```suggestion
import { LitElement } from 'lit';
import { html, unsafeStatic } from 'lit/static-html.js';
import { useArgs } from '@storybook/preview-api';
import { BADGE } from '../../../../.storybook/constants';
```

packages/documentation-v7/src/stories/components/select/select.stories.ts
```diff
@@ -0,0 +1,328 @@
...
+        @change=${(e: Event) => {
+          updateArgs({ value: (e.target as HTMLSelectElement).value });
+
+          if (document.activeElement === e.target) {
+            setTimeout(() => {
+              const element: HTMLSelectElement | null = document.querySelector(`#${id}`);
+              if (element) element.focus();
+            }, 25);
+          }
+        }}
```
oliverschuerch, line 241: To reduce complexity, you can extract this function:
- Define the function somewhere beside the `control` definition, like so: `function onChange(e: Event) { ... }`
- Bind the function like so `@change="${onChange}"`.

oliverschuerch (changes requested):

packages/documentation-v7/src/stories/components/select/select.stories.ts
```diff
@@ -0,0 +1,328 @@
...
+    ];
+    function onChange(e: Event) {
+      updateArgs({ value: (e.target as HTMLSelectElement).value });
+
+      if (document.activeElement === e.target) {
+        setTimeout(() => {
+          const element: HTMLSelectElement | null = document.querySelector(`#${id}`);
+          if (element) element.focus();
+        }, 25);
+      }
```
oliverschuerch, line 229: These lines are setting the focus programmatically to the select field. It's probably not needed anymore in Storybook version 7.
Remove them and check if the focus stays on the select after changing its value in the story. If this is the case, we can get rid of these lines and inline the onChange function again into the control as a one-liner:

```ts
@change="${(e: Event) => updateArgs({ value: (e.target as HTMLSelectElement).value })}"
```

oliverschuerch (changes requested):

packages/documentation-v7/src/stories/components/select/select.stories.ts
```diff
@@ -0,0 +1,328 @@
...
+    }
+    const control = html`
+      <select
+        id=${id}
+        class=${classes}
+        defaultValue=${args.value}
+        ?multiple=${args.multiple}
+        size=${args.multipleSize}
```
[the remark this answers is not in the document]
b1aserlu: I found a way to do this here: [https://lit.dev/docs/templates/expressions/#removing-child]

`size=${args.mulipleSize ?? nothing }`

packages/documentation-v7/src/stories/components/select/select.stories.ts
b1aserlu: You are correct it works without this part of the function.

oliverschuerch (approved):

PR #1773, opened 2023-08-09
chore(components): remove unused puppeteer dependency

imagoiq:


oliverschuerch:
@imagoiq is everything still working without the puppeteer package? Did you run the tests on the components package?

imagoiq:
> @imagoiq is everything still working without the puppeteer package? Did you run the tests on the components package? Or is this PR more like a draft?

I ran successfully the command test and unit on the components and on all packages. And I couldn't find a place where Puppeteer is used, actually. I discussed with @gfellerph, but it seems that he couldn't as well.

PR #1779, opened 2023-08-10
chore(release): automate versions.json

gfellerph:
This workflow update runs whenever something is pushed to `main`. It does the same as before, when there are changesets, open a release PR, if there are no changesets, do a release.

If there are changesets, it will add an updated versions.json to the release PR.

Out of scope: releases and versions.json updates for older release branches.


gfellerph:
- [ ] Current version should always go to https://design-system.post.ch and not to https://design-system.post.ch/vCURRENTMAJOR
- [ ] Insert date created and date updated

.github/actions/update-versions/update-versions.mjs
```diff
@@ -0,0 +1,69 @@
...
+        .filter(k => interestingKeywords.some(i => k.includes(i)))
+        .reduce((acc, key) => {
+          acc[key] = allDependencies[key];
+          return acc;
+        }, {}),
+    },
+  ].sort((a, b) => (a.version > b.version ? -1 : 1));
+}
+
+fs.writeFileSync(versionsPath, JSON.stringify(updatedVersions, null, 2));
```
oliverschuerch, line 69: The const versionJSON is allready a copy of the version file.
So why you not just search and replace what you really need to change and write the edited versionJSON to the file afterwards?

Please read at least the comments about the 'workspace' keywords!

```mjs
if (hasVersionEntry) {
  // Update the existing entry
  const current = versionJSON.find(version => 'version' in version && version.version === currentMajorVersion);
  current.version = packageJSON.version;
  current.dependencies = 'dependencies' in version ?
    // The 'workspace' keyword could be removed before from the 'allDependencies' variable, because we probably don't want those keyords anywhere in the version.json
    Object.keys(version.dependencies).reduce((acc, key) => Object.assign(acc, { [key]: allDependencies[key].replace('workspace:', '') }), {})
   : null;
} else {
  // Add a new entry and sort based on version number
  versionJSON.push({
    title: `Version ${currentMajorVersion}`,
    version: packageJSON.version,
    description: packageJSON.description,
    url: `https://design-system.post.ch/v${currentMajorVersion}`,
    dependencies: Object.keys(allDependencies)
        .filter(k => interestingKeywords.some(i => k.includes(i)))
        // If you do not replace 'workspace' keyword in allDependencies[key] here, it will probably show up in the versions.json output
        .reduce((acc, key) => Object.assign(acc, { [key]: allDependencies[key] }), {})
  })
  .sort((a, b) => (a.version > b.version ? -1 : 1));

  // You probably want to update the previous current version here.
}

fs.writeFileSync(versionsPath, JSON.stringify(versionJSON, null, 2));
```

.github/workflows/release.yaml
```diff
@@ -44,3 +61,68 @@ jobs:
...
+    if: needs.release.outputs.main == true && needs.release.outputs.hasChangesets == true
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: changeset-release/${{ github.ref_name }}
+
+      # Update versions.json
+      - name: Update versions.json
+        uses: ./.github/actions/update-versions
```
oliverschuerch, line 84: This action has the required inputs `version_file_path` and `package_file_path`.

.github/workflows/release.yaml
```diff
@@ -44,3 +61,68 @@ jobs:
...
+    if: needs.release.outputs.main == false && needs.release.outputs.hasChangesets == false
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: changeset-release/${{ github.ref_name }}
+
+      # Update versions.json
+      - name: Update versions.json
+        uses: ./.github/actions/update-versions
```
oliverschuerch, line 105: This action has the required inputs `version_file_path` and `package_file_path`.

packages/demo/src/assets/versions.json
oliverschuerch, line 1: We could add the version 'next' to the first entry, to make sure we do not need to check for a `version` prop everywhere in the `update-versions.mjs` file.

.github/actions/update-versions/update-versions.mjs
oliverschuerch, line 1: You only create or update the current version in the `versions.json` file.
What about updating the previously current versions url?

Consider the following:
1. We have two entries in the file. V5 and V6.
2. V6 is current version and has set its url to `https://design-system.post.ch`.
3. Now we add a new major version V7.
4. There is no entry for this version, so we create a new one, but the url of V6 is not updated in this case and therefore has still the value of `https://design-system.post.ch`.

oliverschuerch (changes requested):

.github/actions/update-versions/update-versions.mjs
gfellerph: Update the logic to also update title and url on other versions if a new one gets added (with latest major detection).

gfellerph:
- [ ] Updating versions.json from an old version needs to run on the main branch, because the old documentation does not know where the versions.json is on main, but the old documentation needs to provide it's package.json as input to the update workflow

oliverschuerch (changes requested):

.github/actions/update-versions/update-versions.mjs
oliverschuerch: Not really what I meant, but as long as it's running, I'm fine with it.

gfellerph:
For the moment, it's easier to do major version changes by hand. Instructions will follow.

PR #1780, opened 2023-08-10
chore: use docs mode

gfellerph:


imagoiq (approved):

packages/documentation-v7/package.json
oliverschuerch, line 1: As long as we want to use the env mode for every environment, we could also add it to the main.ts file.

oliverschuerch (changes requested):
As long as we want to use the env mode for every environment, we could also add it to the main.ts file.
Otherwise we should add the flag to the `start:headless` script as well.

https://storybook.js.org/docs/react/api/main-config-docs#docsmode

oliverschuerch (approved):

PR #1786, opened 2023-08-11
chore(documentation-v7): remove demo stories

alizedebray:
All existing demo stories are now standard stories that way badges are shown correctly.


imagoiq (approved):
No preview URL, but it sounds alright

PR #1788, opened 2023-08-11
chore(documentation-v7): add stories for all doc pages

alizedebray:
All meta are now declared in `.stories.ts` files so that badges are displayed correctly.


packages/documentation-v7/.storybook/preview.ts
```diff
@@ -42,7 +42,7 @@ const preview: Preview = {
           'Templates',
           'Utilities',
           'Misc',
-          ['Migration Guide', 'ChangeLog', 'Versions'],
+          ['Migration Guide', 'Change Log', 'Versions'],
```
imagoiq, line 45: I'm more used to Changelog. I think both are right though, not sure.

imagoiq (approved):

gfellerph (changes requested):
There is an issue with importing markdown as raw https://github.com/swisspost/design-system/actions/runs/5834611361/job/15824527873?pr=1788#step:5:97. Maybe it's due to using vite instead of webpack with storybook v7.

packages/documentation-v7/.storybook/preview.ts
alizedebray: I renamed "Change Log" to "Changelog"

alizedebray:
> There is an issue with importing markdown as raw https://github.com/swisspost/design-system/actions/runs/5834611361/job/15824527873?pr=1788#step:5:97. Maybe it's due to using vite instead of webpack with storybook v7.

The URL wasn't correct, it should be okay now.

alizedebray:
> There is an issue with importing markdown as raw https://github.com/swisspost/design-system/actions/runs/5834611361/job/15824527873?pr=1788#step:5:97. Maybe it's due to using vite instead of webpack with storybook v7.

The path wasn't correct, I just updated it.

gfellerph (approved):

PR #1810, opened 2023-08-16
fix(documentation-v7): overwrite hcm styles for .innerZoomElementWrapper border-color

oliverschuerch:
To test:
- go to any story
- activate high-contrast-mode
- there should be no 10px border around the story anymore

### DIFF
--- a/packages/documentation-v7/.storybook/styles/preview.scss
+++ b/packages/documentation-v7/.storybook/styles/preview.scss
@@ -148,4 +148,12 @@
       }
     }
   }
+
+  @include post.high-contrast-mode() {
+    .innerZoomElementWrapper {
+      > * {
+        border-color: Canvas;
+      }
+    }
+  }
 }


packages/documentation-v7/.storybook/styles/preview.scss
```diff
@@ -148,4 +148,12 @@
       }
     }
   }
+
+  @include post.high-contrast-mode() {
+    .innerZoomElementWrapper {
+      > * {
+        border-color: Canvas;
```
imagoiq, line 155:
```suggestion
        border-color: Canvas !important;
```

The storybook css rules has an important flag so it doesn't work as it is.

gfellerph: Additionally, anyways.

oliverschuerch: > The storybook css rules has an important flag so it doesn't work as it is.
> 
> Also maybe it would rather be better to make a PR in the storybook repo, no?

This is true and I thought I adressed this with an important flag as well, but maybe I just forgot to commit it.

gfellerph (approved):

Issue #1777, opened 2023-08-10
Remove fix height from multiselect element with class `form-select-rg`


oliverschuerch:
The height of a multiselect element should be defiend with the `size` attribute and not by styles setting a fix height.
If we put the class `form-select` and `form-select-rg` on a multiselect element, the styles defined on a normal select element are defining a fix height and are not overwritten by the multiselect styles.

On all other sizes (sm, md, lg) the styles are working as expected.

![image](https://github.com/swisspost/design-system/assets/9716662/c9ba7606-82a0-41cb-9915-121da6782832)

PR #1830, opened 2023-08-17
fix(styles): removes height from form-control-rg in _form.scss

b1aserlu:
Fixes the problem of textarea and multiselect not behaving as expected when this class was asigned to them.

The select component used form-select-rg wich was pretty much an alias for form-control-rg thus the issue #1777 will also be fixed by this change.

Note: The "Text-Input" component should still have the same fixed height after this change.


gfellerph (changes requested):
The height looks good, just some nitpicks on the changeset.

gfellerph (approved):

Issue #1836, opened 2023-08-21
Update select styles


gfellerph:
The select styles were updated. The size variants are down to two.

**Design**: https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/branch/MVANE2KzNGq2et3q6T15PC/Components-Post?node-id=7802%3A33963&mode=dev

```[tasklist]
### Tasks
- [x] Update styles for floating label and non-floating label selects. The floating label has also a small variant now.
- [x] Update story, including the sizing section of the docs page.
```

gfellerph:
Breaking change is now in a separate issue https://github.com/swisspost/design-system/issues/2226

Issue #1835, opened 2023-08-21
Update text input styles


gfellerph:
Text input styles got an update. Size variants are now down to two.

**Design**: https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/branch/MVANE2KzNGq2et3q6T15PC/Components-Post?node-id=7775%3A33876&mode=dev

```[tasklist]
### Tasks
- [x] Update styles for floating label and non-floating label inputs. The floating label has also a small variant now.
- [x] Update story, including the sizing section of the docs page.
```

gfellerph:
Breaking change is now in a separate issue https://github.com/swisspost/design-system/issues/2225

imagoiq:
Please update this line as well: https://github.com/swisspost/design-system/pull/2312/files#diff-56c2d893fe13b2907fbb6816d687553409157f45087326ee9c8104d00e842cc6R151

Issue #1834, opened 2023-08-21
Update textarea styles


gfellerph:
Textareas styling changed. There are now two size variants.

**Design:** https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/branch/MVANE2KzNGq2et3q6T15PC/Components-Post?node-id=7779%3A34916&mode=dev

```[tasklist]
### Tasks
- [x] Update styles for floating label and non-floating label textareas. The floating label has also a small variant now.
- [x] Update the row control in the story: minimum rows allowed are 2 (otherwise it's hard to distinguish between a regular input field).
- [x] Update story, including the sizing section of the docs page.
```

gfellerph:
Breaking change is now in a separate ticket https://github.com/swisspost/design-system/issues/2224

davidritter-dotcom:
Waiting for #1835 to be merged because there would be some conflicts.

davidritter-dotcom:
I encountered some issues with this ticket, that could be discussed:

- The Label of the Input, Select and Textarea should get a smaller font-size in the small variant. I used the `:has` here, or should we just add a class to the label?
- In Figma there are weird paddings like 7px at the top and no padding at the bottom, which gives four different paddings in the small variant.
-  In the select and input, there are totally different paddings and also some strange ones like the top and bottom padding of 17px on the floating select.

PR #1837, opened 2023-08-21
chore: deprecating size variants for text input, textarea and select

gfellerph:
We're deprecating the regular and medium sizes for text input, textarea and select. This will make it easier to select the appropriate size variant: small for internal applications, large for external applications.

Dedicated tickets were created for each component:

- [ ] https://github.com/swisspost/design-system/issues/1834
- [ ] https://github.com/swisspost/design-system/issues/1835
- [ ] https://github.com/swisspost/design-system/issues/1836

### DIFF
--- a/.changeset/calm-swans-begin.md
+++ b/.changeset/calm-swans-begin.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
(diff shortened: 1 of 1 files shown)


alizedebray (approved):
Alerts could be added on the demo app as well.

oliverschuerch:
@gfellerph Input type=file is affected as well and should be updated as well (at least in the documentation).

PR #1841, opened 2023-08-21
chore: improve styles watch performance

gfellerph:
Only copies changed files to dist. This will improve performance for development massively because vite reacted to every change in dist individually and before, every single sass file was copied to dist regardless if it changed or not. Also, map-icons creates a theoretical infinite loop and calling watch again is not necessary within the watch task.

### DIFF
--- a/packages/styles/src/components/badge.scss
+++ b/packages/styles/src/components/badge.scss
@@ -16,6 +16,8 @@
 
 .badge {
   @extend %badge;
+
+  background-color: red;
 }
 
 a,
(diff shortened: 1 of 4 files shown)


packages/styles/src/components/badge.scss
```diff
@@ -16,6 +16,8 @@
 
 .badge {
   @extend %badge;
+
+  background-color: red;
```
imagoiq, line 20: Not for this PR, I suppose

gfellerph: true

imagoiq (approved):

PR #1845, opened 2023-08-22
feat(documentation): add possibility to open all demos full screen

alizedebray:
This PR adds a "View full screen" button to canvas, clicking this button opens the story iframe in a new tab.
Currently the button is added to all canvas, but we could also add it to selected components or even some stories. 
image

@gfellerph @imagoiq 
I need this feature for the internet header stories, but do you think it is useful to have it on all canvas?


imagoiq (approved):
Yes good idea to have it everywhere

gfellerph (commented):
Nice idea @alizedebray! Would it be possible to view the component with the currently selected args as well?

gfellerph (approved):

gfellerph:
> Nice idea @alizedebray! Would it be possible to view the component with the currently selected args as well?

Does not seem to be possible because we're in docs mode where the control state is not represented as query string params.

PR #1846, opened 2023-08-22
feat(documentation): adds visual-regression-test for radio-button

b1aserlu:


packages/documentation/cypress/snapshots/components/radio.snapshot.ts
```diff
@@ -0,0 +1,6 @@
+describe('Radio', () => {
+  it('default', () => {
+    cy.visit('./iframe.html?id=snapshots--card');
```
gfellerph, line 3:
```suggestion
    cy.visit('./iframe.html?id=snapshots--radio');
```

gfellerph (changes requested):

gfellerph (approved):

PR #1849, opened 2023-08-22
feat(documentation): adding the visual regression tests for the topic-teaser component

davidritter-dotcom:


packages/documentation/src/stories/components/topic-teaser/topic-teaser.snapshot.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+    return html`
+      <div>
+        ${['white', 'dark'].map(
+          bg => html`
+            <div class=${'row bg-' + bg}>${templateVariants}</div>
+          `,
+        )}
+      </div>
+    `;
+  },
```
b1aserlu, line 38:
```suggestion
    render: (_args: Args, context: StoryContext) => {
    const short = 'Loremipsum';
    const long = 'Vero siteos et accusam iretea et justo';

    return html`
      <div>
        ${['white', 'dark'].map(
          bg => html`
            <div class=${'row bg-' + bg}>
              ${bombArgs({
                subtitle: [short, long],
                title: [short, long],
                alignment: context.argTypes.alignment.options,
                backgroundColor: ['bg-nightblue', 'bg-coral-bright'],
                linkCount: [1, 5, 10],
              })
                .filter((args: Args) => args.title === args.subtitle)
                .map((args: Args) => {
                  return html`
                    <div class="p-3">
                      ${Default.render?.({ ...context.args, ...Default.args, ...args }, context)}
                    </div>
                  `;
                })}
            </div>
          `,
        )}
      </div>
    `;
  },
```
Use variables when you use the same value multiple times.

I think we dont need all combinations of long and short titles/subtitles with the filter you only have a quarter of the outputs you had before.

gfellerph (approved):

PR #1857, opened 2023-08-23
fix(internet-header): compatibility with storybook v7 for e2e

imagoiq:
* Extract files to packages\documentation from fixed cypress-storybook (https://github.com/NicholasBoll/cypress-storybook/pull/65) to make it work temporarily until the PR is eventually merged (Using a package with git is not supported on our network and pushing to Artifactory is too much work, see [node-sass example](https://wikit.post.ch/display/DEVPLAT/Add+node-sass+compiled+libraries+to+Artifactory)). To be discussed: as the package doesn't seem really maintained anymore and the content relatively small, we should perhaps integrate it in our code and make the upgrades ourselves when needed?
* Fix specific e2e tests

e2e tests should still fail for "breadcrumbs" and "footer" files as the stories don't exist yet. I won't skip them, as @alizedebray is working on that, and it will be ready soon.


packages/documentation/src/stories/internet-header/components/internet-header.stories.tsx
```diff
@@ -15,11 +15,12 @@ const meta: Meta<HTMLSwisspostInternetHeaderElement> = {
...
     badges: [BADGE.STABLE],
     controls: {
-      exclude: ['config-proxy']
+      exclude: ['config-proxy'],
     },
   },
   args: {
-    'project': 'test'
+    project: 'test',
+    language: 'de',
```
imagoiq, line 23: Custom config doesn't work if you are not providing a value with the language attribute.

According to the language prop description and types, it should be possible to not provide any value attribute: https://github.com/swisspost/design-system/blob/fix/e2e-internet-header-with-storybook-v7/packages/internet-header/src/components/post-internet-header/post-internet-header.tsx#L48-L51

By default, it will not load the custom config here:

https://github.com/swisspost/design-system/blob/fix/e2e-internet-header-with-storybook-v7/packages/internet-header/src/components/post-internet-header/post-internet-header.tsx#L185-L189

Other parts are loaded because `getLocalizedConfig` does the job of finding a default language: https://github.com/swisspost/design-system/blob/main/packages/internet-header/src/services/config.service.ts#L61

On the same topic, the external method `getCurrentLanguage` does have a fallback, but it's not spread to other parts of the component, so it can be wrong.
https://github.com/swisspost/design-system/blob/fix/e2e-internet-header-with-storybook-v7/packages/internet-header/src/components/post-internet-header/post-internet-header.tsx#L134-L136

If possible, we should extract the logic of `getUserLang` from `getLocalizedConfig`.

packages/documentation/src/stories/internet-header/components/internet-header.stories.tsx
```diff
@@ -64,12 +65,12 @@ function mockPage(story: any) {
     <div class="page-wrapper">
       ${story()}
       <main class="container mt-huge-r">
-        <swisspost-internet-breadcrumbs/>
-        <h1 class="mt-huge-r mb-big-r bold">Swiss Post Internet Header</h1>
+        <swisspost-internet-breadcrumbs />
+        <h1 class="mt-huge-r mb-big-r bold">Design System Internet Header</h1>
```
imagoiq, line 69: This makes a test fail

packages/internet-header/cypress/e2e/search.cy.ts
```diff
@@ -4,7 +4,8 @@ import mockStaoCacheTypes from '../fixtures/internet-header/staocache-types.json
 import { copyConfig, prepare } from '../support/prepare-story';
 
 describe('search', () => {
-  const searchButton = '#post-internet-header-search-button';
+  const searchButton = '#post-internet-header-search-button[aria-expanded=false]';
+  const closeButton = '#post-internet-header-search-button[aria-expanded=true]';
```
imagoiq, line 8: For some reason, Cypress clicks multiple time on the same button with the same id.

packages/internet-header/cypress/support/prepare-story.ts
```diff
@@ -15,14 +15,13 @@ export const prepare = (
 ) => {
   installInterceptors(config);
   cy.visitStorybook({
-    onBeforeLoad(win: { navigator: any }) {
-      // Set default browser language explicitly to English
-      Object.defineProperty(win.navigator, 'language', {
-        value: 'en',
-      });
```
imagoiq, line 22: This never works actually because it's the latest priority from the list

packages/internet-header/package.json
```diff
@@ -55,6 +55,7 @@
   },
   "devDependencies": {
     "@babel/core": "7.22.10",
+    "@cypress/webpack-preprocessor": "^5.17.1",
```
gfellerph, line 58:
```suggestion
    "@cypress/webpack-preprocessor": "5.17.1",
```

packages/internet-header/package.json
```diff
@@ -79,6 +80,7 @@
     "sass": "1.65.1",
     "start-server-and-test": "2.0.0",
     "ts-jest": "27.1.5",
+    "ts-loader": "^9.4.4",
```
gfellerph, line 83:
```suggestion
    "ts-loader": "9.4.4",
```

packages/internet-header/cypress.config.js
```diff
@@ -7,6 +7,28 @@ module.exports = defineConfig({
     includeShadowDom: true,
     viewportWidth: 1024,
     viewportHeight: 576,
+    setupNodeEvents (on, config) {
```
gfellerph, line 10: Was this added for loading the cypress-storybook files?

packages/documentation/src/stories/internet-header/components/internet-header.stories.tsx
gfellerph: I opened a new issue to handle this https://github.com/swisspost/design-system/issues/1865

packages/internet-header/cypress/support/prepare-story.ts
gfellerph: This should work if nothing else is specified.

packages/internet-header/cypress/support/prepare-story.ts
```diff
@@ -15,14 +15,13 @@ export const prepare = (
...
+    qs: {
+      lang: 'de',
```
gfellerph, line 19: What does this do? Is this equivalent to getting the language from the navigator?

packages/documentation/src/stories/internet-header/components/internet-header.stories.tsx
imagoiq: I'll close mine then ^^ I forgot to link to the comment.

packages/internet-header/cypress/support/prepare-story.ts
imagoiq: Yes, but in this case, we always get the `document.documentElement.lang` and it has a higher priority.

gfellerph: Only if it is specified on the `<html>` element. I added html documents to the fixtures where it is not specified on purpose (https://github.com/swisspost/design-system/blob/main/packages/internet-header/cypress/fixtures/pages/no-language-set.html). `document.documentElement.lang` is an empty string in those cases.

packages/documentation/src/stories/internet-header/components/internet-header.stories.tsx
alizedebray: Can't we change the test instead?
If not, just keep it as is.

alizedebray (approved):

gfellerph (approved):

packages/documentation/src/stories/internet-header/components/internet-header.stories.tsx
imagoiq: Yes you're totally right

PR #1879, opened 2023-08-28
feat(components): tooltips

gfellerph:
Adds a `post-tooltip` component.

- Why `[popover]`?: using the popover attribute (with polyfill) has several advantages. In supporting browsers, the tooltip will be lifted to the top-layer, ignoring `overflow: hidden` and the `z-index` order. This ensures that tooltips are always displayed on top of everything and never cut off. Until all browsers support the popover API (ETA is end of 2023), a polyfill without these qualities (but some decent standards) is needed. 
- Why `[data-tooltip-trigger]`? This is a custom attribute in place of `popovertargetaction`, an attribute that can be used to trigger a popover, but reacts to clicks only. In this situation we want to show the popover on hover, focus and long-press. This behavior is in consideration with `popovertargetaction="interest"`, but is not implemented at the moment. To be able to patch this functionality without interference from clicks, there is no `popovertargetaction` on the trigger element.
- floating-ui: this component uses [floating-ui](https://floating-ui.com/docs/getting-started), the successor to `@popperjs/core`, for positioning. This low-level library can be used to dynamically position the tooltip and the corresponding arrow and can be customised in every imaginable way while leaving a pretty small footprint.
- No snapshot tests: not sure how to test something that only shows up on user interaction
- No animation: had some trouble implementing an animation that was compatible with the polyfill as well as the default API. Suggestion: wait till Design provides a concept and the popover API is standardised


packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -0,0 +1,178 @@
...
+        'aria-describedby',
+        [...describedBy.split(' '), this.host.id].join(' '),
+      );
+    }
+  }
+
+  componentDidRender() {
+    // Can't figure out how to extend HTMLAttributes<HTMLParagraphElement> to support the popover attribute
+    this.tooltipRef.setAttribute('popover', '');
+  }
```
gfellerph, line 82: This is a really tricky TS situation where builds fail if a `<p popover>` is present. Temporary workaround, posted a question here https://discord.com/channels/520266681499779082/1145691780625473628/1145691780625473628 and will update when answered.

oliverschuerch: Is this still an issue?

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -0,0 +1,231 @@
...
+          ref={(el: HTMLParagraphElement & IPopoverElement) => (this.tooltipRef = el)}
+        >
+          <span
+            class="arrow"
+            ref={el => {
+              this.arrowRef = el;
+            }}
+          ></span>
+          <slot>Hi there 👋</slot>
+        </p>
```
oliverschuerch, line 227: Why we're not simply using a div as the tooltip, as we anyway change the role?
If at all, I would wrap only the slot with a paragraph tag:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role#examples

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
```diff
@@ -0,0 +1,77 @@
...
+        'The tooltip default slot accepts any <a href="https://developer.mozilla.org/en-US/docs/Glossary/Inline-level_content">inline content</a> like `<span>` or `<post-icon>`, but no interactive content like `<a>` or `<button>` and no block level content like `<div>` or `p`.',
+    },
+  },
+  // Provide some margin for the tooltip to fit comfortably inside the iFrame with overflow: hidden
+  decorators: [
+    story =>
+      html`
+        <div style="margin-block: 5rem;">${story()}</div>
+      `,
+  ],
```
oliverschuerch, line 55: Removing this decorator would clearly show, that the tooltip is placed on top of the page content and does not take any space in the layout:
image

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
```diff
@@ -0,0 +1,77 @@
...
+};
+
+function render(args: Args) {
+  // Just for fun
+  const message = args.background === 'bg-yellow' ? args.message.replace('👋', '👋🏿') : args.message;
```
oliverschuerch, line 60: The dark icons contrast is not ideal (at least on windows) and not reflected in the control table.
image

Try to use one of the other shades: https://emojipedia.org/de/winkende-hand-medium-til-lys-teint

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  return html`
+    <button
+      class="btn btn-secondary btn-large"
+      data-tooltip-target="tooltip-one"
+      aria-describedby="tooltip-one"
+    >
+      Button
+    </button>
+    <post-tooltip class="hydrated ${args.background}" id="tooltip-one" placement=${args.placement}>
```
oliverschuerch, line 70: Is the hydrated class necessary?

oliverschuerch (changes requested):

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -0,0 +1,231 @@
...
+  private clearAutoupdate: () => void;
+
+  connectedCallback() {
+    if (!this.host.id) {
+      throw new Error(
+        'No id set: <post-tooltip> must have an id, linking it to it\'s target element using the data-tooltip-target attribute.',
+      );
+    }
+
+    this.trigger = document.querySelector(`[data-tooltip-target="${this.host.id}"]`);
```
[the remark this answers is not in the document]
gfellerph: For the same tooltip? Highly unlikely IMO. Also, it's possible to trigger it programmatically from anywhere.

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
gfellerph: That does not work in Firefox and Safari currently, as they are not supporting the `popover` attribute yet and Storybook has a lot of `overflow: hidden` in their maze of divs surrounding a story.

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
gfellerph: Yes, it's being set by stencil initially. If not keeping it here, the args.background removes it and the tooltip disappears.

packages/components/src/components/post-tooltip/post-tooltip.tsx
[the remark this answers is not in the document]
oliverschuerch: What if you have a table with many rows and in each row is a info icon which will show you (for example) one of three diffrent static tooltips? If you could specify those three tooltips only once and add as much triggers to shos/hide them as you like, this could be a nice way to reduce code.
Otherwhise you would need to define a tooltip for every row or am I missing something?

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
oliverschuerch: That's really sad, because it means, that our tooltip component will not work in several browsers in a real situation, where we do not have control of parent elements with the css prop `overflow: hidden`.
I could imagine using the fully supported `dialog` element here - instead of the `popover` attribute - and restyling it to our needs could be a good solution for the moment, as noone can really say, when FF, Safari and Samsung Internet will support the popover API.
wdyt?

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -0,0 +1,231 @@
...
+  /**
+   * Remove a bunch of event listeners if the tooltip gets removed from the DOM
+   */
+  disconnectedCallback() {
+    this.trigger.removeEventListener('mouseenter', this.showTooltip);
+    this.trigger.removeEventListener('mouseleave', this.hideTooltip);
+    this.trigger.removeEventListener('focus', this.showTooltip);
+    this.trigger.removeEventListener('blur', this.hideTooltip);
+    this.trigger.removeEventListener('long-press', this.showTooltip);
+    this.tooltipRef.removeEventListener('toggle', this.handleToggle);
```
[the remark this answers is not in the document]
gfellerph: They did actually not get removed, thanks for the hint. Creating local versions of the handlers is the solution here https://stackoverflow.com/questions/33859113/javascript-removeeventlistener-not-working-inside-a-class.

packages/components/src/components/post-tooltip/post-tooltip.tsx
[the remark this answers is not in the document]
gfellerph: I update the component and docs to support multiple triggers

packages/components/src/components/post-tooltip/post-tooltip.tsx
gfellerph: Using a div now (using a dialog is really not working due to the specific semantics of dialog)

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
gfellerph: Moving all tooltips to the end of the document now, this makes it work in safari and firefox similar too.

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
oliverschuerch: Maybe it was only an issue on my screen, but I couldn't clearly see the dark icon on the yellow background.
Anyway, the value of the message control (in the controls table) is still not updated when we switch the background from `Primary` to `Post yellow`.
But I think we can ignore it in this case.

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
oliverschuerch: Oh man, I feel bad about that! Like I've ruined your great solution. 😢

On the other hand, I think it makes sense to create stable solutions for all supported browsers.
Still, I'm looking forward when we're able to implement the popover API here (hopefully soon).

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -0,0 +1,260 @@
...
+          tabindex="-1"
+          ref={(el: HTMLDivElement & IPopoverElement) => (this.tooltipRef = el)}
+        >
+          <span
+            class="arrow"
+            ref={el => {
+              this.arrowRef = el;
+            }}
+          ></span>
+          <slot>Hi there 👋</slot>
```
oliverschuerch, line 255:
```suggestion
          <slot></slot>
```

oliverschuerch (changes requested):

oliverschuerch (approved):

packages/components/src/components/post-tooltip/post-tooltip.tsx
alizedebray, line 1: Please order properties and methods in accordance with Stencil's style guide: https://stenciljs.com/docs/style-guide#high-level-example-commented (I am looking for a linting rule)

packages/documentation/src/stories/components/tooltip/tooltip.docs.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+<Canvas of={TooltipStories.Paragraph} />
+
+## One tooltip on multiple elements
+
+The same tooltip can be displayed on multiple elements. This reduces the amount of markup necessary.
+
+<Canvas of={TooltipStories.Multiple} />
+
+## Installation
```
alizedebray, line 26: We usually place the installation instruction right after the controls in the docs.

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -0,0 +1,261 @@
...
+   * Define the placement of the tooltip according to the floating-ui options available at https://floating-ui.com/docs/computePosition#placement.
+   * Tooltips are automatically flipped to the opposite side if there is not enough available space and are shifted
+   * towards the viewport if they would overlap edge boundaries.
+   */
+  @Prop() readonly placement?: Placement = 'top';
+
+  /**
+   * Internally used to track changes to the class attribute on the host element
+   */
+  @Prop() readonly class: string;
```
alizedebray, line 53: Where is this used?

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+  component: 'post-tooltip',
+  parameters: {
+    badges: [BADGE.NEEDS_REVISION],
+    controls: {
+      exclude: ['class'],
+    },
+  },
+  render,
+  args: {
+    placement: 'top',
```
alizedebray, line 16: I would remove this line to show in the snippet that the "placement" attribute is not mandatory.

```suggestion
```

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+    background: 'bg-primary',
+    message: 'Hi there 👋',
+  },
+  argTypes: {
+    message: {
```
alizedebray, line 21: The "message" and "slot" controls, refer to the same thing. It is probably not necessary to have both.

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+      control: false,
+      description:
+        'The tooltip default slot accepts any <a href="https://developer.mozilla.org/en-US/docs/Glossary/Inline-level_content">inline content</a> like `<span>` or `<post-icon>`, but no interactive content like `<a>` or `<button>`.',
+    },
+  },
+};
+
+function render(args: Args) {
+  // Just for fun
+  const message = args.background === 'bg-yellow' ? args.message.replace('👋', '🤘🏾') : args.message;
```
alizedebray, line 53: It would be good to use `useArgs` here for the message control to be updated accordingly.

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  return html`
+    <button
+      class="btn btn-secondary btn-large"
+      data-tooltip-target="tooltip-one"
+      aria-describedby="tooltip-one"
```
alizedebray, line 59: This is not necessary as it is set automatically. 

```suggestion
```

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
alizedebray: Can we use a "background-color" property to ease the implementation?

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export default meta;
+export const Default: StoryObj = {};
+
+export const Paragraph: StoryObj = {
+  render: (args: Args) => {
+    return html`
+      <p>
+        This is a paragraph element with a
+        <a href="#" data-tooltip-target="tooltip-two">tooltipped link</a>
```
alizedebray, line 77: I do not get this story: "Tooltip on non-focusable elements" but here the tooltip is on a anchor that is actually focusable.

packages/components/cypress/fixtures/tooltip/tooltip-non-focusable.html
alizedebray, line 1: Why not using the stories to test the tooltips?

alizedebray (changes requested):

alizedebray:
> * No snapshot tests: not sure how to test something that only shows up on user interaction

Probably something like: 
```
cy.get('button[data-tooltip-target="tooltip"]', { timeout: 30000 }).should('be.visible'); // wait for page load
cy.get('button[data-tooltip-target="tooltip"]').focus();
cy.get('#tooltip').should('be.visible');
cy.percySnapshot('Alerts', { widths: [1440] });
```

packages/components/src/components/post-tooltip/post-tooltip.tsx
imagoiq: I couldn't find either any usage, so I removed it.

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
imagoiq: For the hydrated class, I moved this feature into attribute with https://github.com/ionic-team/stencil/issues/2109#issuecomment-653114380

imagoiq:
Thanks @alizedebray for the great review !

imagoiq:
> > * No snapshot tests: not sure how to test something that only shows up on user interaction
> 
> Probably something like:
> 
> ```
> cy.get('button[data-tooltip-target="tooltip"]', { timeout: 30000 }).should('be.visible'); // wait for page load
> cy.get('button[data-tooltip-target="tooltip"]').focus();
> cy.get('#tooltip').should('be.visible');
> cy.percySnapshot('Alerts', { widths: [1440] });
> ```

You are right about the interaction, but I'll still avoid doing the snapshot tests as we are limited by the number of snapshots. In this case, we would need to make one snapshot per variant, as we cannot focus multiple elements at the same time.

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
imagoiq: Actually, I let as it is in this PR, as the hydrated class is used in several tests on other components. It requires another PR.

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -0,0 +1,277 @@
...
+  /**
+   * Defines the placement of the tooltip according to the floating-ui options available at https://floating-ui.com/docs/computePosition#placement.
+   * Tooltips are automatically flipped to the opposite side if there is not enough available space and are shifted
+   * towards the viewport if they would overlap edge boundaries.
+   */
+  @Prop() readonly placement?: Placement = 'top';
+
+  @Watch('backgroundColor')
+  validateBackgroundColor(newValue = this.backgroundColor) {
+    checkType(newValue, 'string', 'The post-tooltip "background-color" prop should be a string.');
```
alizedebray, line 71: Please use the `checkOneOf` checker here, for proper typing and documentation you will probably have to create a constant in an external file as I did for the [alert types](https://github.com/swisspost/design-system/blob/3bcdd5970154a81dc1802a8b99baa0dbb6327e8e/packages/components/src/components/post-alert/post-alert.tsx#L72).

packages/documentation/src/stories/components/tooltip/tooltip.docs.mdx
```diff
@@ -0,0 +1,30 @@
...
+<Meta of={TooltipStories} />
+
+# Tooltip
+
+<div className="lead">For short, informative messages or explanations.</div>
+
+<Canvas of={TooltipStories.Default} />
+<Controls of={TooltipStories.Default} />
+
+## Installation
```
alizedebray, line 14: Thanks for moving the section, can you update the title structure to: 

- Installation
- Examples
  - Tooltip on non-focusable elements
  - One tooltip on multiple elements

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -0,0 +1,277 @@
...
+
+    if (newValue === 'brand-yellow') {
```
alizedebray, line 73: I would just go for "yellow" as both the SASS variable and the background class are simply called "yellow" as well.

```suggestion
    if (newValue === 'yellow') {
```

alizedebray (approved):

PR #1884, opened 2023-08-29
fix(documentation, styles): Fix radio and checkbox invalid styles (are not red on hover and when checked)

davidritter-dotcom:


gfellerph (approved):
Nice fix!

PR #1892, opened 2023-08-31
fix(styles): define max-width for badges

davidritter-dotcom:


packages/styles/src/placeholders/badge.scss
gfellerph, line 1: Let's discuss this! Setting a max width on the badge is the right approach, but the overflow hidden can be applied to the `<span>` containing the text inside the badge. In combination with `text-overflow: ellipsis` we also get the three dots that indicate cut text. This way, the whole positioning and padding magic is not needed.

gfellerph (changes requested):

packages/styles/src/placeholders/badge.scss
oliverschuerch: There is no `span` element inside of the `.badge` element containing the text. And to add one would be a breaking change, because it's a CSS-only component, isn't it?

packages/styles/src/placeholders/badge.scss
```diff
@@ -17,12 +17,27 @@
   text-align: center;
   white-space: nowrap;
   vertical-align: baseline;
+  max-width: 100%;
+  overflow: hidden;
+  position: relative;
```
oliverschuerch, line 22:
```suggestion
```

packages/styles/src/placeholders/badge.scss
```diff
@@ -17,12 +17,27 @@
   text-align: center;
   white-space: nowrap;
   vertical-align: baseline;
+  max-width: 100%;
```
oliverschuerch, line 20: Switch `justify-content: center` to `justify-content: start`

packages/styles/src/placeholders/badge.scss
```diff
@@ -17,12 +17,27 @@
...
+  }
+
+  &:has(> .badge) {
+    padding-right: 24px/16px + badge.$badge-padding-x;
+  }
+
+  &:has(> .badge + .btn-close) {
+    padding-right: 48px/16px + badge.$badge-padding-x;
+  }
+
```
oliverschuerch, line 40:
```suggestion
```

packages/styles/src/placeholders/badge.scss
```diff
@@ -31,11 +46,16 @@
     background-color: badge.$badge-nested-bg-color;
     border-color: badge.$badge-nested-border-color;
     font-size: badge.$badge-nested-font-size;
+    &:has(+ .btn-close) {
+      right: badge.$badge-padding-x + 24px/16px;
+    }
```
oliverschuerch, line 51:
```suggestion
```

packages/styles/src/placeholders/badge.scss
```diff
@@ -31,11 +46,16 @@
...
   }
 
   > .badge,
   > .btn-close {
     margin-right: -1 * (badge.$badge-padding-x - badge.$badge-nested-translate-x);
+    position: absolute;
+    right: badge.$badge-padding-x;
```
oliverschuerch, line 58:
```suggestion
```

packages/styles/src/placeholders/badge.scss
oliverschuerch: @davidritter-dotcom Philipp and I had a conversation about this and here are the outputs:
- [ ] Revert your changes.
- [ ] A Badge with a nested element (e.g. `.badge`, `.btn-close` or `.badge-check-input`) should have a `<span>` around the text. See [demo app](https://design-system.post.ch/#/bootstrap-samples/badge) for examples.
This span should not have a class on it and is not needed, if there is no other nested element.
- [ ] The outer element `<div class="badge">` should be styled with `display: inline-flex;`, `justify-content: flex-start` and `max-width: 100%`, so inner elements get aligned next to each other and the outer element can not be wider as its parent element.
- [ ] The inner elements should not be placed absolutely, so we do not need to adjust the property `padding-right` and we do also not use the `:has` selector.
- [ ] The text wrapping `<span>` needs to be styled with the css props `text-overflow: ellipsis`, `overflow: hidden` and `white-space: nowrap`, to avoid text-break and show an ellipsis when the text is to long.

- [ ] Use the same styles as used for `.badge` on the `.badge-check-label` and as used for `.badge > span` on the `.badge-check-label > span`.

oliverschuerch (commented):
See comment on Philipps initial CR.

packages/styles/src/placeholders/badge.scss
oliverschuerch, line 1: As we do manage the text-overflow on the inner `<span>` now, we can remove the `white-space: nowrap` property from the `.badge` element.

oliverschuerch (changes requested):
Please add a changeset!

packages/documentation/src/stories/components/badge/badge.stories.ts
```diff
@@ -126,7 +126,7 @@ function externalControl(story: any, { args }: StoryContext) {
 
 // RENDERER
 function getDefaultContent(args: Args) {
-  if (!args.nestedBadge) return html`${args.text}`;
+  if (!args.nestedBadge) return html`<span>${html`${args.text}`}</span>`;
```
oliverschuerch, line 129:
```suggestion
  if (!args.nestedBadge) return html`<span>${args.text}</span>`;
```

oliverschuerch (changes requested):

gfellerph (changes requested):
The overflow behaviour seems good now, there are just two suggestions I'd like to address.

packages/documentation/src/stories/components/badge/badge.stories.ts
```diff
@@ -126,7 +126,7 @@ function externalControl(story: any, { args }: StoryContext) {
 
 // RENDERER
 function getDefaultContent(args: Args) {
-  if (!args.nestedBadge) return html`${args.text}`;
+  if (!args.nestedBadge) return html`<span>${args.text}</span>`;
```
oliverschuerch, line 129: Do we still need that?
I would suggest to implement the if statement within the return value below.
Something like

```jsx
return html`
  <span>${html`${args.text}`}</span>
  ${args.nestedBadge ? <span class="badge">10</span> : nothing}
`;
```

oliverschuerch (changes requested):

oliverschuerch (approved):
Great work!

gfellerph (approved):

PR #1898, opened 2023-08-31
feat(documentation): add documentation for Grid

b1aserlu:
Gap messes up the grid so there are no longer 12 columns awailable for elements per row.

There are unusually large padding evey 4th padding startng from the first when you use .border-gutters
image

What class could I use to make the Elements in the Grid look better?
How can I use the colors from the sass library? do I have to use bg-[color classes]?

How do I add Screenshots or imbed figma figures kinda like in the Experience Hub https://www.experience-hub.ch/document/2803/en#/grundlagen/raster-und-abstaende
Then i could add those for illustration?

THX for feedback


b1aserlu:
How do I add Screenshots or imbed figma figures kinda like in the Experience Hub https://www.experience-hub.ch/document/2803/en#/grundlagen/raster-und-abstaende
Then i could add those for illustration.

oliverschuerch:
@b1aserlu thanks for your work, but I think we're far away from what we need as a documentation for the grid.
I would suggest to copy&paste the hole bootstrap grid documentation from [here](https://getbootstrap.com/docs/5.3/layout/grid/), remove everything bootstrap related (ads, keywords like "bootstrap", etc.) and create our own examples, which do not need to be interactive, as the grid is a css-only component (without any dynamic functionality).
We also have more breakpoints than bootstrap provides by default. Please have a look into the grid implementation in the styles package.

If this task feels a bit overwhelming for you, I can take over and you can make the review it at the end to get a feeling how it could/should be done. Just give me a sign about that.

oliverschuerch:
@b1aserlu I have created a counter proposal for the grid docs page. Please have a look: #1947.

oliverschuerch:
Allready merged with: https://github.com/swisspost/design-system/pull/1947

PR #1917, opened 2023-09-04
fix(docs): add checkbox and radio button card code snippet

gfellerph:


packages/documentation/src/stories/components/choice-card/radiobutton-card.docs.mdx
```diff
@@ -38,6 +38,33 @@ import * as RadiobuttonCardStories from './radiobutton-card.stories';
...
+      parent.classList.remove('focused');
+      break;
+    case 'input':
+      parent.classList.toggle('checked', (e.target as HTMLInputElement).checked);
+      break;
+  }
+}));
+`}
+  language="typescript"
+  dark="true"
```
imagoiq, line 65: I think we said that we don't use dark mode as it's not supported by all type of code snippet?

oliverschuerch (approved):

PR #1927, opened 2023-09-06
fix(styles): updates gap in topic-teaser

b1aserlu:
updates/adds the gap between the title and link-list of the topic-teaser-content class to comply with figma design


.changeset/six-clouds-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+updated and added the gaps between the title and the link-list in the topic-teaser-content class to comply with the figma design
```
oliverschuerch, line 5:
```suggestion
Updated and added the gaps between the title and the link-list in the `.topic-teaser-content` class to comply with the figma design.
```

packages/styles/src/variables/components/_topic-teaser.scss
```diff
@@ -9,6 +9,12 @@ $content-cols: (
   xl: 7,
 );
 
+$content-gap: (
+  xs: 10px,
```
oliverschuerch, line 13: If the gap between the `.topic-teaser-title` and the `.link-list` inside of `.topic-teaser-content` really only should be `10px`, you need to remove the `margin-bottom` from the title.
Otherwise you will have a bigger gap on `xs` breakpoint:
image
blue: margin from title
purple: gap

oliverschuerch (changes requested):

packages/styles/src/variables/components/_topic-teaser.scss
b1aserlu: After double checking the gap or rather the space in between the 2 elements is supposed to be 24px wich is exactly the size of the title margin. Therefore I removed the gap for xs and sm completely.

oliverschuerch (approved):

PR #1928, opened 2023-09-06
fix(documentation): Convert fenced code snippet to Source and use dark theme

### LINKED ISSUE #1677: Use the same theme on all codeblock
1. There seems to be and inconsistency between usage of dark mode on code snippet

- Example without dark mode: https://next.design-system.post.ch/?path=/docs/internet-header-getting-started--page
- Example with dark mode: https://next.design-system.post.ch/?path=/docs/getting-started-styles--page

I guess, it's easy to forget to add the `dark` attribute.

2. The dark mode is not dynamic and doesn't adapt 

This was already an issue with Storybook 6. It's usable as is, though.

3. Codeblock used within Markdown and converted (like the changelog) cannot bet set with dark theme.

4. With Storybook 7, this `dark` attribute that we use in the markdown flavor is not working anymore, and there won't be a fix for that, see https://github.com/storybookjs/storybook/pull/21626

I'm talking about this syntax:

\```scss dark
@use '@swisspost/design-system-styles/<bundle-name>.<extension>';
\```

The "new" way is to use the Source element like:

```
<Source
  code={`@use '@swisspost/design-system-styles/basics.scss';
@use '@swisspost/design-system-styles/components/alert.scss';`}
  language="css"
  dark={true}
/>
```

… but it's not really tasty to put code in attribute and it's still suffering from the first and second points.
Unfortunately, it seems there is no way to apply dark theme globally.

imagoiq:
After Storybook 7.4.0, the documentation served (and not build) doesn't display code snippet properly:
Teams_emZES03vNY

This PR convert fenced code snippet to Source element, officially supported. Most code snippets are available in an external file (one code snippet per file) and imported into the Source element.
This PR also set dark theme as default for code snippets (and closes #1677).


packages/documentation/src/shared/styles-package-import-individual.sample.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export function getStylesPackageImportIndividual(props: Props) {
+  const basics = `@use '@swisspost/design-system-styles/basics.scss';`;
+  const generateComponentLine = (name: string) =>
+    `@use '@swisspost/design-system-styles/components/${name}.scss';`;
+  const components = props.components.map(generateComponentLine).join('\n');
+
+  const required = Object.keys(props?.required || {}).reduce(
+    (acc: string, curr: string) =>
+      `${acc}\n//${requiredLabels[curr]}\n${generateComponentLine(curr)}`,
```
oliverschuerch, line 20:
```suggestion
      `${acc}\n// ${requiredLabels[curr]}\n${generateComponentLine(curr)}`,
```

packages/documentation/src/stories/intranet-header/intranet-header.docs.mdx
```diff
@@ -21,23 +26,20 @@ The Header Angular component for internal usage.
...
 ## Usage
 
 1. In `app.module.ts`, import `SwissPostIntranetHeaderModule` and add it to the `imports` array property.
 
-```js dark
-// Other imports ....
-import { SwissPostIntranetHeaderModule } from '@swisspost/design-system-intranet-header';
+<Source
+  code={SampleUse}
+  language="bash"
```
oliverschuerch, line 40:
```suggestion
  language="jsx"
```

oliverschuerch (changes requested):
@imagoiq I like, great job!

oliverschuerch (approved):

Issue #1215, opened 2023-03-03
Deprecations for styles v9


gfellerph:
This ticket lists all deprecations or breaking changes for @swisspost/design-system-styles v7.

```[tasklist]
### Deprecated
- [ ] https://github.com/swisspost/design-system/issues/825
- [ ] https://github.com/swisspost/design-system/issues/952
- [ ] https://github.com/swisspost/design-system/issues/1096
- [ ] https://github.com/swisspost/design-system/issues/1248
- [ ] https://github.com/swisspost/design-system/issues/1194
- [ ] https://github.com/swisspost/design-system/issues/1294
- [ ] https://github.com/swisspost/design-system/pull/1551
- [ ] https://github.com/swisspost/design-system/issues/352
- [ ] https://github.com/swisspost/design-system/issues/1785
- [ ] https://github.com/swisspost/design-system/pull/1837
- [ ] https://github.com/swisspost/design-system/pull/1945
- [ ] https://github.com/swisspost/design-system/issues/2014
- [ ] https://github.com/swisspost/design-system/pull/2045
- [ ] https://github.com/swisspost/design-system/issues/2041
- [ ] https://github.com/swisspost/design-system/issues/2096
- [ ] https://github.com/swisspost/design-system/issues/2196
- [ ] https://github.com/swisspost/design-system/issues/2863
- [ ] https://github.com/swisspost/design-system/issues/2889
- [ ] https://github.com/swisspost/design-system/issues/3045
```

```[tasklist]
### Breaking changes
- [ ] https://github.com/swisspost/design-system/issues/1195
- [ ] https://github.com/swisspost/design-system/issues/2016
- [ ] https://github.com/swisspost/design-system/issues/2032
- [ ] https://github.com/swisspost/design-system/issues/826
- [ ] https://github.com/swisspost/design-system/issues/1274
- [ ] https://github.com/swisspost/design-system/issues/1301
- [ ] https://github.com/swisspost/design-system/issues/1302
- [ ] https://github.com/swisspost/design-system/issues/1303
- [ ] https://github.com/swisspost/design-system/issues/1304
- [ ] https://github.com/swisspost/design-system/issues/1305
- [ ] https://github.com/swisspost/design-system/issues/1991
- [ ] https://github.com/swisspost/design-system/issues/1992
- [ ] https://github.com/swisspost/design-system/issues/2202
- [ ] https://github.com/swisspost/design-system/issues/2224
[...]

PR #1945, opened 2023-09-14
chore(styles, docs): re-apply mask icon changes

alizedebray:
Used mask-image icons for all components except form checks that will be handled separately.


packages/styles/src/components/_index.scss
```diff
@@ -45,4 +45,4 @@
 @use 'datepicker';
 @use 'dropdown';
 @use 'floating-label';
-@use 'form-feedback';
+@use 'form-validation';
```
oliverschuerch, line 48: Is this part of this PR or should we move it to an own PR?

packages/styles/src/components/datatable.scss
```diff
@@ -67,22 +68,18 @@
...
-    background-position: center top, center bottom;
+    @include icons-mx.icon(2127);
   }
 
-  &.datatable-icon-up {
-    background-image: datatable.$datatable-sort-asc-icon;
-    background-position: center center;
+  &.datatable-icon-up,
+  &.datatable-icon-down {
+    @include icons-mx.icon(2112);
```
oliverschuerch, line 78: What do you think, should we document this mixin on the icons docs page?

packages/styles/src/components/datepicker.scss
```diff
@@ -133,7 +136,8 @@ span.ngb-dp-navigation-chevron {
   background: transparent;
 
   .pi-3203,
-  .pi-calendar {
+  .pi-calendar,
+  post-icon {
     @include icons-mx.pi(3203);
```
oliverschuerch, line 141: Can we change this to `icons-mx.icon(3203)`?

packages/styles/src/mixins/_animation.scss
```diff
@@ -31,6 +31,7 @@
   }
 }
 
+// DEPRECATED
```
oliverschuerch, line 34: Did you add this here: https://github.com/swisspost/design-system/issues/1215?

packages/styles/src/variables/components/_datatable.scss
```diff
@@ -45,9 +38,12 @@ $datatable-button-width: calc(100% - 2 * #{$datatable-button-margin});
...
 $datatable-button-focus-shadow: button.$btn-focus-box-shadow;
 
-$_pencil: icons.get-colored-svg-url('3193', color.$gray-80);
-$datatable-editable-icon: url($_pencil);
-
 $datatable-progressbar-height: spacing.$size-micro;
 $datatable-progressbar-backgroundcolor: color.$gray-10;
 $datatable-progressbar-activecolor: color.$yellow;
+
+// DEPRECATED
```
oliverschuerch, line 45: Did you add this here: https://github.com/swisspost/design-system/issues/1215?

oliverschuerch (dismissed):
There are lot of changes, which do not directly have a relation to the re-apply mask icon changes in my opinion.
For example the form-feedback/form-validation renaming.
It would be nice if we could split this PR up in smaller ones, so it's easier to review and we can add suitable changesets.

packages/documentation/src/stories/components/card/card.stories.ts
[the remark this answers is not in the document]
alizedebray: I don't think it's necessary here since this story is just to show that you can use any content from a card.
However, I added this hint to other components having the "Icon" control (buttons and toats).

packages/styles/src/components/datatable.scss
alizedebray: I don't think this mixin should be used externally.
We use it internally because styles need to display icons without being able to modify the HTML structure, but this is not the case for Design System users who need to use the post-icon component.

packages/styles/src/components/_index.scss
alizedebray: This PR is to merge back changes from this branch: https://github.com/swisspost/design-system/compare/main...re-applied-icon-changes. It includes the renaming of "form-feedback" to "form-validation".

alizedebray:
> There are lot of changes, which do not directly have a relation to the re-apply mask icon changes in my opinion. For example the form-feedback/form-validation renaming. It would be nice if we could split this PR up in smaller ones, so it's easier to review and we can add suitable changesets.

@oliverschuerch The goal of this PR is to merge back changes that were reverted a few weeks back and are now contained on this branch https://github.com/swisspost/design-system/compare/main...re-applied-icon-changes. There were several PRs originally which explains why the changes are disparate. 

The rollback was only due to issues found on the form check styles, so we decided during a Roundtable to address them in a separate ticket (#1953) and re-apply all other changes in a single PR (this one).

imagoiq (approved):
Maybe just fix the 3 code smells. e2e seem to failing in other PRs as well I'm having a look

PR #2024, opened 2023-10-02
chore: downgrade other packages to cypress 12

gfellerph:
This change is not strictly necessary, as for these packages, cypress 13 is working just fine. However, for the local binary installation, it makes sense to sync cypress versions across packages.


imagoiq (approved):

imagoiq:
Closes in favor of https://github.com/swisspost/design-system/pull/2055

PR #2026, opened 2023-10-03
chore(documentation): display code by default on first canvas

davidritter-dotcom:


imagoiq (changes requested):
Thanks. I've just tried it and it seems that you can do it by adding this into "preview.ts" file:

```
canvas: {
   sourceState: 'shown',
}
```

imagoiq (approved):
Nevermind I missread it, everything is alright

PR #2027, opened 2023-10-03
feat(internet-header): Add possibility to hide buttons in the breadcrumbs

imagoiq:


packages/internet-header/src/components/post-internet-breadcrumbs/post-internet-breadcrumbs.scss
```diff
@@ -18,6 +19,7 @@
   align-items: center;
   justify-content: space-between;
   gap: 1rem;
+  min-height: spacing.$size-small-huge; // Avoid visual shift when buttons are hidden.
```
imagoiq, line 22: Maybe there is a cleaner way…

packages/internet-header/src/components/post-internet-breadcrumbs/post-internet-breadcrumbs.tsx
```diff
@@ -34,14 +39,14 @@ export class PostInternetBreadcrumbs {
...
+    if (!overlay) {
       console.warn(
         `Internet Header: Failed to toggle overlay with id #${overlayId} as it was not found in the breadcrumb buttons config.`,
       );
       return;
     }
 
-    this.toggleOverlay(overlay)
+    this.toggleOverlay(overlay);
   }
```
imagoiq, line 50: Some linting

packages/internet-header/src/components/post-internet-breadcrumbs/post-internet-breadcrumbs.tsx
```diff
@@ -295,19 +300,23 @@ export class PostInternetBreadcrumbs {
...
-              </button>
-            ))}
-          </div>
+          {!this.hideButtons && (
+            <div class="breadcrumb-buttons">
+              {breadcrumbConfig.buttons.map(button => (
+                <button
+                  class="btn btn-secondary btn-icon"
+                  key={button.text}
+                  aria-expanded={`${Boolean(
```
imagoiq, line 309: Boolean was added as well to prevent having "undefined" inside aria-expanded attribute.

packages/internet-header/src/components/post-internet-breadcrumbs/post-internet-breadcrumbs.scss
oliverschuerch: Not sure if this is really needed? DEVs will probably not toggeling these buttons during a user visit.
Instead they will decide, that in their specific project, those buttons are not needed at all or only on some pages.

Or at least we could use the button height variable `$btn-height-md` (3rem) from styles/src/variables/components/_button.scss.

oliverschuerch (changes requested):

packages/internet-header/src/components/post-internet-breadcrumbs/post-internet-breadcrumbs.scss
imagoiq: I agree with the fact that they will not be toggling between those two states, but I think it should be the same size across projects for homogeneity and to make it easier to reason about with mockup.

imagoiq: Yep it's done

oliverschuerch (approved):

PR #2029, opened 2023-10-03
fix(styles): border color on invalid inputs

imagoiq:


packages/styles/src/components/form-validation.scss
```diff
@@ -81,7 +81,8 @@
   }
 }
 
-.form-control {
+.form-control,
+.form-select {
```
imagoiq, line 85: Was removed here: https://github.com/swisspost/design-system/pull/1945/files#diff-b472b8772f40464a972483363af6a51eb13c948bfa33cf1cef28303d0a5aeb78L58

alizedebray (approved):

PR #2036, opened 2023-10-04
feat(internet-header): adds a clear search button

b1aserlu:


packages/internet-header/src/components/post-search/post-search.tsx
```diff
@@ -377,6 +396,17 @@ export class PostSearch implements HasDropdown, IsFocusable {
                         onKeyDown={e => this.handleKeyDown(e)}
                       />
                       <label htmlFor="searchBox">{translations.flyoutSearchBoxFloatingLabel}</label>
+                      <button
+                        onClick={() => this.handleClearSearchbar()}
+                        class="visually-hidden clear-search-button"
+                        type="reset"
+                        aria-label="Clear input"
```
alizedebray, line 403: This label could be more explicit, for example “Clear search term”.
It also needs to be translated so that it is accessible to all users, check bellow `searchSubmit` label for an example.

packages/internet-header/src/components/post-search/post-search.tsx
```diff
@@ -377,6 +396,17 @@ export class PostSearch implements HasDropdown, IsFocusable {
...
+                        id="clearButton"
+                        ref={el => (this.clearButton = el)}
+                        title="Clear input"
```
alizedebray, line 406: A title is a good idea, but the header currently shows no titles on other icon buttons.

For consistency I suggest you remove the title from this PR and perhaps create a ticket to add titles to all buttons at once.

packages/internet-header/src/components/post-search/post-search.tsx
```diff
@@ -191,6 +199,17 @@ export class PostSearch implements HasDropdown, IsFocusable {
     this.deselectSuggestion();
   }
 
+  /**
+   * clear Searchbar
+   * calls handleSearchInput to update/remove suggestions
+   */
+  private handleClearSearchbar() {
```
alizedebray, line 206: This is a detail but this file uses the term "searchBox" and not "searchbar". It would be nice if you could use the same term in the code you added.

packages/internet-header/src/components/post-search/post-search.tsx
```diff
@@ -168,6 +169,13 @@ export class PostSearch implements HasDropdown, IsFocusable {
     }
     const query = this.searchBox.value.trim();
 
+    // shows or hides clearButton depending on the content of the searchbar
+    if (query !== '') {
+      this.clearButton?.classList.remove('visually-hidden');
```
alizedebray, line 174: Toggling the `.visually-hidden` class may not be the best choice. In fact, this class is used to visually hide an element while leaving it accessible to screen readers.
This means that visually impaired users will still have access to the clear button when the search is empty.

To hide the button for everyone you can use the `hidden` property of HTMLElements: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden.

packages/internet-header/src/components/post-search/post-search.scss
```diff
@@ -111,6 +112,10 @@
     height: 1.5rem;
   }
 }
+.clear-search-button {
```
alizedebray, line 115:
```suggestion

.clear-search-button {
```

alizedebray (changes requested):

PR #2044, opened 2023-10-04
fix(intranet-header): Remove ellipsis from language selector

imagoiq:


packages/styles/src/components/intranet-header/_language-chooser.scss
```diff
@@ -10,7 +10,7 @@
   overflow-y: visible;
 
   .dropdown-item {
-    padding: spacing.$size-small-regular spacing.$size-large;
+    padding: spacing.$size-small-regular 0;
```
imagoiq, line 13: Padding is the same as actually as the space is shared between items, but if we add new items, the padding will change accordingly so it's a bit better for the future. Ideally, we should enlarge the menu according to the number of items or wrap language items on multiple lines.

gfellerph (approved):

PR #2045, opened 2023-10-04
chore: update grid behavior

oliverschuerch:

Issue #2032, opened 2023-10-03
Remove `regular` breakpoint


oliverschuerch:
```[tasklist]
### Tasks
- [ ] #1225
- [ ] Remove `regular` breakpoint from `$grid-breakpoints`
- [ ] Remove `rg` breakpoint from all relevant scss/js maps and docs
- [ ] Adjust `sm` breakpoint to start at a viewport width of `540px`
- [ ] Update Layout documentation (grid, columns, etc.)
```
See "Future Implementation" on https://design-system-grid-visualization.netlify.app/
and https://github.com/swisspost/design-system-planning/issues/91#issuecomment-1746762520

gfellerph:
- [ ] Check with @Cian77 if the Design 2.0 breakpoints line up with the planned changes from this ticket.

oliverschuerch:
After some discussions, we decided to not do this, but instead wait for the next bigger update of the grid system.

PR #2046, opened 2023-10-05
feat(styles): adjust breakpoints and visually get rid of the deprecated `rg` breakpoint

oliverschuerch:
Dependant on #2045

We need to talk about this with @rouvenpost again, before merging!!!


oliverschuerch:
We will do this with https://github.com/swisspost/design-system/issues/2032

PR #2057, opened 2023-10-09
chore(*): Remove protractor and fix vulnerabilities

imagoiq:
Protractor is deprecated since september: https://www.npmjs.com/package/protractor


package.json
```diff
@@ -75,6 +75,23 @@
...
+      "@cypress/request@< 3.0.0": "3.0.0",
+      "ejs@3.1.9": "",
+      "semver@< 5.7.2||6.0.0 ≤ Version < 6.3.1||7.0.0 ≤ Version < 7.5.2": "5.7.2||6.3.1||7.5.2",
+      "merge@<2.1.1": ">=2.1.1",
+      "glob-parent@<5.1.2": ">=5.1.2",
+      "socket.io-parser@>=4.0.4 <4.2.3": ">=4.2.3",
+      "@cypress/request@<=2.88.12": ">=3.0.0",
+      "semver@<5.7.2": ">=5.7.2",
+      "semver@>=7.0.0 <7.5.2": ">=7.5.2",
+      "postcss@<8.4.31": ">=8.4.31"
```
imagoiq, line 94: Result of `pnpm audit --fix` to fix dependabot alerts: https://github.com/swisspost/design-system/security/dependabot

gfellerph: Is this something we should do regularly, also to maintain this newly added section? 

If yes, can you extract this change and write a simple GitHub action that runs once a week or month and creates a PR after running this command?

gfellerph (changes requested):

package.json
imagoiq: I'm not sure what's best. Let's forget it at the moment and try to fix those vulnerabilities by updating packages.

gfellerph (approved):

Issue #1982, opened 2023-09-20
Component: Datepicker


wueestd:
**Figma**
https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?type=design&node-id=16545%3A53350&mode=design&t=UhNqX9sEMKDoYiIW-1

POP
https://www.figma.com/file/Uwug646a2z3s5Qwa2CZ4vV/POP-Design-System?type=design&node-id=2%3A1145&mode=design&t=7PM2KakWhMGDZ1iO-1

**Code**
https://design-system.post.ch/#/ng-bootstrap-samples/datepicker

**Doc**
https://www.experience-hub.ch/document/2803#/atoms-basis-elemente/form-inputs/datepicker-timepicker

wueestd:
@kaloyanso handing this over to you since you are in contact with Ayline

gfellerph:
Closing as done

PR #2061, opened 2023-10-10
fix(styles): add focus color on datepicker component

davidritter-dotcom:


packages/styles/src/components/datepicker.scss
```diff
@@ -98,6 +98,9 @@ span.ngb-dp-navigation-chevron {
   width: 40px !important;
   height: 40px !important;
   line-height: 40px;
+  &:focus-visible{
+    outline: forms.$input-focus-color solid forms.$input-focus-outline-thickness;
+  }
```
imagoiq, line 103: With keyboard navigation, it doesn't work (check Firefox as the default color is blue, it's more obvious).

imagoiq (changes requested):
Also, don't forget to lint. I would advise for doing it automatically with your IDE to avoid forgetting about it.

packages/styles/src/components/datepicker.scss
```diff
@@ -139,6 +146,9 @@ span.ngb-dp-navigation-chevron {
     height: datepicker.$ngb-dp-icon-size;
     transform: none;
   }
+  &:focus-visible {
+    outline: forms.$input-focus-outline-thickness solid var(--post-contrast-color);
+  }
```
imagoiq, line 151: This one is not compliant with WCAG. Check this article: https://www.sarasoueidan.com/blog/focus-indicators/

You need the same style as buttons.

imagoiq (changes requested):
Need another style for the trigger button.

gfellerph:
Wait for [#1982](https://github.com/swisspost/design-system/issues/1982)

The style should be updated according to https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?type=design&node-id=16545%3A53350&mode=design&t=loOePihAoe8Rf8Bz-1

gfellerph (changes requested):
Wait for https://github.com/swisspost/design-system/issues/1982

The style should be updated according to https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?type=design&node-id=16545%3A53350&mode=design&t=loOePihAoe8Rf8Bz-1

gfellerph:
@davidritter-dotcom implement with the current definition in Figma.

gfellerph:
@davidritter-dotcom design is ready. This is no longer blocked.

gfellerph (changes requested):
Fix linting errors and update according to current Design. If that's done, please re-request review.

davidritter-dotcom:
Waiting for #2855 because of the change in the foucs-style mixin.

gfellerph (changes requested):
There some points I noticed that don't seem to be working right now
- Focus ring should not be visible when clicking with a mouse, 
- Focus styles are not visible if a date is selected (select a date and open the picker again, the focus ring should be visible when interacting with keyboard)
- When the picker is open, the date field (and in the demo some more components) get a focus ring as well image. 
  This should not be the case

oliverschuerch:
@davidritter-dotcom: This task is in progress quite long now. Do you need help with something?
Do you need preciser instructions from one of the reviewer (@imagoiq and @gfellerph)?

Let me know if we can help.

davidritter-dotcom:
The focus on the select itself gets fixed when a decision is reached on how to apply focus styles to form-select and form-control.

gfellerph (approved):

PR #2062, opened 2023-10-10
feat(documentation): add docs page on how to use columns

davidritter-dotcom:


packages/documentation/src/stories/foundation/layout/Columns/columns.stories.ts
```diff
@@ -0,0 +1,255 @@
...
+type Story = StoryObj;
+
+export const VerticalExample: Story = {
+  parameters: {
+    controls: { exclude: ['Horizontal Alignement', 'Offset classes'] },
+  },
+  render: (args: Args) => html`
+    <div class="container">
+      <div class="row-height row ${args.alignItems}">
+        <div class="col ${args.alignSelf === 'no self alignment' ? '' : args.alignSelf}">
```
imagoiq, line 96:
```suggestion
        <div class="col${args.alignSelf === 'no self alignment' ? '' : ` ${args.alignSelf`}">
```
Nitpick, but when there is no arg, there is a space at the end (see demo with default behavior).

imagoiq (approved):

imagoiq:
I'm not sure what we would like to see in this page, perhaps @oliverschuerch you are better suited to review this PR?

packages/documentation/src/stories/foundation/layout/Columns/columns.stories.ts
```diff
@@ -0,0 +1,255 @@
...
+  },
+  decorators: [
+    (story: StoryFn, { args, context }: StoryContext) => html`
+      ${story(args, context)}
+      <p class="mt-regular"><small>Resize the browser window to see changes.</small></p>
+    `,
+  ],
+  render: (args: Args) => html`
+    <div class="row">
+      <div class="col-md-4 ${args.offsetItem}">.col-md-4 .${args.offsetItem}</div>
```
oliverschuerch, line 156: Add all available offset-classes to the controls (e.g. `.offset-1`, ... `.offset-11`).

```suggestion
      <div class="col-1 ${args.offsetItem}">.col-1 .${args.offsetItem}</div>
```

oliverschuerch (changes requested):
There are a lot of difficult to understand examples at the end of the doc.
I would try to simplify them by removing breakpoint infixes (e.g. remove `-md-` where possible).
And maybe we should remove the "Margin Utilities" example completly.

packages/documentation/src/stories/foundation/layout/Columns/columns.stories.ts
```diff
@@ -0,0 +1,252 @@
...
+    <div class="row">
+      <div class="col-md-3 ms-md-auto">.col-md-3 .ms-md-auto</div>
+      <div class="col-md-3 ms-md-auto">.col-md-3 .ms-md-auto</div>
+    </div>
+    <div class="row">
+      <div class="col-auto me-auto">.col-auto .me-auto</div>
+      <div class="col-auto">.col-auto</div>
+    </div>
+  `,
+};
```
oliverschuerch, line 217: I would recommend to remove this example. Because yes, it's possible to do such things with the margin-utilities, but I would not tell any one to do it this way. Instead DEVs should use the justify, alignment and/or offset classes.

packages/documentation/src/stories/foundation/layout/Columns/columns.stories.ts
```diff
@@ -0,0 +1,252 @@
...
+        <p class="mt-regular"><small>Resize the browser window to see changes.</small></p>
+      </div>
+    `,
+  ],
+  render: () => html`
+    <div class="col-3 p-3 mb-2">.col-3: width of 25%</div>
+
+    <div class="col-md-9 p-3">.col-md-9: width of 75% above md breakpoint</div>
+  `,
+};
```
oliverschuerch, line 233: I would recommend to remove this example. Because DEVs can still use the sizing utilities to do something similar (https://getbootstrap.com/docs/5.3/utilities/sizing/).

packages/documentation/src/stories/foundation/layout/Columns/columns.stories.ts
```diff
@@ -0,0 +1,252 @@
...
+      <div class="col-3">.col-3</div>
+
+      <!-- Force next columns to break to new line -->
+      <div class="w-100"></div>
+
+      <div class="col-3">.col-3</div>
+      <div class="col-3">.col-3</div>
+    </div>
+  `,
+};
```
oliverschuerch, line 171: Can we add controls to show how it would look, without the `div.w-100` element, this would be great!
For example:

```ts
export const ColumnBreakExample: Story = {
  args: {
    renderBreakingElement: true
  },
  argTypes: {
    renderBreakingElement: {
      name: 'Render breaking element',
      description: 'Toggle rendering of breaking element',
      control: {
        type: 'boolean',
      }
    }
  },
  render: (args: Args) => html`
    <div class="row">
      <div class="col-3">.col-3</div>
      <div class="col-3">.col-3</div>
      <!-- Force next columns to break to new line -->
      ${args.renderBreakingElement ? html`<div class="w-100"></div>` : nothing}
      <div class="col-3">.col-3</div>
      <div class="col-3">.col-3</div>
    </div>
  `,
};
```

packages/documentation/src/stories/foundation/layout/Columns/columns.stories.ts
oliverschuerch, line 1: I think we can remove all the table categories from the argTypes, as they make not much sence for only one or two args.

packages/documentation/src/stories/foundation/layout/Columns/columns.stories.ts
```diff
@@ -0,0 +1,252 @@
...
+  render: () => html`
+    <div class="container">
+      <div class="row">
+        <div class="col">First in DOM, no order applied</div>
+        <div class="col order-5">Second in DOM, with a larger order</div>
+        <div class="col order-1">Third in DOM, with an order of 1</div>
+      </div>
+    </div>
+  `,
+};
```
oliverschuerch, line 133: Can we add some controls to make this interactive? This would be much easier to understand, if user can switch some of the `order` classes.

We could for example add a dropdown select control to set the order-class on each column.
e.g. Column1Order, Column2Order, Column3Order...

packages/documentation/src/stories/foundation/layout/Columns/columns.stories.ts
```diff
@@ -0,0 +1,252 @@
...
+
+export const OffsetExample: Story = {
+  parameters: {
+    controls: { exclude: ['Align Items', 'Align Item 1', 'Horizontal Alignement'] },
+  },
+  render: (args: Args) => html`
+    <div class="row">
+      <div class="col-1 ${args.offsetItem}">.col-1 .${args.offsetItem}</div>
```
oliverschuerch, line 153:
```suggestion
      <div class="col-1 ${args.offsetItem}">.${args.offsetItem}</div>
```

packages/documentation/src/stories/foundation/layout/Columns/columns.stories.ts
```diff
@@ -0,0 +1,252 @@
...
+      <div class="col-sm-5 offset-sm-2 col-md-6 offset-md-0">
+        .col-sm-5 .offset-sm-2 .col-md-6 .offset-md-0
+      </div>
+    </div>
+    <div class="row">
+      <div class="col-sm-6 col-md-5 col-lg-6">.col-sm-6 .col-md-5 .col-lg-6</div>
+      <div class="col-sm-6 col-md-5 offset-md-2 col-lg-6 offset-lg-0">
+        .col-sm-6 .col-md-5 .offset-md-2 .col-lg-6 .offset-lg-0
+      </div>
+    </div>
```
oliverschuerch, line 192: I think one row is enough as example.
Remove this lines.

packages/documentation/src/stories/foundation/layout/Columns/columns.stories.ts
```diff
@@ -0,0 +1,252 @@
...
+        contiguous unit.
+      </div>
+      <div class="col-6">
+        .col-6
+        <br />
+        Subsequent columns continue along the new line.
+      </div>
+    </div>
+  `,
+};
```
oliverschuerch, line 252: Would be nice to have a interactive example here as well.
We could, for example make the class of the first column changeable by a control, so users can change it's width and see how this changes the breaking behavior.

oliverschuerch (changes requested):

gfellerph (changes requested):
@davidritter-dotcom, this page looks really nice with the interactive explanations. Here are some thoughts:
- The initial text looks very large, I fear nobody is going to read that and it pushes the interesting content below the fold. Can we shorten it?
- The grid is based on the bootstrap grid. I think this info is important at the start, so people that already know the bootstrap one feel right at home. We should also link to the bootstrap grid docs (also on the grid page)

oliverschuerch (approved):

oliverschuerch (commented):
Please add the columns page to the preview.ts order property and update the todo page.

```javascript
...
order: [
  ...
  'Foundations',
  [
    ...
    'Layout',
    ['Breakpoints', 'Containers', 'Grid', 'Column', 'TODOS'],
    ...
  ],
  ...
],
...
```

gfellerph (approved):

PR #2065, opened 2023-10-11
fix(styles): Checked radio buttons is wrong on dark background

imagoiq:
## What

This PR is a bit confusing as it addresses two problems.
1. Checked radio button were creating a visual shift in inline layout. It was fixed with https://github.com/swisspost/design-system/pull/2086 by using a border instead of a margin. But it created the second following problem.
2. Checked radio button on dark background blends the circle indicator with the background. #2170

Other additions:
- Added inline layout to snapshot

### Before

image

image

### After

image

image


packages/styles/src/components/form-check.scss
```diff
@@ -78,7 +78,7 @@
       border-radius: 50%;
 
       &:checked::after {
-        margin: spacing.$size-micro;
+        border: spacing.$size-micro solid color.$white;
```
gfellerph, line 81: If a default background color scheme was defined (https://github.com/swisspost/design-system/issues/2123), we could use 

```suggestion
        border: spacing.$size-micro solid rgb(var(--post-bg-rgb));
```
and have automatic dark/light background compatibility.

gfellerph (changes requested):
Do we want to wait for the changes in #2123?

imagoiq:
> Do we want to wait for the changes in #2123?

I'm not sure anymore as it's a bug. I think that it should be quickly fixed.

.changeset/gorgeous-gorillas-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Fixed offset of radio button when checked and in inline layout.
```
imagoiq, line 5: As the second problem is caused by the first and that the first was not announced in a previous changelog, I'll stick to this sentence about the first problem.

packages/styles/src/components/form-check.scss
```diff
@@ -83,9 +83,10 @@
       border-radius: 50%;
 
       &:checked::after {
-        border: spacing.$size-micro solid color.$white;
```
imagoiq, line 86: The border addition was erroneously introduced in https://github.com/swisspost/design-system/pull/2086

packages/components/.eslintrc.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
   ],
   overrides: [],
   parser: '@typescript-eslint/parser',
+  ignorePatterns: ['.eslintrc.js'],
```
imagoiq, line 14: Fix unit tests from: https://github.com/swisspost/design-system/pull/2238/files#diff-4ade8993db0ab98cef66268e7a71355b158eefa4db4a3e03f623638800612980R14

gfellerph (approved):

PR #2067, opened 2023-10-11
fix(styles): Small variants for checkbox and radio

imagoiq:
@rouvenpost I made all the feedback validation boxes smaller according to the mockups (for all variants, you can see them in snapshots). I was not sure if it was intended or not. Please let me know if it's not the case.

### Checkbox
- [Checkbox mockups](https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?node-id=7750%3A33879&mode=dev)
- [Checkbox documentation page](https://preview-2067--swisspost-design-system-next.netlify.app/?path=/docs/components-checkbox--docs)
- [Checkbox snapshots](https://preview-2067--swisspost-design-system-next.netlify.app/iframe.html?id=snapshots--checkbox&viewMode=story)

### Radio
- [Radio mockups](https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?node-id=7777%3A33876&mode=dev)
- [Radio documentation page](https://preview-2067--swisspost-design-system-next.netlify.app/?path=/docs/components-radio-button--docs)
- [Radio snapshots](https://preview-2067--swisspost-design-system-next.netlify.app/iframe.html?id=snapshots--radio&viewMode=story)

### DIFF
--- a/packages/styles/src/components/form-check.scss
+++ b/packages/styles/src/components/form-check.scss
@@ -1,6 +1,7 @@
 @forward './../variables/options';
 
 @use '../variables/color';
+@use '../variables/type';
 @use '../variables/spacing';
 @use '../variables/animation';
 @use '../variables/components/form-check';
@@ -82,7 +83,7 @@
       border-radius: 50%;
 
       &:checked::after {
-        margin: spacing.$size-micro;
+        border: spacing.$size-micro solid color.$white;
         background-color: currentColor;
         border-radius: inherit;
       }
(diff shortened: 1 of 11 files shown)


packages/styles/src/components/form-check.scss
```diff
@@ -82,7 +83,7 @@
       border-radius: 50%;
 
       &:checked::after {
-        margin: spacing.$size-micro;
+        border: spacing.$size-micro solid color.$white;
```
imagoiq, line 86: This is a fix part of this PR: https://github.com/swisspost/design-system/pull/2065

rouvenpost (approved):

packages/styles/src/components/form-check.scss
```diff
@@ -93,10 +94,25 @@
...
+        }
+      }
+    }
   }
 
   &-label {
     flex: 1;
+
+    .form-check-sm & {
+      font-size: type.$font-size-12;
```
gfellerph, line 114: Eventually, I think it would be nice to just define `<body class="small/intranet">` and every component inherits font sizes and acts accordingly. It's not quite developer friendly to have to specify on every single element that you want small variants.

@imagoiq, should we account for this now or update font-sizes and small/large differentiation later?

gfellerph (approved):

packages/styles/src/components/form-check.scss
imagoiq: I think that it's a good idea. We may want both solution to be sure to be flexible enough. As this touches as well other components, we should do this in another ticket.

Issue #1172, opened 2023-02-23
Styles: Visual regression tests for textarea


gfellerph:
Write visual regression tests for the [textarea component](https://next.design-system.post.ch/?path=/story/components-textarea--default).

Create a storybook page similar to [the button test page](https://next.design-system.post.ch/?path=/story/hidden-button-snapshot-test--page) that shows all variants on one page. In the snapshot test, navigate to this page and take one snapshot of all variants at once.

```[tasklist]
### Properties to test
- [x] Label: short and long text
- [x] Floating label
- [x] Size
- [x] Helper text: short and long text
- [x] Disabled
- [x] Validation
- [x] Viewports: 400
```

Issue #2001, opened 2023-09-27
Create a reusable component for snapshot tests


alizedebray:
The code for snapshot tests is repetitive and quite complicated at the moment. 
Creating a reusable lit component would make the snapshot test easier to write and to maintain.

PR #2076, opened 2023-10-16
feat(documentation): Add pseudo-class to snapshots with storybook-addon-pseudo-states plugin

imagoiq:
### What

Use `storybook-addon-pseudo-states` plugin to simulate pseudo-class like `hover`, `focus`, etc. This plugin rewrite the associated stylesheet and try to replace the defined pseudo-class with a class using the following naming template `.pseudo-${pseudoClassName}`.

### Limitation

- No support for @media, see https://github.com/chromaui/storybook-addon-pseudo-states/pull/75
  - Buttons: background-color is right, but we don't see the animated arrow because of `@media (prefers-reduced-motion: no-preference)` usage.
- No support for web-components built stylesheet, see https://github.com/chromaui/storybook-addon-pseudo-states/issues/10
  - Collapsible
  - Tabs
- Currently, it breaks as https://github.com/chromaui/storybook-addon-pseudo-states/pull/82 it doesn't support `::-moz-range-thumb` or `::-webkit-slider-thumb`. This can be fixed by adding them here: https://github.com/chromaui/storybook-addon-pseudo-states/blob/07abc1bb39ee4f3377955b23860c275de12b872f/src/constants.js#L6
- Currently, it uses the DOM API, so the whole stylesheet transformation appears in the browser and for each load of the page, which is not optimal. Also, for this reason, it's limited to 1000, so for us (~3500 lines at the moment) to be usable we need to remove this limit (and fork the package). Lines to be removed: https://github.com/chromaui/storybook-addon-pseudo-states/blob/baaa167909676b6fcfb778564cecb92636db390a/src/rewriteStyleSheet.js#L68-L70

### Misc
- Topic-teaser uses default browser focus-rules, so we can't simulate, and it's invisible in snapshot
- Components without snapshot: [textarea](https://github.com/swisspost/design-system/issues/1172), tooltip, button-group

### DIFF
--- a/packages/documentation/src/stories/components/input/input.snapshot.stories.ts
+++ b/packages/documentation/src/stories/components/input/input.snapshot.stories.ts
@@ -1,7 +1,8 @@
 import type { Args, StoryContext, StoryObj } from '@storybook/web-components';
 import meta from './input.stories';
 import { html } from 'lit';
-import { getCombinations, COMBINATIONS } from '../../../utils/inputComponentsGetCombinations';
+import { COMBINATIONS } from '../../../utils/inputComponentsGetCombinations';
+import { bombArgs } from '../../../utils/bombArgs';
 
 export default {
   ...meta,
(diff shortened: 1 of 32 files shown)


packages/documentation/src/stories/components/input/input.snapshot.stories.ts
```diff
@@ -10,68 +11,65 @@ export default {
...
                   (args.value &&
                     (context.args.type === 'text' || context.args.type === 'password')),
               )
-              .map(
-                (args: Args) =>
-                  html`
-                    <div>${meta.render?.({ ...context.args, ...args }, context)}</div>
-                  `,
-              )}
+              .map(args => renderVariant(args, context))}
```
imagoiq, line 55: I refactored this part to use `bombArgs` instead of a custom function. So https://github.com/swisspost/design-system/issues/2001 will be easier.

PR #2078, opened 2023-10-16
fix(documentation): Choice-card doesn't have proper input name

imagoiq:
image


packages/documentation/src/stories/components/choice-card/choice-card.ts
```diff
@@ -142,7 +142,7 @@ export const choiceCardMeta: Meta = {
 
 let id_ct = 1;
 
-export const choiceCardDefault = (args: Args, name = 'control') => {
+export const choiceCardDefault = (args: Args) => {
```
imagoiq, line 145: As this function is used to render, the second argument is the StoryContext

gfellerph (approved):

PR #2042, opened 2023-10-04
chore(components): use the web animations api in the post-collapsible…

alizedebray:
… component

PR #2079, opened 2023-10-16
feat(components): add post-accordion component

alizedebray:
- [x] Blocked by #2042


packages/documentation/src/stories/components/accordion/accordion.stories.ts
```diff
@@ -0,0 +1,58 @@
...
+        <span slot="header">Titulum 3</span>
+        <p>Contentus momentus vero siteos et accusam iretea et justo.</p>
+      </post-collapsible>
+    </post-accordion>
+  `
+};
+
+export const CloseOthers: Story = {
+  render: () => html`
+    <post-accordion close-others>
```
imagoiq, line 41: Nitpick: maybe we should display it similarly to what you wrote in the description, so close-others="true"

packages/components/src/components.d.ts
```diff
@@ -14,6 +14,24 @@ export { HeadingLevel } from "./components/post-collapsible/heading-levels";
 export { BackgroundColor } from "./components/post-tooltip/types";
 export { Placement } from "@floating-ui/dom";
 export namespace Components {
+    interface PostAccordion {
+        /**
+          * If `true`, only one `post-collapsible` can be open at a time.
+         */
+        "closeOthers": boolean;
```
imagoiq, line 21: Nitpick: In my opinion, it's something that I would have thing to be the default behavior. I looked at a couple of other similar components and they have rather a `alwaysOpen` prop.

imagoiq (approved):

packages/components/src/components.d.ts
alizedebray: Sure, we can update that. I'll check with the team

alizedebray: I just changed the default behavior and called the prop `multiple` in https://github.com/swisspost/design-system/pull/2079/commits/326891f3a3cc3d8f6bbdf04b0eff1c35f74630d9.

PR #2090, opened 2023-10-18
fix(styles): fix active switch in Chrome

alizedebray:


packages/styles/src/mixins/_icons.scss
```diff
@@ -14,6 +14,12 @@
   background-color: currentColor;
 }
 
+@mixin remove-icon() {
+  /* stylelint-disable-next-line property-no-vendor-prefix */
+  -webkit-mask-image: none;
```
imagoiq, line 19: Shouldn't it be managed by https://github.com/swisspost/design-system/blob/main/packages/styles/gulpfile.js#L117 ?

imagoiq: image

It looks like to work with this playground, our browserslist and an older version of Autoprefixer: https://autoprefixer.github.io/

alizedebray: I also checked the playground but it is currently not working for our package.

The quick fix may be temporary, but it would be nice if we could merge it soon since the bug is in production and we have already received support requests about it. I created ticket #2103 to fix the autoprefixer in another release.

imagoiq (approved):

Issue #2084, opened 2023-10-17
Styles: Visual regression tests for the button group


imagoiq:
Write visual regression tests for the [button group component](https://next.design-system.post.ch/?path=/story/components-text-input--default).

Create a storybook page similar to [the button test page](https://next.design-system.post.ch/?path=/story/hidden-button-snapshot-test--page) that shows all variants on one page. In the snapshot test, navigate to this page and take one snapshot of all variants at once.

```[tasklist]
### Properties to test
- [x] elements
- [x] Size
- [x] Checked state
- [ ] Pseudo-class
```

PR #2092, opened 2023-10-18
feat(documentation): Add pseudo-class to snapshots with postcss plugin

imagoiq:
### What

Use `postcss-pseudo-classes` plugin to simulate pseudo-class like `hover`, `focus`, etc. This plugin rewrite the associated stylesheet and tries to replace the defined pseudo-class with a class using the following naming template `:${pseudoClassName}`.

### Limitations

- No support for web-components built stylesheet. https://github.com/swisspost/design-system/issues/2102

### Misc

Snapshots that haven't been changed because they don't have a snapshot file currently:
 - tooltip
 - [textarea](https://github.com/swisspost/design-system/issues/1172)
 - [button-group](https://github.com/swisspost/design-system/issues/2084)


packages/documentation/postcss.config.js
```diff
@@ -0,0 +1,7 @@
+const postcssPseudoClasses = require('./postcss-pseudo-classes.js');
+
+module.exports = () => {
+  return {
+    plugins: [postcssPseudoClasses],
```
imagoiq, line 5: I wasn't able to change the Vite config to use this plugin like https://github.com/vitejs/vite/discussions/8693

imagoiq:
After https://github.com/swisspost/design-system/pull/2133, we should add pseudo-class

imagoiq:
Button-group snapshots have been added. @alizedebray you can review

PR #2093, opened 2023-10-18
feat(components): improve tabs keyboard navigation

alizedebray:


packages/components/src/components/post-tab-header/post-tab-header.tsx
```diff
@@ -29,9 +29,9 @@ export class PostTabHeader {
   render() {
     return (
       <Host data-version={version}>
-        <a aria-selected="false" role="tab" class="tab-title" href="#" id={this.tabId}>
+        <button aria-selected="false" class="tab-title" id={this.tabId} role="tab" tabindex="-1">
```
imagoiq, line 32:
```suggestion
        <button aria-selected="false" class="tab-title" id={this.tabId} role="tab" tabindex="-1" type="button">
```

imagoiq, line 32: Could you explain this change (for my own curiosity)? Is there a case where link would be more appropriate?

imagoiq (approved):

packages/components/src/components/post-tab-header/post-tab-header.tsx
alizedebray: Buttons activate with both `Enter` and `Space` while links only activate with `Space`.
Thus, I used a button since the keyboard interaction guidelines state:
> `Space or Enter`: Activates the tab if it was not activated automatically on focus.

The change should have no other impact. As the element has a "tab" role, it is not displayed as a button or link in the list of elements (`Insert + F7` with NVDA).

PR #2094, opened 2023-10-18
fix(docs): fix alert snapshot tests

alizedebray:


imagoiq:
> > It's not about loading, but about the config.
> 
> The stories are now fixed, but I think waiting for the post-icon in snapshot tests is still necessary.

It works on the snapshot. But there is a weird flash switching between icons on the two last elements of each color group.

imagoiq (approved):

Issue #352, opened 2022-08-17
Clean up alert/toast styles


gfellerph:
Alert and toast styling relies on a series of complicated placeholders. This makes it hard to update or change styles for these components. Clean up and simplify the styling, maybe use a mixin instead of a placeholder.

PR #2106, opened 2023-10-19
feat(styles): Table mono variant

imagoiq:


packages/styles/src/components/tables.scss
```diff
@@ -8,3 +9,54 @@
...
+  }
+
+  > tbody > tr {
+    background-color: color.$gray-background;
+  }
+}
+
+%no-border {
+  border: 0;
+}
```
gfellerph, line 62: This creates one hell of a selector and can lead to complex issues if it's used anywhere else. We had weird issues with our complex setup and opted to go with mixins wherever possible. I think here we don't loose much by using mixins instead.

Reference (one of the cases where it got complicated): https://github.com/swisspost/design-system/issues/352

Whether we use mixins or placeholders, it would be good to outsource it into it's own file.

gfellerph (changes requested):

packages/styles/src/components/tables.scss
imagoiq: Yes, I can imagine that's why I thought about using this placeholder only in the same file and available for sharing. But anyway, I can simply remove it. A mixin here doesn't bring anything, I think.

gfellerph (approved):

PR #2109, opened 2023-10-20
feat(components): popover

gfellerph:
Adds a `post-popover` and a `post-popovercontainer` component.

image

This is a draft PR because the component names are not definitive yet. 

**Why handle mouse, touch and keyboard events manually?**
Cross shadow boundary `[popovertarget]` does not work at this time. Therefore it's necessary to add custom event listeners for triggering the functionality. We can still profit from toplayer positioning in supporting browsers (support will improve over time) and as soon as cross shadow boundary `[popovertarget]` is implemented, this component can be simplified.

**Why is there an undocumented `<post-popovercontainer>` component?**
This is a container component used to display content displayed in front of the page content and anchored to their trigger. It's usage is intended to be internally for components like `<post-tooltip>` and `<post-popover>`. Other components like `<post-split-button>` and `<post-selectlist>` can profit from this component as well. It encapsulates positioning logic from the [floating-ui library](https://floating-ui.com).

```[tasklist]
### Todos
- [x] Add tests for the `<post-popovercontainer>` component
```


packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -0,0 +1,61 @@
+describe('popover', () => {
+  describe('default', () => {
+    beforeEach(() => {
+      cy.getComponent('popover');
+      // Aria-expanded is set by the web component, therefor it's a good measure to indicate the component is ready
```
imagoiq, line 5:
```suggestion
      // Aria-expanded is set by the web component, therefore it's a good measure to indicate the component is ready
```
Nitpick: Therefor exists but I think you mean more therefore

packages/components/src/components/post-popover/post-popover.tsx
```diff
@@ -0,0 +1,154 @@
...
+   * Defines the placement of the popover according to the floating-ui options available at https://floating-ui.com/docs/computePosition#placement.
+   * Popoverss are automatically flipped to the opposite side if there is not enough available space and are shifted
+   * towards the viewport if they would overlap edge boundaries.
+   */
+  @Prop() readonly placement?: Placement = 'right-end';
+
+  /**
+   * Define the caption of the close button for assistive technology
+   */
+  @Prop() readonly closeButtonCaption: string;
```
imagoiq, line 29: It would be great if you can display it as required in Storybook docs.

packages/components/src/components/post-popovercontainer/post-popovercontainer.scss
```diff
@@ -0,0 +1,52 @@
...
+
+  // Keeps the little arrow visible
+  overflow: visible;
+}
+
+.arrow {
+  $arrow-size: math.div(spacing.$spacer, math.sqrt(2));
+  position: absolute;
+  // Diagonale of 16px -> 1rem -> 1/1.41 = ~0.7
+  // https://www.omnicalculator.com/math/square-diagonal?c=CHF&v=hide:0,diagonal:16!cm
```
imagoiq, line 41:
```suggestion
  // Diagonale of 16px -> 1rem -> 1/1.41 = ~0.7
  // https://www.omnicalculator.com/math/square-diagonal?c=CHF&v=hide:0,diagonal:16!cm
  $arrow-size: math.div(spacing.$spacer, math.sqrt(2));
  position: absolute;
```

packages/components/src/components/post-popovercontainer/post-popovercontainer.scss
```diff
@@ -0,0 +1,52 @@
...
+  width: $arrow-size;
+  height: $arrow-size;
+  background-color: inherit;
+  rotate: 45deg;
+  pointer-events: none;
+  z-index: -1;
+
+  // High contrast mode borders
```
imagoiq, line 49:
```suggestion
  // Create transparent border to be styled by and for the high contrast mode
```
Maybe, it's a bit more explicit for people that are not aware.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -0,0 +1,210 @@
...
+          // Prevents shifting away from the anchor too far, while shifting as far as possible
+          // https://floating-ui.com/docs/shift#limiter
+          limiter: limitShift({
+            offset: 32,
+          }),
+        }),
+        size({
+          padding: 8,
+          apply({ availableWidth, elements }) {
+            // Do things with the data, e.g.
```
imagoiq, line 160:
```suggestion
```

imagoiq (changes requested):
1 issue and 1 question:
- When clicking anywhere on the popover, it closes it. So I'm not able to select or do right click on the link.
- Default behavior is that it closes when we click outside. Should we have a prop to be able to change that behavior?

gfellerph:
- [ ] Default positioning -> top centered
- [ ] add variant without arrow (default) and make it configurable
- [ ] update paddings and min height

Design: https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?type=design&node-id=18199%3A6302&mode=design&t=tKgyiB6TDkqu9sYU-1

imagoiq:
> 1 issue and 1 question:
> 
>     * When clicking anywhere on the popover, it closes it. So I'm not able to select or do right click on the link.
> 
>     * Default behavior is that it closes when we click outside. Should we have a prop to be able to change that behavior?

I still have the same issue in Firefox (it works on Chrome, Windows)

imagoiq (changes requested):
>     * When clicking anywhere on the popover, it closes it. So I'm not able to select or do right click on the link.

I still have the same issue in Firefox (it works on Chrome, Windows)

packages/components/src/components/post-popover/post-popover.tsx
```diff
@@ -0,0 +1,169 @@
...
+      // Handle missing re-focusing logic after close. Can be removed as soon as popovertarget works correctly
+      if (this.currentTarget) {
+        this.currentTarget.focus();
+        this.currentTarget = null;
+      }
+    }
+  }
+
+  // Fix for firefox to prevent the following lines from triggering
+  // https://github.com/oddbird/popover-polyfill/blob/main/src/popover.ts#L338
```
imagoiq, line 139: Can you detail why we need it, please? When I try the demos of oddbird/popover-polyfill, I have no problem, so I guess this has to do with cross-shadow-boundary?

packages/documentation/src/stories/components/popover/popover.stories.ts
```diff
@@ -0,0 +1,94 @@
...
+    <button class="btn btn-secondary btn-large" data-popover-target="${args.id}">
+      Click here to see a popover
+    </button>
+    <post-popover
+      class="hydrated bg-${args.backgroundColor}"
+      id="${args.id}"
+      placement="${args.placement}"
+      ?arrow="${args.arrow}"
+    >
+      ${unsafeHTML(args.innerHtml)}
```
imagoiq, line 88: If you don't wrap the content with an element, the selection of the text is limited in Firefox (and chrome, but differently). Triple click doesn't select all content in Firefox. In chrome, it will only select a sentence.

gfellerph: Done

packages/components/src/components/post-popover/post-popover.tsx
gfellerph: I have no idea what the issue is. Event bubbling through shadow dom is a bit tricky, that may be the cause. The standard API for triggering will also not support cross shadow root triggers (which is a shame: https://codepen.io/tuelsch/pen/LYMvmva?editors=1000).

package.json
```diff
@@ -14,6 +14,7 @@
   "scripts": {
     "bootstrap": "pnpm install && pnpm -r build",
     "start": "pnpm docs:start",
+    "start:clean": "pnpm bootstrap && pnpm start",
```
imagoiq, line 17: It shouldn't be part of this PR.

imagoiq (approved):

imagoiq (changes requested):
Tooltip doesn't work anymore on Firefox.

imagoiq (approved):

PR #2115, opened 2023-10-24
chore(styles): Run autoprefixer on scss files

imagoiq:
Currently, we autoprefix CSS compressed output files, but not SCSS files that can also be used through the packages.

### DIFF
--- a/.changeset/spicy-cougars-buy.md
+++ b/.changeset/spicy-cougars-buy.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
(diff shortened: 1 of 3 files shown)


gfellerph (approved):
Good catch! I was wondering why this did not work properly.

gfellerph:
@imagoiq, can you add a changeset? This might have impact on almost every project because most of the are importing sass directly instead of using the compiled CSS output. Something like

"Applied autoprefixer to distributed sass files. Sass files will now contain prefixes for supported browsers according to the browserslist file."

Issue #1201, opened 2023-03-01
Link figma design files to stories


gfellerph:
The `storybook-addon-designs` allows to display figma files in a storybook tab. Create a prototype integration and check what requirements have to be met for a smooth integration for all stories.

Addon: https://storybook.js.org/addons/storybook-addon-designs
Demo: https://pocka.github.io/storybook-addon-designs/?path=/story/docs-figma-examples--embed-file

```[tasklist]
### Tasks
- [ ] Install addon
- [ ] Create a proof of concept for one simple component (not buttons)
- [ ] Gather requirements for figma (do we need special frames containing the designs or can existing ones be used for integration, are all permissions correctly set, ...)
- [ ] Can the figma iframes be integrated into the docs page in a separate "Design" Tab?
```

b1aserlu:
Notes from Pattern Discussio Meeting 24.10.23

- Figma is not open to externals, links could lead to a lot of exernal people trying to access figma from outside the post
- Figma files can be published to the community, then it would be open and accessible
- Embeds have limited usefulness and performance
- Link between storybook and figma would be helpful, a link from figma to storybook is already being implemented


Resolution: put links to figma into storybook instead of embeds

b1aserlu:
Created a new Issue to link a first component to figma #2129

PR #2118, opened 2023-10-24
feat(documentation): Figma in Storybook examples

b1aserlu:
I made a couple examples of how figma can be integrated into storybook using the design addon from storybook. 
The examples are in the alert component under the posttab design.


b1aserlu:
Decision was reached not to use the desgin add on after all #1201

PR #2132, opened 2023-10-25
feat(documentation): Improve this page links should point to exact file on GitHub

davidritter-dotcom:


davidritter-dotcom:
Unfortunately i couldn't get the path to the file from the attachedCSFFile property because its always undefined so i still took it out of the storyIdToCSFFile property of the context.

packages/documentation/.storybook/blocks/footer.tsx
```diff
@@ -29,59 +29,69 @@ const DEVELOPERS: Developer[] = [
...
-            <a
-              className="btn-primary btn btn-rg btn-animated"
-              href="https://github.com/swisspost/design-system/issues"
-              target="_blank"
-              rel="noopener"
-            >
-              <span>Submit an issue</span>
-            </a>
+export default (params: { pathToStoryFile: String }) => (
+  <div>
```
alizedebray, line 39: As this `div` is not necessary (just there to wrap the content), you can then use [a fragment](https://react.dev/reference/react/Fragment). The difference is it is not rendered in the page in the end.
```suggestion
  <>
```

alizedebray (approved):

Issue #2060, opened 2023-10-10
Fix warning from new lint rules provided by @stencil-community/eslint-plugin


imagoiq:
Related to this change: https://github.com/swisspost/design-system/pull/2059

There are around 80 warnings.

PR #2143, opened 2023-10-26
chore(header): fix linting issues

gfellerph:
Fixes linting issues in the internet header package that came up with the new rules.


imagoiq (approved):
You can link it to https://github.com/swisspost/design-system/issues/2060 (even if it's not fixing all of them)

PR #2147, opened 2023-10-27
chore(icons): update packages

gfellerph:


oliverschuerch (commented):
Not sure if this is related, but icons can not be loaded in this PR anymore on the docs: https://preview-2147--swisspost-design-system-next.netlify.app/?path=/docs/icons-search-for-icons--docs

PR #2137, opened 2023-10-26
chore(*): Sync linting and IDE config

imagoiq:
### What

Commit essential linting and formatting config from our favourite IDE to the repository. The aim is to have the same config across the team and across IDE so we avoid dealing with changes that shouldn't be part of our PR.

* Add config for Intellij and vscode
  - Remove gitignore on .idea config
  - Disable prettier on MDX files
  - Use LF as default line separator
  - Enable format on save
* Run `prettier --write "**/*.{**/*,*}.{js,ts,jsx,tsx,scss,css,html,json}"`
* Remove prettier-ignore in mdx files

About MDX:
Disable formatting totally, as the maximum printWidth rule conflicts with MDX syntax about new line = `<p>`. 
Latest issue that we had: https://github.com/swisspost/design-system/pull/2128 . Issue on Prettier repository: https://github.com/prettier/prettier/issues/12209

If this wouldn't be enough, the second option is to ensure that our files are checked before committed with a git hook: https://github.com/lint-staged/lint-staged 

For review: you can do it commit by commit.

PR #2150, opened 2023-10-27
chore(styles): update dependencies and fix some things

gfellerph:
This PR is a little more that dependency updates:
- Updates all possible dependencies of the styles package
- Fixes new linting errors after the update
- Fixes unit tests not running on the styles package
- Fixes an issue with prettier after the update in the migrations


imagoiq (approved):
It would be great to first merge https://github.com/swisspost/design-system/pull/2137 (and the base https://github.com/swisspost/design-system/pull/2140) before as otherwise it will make potentially some merge issues.

Issue #1721, opened 2023-08-02
Stencil config: empty index files


gfellerph:
The internet header package has empty index files in the dist, e.g. `dist/esm/index.js` is empty safe for a link to sourcemaps.

```js


//# sourceMappingURL=index.js.map
```

The recommended package.json setup for main, module, etc. is linking to these files instead of files in the /loader folder. In order for these index files to work, they should actually export or contain the correct code.

For reference, the baloise Design System has a working setup: https://github.com/baloise/design-system/blob/main/packages/components/config/stencil.basic.config.ts (a bit more nested than our approach).

oliverschuerch:
Fixed with #2149

PR #2152, opened 2023-10-27
fix: add components to stencil `dist` outputtargets

oliverschuerch:


imagoiq (approved):
I'm not entirely sure that we should or need to do that according to my comment: https://github.com/swisspost/design-system/issues/1721#issuecomment-1673529767

But anyways, I'm ok with this. Please close this issue as well.

gfellerph:
@oliverschuerch, I'd put the documentation for bundlers first. I think it's the 80% use-case. Also not sure about the aliases in the import statements. What's the reason for those?

PR #2157, opened 2023-10-30
fix(components): add e2e timeout when trying to get component

imagoiq:
Some of the earliest tests are failing in watch mode, so extend timeout from default 4s to 30s to avoid this issue (like we are doing it for e2e in storybook).


gfellerph:
It's already an improvement, even if the tests are not green yet.

PR #2165, opened 2023-10-31
fix(intranet-header): use svg as default user image

alizedebray:


gfellerph (approved):
Can you add a changeset @alizedebray?

PR #2172, opened 2023-10-31
feat(documentation): adds new folder "forms" for form-control based components

b1aserlu:
Version From folder on the same level as the Components folder. Bootstrap does this as well https://getbootstrap.com/docs/5.3/forms/overview/

PS:
I am not sure if the following components: [radio-card, checkbox-card] would be considered part of forms or part of the normal components. I considered them as part of the forms group for now.


b1aserlu:
This option was discarded

PR #2175, opened 2023-11-01
feat(documentation): restructurs components folder now contains subfolders Forms, Ions, Internet and Intranet Header

b1aserlu:
Alternative Version to #2172 .
Was chosen and as discussed the headers and the icons documentation are now also subfolders of components.

image


imagoiq (commented):
Nice! This is much better. The only thing bothering me is that folders inside components are listed at the top. I would prefer them mixed with single file and that everything is orderer alphabetically.

b1aserlu:
To prevent manually maintaining order we could think about renaming "Getting Started" stories to "_Getting Started" or sth similar that way we could have the components folder sort without interferance. The way it is right now the previwe.ts fille needs to ne manually updated every time a story with a lower asci value than "Intranet Header" is added.

imagoiq (approved):

Issue #86, opened 2022-04-27
Select inputs with floating labels


gfellerph:
<p>Bootstrap v5 now provides floating labels, but the behavior of <a href="https://getbootstrap.com/docs/5.1/forms/floating-labels/#selects">selected inputs with a floating label</a>&nbsp;does not match our home version when it comes to the following use case:</p>

<p>With Angular, it happens that the select inputs do not have a selected value. In this case, the floating label should not be "floated".</p>

<p>It is necessary to add this behavior to both the&nbsp;basic&nbsp;and the custom select inputs.</p>

gfellerph:
Show floating label centered when:
- [ ] Standard select with no <option> elements
- [ ] ngb-dropdown when no option is pre-selected

PR #2200, opened 2023-11-01
feat(documentation): Add placeholder feature for floating-label select

imagoiq:
Add a way to see label as a placeholder for select element when `floating-label` is enabled.

Currently, we can use the `select:empty` selector, remove all options (so all the children of the select element) and repopulate them when the user click on the select. But it's quite cumbersome for the end developer. It's much easier to add a class. In the future, we'll be able to do that using `:has` instead of class.

Linked to this old issue: https://github.com/swisspost/design-system/issues/86

This issue fixes as well the color to match the input one.


packages/documentation/src/stories/components/select/select.docs.mdx
```diff
@@ -31,6 +31,13 @@ But note that the `<select>` must come first, so we can ensure the correct style
 
 <Canvas of={SelectStories.FloatingLabel} />
 
+#### With label as a placeholder
+
+If you use an empty option in your select and you want that your label acts as placeholder,
+you can add dynamically the class `.form-select-empty` to the `<select>` when there are no value defined or no option selected.
```
alizedebray, line 37:
```suggestion
you can dynamically add the class `.form-select-empty` to the `<select>` when there are no value defined or no option selected.
```

alizedebray (changes requested):
```
The label seems a off center:
![image](https://github.com/swisspost/design-system/assets/33580481/21294923-41d8-41aa-8b60-0c9e673d7f05)

```

imagoiq:
The snapshots look rather good, but it's quite hard to see as they encompass a lot of changes from other PRs.

alizedebray (approved):

PR #2205, opened 2023-11-02
feat(documentation): adds open graph tags to storybook header

b1aserlu:
There is an anomaly. On the frist page of the review the home--docs  the open graph tags seem to have no effect.


gfellerph (changes requested):
I think we should use a different image as the preview. The one used now has overflowing text: https://www.opengraph.xyz/url/https%3A%2F%2Fpreview-2205--swisspost-design-system-next.netlify.app%2F

Here's a collection of splash screen variants, recommended is an image in the format 1200x630.

**Figma** https://www.figma.com/file/MsZXoWnhUoV1tPg3HboSlm/Design-System-Splash-Screen?type=design&node-id=6439%3A734&mode=design&t=vXlCbLHZjMRdH4AE-1

b1aserlu:
I made the image be a reference to our github code repository becasue a relative path cannot be used in the url. This should work in theory. However it cannot yet work in this preview as there is no image in the location atm as it would be added by this PR as well.

gfellerph:
> I just realised It would probably better to convert the image to a datatype so it has a smaller file size an svg would be ideal but I do not know how to convert a png to a svg. I mean there are convertors out there but I do not know which ones I am allowed to use.

You can choose the file type when exporting from figma. I don't know if all services support svg, I'd stick to png. The png can be further optimised by using https://tinypng.com/.

packages/documentation/.storybook/manager-head.html
```diff
@@ -2,6 +2,35 @@
...
+<meta property="og:title" content="Swiss Post Design System" />
+<meta property="og:locale" content="en_US" />
+<meta
+  property="og:description"
+  content="The Swiss Post Design System pattern library for a unified and accessible user experience across the web platform."
+/>
+<meta property="og:url" content="https://next.design-system.post.ch/" />
+<meta
+  property="og:image"
+  content="https://github.com/swisspost/design-system/packages/documentation/public/images/DesignSystemPreview1280x640.png"
```
gfellerph, line 16: It would be better use https://next.design-system.post.ch/images as image source instead of GitHub.

gfellerph (changes requested):

packages/documentation/.storybook/manager-head.html
imagoiq: @b1aserlu I don't think that the resource is that critical, so it's not a blocker that it will work only when released. What could be more an issue is that we forget to change the URL when we'll change next.design-system.post.ch to design-system.post.ch. Maybe you could have a look, how to inject our base URL. It could be, for example, with environment variables. Can you think of alternative ways?

b1aserlu: I try it with JS now. But I need the preview to work so I can check if it works.

imagoiq: Solving it with JS might be a problem for bots that will try to parse the HTML for this information. I'm having doubt that it will work.

b1aserlu: You are right, it did not work. But I am out of useful ideas how to do it.

gfellerph (changes requested):
Looks good so far, but the splash screen has been changed by Rouven to reflect the accent colors change. Can you update the splash screen from the figma file?

gfellerph (approved):

PR #2208, opened 2023-11-02
chore: revert internet header output targets

gfellerph:

PR #2071, opened 2023-10-13
feat: add components-angular as output-target from components package

oliverschuerch:
Renamed the previous `components-angular` package, which was used as angular workspace for the `@swisspost/design-system-intranet-header` package to `intranet-header-workspace`.

- [x] depending on #2072
- [x] depending on #2214
- [x] create tickets to add components tests (#2215)

PR #2214, opened 2023-11-03
fix: stencil empty output files

oliverschuerch:
This time without the breaking changes, which we had to revert the last time ;) #2208

### DIFF
--- a/packages/documentation/src/stories/getting-started/components/components.docs.mdx
+++ b/packages/documentation/src/stories/getting-started/components/components.docs.mdx
@@ -1,7 +1,10 @@
 import { Meta, Source } from '@storybook/blocks';
 import * as GettingStartedStories from './components.stories';
-import SampleUseAll from './components-use-all.sample?raw';
-import SampleUseSpecific from './components-use-specific.sample?raw';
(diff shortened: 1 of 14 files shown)


packages/documentation/src/stories/getting-started/components/components.docs.mdx
gfellerph, line 1: It would be nice to re-structure this page a little and maybe use tabs. This would clear things up a bit and I think 95% of use cases will be for the framework integration.

```
__________________________________________________
| Framework integration | JavaScript integration |______________
```

gfellerph (changes requested):

packages/documentation/src/stories/getting-started/components/components.docs.mdx
oliverschuerch: This is coming with the PR #2071

packages/internet-header/stencil.config.ts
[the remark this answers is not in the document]
oliverschuerch: Added it back in both, internet-header and components package

gfellerph (approved):
Nice work!

PR #2030, opened 2023-10-03
feat(documentation): adds container stories

b1aserlu:

PR #2238, opened 2023-11-07
feat(styles): deprecate regular breakpoint

oliverschuerch:
- [x] depends on #2030


packages/documentation/src/stories/foundation/layout/grid/grid.blocks.tsx
```diff
@@ -1,122 +1,60 @@
...
+      <th>Gutter width</th>
+      {forEach(
+        SCSS_VARIABLES.grid,
+        (data: { key: string; value: { ['gutter-width']: string } }) => (
+          <td key={data.key}>{data.value['gutter-width']}</td>
+        ),
+      )}
+    </tr>
+
+    <tr>
```
alizedebray, line 30: I don't really understand these 4 lines: 
image

oliverschuerch: @alizedebray which 4 lines are we talking about exactly?

alizedebray (approved):

PR #2258, opened 2023-11-10
chore: zip build artifacts

gfellerph:
This change will centrally handle zipping and unzipping of build artifacts for faster upload/download. There is one catch to this PR: the download action always runs on the main branch (for security reasons) and cannot be tested, therefore the documentation publishing workflow fails for this PR. To actually test it, we'd need to merge this PR and push a change to the documentation.

### DIFF
--- a/.github/actions/artifact-download/action.yaml
+++ b/.github/actions/artifact-download/action.yaml
@@ -49,6 +49,7 @@ runs:
         workflow_conclusion: success
 
     - name: Unzip artifacts
+      shell: bash
       run: unzip artifacts.zip
 
     - name: Create outputs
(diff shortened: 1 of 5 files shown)


imagoiq (approved):

.github/actions/artifact-download/action.yaml
```diff
@@ -48,6 +48,10 @@ runs:
         run_id: ${{ github.event.workflow_run.id }}
         workflow_conclusion: success
 
+    - name: Unzip artifacts
+      shell: bash
+      run: unzip artifacts.zip
```
imagoiq, line 53: `unzip` might need to be installed. Or is it already included somewhere?

PR #2267, opened 2023-11-13
fix(documentation): edit on Github breaks page without stories attached

imagoiq:
Fix Changelog and search for icons page after https://github.com/swisspost/design-system/pull/2132. If `storyIdToCSFFile` is not defined for example on docs page without stories attached, it breaks the whole page.
image


alizedebray (approved):

alizedebray:
The changeset is missing

PR #2270, opened 2023-11-13
chore: add peer dependencies to the style package

alizedebray:


packages/demo/src/app/home/home.component.html
```diff
@@ -290,7 +290,7 @@ <h4 class="h5 bold">Automatic Migration ⚙️</h4>
                 appCopyToClipboard
                 class="d-block mt-1 p-3"
                 [languages]="['bash']"
-                highlight="npx ng update @swisspost/design-system-styles --from=5 --to=6 --migrate-only"
+                highlight="npx ng update @swisspost/design-system-migrations --from=5 --to=6 --migrate-only"
```
imagoiq, line 293: Can you a line somewhere to install/update the package as well?

imagoiq (approved):
Looks good

packages/demo/src/app/home/home.component.html
alizedebray: This package should not be installed into projects, I wrote about that is the README: https://github.com/swisspost/design-system/pull/2270#discussion_r1392017815.

imagoiq: Ah true, good!

packages/documentation/package.json
```diff
@@ -39,32 +39,32 @@
...
+    "lit": "3.0.2",
+    "lit-html": "3.0.2",
     "prettier": "2.8.8",
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "react-syntax-highlighter": "15.5.0",
     "rimraf": "5.0.5",
     "sass": "1.69.5",
-    "storybook": "7.5.3",
+    "storybook": "7.5.2",
```
imagoiq, line 67: Are all those downgrades (across all packages) desired?

packages/styles/package.json
```diff
@@ -18,30 +18,30 @@
...
+    }
   },
   "dependencies": {
-    "@angular-devkit/core": "=15.0.4",
-    "@angular-devkit/schematics": "=15.0.4",
-    "@angular/core": "=15.0.4",
     "@popperjs/core": "2.11.8",
     "bootstrap": "5.3.2",
     "cheerio": "1.0.0-rc.12",
     "prettier": "3.1.0"
```
gfellerph, line 42: These dependencies are also connected to the migrations, prettier for formatting output and cheerio for finding elements. I'm not sure if they are needed in the styles anymore, especially cheerio. Prettier could become a dev dependency so it's not delivered to projects.

packages/styles/package.json
```diff
@@ -18,30 +18,30 @@
...
   },
   "devDependencies": {
-    "@angular/compiler": "=15.0.4",
     "@percy/cli": "1.27.4",
     "@percy/cypress": "3.1.2",
```
gfellerph, line 46: These could be cleaned up as well, when we're already at it.

packages/styles/package.json
```diff
@@ -74,9 +74,5 @@
     ],
     "outputDir": "./dist",
     "tscOutDir": "./out-tsc/lib"
```
gfellerph, line 76:
```suggestion
```

packages/styles/tsconfig.json
gfellerph, line 1: This file can be deleted/moved to the schematics package, there's no more typescript in the styles. The cypress folder and the cypress config file can be removed as well.

gfellerph (changes requested):
Awesome clean up, loving it.

packages/styles/tsconfig.json
alizedebray: I just cleaned up the dependencies.

imagoiq (approved):

alizedebray:
I added "packages/migrations/src/migrations/**/*" in the Duplication Exclusions of SonarCloud.

PR #2273, opened 2023-11-14
2134 button group do not have focus style for checkbox/radio variants

davidritter-dotcom:
The solution works for the checkbox. For the radio variant, since with tab you can only focus the first element of the radio button group, it also only shows the focus style of the first label. If you want to focus the others, you have to use the arrow keys. And if I would put all the radio buttons in a seperate group, we would loose the whole functionality of the component.

### DIFF
--- a/packages/styles/src/components/button-group.scss
+++ b/packages/styles/src/components/button-group.scss
@@ -51,7 +51,7 @@
   position: absolute;
   clip: rect(0, 0, 0, 0);
   pointer-events: none;
-  &:focus-visible + .btn {
+  &:is(:focus-visible, :focus-within, .pretend-focus) + .btn {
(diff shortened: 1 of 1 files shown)


imagoiq:
> The solution works for the checkbox. For the radio variant, since with tab you can only focus the first element of the radio button group, it also only shows the focus style of the first label. If you want to focus the others, you have to use the arrow keys. And if I would put all the radio buttons in a seperate group, we would loose the whole functionality of the component. At least for chrome its like that, in firefox it works.

@davidritter-dotcom I'm not sur what you want to say with your description. Everything looks normal to me. Also, you can tab through radio buttons if none is checked.

imagoiq (approved):

gfellerph (approved):

PR #2280, opened 2023-11-17
chore(internet-header,docs): improve and document the internet header main slot

alizedebray:


packages/internet-header/src/components/post-internet-header/post-internet-header.scss
```diff
@@ -108,6 +109,23 @@
...
+  align-items: center;
+  overflow: hidden;
+  flex-shrink: 1;
+
+  ::slotted(*) {
+    padding-inline: 1rem;
+
+    @include mixins.min(lg) {
+      padding-inline: 1.125rem;
+    }
```
gfellerph, line 128: I'm not sure that this is a good idea. The responsibility for the slot is by the project. If somebody puts two sibling elements into the slot, both will have the padding. That might not be desired.

A better approach would be to document what we use for padding, so people know how to re-create the look in their custom integration.

packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
```diff
@@ -505,6 +511,11 @@ export class PostInternetHeader {
               </If>
             </post-main-navigation>
             <div class="main-navigation-controls">
+              <If condition={!this.isMainSlotEmpty}>
+                <div class="main-navigation-custom-content">
+                  <slot name="main" onSlotchange={e => this.handleMainSlotChange(e)}></slot>
+                </div>
+              </If>
```
gfellerph, line 518: This seems suspicious, because it won't render the slot if it's empty, but how do you know it's empty if it's not there? While it's working with the current implementation, I'm not sure what will happen if the content comes from Angular, meaning it's rendered client side and maybe after the header initialized.

gfellerph (changes requested):

packages/internet-header/src/components/post-internet-header/post-internet-header.scss
alizedebray: I added this default padding because 1.125rem (18px) is used for the KLP login button but not reproducible with our utility classes whereas `.mx-0` is. 

Just removed it but should we document this exact 1.125rem value, update the KLP button or create a class specific to the header buttons?

gfellerph: I think we'd need to update the button padding. Kaloyan documented the component in Figma and used a different padding and that one looks better IMO. But that can be handled in a separate ticket. Documenting the padding for others would be a good idea.

packages/internet-header/src/components/post-internet-header/post-internet-header.scss
```diff
@@ -72,7 +72,8 @@
   display: flex;
 
   & > .main-navigation-controls {
-    flex: 0 0 auto;
+    flex: 0 1 auto;
+    overflow: hidden;
```
gfellerph, line 76: Not sure if this is necessary, slotted content does not seem to be affected by it anyways (tried to apply `position: absolute;top: 200px` to the `<p>` and it was still visible).

```suggestion
```

packages/internet-header/src/components/post-internet-header/post-internet-header.scss
```diff
@@ -108,9 +109,18 @@
...
     border-left: 1px solid color.$gray-10;
+    flex: 0 0 auto;
   }
 }
 
+.main-navigation-custom-content {
+  max-height: var(--header-height);
+  display: flex;
+  align-items: center;
+  overflow: hidden;
```
gfellerph, line 120: Same here, also projects will want to add dropdowns and stuff.

```suggestion
```

packages/documentation/src/stories/internet-header/components/header/header.stories.ts
```diff
@@ -217,3 +215,14 @@ export const OSFlyoutOverrides: Story = {
...
+
+export const CustomContent: Story = {
+  args: {
+    innerHTML: `<p class="mx-regular mb-0" slot="main">Hello, User!</p>`,
+  },
+  parameters: {
+    controls: {
+      include: ['innerHTML'],
+    },
+  },
```
gfellerph, line 227: I tried adding custom content a bit, re-styling the `<p>` etc. and the preview breaks fairly often, because while writing, the HTML is invalid and starts to throw errors. Also copy/pasting does not work always.

I'm not sure if we should actually show this control, in this state it adds little value and fixing it might be a rabbit hole.

gfellerph (changes requested):

packages/internet-header/src/components/post-internet-header/post-internet-header.scss
alizedebray: I created the ticket #2390

gfellerph (approved):
I think this is good to go now. Please have a look at the pnpm-lock.yaml, this PR seems to reset some dependencies to previous versions. Please update from main before merging.

PR #2290, opened 2023-11-21
feat(documenation): add since v1 label on components

davidritter-dotcom:
Added label in all the docs of the components from the components package (accordion, alert, collapsible, tabs, tooltip, icon).


gfellerph (changes requested):
I think it would be better to put the tag up with the others. This way it does not push the main content down.

image

See https://github.com/swisspost/design-system/blob/main/packages/documentation/.storybook/helpers/bages-config.ts for adding new badges.

gfellerph (changes requested):
Please also add a changeset for the documentation to inform users about why this label was added.

gfellerph (approved):

PR #2294, opened 2023-11-22
fix(documentation): remove unneeded badge size class

imagoiq:
###

Remove unneeded badge size class at wrapper level to match https://design-system.post.ch/#/bootstrap-samples/badge

#### Before

image

#### After

image


alizedebray (changes requested):
The "Size" control no longer works when "Interaction Type" is "None".

alizedebray (approved):

Issue #2122, opened 2023-10-24
Collapsible shows focus outline when interacting with a pointer


gfellerph:
<img width="1065" alt="image" src="https://github.com/swisspost/design-system/assets/1659006/d625e787-375e-440b-ba46-cdaa5352c4dd">

When toggling the collapsible with a mouse pointer, the focus outline should not be visible.

gfellerph:
Possible duplicate of #1087

PR #2310, opened 2023-11-27
chore(styles): update accordion

alizedebray:


imagoiq (approved):
Looks good. The only remaining part is https://github.com/swisspost/design-system/issues/2122. Perhaps you want to address it in this PR as well?

PR #2312, opened 2023-11-28
feat(styles, docs): Update select styles

imagoiq:


packages/styles/src/variables/components/_forms.scss
```diff
@@ -74,7 +81,9 @@ $input-height-content-lg: type.$font-size-medium * $input-line-height-lg !defaul
 $input-height-inner-lg: $input-height-content-lg + ($input-padding-y-lg * 2) !default; // Design System
 $input-height-lg: calc(#{$input-height-inner-lg} + #{$input-height-border}) !default;
 
-$input-transition: border-color 0.15s ease-in-out, box-shadow 0.15s ease-in-out !default;
+$input-transition:
+  border-color 0.15s ease-in-out,
+  box-shadow 0.15s ease-in-out !default;
```
imagoiq, line 86: Just some linting as in some other lines

packages/documentation/src/stories/components/select/select.stories.ts
gfellerph, line 1: You could add `success: false` to the args settings, then there will be a switch by default instead of this button "set boolean".

packages/styles/src/components/form-select.scss
```diff
@@ -1,20 +1,64 @@
...
+  }
+
+  &:disabled {
+    background-image: b.escape-svg(form-select.$form-select-indicator-disabled),
+      var(--bs-form-select-bg-img), var(--bs-form-select-bg-icon, none);
+  }
+
+  &.success {
+    border-color: var(--post-success-green);
+    background-image: b.escape-svg(form-select.$form-select-indicator-success),
```
gfellerph, line 51: image

The distance between the two icons is not big enough. As they look really similar it's confusing to look at.

gfellerph (changes requested):
Looks good! Just small things to fix.

gfellerph (approved):

PR #2349, opened 2023-12-04
fix(docs): Reapply styles for vertical rhythm

imagoiq:
### What

- Fixed rules for vertical rhythm that didn't apply anymore since Storybook v7.
- A couple of fixes about the documentation (link, heading).
- Removed padding-bottom on tabs content.


packages/styles/src/components/tabs/_tabs-wrapper.scss
```diff
@@ -56,7 +56,6 @@
 
 .tab-content {
   padding-top: spacing.$spacer;
-  padding-bottom: spacing.$spacer;
```
imagoiq, line 59: This one is important. I'm not sure why it was added, but in my case it adds unwanted spacing. We should either remove it completely and let the user be responsible for the spacing, or convert it to a margin so that it collapses with margin-top of the next element and don't disturb vertical rhythm.

alizedebray: As per design, we want the accordion content to have a default spacing. But sure it would be better with a `margin-block` instead of these padding

PR #2356, opened 2023-12-04
fix(styles): Regression: padding-left for labels

imagoiq:


imagoiq:
Looks alright for switch: https://percy.io/3f971362/design-system/builds/31518670/search/preview/1735126356?browser=safari&searchParam=switch&viewLayout=side-by-side&viewMode=new&width=600

packages/styles/src/components/form-check.scss
```diff
@@ -109,7 +109,10 @@
 
   &-label {
     flex: 1;
-    padding-inline-start: form-check.$form-check-column-gap;
+
+    :not(.form-switch) > & {
```
gfellerph, line 113:
```suggestion
    :not(.form-switch, .radio-button-card, .checkbox-button-card) > & {
```

gfellerph (changes requested):

packages/styles/src/components/form-check.scss
imagoiq: It's becoming a bit convoluted, but at least the intend is clear. Maybe we should not have a `.form-check` that is shared with `.form-switch`

gfellerph (approved):

PR #2358, opened 2023-12-05
fix(styles): changes success-green to be treated as a dark background color

b1aserlu:


packages/styles/src/functions/_contrast.scss
```diff
@@ -48,6 +48,15 @@
 }
 
 @function light-or-dark($background, $light-color: #fff, $dark-color: #000) {
+  /*
+  exception for Success-color issue #2227
+  the color is statically defined because the reference to the variable from here
+  would create a loop
+  */
```
imagoiq, line 55: Can you add a comment on the $success color about this virtual link so that we now that we have to change this static value here as well if we change the value of the success color.

imagoiq (changes requested):

imagoiq (approved):

PR #2369, opened 2023-12-06
chore: dry run publish before starting to release

gfellerph:
This will prevent changesets from starting to release packages and then failing because of an invalid setup on one of the packages. Early failure will prevent some packages from being published without the complete workflow to finish (pushing git tags, notifying teams channel).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -31,7 +31,6 @@ jobs:
       # This will fail the build if something in the publish setup is not correct
       # before changeset magic is starting to run
       - name: Test publish step
-        id: publishtest
         run: pnpm -r publish --dry-run
 
       # The changeset action will behave differently based on whether there are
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -28,6 +28,12 @@ jobs:
       - name: Build changelog formatter
         run: pnpm --filter design-system-changelog-github build
 
+      # This will fail the build if something in the publish setup is not correct
+      # before changeset magic is starting to run
+      - name: Test publish step
+        id: publishtest
```
imagoiq, line 34: Just for my own curiosity, for what the id is used?

imagoiq (approved):

.github/workflows/release.yaml
gfellerph: It's a leftover. You can assign IDs to steps so you can refer to outputs from that step. I first thought that I'd need to check the output of the test but the dry run just errors if anything is wrong and that error will stop the whole action, so it's actually not necessary there.

PR #2371, opened 2023-12-07
fix(documentation): Vite XSS vulnerability in 4.4.7

imagoiq:
Run 

```
pnpm dedupe
pnpm update "@storybook/*
```

to update storybook to latest peer dependency, especially [vite to v5 which is supported since Storybook 7.5.](https://storybook.js.org/blog/storybook-7-5/) We may see also better performance


gfellerph (approved):
Awesome, thanks

PR #2376, opened 2023-12-07
feat(documentation): adds contrastValues and an example text for colors story

b1aserlu:
Different than suggested in the ticket I used the actual function that is used in the design system to calculate the contrast.
In Terms of presentation I do not have a good idea how it should be done.

There is however already a page that shows the proper text for each backround.
https://next.design-system.post.ch/?path=/docs/utilities-background--docs
Maybe this would also be the more appropriate place to add the contrast information.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -34,20 +34,11 @@ jobs:
       - name: Install & build dependencies
         run: pnpm bootstrap
 
-      - name: Prepare Cypress – Spawn Xvfb server
-        run: Xvfb -screen 0 1024x768x24 :99 &
-
-      - name: Prepare Cypress – Share Xvfb server as environment variable
-        run: export DISPLAY=:99
-
       - name: Cypress info
         run: pnpm --filter design-system-documentation exec cypress info
 
       - name: Run tests
         run: pnpm e2e:ci
-
-      - name: Post Cypress – Stop Xvfb server
-        run: pkill Xvfb
 
       - name: Upload screenshots
         uses: actions/upload-artifact@v4.3.0
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "test": "pnpm -r test",
     "unit": "pnpm -r unit",
     "e2e": "start-server-and-test docs:headless 9001 'pnpm -r --stream e2e'",
-    "e2e:ci": "start-server-and-test docs:headless 9001 'pnpm --filter \"...[origin/main]\" --stream e2e'",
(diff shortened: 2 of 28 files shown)


packages/documentation/src/stories/foundation/color/color.docs.mdx
```diff
@@ -17,10 +17,8 @@ They theme our elements and components.
 
 <div className="alert alert-info alert-md mt-regular">
   For accessibility reasons, you should always make sure that your individually created color combinations are
-  sufficiently contrasting. There may be a better solution than coloring elements individually.
+  sufficiently contrasting. In the colour tiles below, we have inserted the corresponding contrast ratios against white and black.
```
oliverschuerch, line 20:
```suggestion
  sufficiently contrasting. In the color tiles below, we have inserted the corresponding contrast ratios against white and black.
```

packages/documentation/src/stories/foundation/color/color.module.scss
```diff
@@ -1,5 +1,25 @@
+@use 'sass:map';
+@use 'sass:list';
 @use '@swisspost/design-system-styles/core' as post;
 
+$all_exported_colors: () !default;
+$all_exported_colors: map-merge(
```
oliverschuerch, line 6:
```suggestion
$all_exported_colors: map.merge(
```
Please update all the other occurences as well.

oliverschuerch (changes requested):

b1aserlu:
found an oversight from #2358 where the exeption was created for the wrong success color.

If we didnt have a PR affected by it before we definitely have now.
image

packages/documentation/src/stories/foundation/color/color.docs.mdx
```diff
@@ -49,6 +53,13 @@ with a specific background color and the best possible contrast color for its co
 
 ## Signal colors
 
+<PostAlert type="info">
+<p slot="heading">Why does `Success-green` have white text?</p>
+<p>
+  The text of “Success-green” is white, despite the displayed contrast ratios, because with WGAC 3.0, the contrast with white is significantly greater, while it hardly differs in WCAG 2.0.
+  </p>
+</PostAlert>
+
```
oliverschuerch, line 62: This Message can be removed, as it does not make sense to explain here, why our background utility works like this. Do not add it in the background utility story as well, because we do not show contrast values there.
And as soon as we implement the new contrast algorithm, the contrast values will change anyway.

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #2379, opened 2023-12-07
feat(components): split post-collapsible and post-accordion-item

alizedebray:


packages/components/src/components/post-accordion-item/post-accordion-item.scss
```diff
@@ -0,0 +1,22 @@
...
+
+  > ::slotted(.text-truncate) {
+    display: block;
+  }
+}
+
+post-accordion-item::part(accordion-item) {
+  @include accordion-mx.background-color;
+
+  post-accordion-item + & {
```
gfellerph, line 19:
```suggestion
  & + post-accordion-item {
```

packages/components/src/components/post-accordion-item/post-accordion-item.tsx
```diff
@@ -0,0 +1,90 @@
...
+  connectedCallback() {
+    this.validateHeadingLevel();
+  }
+
+  componentWillLoad() {
+    this.isOpen = !this.collapsed;
+  }
+
+  componentWillRender() {
+    this.id = this.host.id || `a${crypto.randomUUID()}`;
```
gfellerph, line 48: Can you move this to the componentWillLoad? It evaluates on every re-render.

https://stenciljs.com/docs/component-lifecycle

packages/components/src/components/post-accordion-item/post-accordion-item.tsx
```diff
@@ -0,0 +1,90 @@
...
+    this.headingTag = `h${this.headingLevel ?? 2}`;
```
gfellerph, line 49: This does not need to be a state if you move the assignment into the render function.

packages/components/src/components/post-accordion-item/post-accordion-item.tsx
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  @Listen('collapseChange')
+  onCollapseChange(): void {
+    this.isOpen = !this.isOpen;
```
gfellerph, line 54: Are you sure that a collapse change always inverts the state? Does the collapseChange event not send the actual current state as a parameter? Would be a good idea for stability to always assign what the event sends.

Also, how does this behave on nested accordions? Is the collapseChange event bubbling?

packages/components/src/components/post-accordion-item/post-accordion-item.tsx
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  /**
+   * Triggers the collapse programmatically.
+   */
+  @Method()
+  async toggle(open?: boolean): Promise<boolean> {
+    return this.collapsible.toggle(open);
+  }
```
gfellerph, line 63: Just a nitpick, it's usually called `force`. https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute

```suggestion
  async toggle(force?: boolean): Promise<boolean> {
    return this.collapsible.toggle(force);
  }
```

packages/styles/src/mixins/_accordion.scss
gfellerph, line 1: The accordion can be white on white as well https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?type=design&node-id=42%3A358&mode=design&t=QZ4xXVf9Ro4HqV1X-1

Not sure if we should pre-define the backgrounds like this or just use classes and offer all variations.

```html
<div class="bg-white">
  <post-accordion class="bg-light"></post-accordion>
</div>

<div class="bg-light">
  <post-accordion class="bg-white"></post-accordion>
</div>

**Review by gfellerph — CHANGES_REQUESTED** [review]

File: packages/styles/src/mixins/_accordion.scss
Comment (alizedebray) [reply packages/styles/src/mixins/_accordion.scss:1]: It shouldn't be white on white: 
image

File: packages/components/src/components/post-accordion-item/post-accordion-item.scss (hunk 1 above)
Comment (alizedebray) [reply packages/components/src/components/post-accordion-item/post-accordion-item.scss:19]: I want `post-accordion-item + post-accordion-item::part(accordion-item)` not `post-accordion-item::part(accordion-item) + post-accordion-item`, it is the `accordion-item` part that should not have a border.

File: packages/components/src/components/post-accordion-item/post-accordion-item.tsx (hunk 4 above)
Comment (alizedebray) [reply packages/components/src/components/post-accordion-item/post-accordion-item.tsx:49]: For now, the event has no payload and is sent only when there is an actual change in the collapse. Should this be change?

 image

Also, I stopped the event propagation so that nested accordions does not interfere.

Comment (gfellerph) [reply packages/components/src/components/post-accordion-item/post-accordion-item.tsx:49]: I would send the current state, this could be very helpful for other applications as well.

File: packages/styles/src/mixins/_accordion.scss
Comment (gfellerph) [reply packages/styles/src/mixins/_accordion.scss:1]: True. In this case, using classes does not make any sense.

File: packages/components/src/components/post-accordion-item/post-accordion-item.tsx (hunk 4 above)
Comment (alizedebray) [reply packages/components/src/components/post-accordion-item/post-accordion-item.tsx]: I just updated the event.

**Comment by alizedebray** [comment]
@Cian77 Could you have a look to the accordion?
https://preview-2379--swisspost-design-system-next.netlify.app/?path=/docs/components-accordion--docs

File: .changeset/flat-parents-refuse.md [hunk 6]
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-components': patch
+'@swisspost/design-system-styles': patch
+---
+
+Update the background color of accordion to be white on any background color other than white. The accordion background remain gray on white backgrounds.
```
gfellerph, line 6:
```suggestion
Update the background color of accordion to be white on any background color other than white. The accordion background remains gray on white backgrounds.
```

gfellerph (approved):

PR #2389, opened 2023-12-11
fix(pnpm-lock.yaml): Vite not importing stencil entry files

imagoiq:


imagoiq:
Didn't work. In local it's working, but on ci with the exact same command `pnpm --filter design-system-documentation... build`, it has another issue: https://github.com/swisspost/design-system/actions/runs/7168785970/job/19517718198#step:5:89

PR #2391, opened 2023-12-11
feat(styles): Update valid state style for input text

imagoiq:


packages/styles/src/components/form-validation.scss
```diff
@@ -90,6 +90,14 @@
   display: block;
 }
 
+.success {
```
alizedebray, line 93: Why did you create a new class instead of updating the `.is-valid` styles directly?

packages/styles/src/variables/components/_form-validation.scss
```diff
@@ -10,6 +10,7 @@ $form-feedback-font-size: type.$font-size-12 !default;
...
 $form-feedback-invalid-color: color.$error !default;
 $form-feedback-icon-valid: url("data:image/svg+xml,%3C%3Fxml version='1.0' encoding='UTF-8'%3F%3E%3Csvg xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink' width='24pt' height='24pt' viewBox='0 0 24 24' version='1.1'%3E%3Cg id='surface1'%3E%3Cpath style=' stroke:none;fill-rule:nonzero;fill:rgb(0%25,50.588235%25,22.745098%25);fill-opacity:1;' d='M 9.5 18.398438 L 3.800781 12.699219 L 5.199219 11.300781 L 9.5 15.597656 L 18.800781 6.300781 L 20.199219 7.699219 Z M 9.5 18.398438 '/%3E%3C/g%3E%3C/svg%3E%0A") !default;
+$form-feedback-icon-success: url("data:image/svg+xml,%3C%3Fxml version='1.0' encoding='UTF-8'%3F%3E%3Csvg xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink' width='24pt' height='24pt' viewBox='0 0 24 24' version='1.1'%3E%3Cg id='surface1'%3E%3Cpath style=' stroke:none;fill-rule:nonzero;fill:rgb(0%25,50.588235%25,22.745098%25);fill-opacity:1;' d='M 9.5 18.398438 L 3.800781 12.699219 L 5.199219 11.300781 L 9.5 15.597656 L 18.800781 6.300781 L 20.199219 7.699219 Z M 9.5 18.398438 '/%3E%3C/g%3E%3C/svg%3E%0A") !default;
```
alizedebray, line 13: Isn't is the same SVG as `$form-feedback-icon-valid`?

packages/documentation/src/stories/components/forms/input/input.stories.ts
```diff
@@ -98,11 +99,7 @@ const meta: Meta = {
     },
     size: {
       name: 'Size',
-      description: "Sets the size of the component's appearance.",
-      if: {
```
alizedebray, line 102: This condition still seems necessary: the sizes do not apply to floating labelled input.

packages/styles/src/components/form-validation.scss
imagoiq: Good question. I separated both to match select style as in https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/branch/MVANE2KzNGq2et3q6T15PC/Components-Post?node-id=7802%3A33963&mode=dev
It's true that in the input mockup there is no such indication though.

imagoiq: Good question. I separated both to match select style as in https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/branch/MVANE2KzNGq2et3q6T15PC/Components-Post?node-id=7802%3A33963&mode=dev
It's true that in the input mockup there is no such indication though.

packages/documentation/src/stories/components/forms/input/input.stories.ts
imagoiq: Good catch, it will be part of the next PR

packages/styles/src/components/form-validation.scss
alizedebray: Select styles were updated, there is now only one success state which should probably be handled with the `.is-valid` class.

alizedebray (changes requested):

packages/styles/src/components/form-validation.scss
imagoiq: Ok, thanks for your investigation.

.changeset/eleven-ghosts-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Update valid state style for input text.
```
alizedebray, line 5:
```suggestion
Updated styles for the `.is-valid` class.
```

alizedebray (approved):

PR #2394, opened 2023-12-12
feat(components): updates tooltip component

b1aserlu:


.changeset/swift-tips-kiss.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': minor
+'@swisspost/design-system-components': minor
+---
+
+Update tooltip minHeight and maxWidth and added a new property arrow to tooltip that defines wheter or not the pointer arrow is displayed.
```
oliverschuerch, line 6:
```suggestion
Updated the tooltip `min-height` and `max-width` and added a new property arrow to the tooltip, that defines wheter or not the pointer arrow is displayed.
```

packages/components/src/components/post-tooltip/post-tooltip.scss
oliverschuerch, line 1: What about the so defined padding update? Is everything done as it should be?

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
```diff
@@ -75,6 +75,7 @@ function render(args: Args) {
       id="${args.id}"
       class="hydrated bg-${args.backgroundColor}"
       placement="${ifDefined(args.placement)}"
+      arrow="${ifDefined(args.arrow)}"
```
oliverschuerch, line 78: You should import the `nothing` object from lit, set `args.arrow` to `false` in the meta object and then define the arrow binding like this:
`arrow="${args.arrow ? args.arrow : nothing}"`
This will make the arrowbinding appear only if `arrow` is set to `true` and show the false/true toggler in the storybook control table from the very beginning.

oliverschuerch (changes requested):

packages/components/src/components/post-tooltip/post-tooltip.scss
b1aserlu: The padding was already correct

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -25,6 +25,11 @@ export class PostTooltip {
    */
   @Prop() readonly placement?: Placement = 'top';
 
+  /**
+   * Wheter or not to display a little pointer arrow
+   */
+  @Prop() readonly arrow?: boolean = false;
```
oliverschuerch, line 31:
```suggestion
  @Prop() readonly arrow?: boolean = true;
```

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
```diff
@@ -17,6 +17,7 @@ const meta: Meta = {
     innerHTML: 'Hi there 👋',
     backgroundColor: 'primary',
     placement: 'top',
+    arrow: false,
```
oliverschuerch, line 20:
```suggestion
    arrow: null,
```

If the default of the component is "true", not setting it via story attributes or setting it to "true", should result in a value of "true" (arrow is visible).
If we set the arrow property to "false" the arrow should not be visible.

oliverschuerch (changes requested):

oliverschuerch (approved):

Issue #2399, opened 2023-12-13
Remove grid container helper classes


b1aserlu:
Remove the following in #2041 deprecated classes.

```[tasklist]
### Tasks
- [ ] Remove `.vertical-gutters` (use gutter classes instead, e.g. `.g-*`, `.gy-*` or `.gx-*`)
- [ ] Remove `.row.border-gutters`
- [ ] Remove `.container-reset`
- [ ] Remove `.container-reset-left` and `.container-reset-right`
- [ ] Remove `.container-fluid-#{$breakpoint}`
```

PR #2400, opened 2023-12-13
chore(styles): adds comments for deprecation

b1aserlu:
Also created the new issue #2399 that will remove the deprecated classes.

Wait for #2503 to me merged.


packages/styles/src/components/grid.scss
```diff
@@ -89,7 +89,9 @@ $gutter-x-cache: 0;
 }
 
 // legacy code
```
oliverschuerch, line 91:
```suggestion
```

oliverschuerch (changes requested):
Add a changeset, to inform users about the deprecations.

oliverschuerch (approved):

PR #2402, opened 2023-12-13
fix(demo): Links to new icons documentation are broken

imagoiq:


packages/demo/src/app/post-sample/components/icons/icons-demo-page/icons-demo-page.component.html
oliverschuerch, line 1: To me it looks, like we should switch the links:
brand new icons documentation: links over to the icon search, not to the getting started page
discover the new icons: links over to the getting started page, not to the icon search.

What do you think?

oliverschuerch (changes requested):

packages/demo/src/app/post-sample/components/icons/icons-demo-page/icons-demo-page.component.html
imagoiq: Here it is

oliverschuerch (approved):

Issue #1931, opened 2023-09-07
[Notification-overlay] Title classes get margin-top on modal with internet styles


imagoiq:
When using `@use "@swisspost/design-system-styles/index.scss";` and code example from https://design-system.post.ch/#/post-samples/notification-overlay , we get unwanted margin-top with all titles:

![image](https://github.com/swisspost/design-system/assets/12294151/f0ece141-9414-4ebe-af96-6a9105c30e51)

imagoiq:
So the issue seems present since the beginning, here is the culprit and PR: https://github.com/swisspost/design-system/blob/d69e86bce25d947913cc316476797b2f6d1283cb/packages/web-styles/src/layouts/portal/_type.scss#L53

This type for the portal style is only loaded on external apps (and so index.scss). Source: https://github.com/swisspost/design-system/blob/d69e86bce25d947913cc316476797b2f6d1283cb/packages/web-styles/src/post-portal.scss#L13

I guess this was never seen as the Notification overlays are not used on external apps? I'm not sure what is the right fix. I don't think those portal´styles should be part of the design system. @gfellerph @alizedebray @oliverschuerch what do you think?

gfellerph:
@imagoiq I would prefer a temporary fix in the portal styles. With the next major version, typographic styles will get a complete overhaul and this theme will be refactored anyways. We're also looking at design tokens which will render these styles obsolete.

PR #2407, opened 2023-12-14
fix(styles): removes unwanted margin for heading in layouts/portal

b1aserlu:


imagoiq (changes requested):
Thanks for the proposal, but it's only a workaround. The issue is that the styles are different between the index and intranet bundles (see https://next.design-system.post.ch/?path=/docs/getting-started-styles--docs). Can you please check with Philipp? Or assign the issue to him. Last time that I talked about this issue, he had some ideas.

b1aserlu:
Asigned Issue to Phillip

gfellerph:
See https://github.com/swisspost/design-system/issues/1931#issuecomment-1948478202. A temporary resolution to this problem would be my preferred way of dealing with it.

packages/styles/src/layouts/portal/_type.scss
```diff
@@ -53,6 +53,9 @@
   h5,
   .h5 {
     margin-top: map.get(spacing.$post-sizes, 'bigger-big');
+    &:first-child {
+      margin-top: 0;
+    }
```
gfellerph, line 58: This selector is too generic, scope it to the notification overlay title, no matter what heading level.

.changeset/spotty-countries-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Removes unwanted margin from heading when using `@use "@swisspost/design-system-styles/index.scss";`
```
gfellerph, line 5:
```suggestion
Removes unwanted margin from heading within the notification overlay component.
```

gfellerph (changes requested):

PR #2421, opened 2023-12-19
feat(setup): update netifly to v16

b1aserlu:


gfellerph:
@b1aserlu, @oliverschuerch the fact that deploy workflows are run from main is a security feature. This allows us to not use the github secrets for feature branches or pull requests from forks where they could potentially be stolen.

Therefore, the only way to test this, is on main. Not ideal, but secure.

gfellerph (approved):

b1aserlu:
image

It seems that we need to use ` @swisspost/design-system-demo` and ` @swisspost/design-system-documentation` for it to work

PR #2423, opened 2023-12-19
fix(documentation): shows example code for first canvas where it was missing

b1aserlu:


packages/documentation/src/stories/foundation/layout/grid/grid.docs.mdx
```diff
@@ -20,7 +20,7 @@ import SampleScssMixins from './grid-scss-mixins.sample.scss?raw';
 
 Our grid system uses a series of containers, rows, and columns to layout and align content. It’s built with flexbox and is fully responsive. Below is an example and an in-depth explanation for how the grid system comes together.
 
-<Canvas of={GridStories.Basis} />
+<Canvas sourceState="shown" of={GridStories.Basis} />
```
oliverschuerch, line 23: I would recommend expanding the source only on the first canvas on components, but leave foundation pages as they are.

packages/documentation/src/stories/utilities/sizing/sizing.docs.mdx
```diff
@@ -27,7 +27,7 @@ Bootstrap offers sizing classes with the suffixes bellow. They allow you to set
 
 #### Example
 
-<Canvas of={SizingStories.SizesPercent} />
+<Canvas sourceState="shown" of={SizingStories.SizesPercent} />
```
oliverschuerch, line 30: I would recommend expanding the source only on the first canvas on components, but leave foundation pages as they are.

oliverschuerch (changes requested):
set `sourceState="shown"` also for the following pages:
- internet-header breadcrumbs `docs` and `custom-items`
- internet-header footer `docs` and `custom-items`
- internet-header header `docs`, `custom-items`, `language switch overrides` and `os flyout overrides`

packages/documentation/src/stories/foundation/layout/grid/grid.docs.mdx
b1aserlu: For the moment all the pages wil be have the source code shown. Maybe this wil change in the future

oliverschuerch (approved):

PR #2424, opened 2023-12-20
chore(components): export accordion, popover and popovercontainer

alizedebray:


packages/components/src/index.ts
```diff
@@ -1,9 +1,12 @@
 export { Components, JSX } from './components';
 
 // Export every single component so it gets included in the dist output
+export { PostAccordion } from './components/post-accordion/post-accordion';
 export { PostAlert } from './components/post-alert/post-alert';
 export { PostCollapsible } from './components/post-collapsible/post-collapsible';
 export { PostIcon } from './components/post-icon/post-icon';
+export { PostPopover } from './components/post-popover/post-popover';
+export { PostPopovercontainer } from './components/post-popovercontainer/post-popovercontainer';
```
alizedebray, line 9: @oliverschuerch This is actually an undocumented (internal) component, should it still be exported here?

oliverschuerch: In general, I would say "no".
But as long as stencil is generating the types for this component and putting them in the components.d.ts file, we need to export all the components. Otherwise we will not be able to release a bug-free components-angular package.

I have created a ticket to address this problem: #2426.

oliverschuerch (approved):

PR #2428, opened 2023-12-22
chore(components): write contribution guidelines

alizedebray:


packages/components/CONTRIBUTING.md
```diff
@@ -20,6 +20,60 @@ To run the unit tests for the components, run:
...
+All components should be named with the prefix `post-`.
+
+### Use slots
+You can use [the `<slot>` HTML element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) as a placeholder for any user-defined markup.
+
+If the component has only one slot, it should not have a `name` attribute.
+Also, don't use named slots for primary content, but instead reserve them for secondary content.
+
+When you need to access elements assigned to a given slot, always prefer to use the `onSlotChange` event to ensure that all asynchronous changes to the content are taken into account.
+Indeed, the component is not re-rendered if only its content changes.
```
gfellerph, line 40:
```suggestion
Parent components are not re-rendered if only their content changes.
```

packages/components/CONTRIBUTING.md
```diff
@@ -20,6 +20,60 @@ To run the unit tests for the components, run:
 ```bash
 pnpm --filter design-system-components test
 ```
+## Writing Components
+
```
gfellerph, line 24: Include a section about the general styleguide of stencil https://stenciljs.com/docs/style-guide and that ~~we're~~ we should extend it.

packages/components/CONTRIBUTING.md
```diff
@@ -20,6 +20,60 @@ To run the unit tests for the components, run:
...
+Properties, defined with the `@Prop` decorator, are custom attributes publicly exposed on the HTML element.
+To ensure that the value passed by the user matches what is expected, it is necessary to implement validation.
+
+All property validators are available in the `src/utils/property-checkers` folder.
+
+You can use these validators in a function decorated with the `@Watch` decorator to validate the value given to a prop every time it changes, as shown in the [Stencil documentation](https://stenciljs.com/docs/properties#prop-validation).
+Also make sure to call this function in the `connectedCallback` lifecycle hook so that the value is also validated when the component is initially rendered.
+
+### Write Component Styles
+Web components mainly use styles from the `@swisspost/design-system-styles` package.
```
gfellerph, line 52: Components that have a basic HTML & CSS variant have their styles defined in the styles package. Sometimes it's possible to re-use the same styles but in many cases, the shadow DOM structure needs specific selectors. In these cases write custom CSS inside the component styles, using variables, mixins and functions from the styles package.

Don't assume that any CSS from the styles package is actually present on the page. It should be possible to use the component package standalone. Therefore, if you're using CSS Custom Properties (CSS variables), which is highly encouraged, define fallback values for them.

```SCSS
:host {
  --post-accordion-background-color: var(--post-accordion-background-color, #{accordion.$background-color});
}
```

packages/components/CONTRIBUTING.md
```diff
@@ -20,6 +20,60 @@ To run the unit tests for the components, run:
...
+You can use these validators in a function decorated with the `@Watch` decorator to validate the value given to a prop every time it changes, as shown in the [Stencil documentation](https://stenciljs.com/docs/properties#prop-validation).
+Also make sure to call this function in the `connectedCallback` lifecycle hook so that the value is also validated when the component is initially rendered.
+
+### Write Component Styles
+Web components mainly use styles from the `@swisspost/design-system-styles` package. 
+They are included in the component stylesheet using the [Sass `@use` rule](https://sass-lang.com/documentation/at-rules/use/).
+Make sure to only include styles related to the component to ensure it remains lightweight.
+
+It is also possible to add specific styles for slotted elements using the [CSS `::slotted()` selector](https://developer.mozilla.org/en-US/docs/Web/CSS/::slotted),
+or to style parts of the component based on a parent using the [CSS `::part()` selector](https://developer.mozilla.org/en-US/docs/Web/CSS/::part).
```
gfellerph, line 57:
```suggestion
or to style parts of the component based on a parent using the [CSS `::part()` selector](https://developer.mozilla.org/en-US/docs/Web/CSS/::part). Document parts in Storybook as they are also exposed to the host document and people can customize these elements.
```

gfellerph (changes requested):
Nice write up!

packages/components/CONTRIBUTING.md
alizedebray: What about "Any parts intended for customization by users of the component package must be documented in Storybook."? I can imagine that some parts will be created for internal use only.

.changeset/young-impalas-happen.md
gfellerph, line 1: It's not necessary to write a changeset for this, as the component output did not change. We'd end up with two identical versions.

packages/components/CONTRIBUTING.md
```diff
@@ -12,14 +12,92 @@ pnpm components:start
...
 ```bash
-pnpm --filter design-system-components test
+pnpm components:test
 ```
+## Writing Components
+
+### Generate a New Component
+To generate a new component, run:
+```bash
+pnpm components:generate my-new-component
```
gfellerph, line 28: Follow your rules, dammit 😄

```suggestion
pnpm components:generate post-my-new-component
```

packages/components/CONTRIBUTING.md
```diff
@@ -12,14 +12,92 @@ pnpm components:start
...
+
+You can use these validators in a function decorated with the `@Watch` decorator to validate the value given to a prop every time it changes, as shown in the [Stencil documentation](https://stenciljs.com/docs/properties#prop-validation).
+Also make sure to call this function in the `connectedCallback` lifecycle hook so that the value is also validated when the component is initially rendered.
+
+### Write Component Styles
+Components that have a standard HTML variant have their styles defined in the `@swisspost/design-system-styles` package.
+It is sometimes possible to reuse these styles as-is but in many cases, the shadow DOM structure needs specific selectors.
+In these cases write custom CSS inside the component styles, using variables, mixins and functions from the styles package.
+
+
```
gfellerph, line 66:
```suggestion

```

gfellerph (approved):

PR #2453, opened 2024-01-09
fix(icons): fix icon changelog and update icon workflow

alizedebray:


.github/workflows/fetch-icons.yaml
gfellerph, line 1: I don't want to ruin the fun, but thinking about it, adding icons would be a feature 🤐

alizedebray: True... I'll update!

PR #2467, opened 2024-01-11
feat(docs): Add autolink headings

imagoiq:
This is proof-of-concept. It has some complexities, and please have no hesitation to say if it's preferable not to add it.

HTML markup for links is added through [remark-autolink-headings](https://github.com/remarkjs/remark-autolink-headings). Remark plugins are officially supported by Storybook Docs and can be easily added.

### First issue (legacy plugin)

However, this plugin is considered as legacy and not updated since 2021. The author recommends using `remark-rehype` and `rehype-autolink-headings` together instead. Unfortunately, it's not possible with our setup as it'll create a double transformation into HTML as Storybook is already doing a transformation. I think that it's not a showstopper, as the plugin is quite simple (1 file) and could be even integrated into our repository if needed.

### Second issue (monkey patch)

Another requirement was to add a script that rewrite the `base` element of the preview iframe to be able to get the anchor link related to the main URL and not the iframe. Without it, it would work, but let the user without sidebar and toolbar, and it would be just confusing for the user.

I think, it's quite unlikely that Storybook will update themselves the `base` element or let us do it as [the element is included in a basic HTML file.](https://github.com/storybookjs/storybook/blob/3664135b74c589f5c24bbfbad4438fd3fe32cd74/code/lib/core-common/templates/base-preview-head.html#L4)


packages/documentation/package.json
```diff
@@ -39,6 +39,7 @@
     "@percy/cypress": "3.1.2",
     "@pxtrn/storybook-addon-docs-stencil": "6.4.1",
     "@storybook/addon-designs": "7.0.9",
+    "@storybook/addon-docs": "^7.6.7",
```
oliverschuerch, line 42:
```suggestion
    "@storybook/addon-docs": "7.6.7",
```

packages/documentation/package.json
```diff
@@ -62,6 +63,7 @@
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "react-syntax-highlighter": "15.5.0",
+    "remark-autolink-headings": "^7.0.1",
```
oliverschuerch, line 66:
```suggestion
    "remark-autolink-headings": "7.0.1",
```

oliverschuerch (approved):
Good job, I like it a lot!
And the fact that you used a legacy plugin is no problem for me (at the moment).
We can try to replace it with something newer in the future...

There is only one issue left: It's not working on the "Home" page.
On the other hand, it's not really important to have it available there.

PR #2501, opened 2024-01-16
fix(docs): Clean icon component controls from duplicate and null values

imagoiq:
### What

- Fix duplicate `flip-h` and `flip-v` argument
- Change category to make it work with how `flip-h` and `flip-v` are translated from plugin
- Remove args with falsy value so we don't display them at all. Before you could have `animation="null"` or `base=""`


oliverschuerch (approved):
Should we add a changset for this documentation fix?

PR #2503, opened 2024-01-16
fix(internet-header): Fix e2e after improving the main slot

imagoiq:


packages/internet-header/cypress/e2e/meta-navigation.cy.ts
```diff
@@ -47,13 +47,13 @@ describe('meta-navigation', () => {
...
 
-      it(`sets css variable --meta-header-height to 0`, () => {
+      it(`sets css variable --meta-header-height to 3rem`, () => {
         cy.changeArg('meta', false);
         cy.get('swisspost-internet-header').within(() => {
           cy.get('header').then($el => {
             const style = getComputedStyle($el[0]);
             const metaHeaderHight = style.getPropertyValue('--meta-header-height');
-            expect(metaHeaderHight).to.eq('0px');
+            expect(metaHeaderHight).to.eq('3rem');
```
imagoiq, line 56: @alizedebray not sure if it's the right fix, but I'm sure that you know.

alizedebray: It seems the test was correct, only the name was wrong.

With a `meta="false"` attribute set on the `<swisspost-internet-header>`, the custom property should be set to 0px:
grafik

imagoiq: Then something is wrong. Could you handle this PR?

packages/internet-header/cypress/e2e/meta-navigation.cy.ts
```diff
@@ -61,13 +61,13 @@ describe('meta-navigation', () => {
 
     describe('meta: changes during runtime (false => true => false)', () => {
       it(`adds and removes meta navigation`, () => {
-        cy.changeArg('meta', false);
+        cy.changeArg('meta', 'false');
         cy.get('post-meta-navigation').should('not.exist');
 
         cy.changeArg('meta', true);
```
oliverschuerch, line 67: should this not also be `cy.changeArg('meta', 'true')`, if on the other hand it must be `'false'` instead of `false`?

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #2504, opened 2024-01-16
chore(styles): Upgrade stylelint to v16

imagoiq:
Upgrade manually stylelint to v16 because of relation with the following packages:

https://github.com/swisspost/design-system/pull/2418
https://github.com/swisspost/design-system/pull/2419
https://github.com/swisspost/design-system/pull/2431
https://github.com/swisspost/design-system/pull/2434

We could have a special rule in renovate to group them, but I'm not sure that they totally aligned to a major version. I think this upgrade of stylelint is pretty unique.

https://github.com/prettier/stylelint-config-prettier was removed as it is deprecated since v15.

Fixes in SCSS code are not due to the upgrade.


imagoiq:
e2e failing tests are not related to this PR. Fix: https://github.com/swisspost/design-system/pull/2503

imagoiq:
e2e Xvfb issue seems to be linked with: https://github.com/cypress-io/cypress-docker-images/issues/54

alizedebray (approved):

PR #2517, opened 2024-01-19
feat(demo): provide internationalization guidelines for ng-bootstrap components

alizedebray:


imagoiq:
The build fails, let me know when I can review

alizedebray:
> The build fails, let me know when I can review

@imagoiq
The build in fixed, you can now review the PR.

imagoiq (changes requested):
Could you change the demo app so it works with the first example here https://preview-2517--swisspost-web-frontend.netlify.app/#/ng-bootstrap-samples/datepicker ?

alizedebray:
> Could you change the demo app so it works with the first example here https://preview-2517--swisspost-web-frontend.netlify.app/#/ng-bootstrap-samples/datepicker ?

@imagoiq
Not realy: the translations are injected when the component is created and they cannot be changed afterwards.

The datepicker is created the first time it is opened, which means that if I select a language before it is properly applied.
Howerver, it is not destroyed when closed, the same instance is used for all further opening. Therefore, if I change the language again is not applied.

imagoiq (approved):

PR #2519, opened 2024-01-20
fix(documentation): fix path to entry point for lazy loading components

christian-bromann:
Hey there 👋 

I am developer for the Stencil team and have been looking into how Stencil suggests to distribute components within design systems like this one. During the research I stumbled upon this bug and I would like to send this patch to fix it. We will hopefully soon have some updates on this and I will come back here to report 😉 

The path in the current docs point to a file that only exports the web components. In the context of the docs though you want to point to the file that lazy loads all components so someone can use them in the HTML.

Have a nice weekend!

### DIFF
--- a/.changeset/spicy-swans-punch.md
+++ b/.changeset/spicy-swans-punch.md
@@ -1,0 +1,5 @@
+---
+"@swisspost/design-system-documentation": patch
+---
+
+fix(documentation): fix path to entry point for lazy loading components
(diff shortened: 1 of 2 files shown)


gfellerph:
@christian-bromann thanks for reaching out and submitting a PR! That's an amazing way to connect with the community. We're very much looking forward to hearing more about the topic as we had some confusion in the past about stencil outputs and where to use what exactly.

Have a nice week 😎

oliverschuerch:
In our docs, we are pointing to the `loader/index.js`, which contains the lazy loading components.

PR #2529, opened 2024-01-23
fix(documentation): Autolink with base url disrupt other anchor links

imagoiq:
Fix over https://github.com/swisspost/design-system/pull/2467

In the previous PR, autolink anchor link associated to heading were formed by changing the  `<base>` element href attribute so that the absolute URL would be created by using the parent frame URL (the main URL that the user navigate with). However, it appears that other anchor links, mostly the one used for the table of content, will use this absolute URL as well. They are correct, but the transition to an anchor is managed by Storybook and not the browser so in this case as the URL used is not the same as the iframe, it doesn't work if the page is already loaded (it works when the page is refreshed).

To avoid this issue, I propose a progressive enhancement to capture all the autolink anchor links and change them (but only them) to use the absolute URL from the main frame. If the script fails, the link still works but the manager (sidebar, toolbar) of Storybook will not appear.

Ideally, we should generate those links in Storybook. But I couldn't find a way to get the context in the rehype plugin and hooking into the layout.tsx children don't seem feasible. 

### Testing

You can find table of contents in https://preview-2529--swisspost-design-system-next.netlify.app/?path=/docs/components-icons-getting-started--docs (although some links are broken, but fixed with https://github.com/swisspost/design-system/pull/2496)


packages/documentation/.storybook/main.ts
```diff
@@ -1,7 +1,7 @@
 import type { StorybookConfig } from '@storybook/web-components-vite';
 import pkg from '../package.json';
-import remarkAutolinkHeadings from 'remark-autolink-headings';
 import { mergeConfig } from 'vite';
+import rehypeAutolinkHeadings from 'rehype-autolink-headings';
```
imagoiq, line 4: I had another try and I managed to make it work with this plugin which replace the legacy `remark-autolink-headings` (see https://github.com/remarkjs/remark-autolink-headings)

packages/documentation/.storybook/preview-body.html
```diff
@@ -0,0 +1 @@
+<script src="/assets/scripts/storybook-preview-events.js"></script>
```
imagoiq, line 1: It could have been part of the manager, but it would have required more work on the observer and I think this event could be beneficial as well for e2e tests, for example, where we only load the iframe not the whole storybook.

packages/documentation/public/assets/scripts/storybook-preview-events.js
```diff
@@ -0,0 +1,30 @@
+const previewIframe = document.body;
+let footerEl = document.querySelector('.docs-footer');
+let currentPage = null;
+
+if (footerEl) {
+  // if footer already exists, just emit ready event and listen for route-changes
+  contentReady();
+} else {
+  // if footer does not exist yet, wait until its rendered, then emit contentReady event and listen for route-changes
+  new MutationObserver(function () {
```
imagoiq, line 10: Unfortunately, we have to make all this code to wait until the footer is available. Maybe after https://github.com/storybookjs/storybook/issues/24429, we could simply put the script at the end of the body and simplify it?

alizedebray (approved):

PR #2530, opened 2024-01-23
fix(styles): Regression on input valid state

imagoiq:


.changeset/spicy-cars-act.md
gfellerph, line 1: This file is not necessary, since the regression never got published.

imagoiq:
@gfellerph looks ready for review.

gfellerph (approved):

PR #2531, opened 2024-01-23
feat(docs): add modal documentation

alizedebray:


imagoiq (approved):

imagoiq (changes requested):
I'm not sure what to do, but I just noticed that the page in the sidebar is at the right position, a bit by chance. Currently in `preview.ts`, only pages until "Icons" are defined. The rest are added alphabetically after it. Perhaps to be coherent, we should define all the pages there, no?

imagoiq (approved):

PR #2532, opened 2024-01-23
feat(styles): Tag component but .tag class name is also used for code examples

b1aserlu:
The code examples also use a class called .tag and when I named my class .tag as well it overwrote the styles for the code examples.


b1aserlu:
closed as decision was reached in Roundtable meeting of 25.01.2024

PR #2540, opened 2024-01-24
feat(docs): Migrate Pagination documentation

imagoiq:


gfellerph (approved):
Not sure why the e2e tests are failing here, but the result looks good.

PR #2542, opened 2024-01-24
chore(.npmrc): Use save-exact for new dependencies

imagoiq:


.npmrc
```diff
@@ -1,2 +1,4 @@
 # Added for storybook peer dependencies to react
-strict-peer-dependencies=false
\ No newline at end of file
+strict-peer-dependencies=false
+# Install new dependency with exact version to let renovate handle the update
```
imagoiq, line 3:
```suggestion
# Install new dependencies with exact version to let renovate handle the update
```

alizedebray (approved):

PR #2543, opened 2024-01-24
fix(internet-header): Header Logo covering content on postauto.ch

imagoiq:


gfellerph (approved):
Was a little silly to use aspect ratio in retrospect. Thanks for the fix.

PR #2552, opened 2024-01-25
feat(components, documentation, styles): new tag styles and post-tag webcomponent

b1aserlu:


packages/styles/src/variables/components/_tag.scss
```diff
@@ -0,0 +1,20 @@
+@use './../color' as col;
+@use './../commons' as commons;
+@use './../spacing' as spacing;
+@use './../type' as type;
+
+$tag-max-width: 296px;
```
alizedebray, line 6: Why 296 ? Also can you use `rem` instead of `px`?

packages/styles/src/variables/components/_tag.scss
```diff
@@ -0,0 +1,20 @@
+@use './../color' as col;
+@use './../commons' as commons;
```
alizedebray, line 2: The `as` is not necessary, we usually only use an alias for the mixins, functions and placeholders but leave the variable namespaces as is.

```suggestion
@use './../commons';
```

packages/styles/src/components/tag.scss
```diff
@@ -0,0 +1,53 @@
+@forward './../variables/options';
+@forward './utilities';
+
+@use './../variables/components/tag' as tag;
+
+.tag:has(:not(pre)), // is only a temporary fix till I know how to do it wihtout causing future issues. wihtout this the Code sample would break currently.
```
alizedebray, line 6: The `:has(:not(pre))` selector fixes a problem that only exists in Storybook, therefore it should not be in the styles package.

It would be better to override the `.tag` styles in preview.scss.

packages/styles/src/components/tag.scss
```diff
@@ -0,0 +1,53 @@
...
+.tag-sm {
```
alizedebray, line 7: Usually we define the size classes as modifiers. 
This means the `.tag` class adds all the default styles for a tag and the `.tag-sm` only modifies some properties to make the tag smaller. Then you would use `tag` for a default tag and `tag tag-sm` for a small tag. 

You can check how this is done for inputs and buttons for examples.

packages/styles/src/components/tag.scss
```diff
@@ -0,0 +1,53 @@
...
+  display: inline-flex;
+  justify-content: flex-start;
+  justify-items: center;
```
alizedebray, line 10: The `justify-items` property is for grid containers, it is ignored in flexbox layouts: https://developer.mozilla.org/en-US/docs/Web/CSS/justify-items?utm_source=devtools&utm_medium=inspector-inactive-css
```suggestion
```

packages/styles/src/components/tag.scss
```diff
@@ -0,0 +1,53 @@
...
+  align-items: center;
+  gap: tag.$tag-gap;
+  padding: tag.$tag-vertical-padding tag.$tag-horizontal-padding;
+  font-size: tag.$tag-font-size;
+  border: tag.$tag-border-width solid transparent;
+  border-radius: tag.$tag-border-radius;
+  line-height: inherit;
```
alizedebray, line 17: Not necessary, this is the default behavior.
```suggestion
```

packages/styles/src/components/tag.scss
```diff
@@ -0,0 +1,53 @@
...
+  text-align: center;
+  vertical-align: baseline;
```
alizedebray, line 19: Not necessary, this is the default behavior.
```suggestion
```

packages/styles/src/variables/components/_tag.scss
```diff
@@ -0,0 +1,20 @@
...
+@use './../spacing' as spacing;
+@use './../type' as type;
+
+$tag-max-width: 296px;
+$tag-border-radius: spacing.$size-micro;
+$tag-border-width: commons.$border-width;
+$tag-gap: spacing.$size-micro;
+$tag-font-size: type.$font-size-16;
+$tag-horizontal-padding: spacing.$size-mini;
```
alizedebray, line 11: We usually use "padding-x" and "padding-y":
```suggestion
$tag-padding-x: spacing.$size-mini;
```

packages/styles/src/components/tag.scss
```diff
@@ -0,0 +1,53 @@
...
+  justify-content: flex-start;
+  justify-items: center;
+  align-items: center;
+  gap: tag.$tag-gap;
+  padding: tag.$tag-vertical-padding tag.$tag-horizontal-padding;
+  font-size: tag.$tag-font-size;
+  border: tag.$tag-border-width solid transparent;
+  border-radius: tag.$tag-border-radius;
+  line-height: inherit;
+  text-align: center;
```
alizedebray, line 18: Not necessary since you position content using a flexbox layout:
```suggestion
```

packages/styles/src/components/tag.scss
```diff
@@ -0,0 +1,53 @@
...
+    text-overflow: ellipsis;
+    overflow: hidden;
+    white-space: nowrap;
+  }
+}
+
+.tag:has(:not(pre)) {
+  height: tag.$tag-height;
+
+  .tag-icon {
```
alizedebray, line 36: This additional class is not necessary, you can simply select the `post-icon` with its tag name:
```suggestion
  post-icon {
```

alizedebray (changes requested):

packages/styles/src/components/tag.scss
oliverschuerch: Yes, but...
The same problem could happen on project using code highlighting tools as well.

packages/components/src/components/post-tag/post-tag.tsx
```diff
@@ -0,0 +1,77 @@
...
+   * <span className="alert alert-sm alert-info">If not set the icon will not show up.</span>
+   * To learn which icons are available, please visit our <a href="/?path=/docs/5704bdc4-c5b5-45e6-b123-c54d01fce2f1--docs" target="_blank">icon library</a>.
+   */
+  @Prop() readonly icon: null | string = null;
+
+  /**
+   * Defines the text of the component.
+   * Most of the time this will fit your needs, if you need to add custom content, use the default slot instead.
+   */
+  @Prop() readonly text: string;
```
alizedebray, line 39: Do we keep both the prop and slot in the end?

alizedebray (approved):

packages/components/src/components/post-tag/post-tag.tsx
oliverschuerch: Legitimate question! I'll discuss it with @gfellerph.

PR #2561, opened 2024-01-30
test if fix worked

b1aserlu:


b1aserlu:
The deployements with netifly v16 are working now therefore the test was succesful and this PR is no longer needed

PR #2577, opened 2024-01-31
feat(documentation): Migrate Migration page

imagoiq:
### How

As far as I know, there is no real way to manage state inside MDX (there is an attempt with [mdx-observable](https://github.com/alexkrolick/mdx-observable), but it's alpha, not maintained and quite verbose). So I used web component (lit) without shadom DOM (aka light DOM) for that based on those examples:
https://lamplightdev.com/blog/2019/03/13/how-to-pass-data-between-web-components/
https://www.youtube.com/watch?v=T9mxtnoy9Qw
It makes sense to not use shadow dom as we want that the storybook styles apply to the content of those elements.

### What was not migrated

- Feature to copy code from `<code>` element. 
- Feature to store collapsible state of the accordion.


.prettierrc
```diff
@@ -12,7 +12,7 @@
   "arrowParens": "avoid",
   "insertPragma": false,
   "proseWrap": "preserve",
-  "htmlWhitespaceSensitivity": "ignore",
+  "htmlWhitespaceSensitivity": "css",
```
gfellerph, line 15: Just curious, why did you change this?

gfellerph (approved):
Awesome work 🧨

PR #2578, opened 2024-01-31
chore(*): Remove rel=noopener attribute which is not needed anymore

imagoiq:
The security issue with target="_blank" has been resolved since at least 2020, see : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#browser_compatibility


alizedebray:
@imagoiq It seems [Sonarcloud disagrees with you](https://github.com/swisspost/design-system/pull/2578#issuecomment-1919317545)... We could probably configure it otherwise

imagoiq:
> @imagoiq It seems [Sonarcloud disagrees with you](https://github.com/swisspost/design-system/pull/2578#issuecomment-1919317545)... We could probably configure it otherwise

Yes, they are cautious here as they don't know which browsers version we are supporting. But they still have a small note about it:
image

PR #2587, opened 2024-02-01
chore(workflow): Start Build documentation workflow for any new package

imagoiq:


.github/workflows/build-documentation.yaml
```diff
@@ -14,11 +14,7 @@ name: Build Documentation
 on:
   pull_request:
     paths:
-      - 'packages/documentation/**'
-      - 'packages/styles/**'
-      - 'packages/components/**'
-      - 'packages/internet-header/**'
-      - 'packages/icons/**'
+      - 'packages/**'
```
imagoiq, line 17: According to https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths, it seems to be recursive this way.

gfellerph (approved):

PR #2589, opened 2024-02-01
feat(documentation): Add example and explanation for grid gutters and remove unneeded page

imagoiq:
### What

- Add Gutters example
- Rework how we style the column in the documentation, I'm not sure if it's the best way. 
The Bootstrap documentation is confusing: 

  There is a gutter by default to 24px (check `.row` CSS rule. --bs-gutter-x: equals 1.5rem;). But if you look at [the examples on the column page](https://getbootstrap.com/docs/5.2/layout/columns/). The gutter is not displayed (nor removed with `.g-0`). The gutter is not visible because the applied style (background, padding, border) is on the `.col`.
  Meanwhile, if you look at the [examples on the gutters page](https://getbootstrap.com/docs/5.2/layout/gutters/), the styles are applied on a child element of the column (try to remove `.gx-5` from the first example, and you still get a gutter, the default one). 
  
  So I decided to remove gutters with the `.g-0` class in our examples to match the previous documentation (except for the first example to highlight the default gutter). And to highlight how nesting works, I made an exception on the style to match the column instead of the child element.


packages/documentation/src/stories/foundation/layout/grid/grid.blocks.tsx
```diff
@@ -34,7 +34,9 @@ export const GridTable = () => (
     <tr>
       <th>Custom gutters</th>
       <td colSpan={SCSS_VARIABLES['breakpoint-count']}>
-        <a href="/?path=/docs/foundations-layout-gutters--docs">yes</a>
+        <a href="#gutters" target="_self">
+          yes
+        </a>
```
gfellerph, line 39: Is the white space here really intended?
```suggestion
        <a href="#gutters" target="_self">yes</a>
```

imagoiq: Just checked and yes, it's from prettier. But it doesn't matter here as it's tsx and not mdx so no `<p>` wrapper added.

PR #2599, opened 2024-02-01
fix(icons): update keyword parsing

alizedebray:


packages/icons/public/report.json
gfellerph, line 1: Why was this file deleted?

gfellerph (changes requested):

gfellerph (approved):

Issue #2052, opened 2023-10-06
Header: Add logout url parameter


gfellerph:
The logout URL is usually provided by the Portal Config. Some projects started to migrate to cloud and now require a different logout URL for properly forwarding the event to the proxy and their service. In the config however, the logout URL is shared between all projects and can't be configured on a per project level.

By providing a logout-url parameter, we can allow projects to change this URL and let them override the URL from the config. The URL override should happen where options are being passed to the KLP Login Widget.

```[tasklist]
### Tasks
- [ ] Add a new `logout-url` prop to the `<swisspost-internet-header>`
```

PR #2600, opened 2024-02-01
chore(test): update logout test

gfellerph:
No need to acutally click on the logout link, just ensure it's there.


imagoiq (approved):
I guess that you didn't find a way to solve the issue, so I approve it, but in my opinion the test worth it to check if the override URL provided in the config works as it should. It originates from this PR: https://github.com/swisspost/design-system/issues/2052

PR #2603, opened 2024-02-05
feat(documentation): Set id/permalink for stories

imagoiq:
Set an id for each story as in https://storybook.js.org/docs/configure/sidebar-and-urls#permalink-to-stories. This allows to separate the URL generation from the display name. The URL is generated with the id and the name of the story with the title. 
You can change the title without changing the id so we can keep a kind of permalink.

Most of id are duplicated from title, except for foundations.


alizedebray (changes requested):
This is great! 👍 

Can you just update the links to the foundation docs using the new id's? 

Also, I'm not sure that using the titles as identifiers is the best solution because there will be discrepancies as soon as we update a component name... I would probably go for a unique id, maybe with an enum somewhere. What do you think?

imagoiq:
> This is great! 👍
> 
> Can you just update the links to the foundation docs using the new id's?

Good catch!

> Also, I'm not sure that using the titles as identifiers is the best solution because there will be discrepancies as soon as we update a component name... I would probably go for a unique id, maybe with an enum somewhere. What do you think?

I think this is a timeless question between slug vs id/uuid. It can be one or another, or mixed. I'm not really fond of only uuid as it's not user-friendly (and a bit prone to security issue). Using a mix doesn't make it more permanent except if we say that the slug is only here for human (e.g. /bzasdh/accessibility) and is ignored in the redirect process, but this is not how Storybook works, and it doesn't allow that. So it's debatable, perhaps we should have survey in the roundtable meeting?

alizedebray:
> So it's debatable, perhaps we should have survey in the roundtable meeting?

Sure, let's talk about that with the team

packages/documentation/src/stories/components/forms/custom-select/custom-select.stories.ts
```diff
@@ -2,6 +2,7 @@ import { Meta, StoryObj } from '@storybook/web-components';
 import { BADGE } from '../../../../../.storybook/constants';
 
 const meta: Meta = {
+  id: 'Components/Forms/Custom-Select',
```
alizedebray, line 5:
```suggestion
  id: 'Components/Forms/Custom Select',
```

packages/documentation/src/stories/components/collapsible/collapsible.stories.ts
```diff
@@ -29,7 +30,7 @@ export default meta;
 // DECORATORS
 function externalControls(story: StoryFn, context: StoryContext) {
   const { args, canvasElement } = context;
-  const togglerId = `${context.id}--button`;
+  const togglerId = `button--${context.id}`;
```
imagoiq, line 33: This is not valid otherwise, as we are targetting an id starting with a number later.

imagoiq:
It was decided on the roundtable to used exclusively uuid.

alizedebray (approved):

PR #2612, opened 2024-02-06
feat(styles, documentation): Update spinner style, animation, accessibility and migrate documentation

imagoiq:
### What

* Adapt spinner code to match style and animation from design ([figma](https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?type=design&node-id=16705-3536&mode=design&t=YBkVhkiOqTJFqBEM-0)).
* Add role and visually hidden content to improve accessibility.
* Migrate documentation page and examples on the new documentation.
* Document best practices on live region
* Deprecated sizes.
* Add snapshots tests.


packages/documentation/src/stories/components/spinner/spinner.stories.ts
```diff
@@ -0,0 +1,96 @@
...
+function generateDecorators(story: any) {
+  return html` <div class="m-5 overflow-hidden">${story()}</div> `;
+}
+
+function render(args: Args) {
+  const classes = ['loader', 'm-auto', args.size].filter(c => c && c !== 'null').join(' ');
+
+  return html` <div class="spinner-bg">
+    <div class="loading-modal">
+      <div class="${classes}" role="status">
```
imagoiq, line 59: About using `role=status`.

We might want to discuss it, there is no clear recommendation. 
- [Bootstrap uses `role=status`.](https://getbootstrap.com/docs/5.2/components/spinners/#about)
- [Access42 recommends `role=alert` for spinner](https://access42.net/chargements-asynchrones-donnees-bonnes-pratiques-accessibilite-numerique/).

For me, it makes more sense to use the `role=status` as the loading aspect is important, but is not important enough to interrupt all other actions, but it might be different from the context.

Links
- [MDN `role=status`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role)
- [MDN `role=alert`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role)
- [MDN `role=progressbar`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role)

.changeset/rude-fishes-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': minor
```
gfellerph, line 2:
```suggestion
'@swisspost/design-system-styles': patch
```

packages/documentation/src/stories/foundation/accessibility/accessibility.docs.mdx
```diff
@@ -24,3 +26,27 @@ TODO: write accessibility guide about form labels (visible and invisible), aria-
...
+They are defined either by using the `aria-live` attribute or some specific `role` attributes such as `role=status`, `role=alert`, `role=log`, etc.
+
+### Add or remove dynamically a live region
+
+As a best practice, live region should be present on the page load and only the content inside it should change.
+
+#### Example with a spinner on Angular
+
+When adding, for example, a spinner above a section, the spinner code should be always present,
+and then you can use `ng-container` to add the content when the spinner is active.
```
gfellerph, line 42:
```suggestion
When adding, for example, a spinner above a section, the live region container should be rendered initially while you can use `ng-container` to show and hide the spinner when you need it.
```

packages/documentation/src/stories/foundation/accessibility/accessibility.docs.mdx
```diff
@@ -24,3 +26,27 @@ TODO: write accessibility guide about form labels (visible and invisible), aria-
...
+
+##### Good example
+
+<Source code={liveRegionGoodAngularSample} language="html" />
+
+##### Bad example
+
+You shouldn't add the whole region dynamically as the accessibility might not be [re-rendered with some screen readers](https://www.scottohara.me/blog/2022/02/05/are-we-live.html#non-normative-rule-for-common-live-region-use-case).
```
gfellerph, line 50:
```suggestion
You shouldn't add the whole region dynamically as the accessibility bindings might not be [updated with some screen readers](https://www.scottohara.me/blog/2022/02/05/are-we-live.html#non-normative-rule-for-common-live-region-use-case).
```

packages/styles/src/components/spinner.scss
gfellerph, line 1: While I'm impressed with the effort for making the graphic look smooth, the animation currently runs janky, I suspect it's due to easing or the distribution of the animation steps. Can you schedule a review with Rouven to discuss this?

gfellerph (changes requested):

packages/styles/src/components/spinner.scss
imagoiq: I think by changing the easing it's already better and match better the original SVG version. What do you think?

packages/documentation/src/stories/components/spinner/spinner.stories.ts
gfellerph: I think going with the MDN recommendation is the right way for us, thanks for the explanation.

packages/styles/src/components/spinner.scss
gfellerph: Looks smoother, thanks!

gfellerph (approved):

PR #2616, opened 2024-02-07
chore(github): update demo release workflow

alizedebray:


.github/workflows/release-demo.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
 
       - name: Setup
         if: steps.check.outputs.changed == 'true'
-        uses: swisspost/design-system/.github/actions/setup-pnpm@main
+        uses: swisspost/design-system/.github/actions/setup-pnpm@release/angular-demo-archive
 
       # Install changesets locally
       - name: Install demo app & dependencies
```
imagoiq, line 39: On step "name: Publish demo app to netlify" perhaps you can already duplicate the environment variable for the site ID and rename with something like "NETLIFY_ARCHIVE_SITE_ID" so it's clearer where it is deployed?

imagoiq (approved):

gfellerph (changes requested):
- [x] This PR needs to go on the `release/angular-demo-archive` branch and not to `main`
- [x] I will add a ticket about cleaning up main after this. This workflow as well as the build and deploy demo app workflows can be deleted on main after we migrated
- [x] The netlify site id and token need to be changed because the archive will be deployed to it's own site. This has to be done in this PR and the tokens have to be prefixed like Loïc proposed https://github.com/swisspost/design-system/pull/2616/files#r1481257653

.github/workflows/release-demo.yaml
```diff
@@ -54,8 +54,8 @@ jobs:
         id: netlify
         if: steps.check.outputs.changed == 'true'
         env:
-          NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
-          NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
+          NETLIFY_ARCHIVE_SITE_ID: ${{ secrets.NETLIFY_ARCHIVE_SITE_ID }}
+          NETLIFY_ARCHIVE_AUTH_TOKEN: ${{ secrets.NETLIFY_ARCHIVE_AUTH_TOKEN }}
```
gfellerph, line 58: These are used by netlify and are predefined, we just need to update the naming of our secrets.
```suggestion
          NETLIFY_SITE_ID: ${{ secrets.NETLIFY_ARCHIVE_SITE_ID }}
          NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_ARCHIVE_AUTH_TOKEN }}
```

.github/workflows/release-demo.yaml
```diff
@@ -54,8 +54,8 @@ jobs:
...
         run: |
-          netlify link --id $NETLIFY_SITE_ID
+          netlify link --id $NETLIFY_ARCHIVE_SITE_ID
```
gfellerph, line 60:
```suggestion
          netlify link --id $NETLIFY_SITE_ID
```

gfellerph (changes requested):

gfellerph (approved):

PR #2620, opened 2024-02-07
feat(documentation): Document collapseChange event on post-accordion

imagoiq:


oliverschuerch (changes requested):
Does this event fire on the `post-accordion` or the `post-accordion-item` component or on both?

There are other things we should document about the `post-accordion-item`, like the `collapsed` or `headingLevel` attribute or the `toggle` method.
In my opinion, it would make more sense, to add the whole control table for the `post-accordion-item` below the control table for the `post-accordion` and add some headings for both tables, so we are sure that everything from both components is documented, whenever we change something on them.

imagoiq:
> Does this event fire on the `post-accordion` or the `post-accordion-item` component or on both?
> 
> There are other things we should document about the `post-accordion-item`, like the `collapsed` or `headingLevel` attribute or the `toggle` method. In my opinion, it would make more sense, to add the whole control table for the `post-accordion-item` below the control table for the `post-accordion` and add some headings for both tables, so we are sure that everything from both components is documented, whenever we change something on them.

It fires on both. For `post-accordion-item`, it's a good idea, I've created a ticket: https://github.com/swisspost/design-system/issues/2654

oliverschuerch (approved):

PR #2621, opened 2024-02-07
fix(tooltip): update a11y patcher, event listeners and de-dupe instance creation in Angular

gfellerph:
More broad capture with pointer events instead of just mouse events.


gfellerph:
@imagoiq, there are some more changes coming to this PR.

gfellerph:
@imagoiq, ready for re-review

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -1,25 +1,106 @@
...
+ */
+const globalHideTooltip = (tooltip: HTMLPostTooltipElement | PostTooltip) => {
+  hideTooltipTimeout = window.setTimeout(() => {
+    tooltip.hide();
+    hideTooltipTimeout = null;
+  }, 42);
+};
+
+/**
+ * Patch some hard to remember or understand accessibility features
```
imagoiq, line 63:
```suggestion
 * Patch some accessibility features that are hard to remember or understand
```

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -35,49 +116,55 @@ export class PostTooltip {
...
       );
     }
+  }
 
-    if (!this.triggers) {
-      throw new Error(
-        `No trigger found for <post-tooltip id="${this.host.id}">, please add the 'data-tooltip-target="${this.host.id}" attribute to the trigger element.`,
-      );
+  /**
+   * Add interest event listeners, but only once, and start
```
imagoiq, line 129: Is an "interest event listeners" a pattern? I never heard about this naming. In my opinion, handlers could be named perhaps differently as for example `globalInterestHandler` doesn't tell me about events.

imagoiq (approved):
LGTM. It would be good to add tests.

packages/components/src/components/post-tooltip/post-tooltip.tsx
gfellerph: Naming comes from here https://open-ui.org/components/interest-invokers.explainer/, it's like a collective term for hover/focus/long press events.

packages/components/src/components/post-popovercontainer/post-popovercontainer.scss
```diff
@@ -36,6 +36,15 @@
...
   overflow: visible;
+
+  &:has(.arrow) {
+    &::after {
+      position: absolute;
+      content: '';
+      inset: -13px;
+      z-index: -1;
+    }
+  }
```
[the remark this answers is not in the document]
gfellerph: It creates a safe space for hover interactions. The mouse pointer leaving the trigger enters the tooltip area without gap, enabling the tooltip to remain displayed. You're right, it probably needs a fallback and a comment.

gfellerph:
@imagoiq I added a test for the mutation observer as well and noticed it did not catch added elements - which I fixed.

imagoiq (approved):

PR #2628, opened 2024-02-12
chore(documentation): Remove input type color

imagoiq:


.changeset/shaggy-pets-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': patch
+---
+
+Removed input type color variant.
```
gfellerph, line 5:
```suggestion
Removed input type color variant as it is not compatible with the current input styles.
```

gfellerph (approved):

PR #2637, opened 2024-02-13
chore(e2e): Fix Cypress issue with xvfb

imagoiq:
This is the recommended way to run cypress in parallel: https://docs.cypress.io/guides/continuous-integration/introduction#Xvfb

This was tested successfully in https://github.com/swisspost/design-system/actions/runs/7889393329/job/21529150647?pr=2552


imagoiq:
Another example where it works: https://github.com/swisspost/design-system/pull/2638

gfellerph (approved):
Thanks, let's test it.

PR #2638, opened 2024-02-14
fix(styles): Display of icons in HCM on Webkit

imagoiq:
Currently, icons for example on button on in an alert do not display in Webkit based browser. (Chrome, Edge, Safari?). This is due to the `currentColor` property whose support had some changes recently, see writting from Edge https://melanie-richards.com/blog/currentcolor-svg-hcm/ or https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/Accessibility/PreserveParentColor/explainer.md


packages/styles/src/mixins/_icons.scss
```diff
@@ -10,6 +10,8 @@
 @mixin icon($name) {
   mask-image: url('#{icon-fn.get-svg-url($name)}');
   background-color: currentColor;
+  color: currentColor;
```
imagoiq, line 13: I'm not sure exactly why it is required, but it doesn't work without it.

imagoiq, line 13:
```suggestion
  color: currentColor; // Required in this case with usage of forced-color-adjust: preserve-parent-color
```

.github/workflows/e2e-tests.yaml
```diff
@@ -46,6 +46,9 @@ jobs:
       - name: Run tests
         run: pnpm e2e:ci
 
+      - name: Post Cypress – Stop Xvfb server
```
alizedebray, line 49: Not part of this PR, still okay anyway

alizedebray (approved):

.github/workflows/e2e-tests.yaml
imagoiq: Yes, it was for testing. It was cherry-picked so it has disappeared now that I updated the branch with main.

PR #2653, opened 2024-02-14
fix(styles,documentation): Improved rendering of the home page for HCM

imagoiq:


packages/styles/src/mixins/_notification.scss
```diff
@@ -72,7 +72,6 @@
 
   @include utilities-mx.high-contrast-mode {
     filter: none !important;
-    background-color: transparent !important;
     border: 2px solid WindowText;
```
oliverschuerch, line 75:
```suggestion
    border: 2px solid CanvasText;
```
https://wiki.mozilla.org/Accessibility/Design_Guide ->
"Components not in the above categories should use CanvasText / Canvas (older implementations may use WindowText / Window)"

oliverschuerch (changes requested):
The SVG package icons are not working in dark-mode (not HCM related).
Do you have any idea how to fix this?
image

packages/styles/src/mixins/_notification.scss
imagoiq: Good catch!

imagoiq:
> The SVG package icons are not working in dark-mode (not HCM related). Do you have any idea how to fix this? image

@oliverschuerch I can't reproduce it. Do you use the rendering panel on Chrome or something else?

oliverschuerch (approved):

PR #2657, opened 2024-02-15
fix(e2e-tests): change e2e test to work with storybook id and fix issue with xvfb

b1aserlu:


packages/internet-header/cypress/e2e/breadcrumbs.cy.ts
```diff
@@ -33,13 +33,13 @@ describe('breadcrumb', () => {
       modifiedConfig.fr.breadcrumb = undefined;
       modifiedConfig.it.breadcrumb = undefined;
 
-      prepare('Components/Internet Header/Breadcrumbs', 'Default', { config: modifiedConfig });
+      prepare('4347e5bf-8bf2-4f44-9075-9faaa53591ed', 'Default', { config: modifiedConfig });
```
imagoiq, line 36: Small improvement, you can extract this string to a variable. You can do it for every component. You can try as well to import the story and get the id from there, but there might be a circle redundancy issue.

imagoiq (approved):

PR #2667, opened 2024-02-15
fix(styles): update form switch HCM styles

alizedebray:


packages/styles/src/components/form-check.scss
```diff
@@ -163,6 +180,16 @@
...
   &:hover > .form-check-input:not([disabled]) {
     background-color: rgba(var(--post-contrast-color-rgb), 0.1);
+
+    @include utility-mx.high-contrast-mode {
+      border-color: Highlight;
+
+      &:checked,
+      &::after,
+      &:checked::after {
+        background-color: Highlight;
```
imagoiq, line 190: Maybe we can use `AccentColor` and as a fallback Highlight. I'm not sure of what I'm saying, but it looks like it's the color to show that an element is active. It's supported by Firefox (though I don't see the difference) and under a flag on blink engine browsers.

packages/styles/src/components/form-check.scss
```diff
@@ -152,6 +153,22 @@
       }
     }
 
+    @include utility-mx.high-contrast-mode {
+      transition: none;
+      border: form-check.$form-check-input-border-width solid CanvasText;
+
+      &:checked {
+        background-color: CanvasText;
```
imagoiq, line 161: Maybe we can use AccentColor and as a fallback to CanvasText. I'm not sure of what I'm saying, but it looks like it's the color to show that an element is active. It's supported by Firefox (though I don't see the difference) and under a flag on blink engine browsers.

imagoiq (approved):

PR #2676, opened 2024-02-16
docs: pattern for sorry pages

gfellerph:
This adds a small CSS part and a template for error pages.


packages/documentation/src/stories/templates/error-page/error-page.stories.ts
```diff
@@ -0,0 +1,64 @@
...
+            </p>
+            <div class="d-flex gap-3 mt-4">
+              <a class="btn btn-primary" href="/">Home page</a>
+              <a class="btn btn-secondary" href="https://www.post.ch/de/pages/suche#t=AllTab">
+                Search
+              </a>
+            </div>
+          </div>
+          <div class="col-6 offset-3 offset-rg-0 col-rg-4 align-self-end">
+            <img class="error-container--image" src="/images/content/404.png" alt="" />
```
oliverschuerch, line 52: What about hosting this picture on post.ch or on a cdn, so we could easily update it in no time, in all projects using it?

packages/documentation/src/stories/templates/error-page/error-page.docs.mdx
```diff
@@ -0,0 +1,12 @@
+import { Meta, Canvas } from '@storybook/blocks';
+import * as ErrorPageStories from './error-page.stories';
+
+<Meta of={ErrorPageStories} />
+
+# Error page template
+
+A generic error page pattern with an image. The image can be exchanged based on the type of error.
+
+- 404 not found: <a download="/images/content/404.png" href="/images/content/404.png">Download "404 not found" image</a>
```
oliverschuerch, line 10: Wrap this content in a `<div class="container">` element.

And as far as I know, the `download` attribute on a link should contain no value or at least only the filename/exension, but not the hole path (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attributes).

oliverschuerch (changes requested):

packages/documentation/src/stories/templates/error-page/error-page.stories.ts
gfellerph: For the moment, the CDN situation is really unclear. We failed to get a CDN for the fonts. I'll have to check with cdn.post.ch if this is intended for hosting images for everybody. In the meantime, it would really be best if everybody was hosting their stuff.

packages/documentation/src/stories/templates/error-page/error-page.docs.mdx
```diff
@@ -0,0 +1,14 @@
...
+import * as ErrorPageStories from './error-page.stories';
+
+<Meta of={ErrorPageStories} />
+
+<div class="container">
+  # Error page template
+
+  A generic error page pattern with an image. The image can be exchanged based on the type of error.
+
+  - 404 not found: <a download href="/images/content/404.png">Download "404 not found" image</a>
```
oliverschuerch, line 11: The preceding text "404 not found:" is kind of irritating me. Wouldn't it be enough to just leave the text inside of the link?

oliverschuerch: The example is very large and therefore (even if I now storybook probably pretty well) I was thinking about how to get the necessary code from the example.
It could be a good idea to add a note, and to tell visitors how to do this here.

packages/documentation/src/stories/templates/error-page/error-page.stories.ts
```diff
@@ -0,0 +1,62 @@
...
+              We kindly ask you to review this once again. It is also possible that we have moved,
+              archived, or renamed the relevant page. Perhaps you can find the content you're
+              looking for on our homepage. Or use the search on our portal to locate the desired
+              page.
+            </p>
+            <div class="d-flex gap-3 mt-4">
+              <a class="btn btn-primary" href="/">Home page</a>
+              <a class="btn btn-secondary" href="https://www.post.ch/de/pages/suche#t=AllTab">
+                Search
+              </a>
```
oliverschuerch, line 46: Add some click event listeners (`@click="${e => e.preventDefault()}"`) on both links, to prevent default behaviour in the example. Because at least the Home page button is misleading us to our storybook root url.

oliverschuerch (approved):

gfellerph:
@oliverschuerch I restructured the page a little to make it easier to navigate. Can you have another look please?

oliverschuerch (approved):

PR #2695, opened 2024-02-19
fix(popovercontainer): prevent immediate re-toggle

gfellerph:
This small timeout prevents the popovercontainer from opening and closing with the same event chain. This happens for example when the popover is open and the trigger button is clicked again. It's because first, an outside click is registered the popover closes. Right after that, the click from the toggle button fires and opens the popover again.

### DIFF
--- a/packages/components/cypress.config.js
+++ b/packages/components/cypress.config.js
@@ -9,7 +9,6 @@ module.exports = defineConfig({
     viewportHeight: 576,
   },
   includeShadowDom: true,
-  chromeWebSecurity: false,
   retries: {
     runMode: 1,
   },
--- a/packages/components/cypress.config.js
+++ b/packages/components/cypress.config.js
@@ -9,6 +9,7 @@ module.exports = defineConfig({
     viewportHeight: 576,
   },
   includeShadowDom: true,
+  chromeWebSecurity: false,
   retries: {
     runMode: 1,
   },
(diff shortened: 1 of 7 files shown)


packages/components/cypress.config.js
```diff
@@ -9,6 +9,7 @@ module.exports = defineConfig({
     viewportHeight: 576,
   },
   includeShadowDom: true,
+  chromeWebSecurity: false,
```
imagoiq, line 12: Can you please explain why it is necessary. Here is enough.

packages/components/src/components/post-popover/post-popover.tsx
```diff
@@ -1,22 +1,39 @@
...
 
+let popoverInstances = 0;
+const popoverTargetAttribute = 'data-popover-target';
+
+const globalToggleHandler = (e: PointerEvent | KeyboardEvent) => {
+  const target = e.target as HTMLElement;
+  if (!target || !('getAttribute' in target)) return;
+  const popoverTarget = target.getAttribute(popoverTargetAttribute);
+  if (!popoverTarget || popoverTarget === '') return;
+  if ('key' in e && e.key !== 'Enter') return;
```
imagoiq, line 18: It's not easy to read. Maybe you can extract those conditions into variables and have a single if condition?

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -47,6 +47,7 @@ export class PostPopovercontainer {
   private arrowRef: HTMLElement;
   private eventTarget: Element;
   private clearAutoUpdate: () => void;
+  private toggleTimeout: number;
```
imagoiq, line 50: At first read, I thought that this variable contain an actual time value, but it's rather the id of the setTimeout. Also it can contain null on line 131

imagoiq (approved):

packages/components/cypress.config.js
gfellerph: For some reason I had CORS issues when testing locally. I might be able to remove it again, will retest.

PR #2696, opened 2024-02-19
fix(styles): HCM Light: Select arrow icon is not visible

imagoiq:


packages/styles/src/variables/components/_form-select.scss
```diff
@@ -18,4 +18,5 @@ $form-select-indicator-success: url(icons.get-colored-svg-url(
   color.$success-green
 )) !default;
 $form-select-indicator-error: url(icons.get-colored-svg-url('2104', color.$error-red)) !default;
-$form-select-indicator-hcm: url(icons.get-colored-svg-url('2113', color.$white)) !default;
+$form-select-indicator-hcm-dark: url(icons.get-colored-svg-url('2113', color.$white)) !default;
```
imagoiq, line 21: Changing the variable name should be normally deprecated, but in this case, I doubt that anyone has changed this variable (or should change it).

alizedebray (approved):

PR #2700, opened 2024-02-20
fix(internet-header): Post-meta navigation is missing accessible labels

gfellerph:


packages/documentation/cypress/e2e/components/header.cy.ts
```diff
@@ -0,0 +1,19 @@
...
+      cy.checkA11y('#root-inner', {
+        rules: {
+          'aria-required-children': {
+            enabled: false,
+          },
+        },
+      });
+    });
+  });
+});
```
imagoiq, line 19: You can remove that as it will not work without the package.

imagoiq (approved):

PR #2704, opened 2024-02-20
chore: use the SWISSPOSTDEVS_ACCESS_TOKEN for the icon workflow

alizedebray:


.github/workflows/fetch-icons.yaml
```diff
@@ -61,8 +61,8 @@ jobs:
 
       - name: Commit Changes
         run: |
-          git config --global user.name "swisspost-bot"
-          git config --global user.email "oss@post.ch"
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
```
gfellerph, line 65: Why did you change this? If we're using the swisspost-bot token, it would seem more logical to also use it's credentials, no?

alizedebray: I changed the credentials to match the release workflow: the commit is made by the [github-actions[bot]](https://github.com/swisspost/design-system/commits?author=github-actions%5Bbot%5D) and the PR is created by the [swisspost-bot](https://github.com/swisspost-bot). 

To me this makes sense because it is actually the GitHub actions that write the changes. What do you think?

gfellerph (approved):

PR #2706, opened 2024-02-21
fix(styles): Badge + switch checked state in HCM

imagoiq:


packages/styles/src/components/form-check.scss
```diff
@@ -158,7 +158,7 @@
       border: form-check.$form-check-input-border-width solid CanvasText;
 
       &:checked {
-        background-color: CanvasText;
+        background-color: SelectedItem;
```
imagoiq, line 161: @alizedebray this is what I was looking for. It makes not difference in Firefox though with Highlight color, but it does make one in blink-based browser.

packages/styles/src/mixins/_badge.scss
```diff
@@ -7,7 +7,10 @@
   border-color: transparent;
 
   @include utilities.high-contrast-mode() {
+    background-color: Highlight;
     border-color: Highlight;
+    color: HighlightText;
+    forced-color-adjust: none; // Disable "readability backplate" on blink browser that interferes with the colors on this case
```
imagoiq, line 13: About this comment, this is the related thingy: https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/Accessibility/HighContrast/explainer.md#ensuring-readability

Basically without forced-color-adjust, you have this intermediate box with a contrasted background-color. This is great in principle, but if you provide a color as I did (HighlightText for color and Highlight for background), the contrast will be inverted on blink-based browser, meanwhile it will work on other browsers not using this "backplate" technic. So instead of targetting a specific browser, we can disable this "backplate" with this property.

oliverschuerch (approved):
It could be a good idea to use the `SelectedItem` value also for the checkbox and radio button elements. For now, we only use `Highlight` but we do not distinguish between the hover and the checked state.

PR #2748, opened 2024-03-04
chore(documentation): Move versions.json from archive to current documentation

imagoiq:


imagoiq:
Versions.json is available at https://preview-2748--swisspost-design-system-next.netlify.app/assets/versions.json so it seems correct

gfellerph (approved):

PR #2774, opened 2024-03-11
fix(styles): Make styling more coherent for forms elements with High contrast mode and focus/hover state

imagoiq:


packages/styles/src/components/form-check.scss
```diff
@@ -77,6 +77,13 @@
       &:indeterminate::after {
         @include icons-mx.icon(2039);
       }
+
+      &:checked,
+      &:indeterminate {
+        @include utility-mx.high-contrast-mode {
+          background-color: SelectedItem;
```
gfellerph, line 84: image

Contrast of the check mark gets really low when setting the background color. I think I'd prefer just setting the border color here (or updating the color of the checkmark).

gfellerph (changes requested):

packages/styles/src/components/form-check.scss
imagoiq: Yes, and it also hide for radio the selected state. Here is a fix.

gfellerph: Checkboxes are good now, the only thing remaining would be the checkbox card, where I can't see the check mark:
image

gfellerph (changes requested):

packages/styles/src/components/form-check.scss
imagoiq: Good catch. Those card control uses more class to define background and color, so I increased specificity for HCM style.

packages/styles/src/components/form-check.scss
```diff
@@ -155,7 +178,7 @@
       border: form-check.$form-check-input-border-width solid CanvasText;
 
       &:checked {
-        background-color: SelectedItem;
+        background-color: SelectedItem !important;
```
gfellerph, line 181: Sorry to pick at this again, but the contrast of the ckecked radio button is low against the black background in this scenario. I believe we should aim for the colors the native elements have
image
vs.
image

gfellerph (changes requested):

packages/styles/src/components/form-check.scss
imagoiq: In my opinion, we should not match native elements. In the past, it has proven that it has some flaws, for example with focus style. 

I used the `SelectedItem` which, [according to W3C](https://www.w3.org/TR/css-color-4/#css-system-colors) can be used as a background for checkbox:

image

In Firefox, the color is almost the same as the `Highlight` (if you look at the W3C list, you'll notice the slight difference as well). You might think this is only because it's not supported, but it seems that it's not the case [according to the tests of the W3C](https://wpt.fyi/results/css/css-color/system-color-support.html?label=experimental&label=master&aligned).

### Browser comparison

On Blink browser (with focus on selected radio): 

msedge_H0lYd3W5wI

On Firefox (with focus on selected radio): 

firefox_H4iMHLAx31

So, in my opinion, we should let browsers/OS handle the color, but use the appropriate one from specification.

gfellerph: This makes sense, although I'm not really happy to release something if we know it's not going to work for everyone.

gfellerph (approved):

packages/styles/src/components/form-select.scss
```diff
@@ -9,14 +9,43 @@
...
+.form-select-wrapper {
+  @include utilities.focus-style;
+}
+
+:not(.form-select-wrapper) > {
+  .form-select {
+    @include utilities.focus-style;
+  }
+}
+
```
imagoiq, line 21: We can either do it like that or add a breaking change so that user needs to add a wrapper element

alizedebray:
Checked checkboxes do not have a border on Firefox currently. Also we could maybe drop the color transition on HCM since the check icon changes from white to black some ms after the checkbox is clicked.

alizedebray (changes requested):
Checked checkboxes do not have a border on Firefox currently.
Also it would be good to drop the color transition on HCM since the check icon changes from white to black some ms after the checkbox is clicked.

imagoiq:
@oliverschuerch Can you check as well? As there are a lot of few changes, it's better to have more reviewers.

alizedebray (approved):

PR #2777, opened 2024-03-11
fix(components): Post-tabs cross-root aria-id

imagoiq:
Workaround or alternative with light DOM (but not ready to use): https://github.com/swisspost/design-system/pull/2776

### DIFF
--- a/packages/components/src/components/post-tab-header/post-tab-header.tsx
+++ b/packages/components/src/components/post-tab-header/post-tab-header.tsx
@@ -32,17 +32,15 @@ export class PostTabHeader {
 
   render() {
     return (
-      <Host data-version={version}>
-        <button
-          aria-selected="false"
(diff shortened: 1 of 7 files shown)


packages/components/src/components/post-tab-header/post-tab-header.tsx
```diff
@@ -32,17 +32,15 @@ export class PostTabHeader {
...
-        </button>
+      <Host
+        id={this.tabId}
+        role="tab"
+        data-version={version}
+        aria-selected="false"
+        tabindex="-1"
+        class="tab-title"
+      >
+        <slot />
```
imagoiq, line 43: Button had to be removed because of https://dequeuniversity.com/rules/axe/4.7/nested-interactive?application=axeAPI

gfellerph (approved):

PR #2779, opened 2024-03-12
feat(documentation): add docs table for post-accordion-item

davidritter-dotcom:


packages/documentation/src/stories/components/accordion/accordion.docs.mdx
```diff
@@ -13,8 +14,14 @@ It is used to manage how content is shown and hidden across a set of collapsible
 
 <Canvas sourceState="shown" of={accordionStories.Default} />
 
+## `<post-accordion>`
+
 <Controls of={accordionStories.Default} />
 
+## `<post-accordion-item>`
+
+<Controls of={accordionItemStories.Default} />
```
imagoiq, line 23: It would be nice if you could control as well from this table the same Canvas, but I don't think you can do it.

imagoiq (approved):

packages/documentation/src/stories/components/accordion/accordion.docs.mdx
davidritter-dotcom: I tried it, but I couldn't find a solution that works.

PR #2781, opened 2024-03-13
feat(styles): Input range progress webkit box-shadow workaround

imagoiq:
### What

So at first, there was already a hack to display the progress of the slider on Webkit: https://codepen.io/thebabydino/pen/goYYrN
It uses JavaScript to update the background-color with the input value and calculate a width ratio.
The FormRangeDirective was responsible for doing that and there was a recommendation to use it in the demo app. This work great, but need JavaScript and a MutationObserver, etc.

An alternative would be to use ask the user to set a width to the input and use it like in this example: https://stackoverflow.com/a/72837476

I think it's also something we don't want to define explicitly and let the user have a responsive input.

The workaround in this PR is using the same box-shadow approach, but with a clip-path. It is based on this https://antvil.github.io/css-sos/sos/progress/
While it might not be pixel perfect (for example, there is no border-radius on the progress and there is a gap between the thumb and the progress), I think it's the best progressive enhancement approach. By that, I mean that several libraries (like Bootstrap, Shoelace) chose not to display any progress bar at all, so the progress bar in itself might not be essential.

### Options for decision poll

1. Not pixel-perfect: we accept this isn't pixel perfect on all browsers
2. More research: we want to spend more time searching for other solutions
3. Javascript: design need to be coherent at all costs, so we should rather use the JavaScript solution
4. Discuss with design: we might want to change the design to get the same focus design like checkboxes, radio or rating for example. (but I guess that with two handles it wouldn't not work)
5. No progress bar: Is progress bar really a thing?

What about dashed border? Should it be "pixel-perfect" as well (you cannot control the number of dashes, and it might be different browser from browser. The alternative would be to use an SVG: https://kovart.github.io/dashed-border-generator/).


packages/styles/src/components/form-range.scss
oliverschuerch, line 1: The clip-path is a great idea!
I've played around with it a little bit and come up with the following suggestion:

```scss
.form-range {
  &::-webikit-slider-thumb {
    clip-path: polygon(
      0 calc(50% - 2px),
      1px calc(50% - 6px),
      1px 0,
      32px 0,
      32px 32px,
      1px 32px,
      1px calc(50% + 6px),
      0 calc(50% + 2px),
      -100vw calc(50% + 2px),
      -100vw calc(50% - 2px)
    );
  }
}

&:not(:disabled, .disabled) {
  &::-webkit-slider-thumb {
    box-shadow: calc(-100vw - 32px) 0 0 100vw #000;
  }
}
```

or with variables (not tested)

```scss
$track-height: 4px;
$webkit-thumb-width: 32px;

.form-range {
  &::-webikit-slider-thumb {
    clip-path: polygon(
      0 calc(50% - $track-height * 0.5),
      1px calc(50% - #{$track-height * 0.5 + 4px}),
      1px 0,
      $webkit-thumb-width 0,
      $webkit-thumb-width $webkit-thumb-width,
      1px $webkit-thumb-width,
      1px calc(50% + #{$track-height * 0.5 + 4px}),
      0 calc(50% + #{$track-height * 0.5}),
      -100vw calc(50% + #{$track-height * 0.5}),
      -100vw calc(50% - #{$track-height * 0.5})
    );
  }
}

&:not(:disabled, .disabled) {
  &::-webkit-slider-thumb {
    box-shadow: calc(-100vw - $webkit-thumb-width) 0 0 100vw color.$black;
  }
}
```

Please give it a try and tell me what you think ;)

oliverschuerch (approved):

packages/styles/src/components/form-range.scss
imagoiq: Look good, let's go with your improvement., thanks!

PR #2783, opened 2024-03-13
fix(docs, demo, migrations): update migration instructions

alizedebray:


packages/demo/src/app/home/home.component.html
```diff
@@ -314,7 +314,9 @@ <h4 class="h5 bold">Automatic Migration ⚙️</h4>
                 appCopyToClipboard
                 class="d-block mt-1 p-3"
                 [languages]="['bash']"
-                highlight="npx ng update @swisspost/design-system-migrations --from=5 --to=6 --migrate-only"
+                highlight="npm install @swisspost/design-system-migrations
```
imagoiq, line 317: Did you try to use npx to have temporarily the package on the system rather than install/uninstall it?

imagoiq (changes requested):

packages/demo/src/app/home/home.component.html
alizedebray: Yes, I tried `npx -p @swisspost/design-system-migrations ng update @swisspost/design-system-migrations --from=5 --to=6 --migrate-only --allow-dirty`.
With this command the migration package is temporarily installed yet Angular CLI cannot find it.

imagoiq: Nope, then let's go like this.

imagoiq (approved):

PR #2794, opened 2024-03-14
feat(documentation): Add Link to Figma for Alert and Button component

imagoiq:
As it's a prototype and it involves design a bit, we can discuss it in the patterns meeting.


packages/documentation/src/utils/figma-link.ts
```diff
@@ -0,0 +1,6 @@
+const fileId = 'xZ0IW0MJO0vnFicmrHiKaY';
+const figmaMode: 'design' | 'dev' = 'dev';
+
+export function getComponentFigmaLink(figmaNodeId: string) {
+  return `https://www.figma.com/file/${fileId}/Components-Post?type=design&node-id=${figmaNodeId}&mode=${figmaMode}&t=j3o3HMwq5nigL5Mj-4`;
```
imagoiq, line 5: Unfortunately, it seems there is no way to zoom-in into the frame (see for example https://forum.figma.com/t/link-to-frame-fit-the-frame-to-screen/35680, there are multiple requests from the community).

packages/documentation/src/stories/components/alert/standard-html/alert.stories.ts
```diff
@@ -17,6 +18,10 @@ const meta: MetaComponent = {
     controls: {
       exclude: ['Title', 'Content'],
     },
+    design: {
+      type: 'figma',
+      url: getComponentFigmaLink('17001-2244'),
+    },
```
imagoiq, line 24: For compatibility, I used the same schema as official figma addons (even if it's rather per story than per docs page): the https://storybook.js.org/docs/sharing/design-integrations#link-figma-components-to-stories

packages/documentation/.storybook/addons/figma-link/FigmaLink.tsx
```diff
@@ -0,0 +1,23 @@
...
+import logoFigma from '../../../public/assets/images/technologies/logo-figma.svg';
+import { getTitleFromPath } from '../../../src/utils';
+
+function FigmaLink() {
+  const api = useStorybookApi();
+  const designParameter = api.getCurrentParameter<{ type: string; url: string }>('design') || [];
+  const titlePath = api.getCurrentStoryData()?.title;
+
+  return designParameter?.type === 'figma' && designParameter?.url ? (
+    <a className="figma-link btn btn-primary btn-sm btn-icon" href={designParameter.url}>
```
imagoiq, line 13:
```suggestion
    <a className="figma-link btn btn-secondary btn-sm btn-icon" href={designParameter.url}>
```
Perhaps, it's slightly better with secondary type.

packages/documentation/src/shared/link-design/link-design.component.ts
```diff
@@ -0,0 +1,41 @@
...
+import { ModuleExport } from '@storybook/types';
+
+interface DesignParameter {
+  type: 'figma';
+  url: string;
+}
+
+@customElement('link-design')
+export class LinkDesignComponent extends LitElement {
+  @property({ type: Object }) of: ModuleExport = {};
```
imagoiq, line 14: ModuleExport (which is used in `<Meta />` is equal to any at the moment, but it might change

imagoiq:
Ready for review with opinion from today pattern's discussion.

alizedebray (approved):

PR #2807, opened 2024-03-18
fix(internet-header): Incomplete multi-character sanitization Code Scanning alert on hours field of internet-header footer

imagoiq:


packages/internet-header/src/components/post-internet-footer/components/post-footer-block-contact.component.tsx
```diff
@@ -44,8 +42,8 @@ export const PostFooterBlockContact = (props: {
               {content.text ? <p class="text">{content.text}</p> : null}
               {content.hours && isLiveSupport && <LiveSupport hours={content.hours} />}
               {content.hours && !isLiveSupport && (
-                // Some values arrive in the form of <p>8&emdash;12</p> and without replace and innerHTML, tags get rendered as text (project="klp" language="en" environment="int02")
-                <p class="hours" innerHTML={getContentHours(content.hours)}></p>
+                // Some values arrive in the form of <p>8&mdash;12</p> and without replace and innerHTML, tags get rendered as text (project="klp" language="en" environment="int02")
```
imagoiq, line 45: &emdash; is not decoded with any library. It seems to never have existed (at least officially), see: https://bugzilla.mozilla.org/show_bug.cgi?id=221606

packages/internet-header/src/components/post-internet-footer/components/post-footer-block-contact.component.tsx
```diff
@@ -1,8 +1,10 @@
 import { h } from '@stencil/core';
 import { BlockEntity } from '../../../models/footer.model';
 
-const getContentHours = (hours: string) => hours.replace(/<[^>]*>?/gm, '');
-
+function stripHtml(html: string): string {
+  const doc = new DOMParser().parseFromString(html, 'text/html');
```
imagoiq, line 5: DOMParser().parseFromString is not executing any script. It can still contain malicious code like onclick event that will work if it's not santitize inside the DOMParser and that the result is injected bad in the current DOM document. In our case, it's safe because using "textContent" will always allow only text. Also, this technic is used by [DOMPurify library](https://github.com/cure53/DOMPurify/tree/main). 

More infos:
* [Discussion about if this technic is safe or not in web-dev repository](https://github.com/GoogleChrome/web.dev/issues/6890)
* [Example of how to sanitize HTML with this technic](https://gomakethings.com/how-to-sanitize-html-strings-with-vanilla-js-to-reduce-your-risk-of-xss-attacks/)

gfellerph (approved):
Today I learned, thanks for the in-depth explanations!

PR #2808, opened 2024-03-18
chore(styles): Replace $success with $success green

imagoiq:


packages/styles/src/functions/_color.scss
gfellerph, line 1: Nice!

gfellerph (approved):

PR #2810, opened 2024-03-18
fix(styles): Hide focus state when unnecessary and make design more coherent

imagoiq:
Currently, forms elements like checkbox, radio, switch show focus styles on `:focus` state while others like buttons show only on `:focus-visible` state. This is an improvement using `:has` to make the design more coherent and simulate `:focus-visible` instead of using `:focus-within` as there is no `:focus-visible:within` unfortunately.

We'll quickly discuss it on the pattern meeting.


packages/styles/src/mixins/_utilities.scss
```diff
@@ -100,10 +101,34 @@
...
+      outline-offset: spacing.$size-line;
+      outline: spacing.$size-line solid var(--post-focus-color);
+      border-radius: $border-radius;
+
+      @include high-contrast-mode() {
+        outline-color: Highlight;
+      }
+
+      // In case rules need to be slightly adjusted
+      @content;
```
imagoiq, line 117: This is a duplicated code, but I don't know how to improve that situation. We cannot use a placeholder here and the `@content` restrict the usage of another @mixin

packages/styles/src/mixins/_utilities.scss
```diff
@@ -88,7 +88,8 @@
 }
 
 @mixin focus-style($border-radius: commons.$border-radius, $vendor-prefix: '') {
-  &:is(:focus-visible, :focus-within, .pretend-focus)#{$vendor-prefix} {
+  // :has(:focus-visible) mimic a focus-visible-within pseudo-class
+  &:is(:focus-visible, :has(:focus-visible), .pretend-focus)#{$vendor-prefix} {
     outline-offset: spacing.$size-line;
```
gfellerph, line 93: If you want to get rid of that duplication as much as possible, you could use the long form definitions:

```scss
@mixin focus {
  outline-style: none;
  outline-offset: 2px;
  outline-width: 2px;
  outline-color: blue;
  
  &:where(:has(:focus-visible), :focus-visible) { outline-style: solid; @content; }
  @supports not selector(:has) { &:where(:focus-within) { outline-style: solid; @content; } }
  @include high-contrast-mode() {
     outline-color: Highlight;
  }
```

gfellerph (changes requested):

packages/styles/src/mixins/_utilities.scss
```diff
@@ -88,7 +88,8 @@
 }
 
 @mixin focus-style($border-radius: commons.$border-radius, $vendor-prefix: '') {
-  &:is(:focus-visible, :focus-within, .pretend-focus)#{$vendor-prefix} {
+  // :has(:focus-visible) mimic a focus-visible-within pseudo-class
+  &:is(:focus-visible, :has(:focus-visible), .pretend-focus)#{$vendor-prefix} {
```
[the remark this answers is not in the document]
imagoiq: For the vendor prefix, it is used here: https://github.com/swisspost/design-system/pull/2781/files#diff-f8aaec5942d0b7bdec7264460de3a7becec65dfa3ca88797a0875320dade340fR84
Previously it was used on more lines. Sure it look a bit odd, but it looks practical right now.

The border-radius could have unwanted side effects, you are right, but we need it to shape the outline. Perhaps, we want to add the outline into a `::after` instead?

packages/styles/src/mixins/_utilities.scss
imagoiq: Good idea, I did that except for hcm where specificity is easier to handle like that. Also `@supports not selector(:has)` is not working.

gfellerph (approved):

PR #2812, opened 2024-03-19
feat(documentation): accessibility docs disabled state

davidritter-dotcom:


packages/documentation/src/stories/foundation/accessibility/accessibility.docs.mdx
```diff
@@ -25,8 +27,94 @@ TODO: write accessibility guide about form labels (visible and invisible), aria-
...
+      <td><Canvas of={AccessibilityStories.FormBadExample}/></td>
+      <td><div class="mt-4"><Source code={formBadSample} language="ts" /></div></td>
+    </tr>
+     <tr>
+      <td colSpan="2" class="fw-bold">Form good example</td>
+    </tr>
+    <tr>
+      <td><Canvas of={AccessibilityStories.FormGoodExample}/></td>
+      <td><div class="mt-4"><Source code={formGoodSample} language="ts" /></div></td>
+    </tr>
```
oliverschuerch, line 115: It's a bit sad how much space we loose, because of the fact, that the actual form and the validator code example are shown in cells beside each other:
image

1. Remove the `<div class="container">...</div>` element which wraps the form.
2. Try to place the good and bad examples beside each other.
3. Try to add the validator code example into a script tag, within the HTML example, so it's only visible if the user expandes the code from story-panel.
I'm not even sure, if the visualisation of this code adds any value to the example. Consider removing them completely, as DEVs should understand the point, by playing around with the story examples itself.

```html
<tr>
  <td class="fw-bold">Form bad example</td>
  <td class="fw-bold">Form good example</td>
</tr>
<tr>
  <td><Canvas of={AccessibilityStories.FormBadExample}/></td>
  <td><Canvas of={AccessibilityStories.FormGoodExample}/></td>
</tr>
```

If the above changes do not lead to a more proper visualisation, try to move the examples out of the table and add them to the HTML directly after the table.

oliverschuerch (changes requested):

packages/documentation/src/stories/foundation/accessibility/accessibility.docs.mdx
oliverschuerch: I think we are still not there.
Can you please move the "Form bad" and "Form good" examples out of the table and implement one after another?

And by the way, please add the `novalidate` attribute to the form tags, to prevent native error messages show up.

oliverschuerch (approved):

PR #2813, opened 2024-03-19
fix(styles): Update disabled styles for form-check (except switch)

imagoiq:


packages/styles/src/variables/components/_form-check.scss
```diff
@@ -14,7 +14,27 @@ $form-check-input-size: 1.375rem !default;
 $form-check-input-size-sm: spacing.$size-regular !default;
 $form-check-input-border-width: forms.$input-border-width !default;
 $form-check-input-focus-box-shadow: forms.$input-focus-box-shadow !default;
+$form-check-input-disabled-background-url-light: "data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg'%3e%3crect width='100%25' height='100%25' fill='none' stroke='" +
+  color.$gray-60 + "' stroke-width='" + $form-check-input-border-width
```
gfellerph, line 18: Very nitpicky, but it looks like the border width is 1px, in Design it's 1.5px for the disabled state and 2px for regular borders. 
image
Asked for clarification (https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY?node-id=17960:14667&mode=design#742433222)

imagoiq: True, I've adjusted the width

packages/styles/src/variables/components/_form-check.scss
```diff
@@ -14,7 +14,28 @@ $form-check-input-size: 1.375rem !default;
 $form-check-input-size-sm: spacing.$size-regular !default;
 $form-check-input-border-width: forms.$input-border-width !default;
 $form-check-input-focus-box-shadow: forms.$input-focus-box-shadow !default;
+$form-check-input-disabled-svg-border-width: $form-check-input-border-width * 2;
+$form-check-input-disabled-background-url-light: "data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg'%3e%3crect width='100%25' height='100%25' fill='none' stroke='" +
+  color.$gray-60 + "' stroke-width='" + $form-check-input-disabled-svg-border-width
+  "' stroke-dasharray='4 2' stroke-vector-effect='non-scaling-stroke' linecap='square'/%3e%3c/svg%3e"; // Source: https://kovart.github.io/dashed-border-generator/ + added stroke-vector-effect
```
oliverschuerch, line 20: `linecap="square"` is for paths, not strokes, but if you add `stroke-linecap="square"` you have to calculate the width of the stroke within every visible dash value, so remove this or change it to its default, which is `stroke-linecap="butt"`.

The dasharray below works only for the exact size of the given checkbox, which is 22x22 pixel.

```suggestion
  "' stroke-dasharray='2 1.5 4 1.5 4 1.5 4 1.5 2 3.5 4 1.5 4 1.5 4 1.5 2 3.5 4 1.5 4 1.5 4 1.5 2 3.5 4 1.5 4 1.5 4' stroke-vector-effect='non-scaling-stroke' /%3e%3c/svg%3e"; // Source: https://kovart.github.io/dashed-border-generator/ + added stroke-vector-effect
```

packages/styles/src/variables/components/_form-check.scss
```diff
@@ -14,7 +14,28 @@ $form-check-input-size: 1.375rem !default;
 $form-check-input-size-sm: spacing.$size-regular !default;
 $form-check-input-border-width: forms.$input-border-width !default;
 $form-check-input-focus-box-shadow: forms.$input-focus-box-shadow !default;
+$form-check-input-disabled-svg-border-width: $form-check-input-border-width * 2;
+$form-check-input-disabled-background-url-light: "data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg'%3e%3crect width='100%25' height='100%25' fill='none' stroke='" +
```
oliverschuerch, line 18: For me it's unclear why there are numbers behind the percentage character. And as far as I can see, they don't do anything.
```suggestion
$form-check-input-disabled-background-url-light: "data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg'%3e%3crect width='100%' height='100%' fill='none' stroke='" +
```

oliverschuerch (changes requested):

packages/styles/src/variables/components/_form-check.scss
imagoiq: It's percent encoding: https://en.wikipedia.org/wiki/Percent-encoding#Reserved_characters
Looks fine to me.

packages/styles/src/variables/components/_form-check.scss
imagoiq: @oliverschuerch I probably made a mistake when copy-paste or optimizing with svgo, I don't know. How did you generate this dasharray (just by curiosity)? We can make it work by using [`pathLength` attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/pathLength), but the visual space is quite full, so I made another pattern for the small variant

packages/styles/src/components/form-check.scss
```diff
@@ -49,10 +50,11 @@
     }
 
     @include utility-mx.focus-style-none();
+
     display: inline-flex;
     flex: 0 auto;
     appearance: none;
-    background: color.$white;
+    background: transparent;
```
imagoiq, line 57: Checkboxes are not visible on dark background. This change was done for card-control

packages/styles/src/variables/components/_form-check.scss
oliverschuerch: The new pattern looks great on a disabled checkbox, both small and large variant.
But for the radio button the outline is not dashed when disabled.

oliverschuerch (approved):
Please have a look at the disabled radio button before merging!

packages/styles/src/variables/components/_form-check.scss
imagoiq: Oops on Chrome, the false value is interpreted as a 0 and infinity. Interesting.

PR #2814, opened 2024-03-19
fix(components): post-tooltip doesn't show up when the pointer is on a child element

imagoiq:
Animation2 4

### DIFF
--- a/packages/components/src/components/post-tooltip/post-tooltip.tsx
+++ b/packages/components/src/components/post-tooltip/post-tooltip.tsx
@@ -15,6 +15,7 @@ import { getAttributeObserver } from '../../utils/attribute-observer';
 let tooltipInstances = 0;
 let hideTooltipTimeout: number = null;
 const tooltipTargetAttribute = 'data-tooltip-target';
+const tooltipTargetAttributeSelector = `[${tooltipTargetAttribute}]`;
 
 /**
  * Global event listener to show tooltips. This is globalized so that triggers that are rendered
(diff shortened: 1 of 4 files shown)


packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -23,12 +24,14 @@ const tooltipTargetAttribute = 'data-tooltip-target';
  * @returns
  */
 const globalInterestHandler = (e: PointerEvent | FocusEvent) => {
-  const target = e.target as HTMLElement;
-  if (!target || !('getAttribute' in target)) return;
-  const tooltipTarget = target.getAttribute(tooltipTargetAttribute);
+  const targetElement = (e.target as HTMLElement).closest(
```
imagoiq, line 27: Using `e.currentTarget` is an option as well, but it will move the tooltip to the child element which is unwanted.

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -23,12 +24,14 @@ const tooltipTargetAttribute = 'data-tooltip-target';
...
+    tooltipTargetAttributeSelector,
+  ) as HTMLElement;
+  if (!targetElement || !('getAttribute' in targetElement)) return;
+  const tooltipTarget = targetElement.getAttribute(tooltipTargetAttribute);
   if (!tooltipTarget || tooltipTarget === '') return;
   const tooltip = document.getElementById(tooltipTarget) as HTMLPostTooltipElement;
-  tooltip?.show(target);
+  void tooltip?.show(targetElement);
```
imagoiq, line 34: Just to make intellij happy.

gfellerph (approved):
Thanks for the fix.

PR #2816, opened 2024-03-19
chore(documentation): Add link to intranet-header archived doc

imagoiq:


gfellerph (approved):

gfellerph:
In theory, this change would need a changeset, otherwise the docs won't get published to prod. But there are enough other changesets that the docs will get published (they only do if the version number in the documentation package.json is being bumped).

PR #2818, opened 2024-03-20
fix(documentation): Product card with dark background do not render correct color

imagoiq:


packages/documentation/src/stories/components/cards/product-card/product-card.stories.ts
```diff
@@ -217,7 +217,7 @@ export const Multipart: Story = {
             </dl>
           </div>
 
-          <div class="card-body">
+          <div class="card-body bg-light">
```
imagoiq, line 220: It uses similar classes as other parts like line 205

gfellerph (approved):

PR #2825, opened 2024-03-20
fix(documentation): Floating label class is added on input even when input has no floating label

imagoiq:


gfellerph (commented):
@imagoiq where exactly does this happen? I can't seem to reproduce on https://design-system.post.ch

imagoiq:
> @imagoiq where exactly does this happen? I can't seem to reproduce on https://design-system.post.ch

Only in the last story, here: https://design-system.post.ch/?path=/docs/2df77c32-5e33-402e-bd2e-54d54271ce19--docs#validation

gfellerph (approved):

PR #2834, opened 2024-03-20
fix(intranet-header): "tabIndex" should only be declared on interactive elements

imagoiq:


.changeset/polite-starfishes-vanish.md
```diff
@@ -0,0 +1,8 @@
+---
+'@swisspost/design-system-intranet-header': minor
+'@swisspost/design-system-documentation': minor
+'@swisspost/design-system-styles': minor
+'@swisspost/design-system-demo': minor
+---
+
+Changed mobile navigation trigger to make it more accessible by default using a button element.
```
imagoiq, line 8: I don't know which package to involve.

It's a minor I think because it's a good change on the behavior. This is not a breaking change. Without a button, the layout is slightly changed to have a smaller padding, but it is still usable.

gfellerph: I think it's ok for this change.

gfellerph (approved):

PR #2838, opened 2024-03-21
fix(demo): Version-switcher is obscured and do not display button label

imagoiq:


packages/demo/src/app/app.component.html
```diff
@@ -7,13 +7,8 @@
       <button class="btn btn-link dropdown-button" id="versionDropdownButton" ngbDropdownToggle>
         {{ (currentVersion$ | async)?.title }}
       </button>
-      <div ngbDropdownMenu aria-labelledby="versionDropdown">
-        <a
-          *ngFor="let version of versions$ | async"
-          [href]="isLatest ? 'https://design-system.post.ch' : version.url"
-          [class.active]="version.isCurrent"
```
imagoiq, line 14: IsLatest and version.isCurrent do not exist. I simplify a bit by removing `[class.active]` as this documentation is already archived, we just need a usable version switcher.

alizedebray (approved):

PR #2845, opened 2024-03-21
fix(styles): Alert actions button color contrast was not met with success background color

imagoiq:
### What

From design files, text color of alert actions should be white on dark background (currently, it's 80% white as it's based on the `--post-black-80` variable, but inverted). To define another color, I needed to add the alert classes and selector (for the web component) into the dark-backgrounds variables. The notification name list was moved to color file to avoid recursion issue (perhaps we could put it somewhere else).

I'll check with design what to do with hover style.


packages/styles/src/variables/_color.scss
```diff
@@ -131,15 +131,41 @@ $background-colors: map.merge(
   )
 );
 
+$notification-colors: () !default;
+$notification-colors: list.join(
+  $notification-colors,
+  (
+    'primary' $gray-80,
+    'success' $success-green,
+    'danger' $error,
```
alizedebray, line 140:
```suggestion
    'danger' $error-red,
```

packages/styles/src/variables/_color.scss
```diff
@@ -131,15 +131,41 @@ $background-colors: map.merge(
...
+    'warning' $warning,
```
alizedebray, line 141:
```suggestion
    'warning' $warning-orange,
```

packages/styles/src/variables/components/_notification.scss
```diff
@@ -70,20 +70,14 @@ $notification-icon-size-map: map.merge(
...
-    'danger' color.$error 2104,
-    'warning' color.$warning 2104,
-    'info' color.$info 2106,
-    'gray' color.$gray-background 2201,
-    // deprecated
-    'notification' color.$gray-80 3134,
-    'error' color.$error 2104
-  )
-);
+$notification-icons: (3134, 2105, 2104, 2104, 2106, 2201, 3134, 2104);
```
alizedebray, line 73: A list makes it difficult to know which color corresponds to which icon.

alizedebray (changes requested):

packages/styles/src/variables/_color.scss
imagoiq: Good catch, I made another PR: https://github.com/swisspost/design-system/pull/2869

packages/styles/src/variables/components/_notification.scss
imagoiq: Update with our discussion's output

PR #2855, opened 2024-03-27
feat(styles, docs, migrations): create the chip component

alizedebray:


oliverschuerch (approved):

oliverschuerch (changes requested):
The nested badgesdo not look like in Figma:
image
image
CSS:
image

What about the HCM styles, will they be added in a separate ticket?
What about the migration guide, will the information be added in a separate ticket?

alizedebray:
> The nested badgesdo not look like in Figma:

Rouven was changing them yesterday just after our discussion.
Since this PR is necessary for the v7 release, we decided together with Philipp and Loïc to merge it as soon as the structure is correct and update the styles as a patch later on (ticket #2918)

> What about the HCM styles, will they be added in a separate ticket?

Same as above, the HCM styles are working for now and we can still update them in a future patch  (ticket #2919) 

> What about the migration guide, will the information be added in a separate ticket?

The migration guide is already merged (PR https://github.com/swisspost/design-system/pull/2915)

oliverschuerch (approved):

PR #2856, opened 2024-03-27
chore(versioning): fix versions of main packages

gfellerph:
This will start to align the version of the fixed packages together. Fixed packages will be published under a new version, even if they did not have any changes (https://github.com/changesets/changesets/blob/main/docs/fixed-packages.md).

There are several packages, that are not included in the list because they should still be versioned individually. The reason for this is, that they are not main deliverables of the Design System - styles and components. For example the documnetation, which can only be released with a version bump. If it would be fixed with the other packages, a correction of a typo would cause a version bump for all the deliverables, even if nothing changed for end users.

### DIFF
--- a/.changeset/fifty-rabbits-flow.md
+++ b/.changeset/fifty-rabbits-flow.md
@@ -4,6 +4,8 @@
 '@swisspost/design-system-components-react': major
 '@swisspost/design-system-styles': major
 '@swisspost/design-system-migrations': major
+'@swisspost/design-system-icons': major
+'@swisspost/design-system-intranet-header': major
 ---
 
 Syncing versions of the following packages:
@@ -13,5 +15,7 @@ Syncing versions of the following packages:
 - @swisspost/design-system-components-react
 - @swisspost/design-system-components-angular
 - @swisspost/design-system-migrations
+- @swisspost/design-system-icons
+- @swisspost/design-system-intranet-header
 
 This will help understanding the dependencies between these packages at a glance but also means that for the individual pacakges, semver is no longer being used. This enables us also to talk about and document Design System versions as a whole instead of documenting the fragmented versions in a complex lookup table.
--- a/.changeset/fifty-rabbits-flow.md
+++ b/.changeset/fifty-rabbits-flow.md
@@ -3,6 +3,7 @@
 '@swisspost/design-system-components-angular': major
 '@swisspost/design-system-components-react': major
 '@swisspost/design-system-styles': major
+'@swisspost/design-system-migrations': major
 ---
 
 Syncing versions of the following packages:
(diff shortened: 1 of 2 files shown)


oliverschuerch (commented):
I would suggest to align the following packages as well:
- `@swisspost/design-system-icons` (because it is a released package, dependancy of the styles package)
- `@swisspost/internet-header` (because it is a released package, dependant on the styles package and could become dependant on the components package in the future)
- `@swisspost/design-system-intranet-header` (because it is a released package, dependant of the styles package)
- `@swisspost/design-system-documentation` (because it documents the packages with a specific version)
Even if this means we must add a version bump for all packages to fix a typo in the docs. I don't think we will ever do that, without any changes in at least one other package, so it's fine.

gfellerph:
> I would suggest to align the following packages as well:
> 
> * `@swisspost/design-system-icons` (because it is a released package, dependancy of the styles package)
> * `@swisspost/internet-header` (because it is a released package, dependant on the styles package and could become dependant on the components package in the future)
> * `@swisspost/design-system-intranet-header` (because it is a released package, dependant of the styles package)
> * `@swisspost/design-system-documentation` (because it documents the packages with a specific version)
>   Even if this means we must add a version bump for all packages to fix a typo in the docs. I don't think we will ever do that, without any changes in at least one other package, so it's fine. And this would enable us to add a version switch to the docs with miner versions, so you can navigate to the exact design-system version you have installed in your project).

- agree for icons
- the internet header is a special package and I'd like to keep strict semver on this because when we're updating to a major styles version internally, it has only patch consequences for projects. If there is a second version with tighter component integration, let's discuss again
- agree for the intranet header
- disagree for the documentation, I'd like to keep the flexibility there, also for doing updates to past major release doc pages. We already have the versions page and the docs package version is almost always irrelevant. I don't see how syncing the versions will help us add minor entries to the version switch (or the real benefit of that feature).

.changeset/fifty-rabbits-flow.md
```diff
@@ -0,0 +1,17 @@
...
+'@swisspost/design-system-migrations': major
+---
+
+Syncing versions of the following packages:
+
+- @swisspost/design-system-styles
+- @swisspost/design-system-components
+- @swisspost/design-system-components-react
+- @swisspost/design-system-components-angular
+- @swisspost/design-system-migrations
```
oliverschuerch, line 15: Don't we need to add newly added synced packages (intranet-header, icons) here as well?

oliverschuerch (approved):

.changeset/fifty-rabbits-flow.md
gfellerph: true, forgot about that

PR #2857, opened 2024-03-27
feat(documentation, styles): Update textarea styles and add floating label small variant

davidritter-dotcom:
I updated the styles to be as close to figma as possible, but since some paddings in figma don't match any post-size, there will be some differences.


oliverschuerch (changes requested):
Please follow this discussion: https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?type=design&node-id=7779-34916&mode=design&t=MZFCh0b62SndmXX9-0

oliverschuerch (approved):

PR #2865, opened 2024-03-28
feat(documentation, styles): Add animated button icon on the left

davidritter-dotcom:


packages/styles/src/components/button.scss
```diff
@@ -134,22 +134,48 @@
 }
 
 // Animated
+.btn-animated,
+.btn-animated-left {
```
alizedebray, line 138: With Bootstrap 5 we updated all classes that used "left" and "right" to "start" and "end", let's use the same wording here too: 
```suggestion
.btn-animated-start {
```

packages/styles/src/components/button.scss
```diff
@@ -134,22 +134,48 @@
...
     right: button.$btn-padding-x-md - button.$btn-animation-distance-md;
-    width: 0.625em;
-    height: 0.625em;
     transform: rotateZ(315deg)
       translate(button.$btn-border-width * -1, button.$btn-border-width * -1);
-    transform-origin: center center;
-    transition:
-      opacity 250ms,
-      transform 250ms;
     border-right: button.$btn-border-width solid currentColor;
```
alizedebray, line 177: This PR would be a good opportunity to use proper chevron icons with `@include icons-mx.icon(2110);` and `@include icons-mx.icon(2111);`. This will simplify the code and help match the design even more, what do you think? 
If you prefer you can simply open a new ticket so that we do it later.

alizedebray (changes requested):

oliverschuerch:
@davidritter-dotcom To fix the e2e error, you can simply merge the main into this branch and push it again.

PR #2874, opened 2024-04-02
refactor(components): prefix all web-component custom-events with `post`

oliverschuerch:


imagoiq (approved):
I'm wondering if using `post` as a prefix is a good idea, as the word in itself can have some other meaning (like the POST method in REST).

oliverschuerch:
> I'm wondering if using `post` as a prefix is a good idea, as the word in itself can have some other meaning (like the POST method in REST).

Same feeling here, but because there is no good alternative right now, I've imlemented it the way we decided in the last discussion...

PR #2877, opened 2024-04-03
feat(docs): add v7 migration guide

alizedebray:


packages/documentation/src/stories/getting-started/migration-guide/setup.component.ts
```diff
@@ -16,13 +17,37 @@ export class SetupComponent extends LitElement {
...
+              <option value="v6-to-v7" ?selected="${this.version === 'v6-to-v7'}">
+                @swisspost/design-system-styles 6.x.x
+              </option>
+              <option value="v5-to-v6" ?selected="${this.version === 'v5-to-v6'}">
+                @swisspost/design-system-styles 5.x.x
+              </option>
+              <option value="v4-to-v5" ?selected="${this.version === 'v4-to-v5'}">
+                @.../common-web-frontend 4.x.x or lower
+              </option>
+            </select>
```
imagoiq, line 39: Nitpick: it looks a bit misleading. As a user I want to upgrade from v4 to v7 and when I select the v4 it tells me to upgrade to v5. I could think that I can't really upgrade to the latest. Wouldn't be easier to get all the three steps in the same page in that case?

packages/documentation/src/stories/getting-started/packages/components-angular/components-angular.stories.ts
```diff
@@ -3,7 +3,7 @@ import { MetaExtended } from '../../../../../types';
 
 const meta: MetaExtended = {
   id: '833ef689-a573-40f5-a6a6-30a999b94733',
-  title: 'Getting Started/Packages/Components for Angular',
+  title: 'Getting Started/Packages/Components-Angular',
```
imagoiq, line 6: If you want to go this way, you should change the tile on the home page and the introduction page.

imagoiq (approved):

packages/documentation/src/stories/getting-started/packages/components-angular/components-angular.stories.ts
alizedebray: Actually I prefer "Components for Angular", I just thought we decided to go with "Components-Angular".

packages/documentation/src/stories/getting-started/migration-guide/setup.component.ts
alizedebray: I just updated the version selector: now if you select v4 as your current version you will see all migration guides to migrate from 4 to 5, then from 5 to 6, and finally from 6 to 7 in that order. What do you think?

packages/documentation/src/stories/getting-started/packages/components-angular/components-angular.stories.ts
imagoiq: I thought for a while that you've changed your mind about it. Then let's go with "for"

packages/documentation/src/stories/getting-started/migration-guide/setup.component.ts
imagoiq: Perfect 👌

imagoiq (approved):

PR #2882, opened 2024-04-04
fix: prevent stencil from cleaning the dist folder before build to prevent vite import errors

oliverschuerch:


oliverschuerch:
Wrong base branch

PR #2886, opened 2024-04-05
chore(components): change package.json main, module and types files-paths to the recommended dist-folder files

oliverschuerch:


packages/components/package.json
```diff
@@ -3,10 +3,9 @@
   "version": "2.1.0",
   "description": "A collection of web components built with Stencil JS for the Swiss Post Design System.",
   "license": "Apache-2.0",
-  "main": "loader/index.cjs.js",
-  "module": "loader/index.js",
-  "es2017": "loader/index.2017.js",
-  "types": "loader/index.d.ts",
+  "main": "dist/index.cjs.js",
+  "module": "dist/index.js",
+  "types": "dist/types/index.d.ts",
```
gfellerph, line 8: Can we do a beta release to check if this is not a breaking change?

packages/components/package.json
```diff
@@ -24,9 +23,9 @@
     "linkDirectory": true
   },
   "scripts": {
-    "dev": "stencil build --serve --port 9200 --watch --docs-readme",
-    "start": "stencil build --watch --docs-readme",
-    "build": "stencil build --docs-readme",
+    "dev": "stencil build --port 9200 --serve --watch",
```
gfellerph, line 26: If you remove docs-readme and push a new/changed component without building first, you'll have changes on the CI/CD that will not be committed.

gfellerph (changes requested):

packages/components/package.json
oliverschuerch: I'will revert that.

gfellerph (approved):

PR #2922, opened 2024-04-10
fix(internet-header): Trap focus is not working on overlays and missing in search overlay

imagoiq:


packages/internet-header/package.json
```diff
@@ -27,7 +27,7 @@
     "loader/"
   ],
   "scripts": {
-    "dev": "stencil build --serve --port 9310 --watch --docs-readme",
+    "dev": "stencil build --serve --port 9310 --watch --docs-readme --dev",
```
imagoiq, line 30: This make it easier to use the browser debugger

packages/internet-header/package.json
```diff
@@ -70,6 +69,7 @@
     "cypress-each": "1.14.0",
     "cypress-storybook": "0.5.1",
     "eslint-plugin-react": "7.34.1",
+    "focus-trap": "7.5.4",
```
imagoiq, line 72: focus-trap is from the same creator as tabbable and uses it inside

packages/internet-header/src/components/post-internet-breadcrumbs/post-internet-breadcrumbs.scss
```diff
@@ -171,7 +171,6 @@
   position: relative;
   background: white;
   box-shadow: 0 0 40px rgba(0, 0, 0, 0.6);
-  visibility: hidden;
```
imagoiq, line 174: This makes the overlay not tabbable, I didn't see any differences without it.

packages/internet-header/cypress/e2e/breadcrumbs.cy.ts
```diff
@@ -14,8 +14,8 @@ describe('breadcrumb', () => {
     cy.get('@breadcrumbs')
       .find('div.overlay')
       .should('exist')
-      .find('div.overlay-container')
-      .should('have.class', 'loaded');
+      .find('[role=dialog].loaded', { timeout: 30000 })
+      .should('exist');
```
imagoiq, line 19: A better test would be to check if the element is visible or not

gfellerph (approved):

PR #2926, opened 2024-04-11
fix(styles): Popover update max-width and some styles

imagoiq:


packages/components/src/components/post-popover/post-popover.scss
```diff
@@ -21,8 +21,7 @@
   align-items: self-start;
   padding: 0.5em;
 
-  // Max 2/3 of the column width
-  max-width: min(66vw, 60rem);
+  max-width: 280px;
```
oliverschuerch, line 24: The Figma design includes a min- and a max-width. Why was this removed?
https://www.figma.com/file/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?type=design&node-id=18199-9350&mode=design&t=1TuHz3HP8n0aZsF9-0

.changeset/nervous-experts-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Updated popover max-width and some styles.
```
oliverschuerch, line 5:
```suggestion
Updated popover styles.
- Removed popover `min-width` and updated `max-width`.
- Simplyfied popover arrow size definition.
- Removed `:focus` selector fom `.text-auto` utility class
```

oliverschuerch (changes requested):

packages/components/src/components/post-popover/post-popover.scss
imagoiq, line 24:
```suggestion
  min-width: 160px;
  max-width: 280px;
```

imagoiq: Because with these dynamically value, you don't get the same result as the design. If you look at the current version. It makes the popover much longer than in the design and the readability is poor.

oliverschuerch:
To fix the e2e tests, simply remove the `aria-prohibited-attr` and only leave the `aria-allowed-attr` statement in the axe test.

oliverschuerch (approved):

PR #2937, opened 2024-04-12
fix(switch): background color mapping

gfellerph:


oliverschuerch (approved):
No changeset needed?

gfellerph:
> No changeset needed?

no, the bug has not been published yet

PR #2953, opened 2024-04-16
fix(styles): Checkbox checkmark are invisible on dark background

imagoiq:


imagoiq:
The change come from here: https://github.com/swisspost/design-system/commit/9c40fd7d23d5d4a437fe3923037c93a9e5c353c5 But I don't see any change in the card-control component @oliverschuerch

gfellerph:
> The change come from here: [9c40fd7](https://github.com/swisspost/design-system/commit/9c40fd7d23d5d4a437fe3923037c93a9e5c353c5) But I don't see any change in the card-control component @oliverschuerch

@imagoiq this line https://github.com/swisspost/design-system/pull/2953/files#diff-c0a3347da70fe4b7be6aa1a19ef1828c327082f825f230e9557dada9e02cee44R185 was also added because of the white background. If it's transparent (or even remove the definition), the line for the switches can be removed as well.

oliverschuerch (approved):

Issue #2863, opened 2024-03-28
Deprecate accent colors


oliverschuerch:
https://www.figma.com/file/ojCcgC5Zd12eUSzq6V5m24/Foundations-Post?type=design&node-id=26-111&mode=design&t=zIpCn0DMQsCONKvo-0

```[tasklist]
### Tasks
- [x] Deprecate accent colors
- [x] Deprecate accent colored buttons
```

PR #2977, opened 2024-04-18
feat(styles): Update buttons colors

imagoiq:


packages/styles/src/mixins/_color.scss
```diff
@@ -24,12 +24,12 @@
 // Selector for styles to be applied on dark backgrounds
 @mixin on-dark-background($contains: false) {
   @if $contains == true {
-    &:is(#{color.$dark-backgrounds}),
-    :is(#{color.$dark-backgrounds}) & {
+    &:where(#{color.$dark-backgrounds}),
+    :where(#{color.$dark-backgrounds}) & {
```
imagoiq, line 28: Big change, this was necessary to be able to keep style for button with specific color background (e.g. `.btn-petrol-bright`). This mixin was changed recently to support `post-aler[type"..."]` and this change the specificity as it adds one element to the list of classes, and then override the style for the specific color background.

imagoiq:
@rouvenpost I need your design review to check if it's correct and if I didn't miss any change.

rouvenpost:
I think we shoul remove the «Accent colors» Section altogether. They should not be used anymore.

imagoiq:
> I think we should remove the «Accent colors» Section altogether. They should not be used anymore.

You're right. They will be deprecated with this ticket: https://github.com/swisspost/design-system/issues/2863

packages/styles/src/components/button.scss
```diff
@@ -41,71 +38,116 @@
...
   text-decoration: none !important; // For <a> tags, !important for hover
   white-space: nowrap; // Long content should never break in buttons
-}
 
-// Primary
-.btn-primary {
-  border-color: transparent;
-  background-color: var(--post-gray-80);
-  color: var(--post-contrast-color-inverted);
+  &:disabled {
```
alizedebray, line 42: It would be good to use the `utilities.disabled-style` mixin here.

rouvenpost (approved):

PR #2978, opened 2024-04-18
feat(styles): invalid message on card-control

davidritter-dotcom:


oliverschuerch:
There is an open PR for merging the card-control stories. Please do not merge this PR before the other!

oliverschuerch (changes requested):
Something looks wrong with the icon placement
image

Tested on the css-component (web-component looks good).

davidritter-dotcom:
I've only added the icon in the message on the form elements who don't have an icon inside in this PR, but since the icon inside gets removed everywhere, the icon would need to be added everywhere in the validation message.

packages/styles/src/components/form-validation.scss
```diff
@@ -25,16 +28,46 @@
...
+  }
+}
+
+input:not([type='radio'], [type='checkbox']) ~ .valid-feedback::before,
+textarea ~ .valid-feedback::before {
+  display: none;
+}
+
+post-card-control ~ .invalid-feedback {
+  display: inline-flex;
```
oliverschuerch, line 70: This should be part of the post-card-control.scss.

oliverschuerch (changes requested):
Please add a changeset!

oliverschuerch (commented):
@davidritter-dotcom did you add the error-message only for the web-component or also for the css-component?
And could we add the error-message also in the default story of both components?

davidritter-dotcom:
I added the icon for all the invalid- and valid-feedback messages

oliverschuerch (approved):

PR #2998, opened 2024-04-25
chore(documentation, components): replace relative import paths with  alias in documentation and components

davidritter-dotcom:


packages/components/tsconfig.json
```diff
@@ -1,5 +1,8 @@
 {
   "compilerOptions": {
+    "paths": {
+      "@/*": ["./src/*"]
```
oliverschuerch, line 4: We could implement also:
```json
"paths": {
  "@root/*": ["./*"],
  "@/*": ["./src/*"]
}
```
Which would enable us to use `import { version } from '@root/package.json';`

What do you think?

BTW, this could be implemented in the storybooks `vite.config.js` file as well.

```js
{
  resolve: {
    alias: [
      { find: '@root/', replacement: fileURLToPath(new URL('./', import.meta.url)) },
      { find: '@/', replacement: fileURLToPath(new URL('./src', import.meta.url)) }
    ]
  },
  ....
}
```

and then 

```ts
import ... from '@root/package.json';
import ... from '@root/types';
import ... from '@root/.storybook';
...
```

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #3003, opened 2024-04-25
chore(): delete target blank in our docs

davidritter-dotcom:


oliverschuerch (changes requested):
Please revert all the updates in the demo package.

PR #3016, opened 2024-04-26
chore(components): implement path aliases

oliverschuerch:
I've implemented path aliases and optimized the import of the package.json, so its definition does not end up in every single component output file.

Needed to update stencil to get a fix, which enables the wanted feature: https://github.com/ionic-team/stencil/issues/2319#issuecomment-2037146011


oliverschuerch:
Duplicates #2998

PR #3038, opened 2024-05-01
chore(workflows): remove branch protection from setup-pnpm action calls

oliverschuerch:


oliverschuerch:
I thought this is needed to do the node@20 update, but it seems to work without.
Since we decided to discuss the removal of the branch protection first, I close this PR.

PR #3039, opened 2024-05-01
fix(demo): add missing @types/node dependency

oliverschuerch:


davidritter-dotcom (approved):

packages/demo/package.json
```diff
@@ -69,6 +69,7 @@
     "sockjs": "0.3.24",
     "ts-node": "10.9.2",
     "typescript": "5.3.3",
+    "undici-types": "6.15.0",
```
oliverschuerch, line 72: Probably a missing installation in `@types/node` package, which caused the demo:build task to crash.
I don't know, why this only was an issue in the demo package, but installing it fixed the problem.
https://support.aspnetzero.com/QA/Questions/11907/Cannot-find-module-%27undici-types%27

gfellerph (approved):

PR #3049, opened 2024-05-02
feat(documentation): add version switch active state to v6

davidritter-dotcom:
Depends on the following PRs:
- [x] #2985


gfellerph:
@davidritter-dotcom and @oliverschuerch, that won't work. The workflows are not in place to release older versions yet. You can still merge it and as soon as the workflows work again, it's going to be released (or you could release manually).

oliverschuerch:
The PR is up to date and I deployed the storybook output from it manually to https://design-system-version-6.netlify.app.
There is a problem with some blocked assets, which get blocked.
I've created also a PR to change the current v6 url to the one mentioned above, so the version switcher can point to the new netlify page.

- [ ] Contact UHD to unblock everything from that URL.
- [ ] #3126

oliverschuerch (approved):

PR #3058, opened 2024-05-03
test(workflow): preview message

oliverschuerch:


oliverschuerch:
PR was only created to finally test the new preview message behaviour.
Looks like it works 🌟

PR #3062, opened 2024-05-03
fix(styles): wrap button group overflow

alizedebray:


packages/styles/src/components/button-group.scss
oliverschuerch, line 1: Would be nice if we could avoid the offset on the second line, caused by `margin-left: -2px`:
image

This could be done like so:

```SCSS
// button-group.scss

@use './../variables/components/button';

// make sure this overrides the bootstrap styles for margin-left
.btn-group {
  margin-inline: button.$btn-border-width * 0.5;

  > :not(.btn-check:first-child),
  > :not(.btn-check:first-child) + .btn {
    margin-inline: button.$btn-border-width * -0.5;
  }
}

```

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #3066, opened 2024-05-06
fix(docs): update the post-accordion documentation

alizedebray:


.changeset/modern-crews-impress.md
oliverschuerch, line 1: Would be nice to describe what really has been changed.

oliverschuerch (approved):

PR #3067, opened 2024-05-06
feat(components): add a heading-level property to the post-accordion

alizedebray:


packages/components/src/components/post-accordion/post-accordion.tsx
```diff
@@ -16,13 +18,34 @@ export class PostAccordion {
 
   @Element() host: HTMLPostAccordionElement;
 
+  /**
+   * Defines the hierarchical level of the `post-accordion-item` headers within the headings structure.
+   */
+  @Prop() readonly headingLevel?: HeadingLevel;
```
alizedebray, line 24: I did not make it required to avoid a breaking change.

packages/documentation/src/stories/components/accordion/accordion.stories.ts
```diff
@@ -59,7 +60,7 @@ export const Default: Story = {
 
 export const MultipleOpenPanels: Story = {
   render: () => html`
-    <post-accordion multiple="true">
+    <post-accordion heading-level="4" multiple="true">
```
oliverschuerch, line 63: Why did you not use the storybook logic to override args on the story level here?

```ts
export const MultipleOpenPanels: Story = {
  args: {
    headingLevel: 4
  },
  render: () => html`
    <post-accordion heading-level=${args.headingLevel} multiple="true">
    ...
  `
};
```

packages/documentation/src/stories/components/accordion/accordion.stories.ts
```diff
@@ -80,7 +81,7 @@ export const MultipleOpenPanels: Story = {
 
 export const DefaultCollapsedPanels: Story = {
   render: () => html`
-    <post-accordion>
+    <post-accordion heading-level="4">
```
oliverschuerch, line 84: Why did you not use the storybook logic to override args on the story level here?

packages/documentation/src/stories/components/accordion/accordion.stories.ts
```diff
@@ -99,39 +100,38 @@ export const DefaultCollapsedPanels: Story = {
   `,
 };
 
-// Used only in testing for now.
 export const Nested: Story = {
   render: () => html`
-    <post-accordion>
+    <post-accordion heading-level="4">
```
oliverschuerch, line 105: Why did you not use the storybook logic to override args on the story level here?

packages/documentation/src/stories/components/accordion/accordion.stories.ts
```diff
@@ -99,39 +100,38 @@ export const DefaultCollapsedPanels: Story = {
...
       <post-accordion-item>
-        <span slot="header">Titulum 1</span>
+        <span slot="header">Titulum 1 <code>h4</code></span>
         <p>Contentus momentus vero siteos et accusam iretea et justo.</p>
 
-        <post-accordion>
+        <post-accordion heading-level="5">
```
oliverschuerch, line 110: If you use the storybook logic to override args on the story level, you could change this line to:
`<post-accordion heading-level=${Math.min(args.headingLevel + 1, 6)}>`

packages/documentation/src/stories/components/accordion/accordion.stories.ts
```diff
@@ -99,39 +100,38 @@ export const DefaultCollapsedPanels: Story = {
...
 };
 
-// Used only in testing for now.
 export const Nested: Story = {
   render: () => html`
-    <post-accordion>
+    <post-accordion heading-level="4">
       <post-accordion-item>
-        <span slot="header">Titulum 1</span>
+        <span slot="header">Titulum 1 <code>h4</code></span>
```
oliverschuerch, line 107: If you use the storybook logic to override args on the story level, you should dynamically output the heading-level here as well.

packages/documentation/src/stories/components/accordion/accordion.stories.ts
```diff
@@ -16,6 +16,7 @@ const meta: MetaComponent<HTMLPostAccordionElement & HTMLPostCollapsibleElementE
   },
   args: {
     multiple: false,
+    headingLevel: 3,
```
oliverschuerch, line 19: Is this the default value of the headingLevel/heading-level? Because there is no default value defined in the component. And also this value is not shown as selected in the controls table on the docs page...

packages/components/src/components/post-accordion/post-accordion.tsx
oliverschuerch: Should we add a default value here? I mean, there was a default value (2), when we had the prop on the post-accordion-item.

packages/documentation/src/stories/components/accordion/accordion-item.stories.ts
```diff
@@ -4,12 +4,17 @@ const meta: Meta = {
   id: '3b86bc9b-3dcd-4788-a436-88fd18a6312d',
   title: 'Components/Accordion Item',
   component: 'post-accordion-item',
+  parameters: {
+    controls: {
+      exclude: [
+        'heading-level', // the heading level should be set on the parent post-accordion
```
oliverschuerch, line 10: Instead of removing it visually from the docs, I would suggest to leave it there, but add a deprecation message.
What do you think?

oliverschuerch (changes requested):

packages/components/src/components/post-accordion/post-accordion.tsx
alizedebray: The post-accordion-item still has a default value that will be used if the property is not defined on the post-accordion.
Do you mean having the default property for the docs?

packages/documentation/src/stories/components/accordion/accordion.stories.ts
alizedebray: I would prefer not to have a default value here to prepare for the future when this property will be mandatory.

The value (3) is now selected in the control.

packages/components/src/components/post-accordion/post-accordion.tsx
oliverschuerch: Does this make sence, our would it be better to get the default value allways from the post-accordion component?

post-accordion headingLevel (default value of 2, not required)
post-accordion-item headingLevel (no default value, required)

Would this make more sence? Would it be a breaking change? What do you think?

packages/documentation/src/stories/components/accordion/accordion.stories.ts
oliverschuerch: I'm fine with that!

packages/documentation/src/stories/components/accordion/accordion-item.stories.ts
```diff
@@ -4,12 +4,26 @@ const meta: Meta = {
...
     },
     headingLevel: {
-      control: false,
+      name: 'heading-level',
+      description:
+        'Defines the hierarchical level of the accordion item header within the headings structure.' +
+        '<span className="mb-mini alert alert-warning alert-sm">' +
+        '<code>heading-level</code> on <code>post-accordion-item</code> is deprecated. ' +
+        'Set it directly on the parent <code>post-accordion</code>.' +
+        '</span>',
```
oliverschuerch, line 25: Can we move this description into the component file (as jsdoc block) or is there a problem with the alert?

oliverschuerch (approved):

packages/components/src/components/post-accordion/post-accordion.tsx
alizedebray: Removing the default title level from the post-accordion element would be a drastic change in my opinion.
We can always add a default to the post-accordion too, but that would be difficult to maintain because the two defaults need to be in sync. Also, it wouldn't fit [this idea](https://github.com/swisspost/design-system/pull/3067#discussion_r1598583123).

packages/documentation/src/stories/components/accordion/accordion-item.stories.ts
alizedebray: Good idea!
I had to add some [custom logic](https://github.com/swisspost/design-system/pull/3067/files#diff-3b36e8ee2b1769b10d210808ac598a870fe45bcc961438ce1f7e1e1af38f70beR33) but the deprecation message is now linked to the jsDocs

PR #3105, opened 2024-05-23
fix(styles): stepper scss

oliverschuerch:


.changeset/serious-badgers-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Replaced the deprecated `/` operator with the recommended `math.div()` function in the stepper styles.
```
alizedebray, line 5: This changeset is probably not necessary as the change has no impact for the users.

alizedebray (approved):

PR #3116, opened 2024-05-28
fix(components-angular): add components package as peer dependency in components-angular package

oliverschuerch:
Installing @swisspost/design-system-components as a direct dependency in all the consumer projects, should not be necessary anymore.

### DIFF
--- a/.changeset/thirty-lemons-remember.md
+++ b/.changeset/thirty-lemons-remember.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-components-angular': patch
+---
+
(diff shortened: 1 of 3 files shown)


gfellerph (approved):

pazoozooCH:
@oliverschuerch , it's added as a peer dependency, not a direct dependency. Peer dependencies will still have to be installed explicitly. Would be good if the documentation would reflect that: https://design-system.post.ch/?path=/docs/833ef689-a573-40f5-a6a6-30a999b94733--docs

PR #3119, opened 2024-05-29
chore: fix code smells

alizedebray:


packages/intranet-header-workspace/projects/intranet-header-showcase/src/samples/navigation/navigation.component.html
```diff
@@ -81,10 +81,18 @@
 </section>
 
 <ng-template #myCustomOptions let-dropdown="dropdownRef">
-  <div class="dropdown-item" (click)="dropdown.close()">Close Dropdown</div>
+  <div class="dropdown-item" (click)="dropdown.close()" (keydown)="dropdown.close()">
```
gfellerph, line 84: Won't this close the dropdown on any keydown? This might not be intended. Also, the [enter] key triggers a click event on buttons for example.

gfellerph (changes requested):

packages/intranet-header-workspace/projects/intranet-header-showcase/src/samples/navigation/navigation.component.html
alizedebray: Right, I just updated this to close the dropdown only on [enter] or [space] key press.

gfellerph (approved):

PR #3127, opened 2024-05-30
chore: add linting for icons and migrations package

oliverschuerch:


packages/icons/src/models/censhare-result-page.model.ts
```diff
@@ -101,6 +101,7 @@ export enum Businessfield {
   Kommunikation = 'kommunikation',
 }
 
+/* eslint-disable-next-line @typescript-eslint/no-empty-interface */
 export interface DetectedColorClass {}
```
gfellerph, line 105: I think wherever `DetectedColorClass` is being used, it should be changed to `unknown` or `any`, because an emtpy interface is euqally helpful.

gfellerph (changes requested):

packages/icons/src/models/censhare-result-page.model.ts
oliverschuerch: I was able to replace it by `DetectedColor` type.

gfellerph (approved):

PR #3151, opened 2024-06-12
chore: add test lit setup and check if ssr works

gfellerph:
This is just a showcase for lit + nextjs SSR. Don't merge.


gfellerph:
Works generally, but will not switch right now.

PR #3160, opened 2024-06-18
chore(components): remove unnecessary collapsible state

alizedebray:


packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -25,12 +25,12 @@ import { checkEmptyOrType, isMotionReduced } from '@/utils';
 })
 export class PostCollapsible {
   private isLoaded = false;
+  private isOpen = true;
   private collapsible: HTMLElement;
 
   @Element() host: HTMLPostCollapsibleElement;
 
   @State() id: string;
-  @State() isOpen = true;
```
alizedebray, line 33: This does not need to trigger a re-rendering, thus should not be a @State().

PR #3166, opened 2024-06-20
fix(documentation): remove form-control-wrapper classes from stories

oliverschuerch:


alizedebray (changes requested):
- Styles should be updated to work with the form validation (using border-radius: 1px on the input and margin-top: -2px on the feedback)
image
- The wrapper should be removed from the Patterns>Form docs as well

oliverschuerch:
> * Styles should be updated to work with the form validation (using border-radius: 1px on the input and margin-top: -2px on the feedback)
>   image
> * The wrapper should be removed from the Patterns>Form docs as well

Can I get some visuals to know exactly what and where I should change someting for the validation styling?
Or are you able to implement the wanted changes by yourself?

oliverschuerch:
We have the "GO" from the Designers!

gfellerph (changes requested):
https://github.com/swisspost/design-system/assets/1659006/e3681b74-f133-4f48-8cdd-9140df6417df

I don't think that this layout shift is what we want either. To my understanding we defined to have a partially low contrasting focus over the validation background and fix the pattern with v2 for good.

oliverschuerch:
> Screen.Recording.2024-07-01.at.13.53.54.mov 
> I don't think that this layout shift is what we want either. To my understanding we defined to have a partially low contrasting focus over the validation background and fix the pattern with v2 for good.

Too bad, because I checked with all the other team members (including Christian) and nobody had a complaint.

gfellerph (approved):

PR #3169, opened 2024-06-20
chore: move workspace peerDependency to dependencies

gfellerph:


oliverschuerch (changes requested):
⚠️ !!! This needs to be tested before we can merge it !!! ⚠️
Because we did that once (last april) and it failed!
The only difference now, is that you've added it to "dependencies" instead of "devDependencies".

Also we should add a comment or README.md somewhere, to clearify, that this is a workaround for a known bug in the changeset package and not the standard way to do it.

- [ ] Release a beta version of the components-angular package (pnpm publish --tag beta).
- [ ] Setup a proper angular project from scratch.
- [ ] Install the beta version of the components-angular package and try to start the project.

oliverschuerch:
@gfellerph When I do a `pnpm bootstrap` or `pnpm components-angular:build`, I get the following error:
image

Should we explicitly allow using the components package as non-peer-dependency?

oliverschuerch:
After adding the package `@swisspost/design-system-components-angular` to `allowedNonPeerDependencies` in the file `ng-package.json`, everything went well.
I was able to release a [beta](https://www.npmjs.com/package/@swisspost/design-system-components-angular/v/7.1.1-beta.1), installed it in a freshly set up angular project (Angular CLI v17 with --no-standalone), import the components-angular module into the `app.module.ts` file and afterwards use our components in the `app.component.html` file.

oliverschuerch (approved):

gfellerph:
> After adding the package `@swisspost/design-system-components-angular` to `allowedNonPeerDependencies` in the file `ng-package.json`, everything went well.
> I was able to...
> - release a [beta](https://www.npmjs.com/package/@swisspost/design-system-components-angular/v/7.1.1-beta.1)
> - install the beta it in a freshly set up angular project (Angular CLI v17 with `--no-standalone`)
> - import the components-angular module into the `app.module.ts` file
> - use our components in the `app.component.html` file
> 
> :tada::tada::tada:

Thanks for the testing and fixing on this one @oliverschuerch

PR #3170, opened 2024-06-20
chore: add packageManager field

gfellerph:
This field tells corepack what package manager with what version should be used for this repo (https://nodejs.org/api/corepack.html).


package.json
```diff
@@ -111,5 +111,6 @@
       "es5-ext@>=0.10.0 <0.10.63": ">=0.10.63",
       "follow-redirects@<=1.15.5": ">=1.15.6"
     }
-  }
+  },
+  "packageManager": "pnpm@9.3.0"
```
oliverschuerch, line 115: Do you realise that this is still experimental?
https://nodejs.org/docs/latest-v18.x/api/all.html#all_packages_packagemanager

If I understand it correctly, this will be a very helpful feature!
On the other side we will loose the possibility to use pnpm as node version manager.

oliverschuerch (approved):

package.json
gfellerph: @oliverschuerch did you test that pnpm does no longer work as a node version manager after this change?

PR #3193, opened 2024-06-27
feat(components): add post-button webcomponent

Meierschlumpf:
image


gfellerph:
Thanks @Meierschlumpf for the initiative. Please mark the PR as ready for review when you're ready for the first feedback 🤓

Meierschlumpf:
I've got two questions regarding the implementation:

1. How to deal with class names? should we just forward them? (Especially for the full width example)
2. How do we deal with buttons that are used as `a` tags or that are using custom components like `Link` in NextJS? Do we just not support them through this component as for example loading will not be necessary for links?

gfellerph:
> I've got two questions regarding the implementation:
> 
> 1. How to deal with class names? should we just forward them? (Especially for the full width example)

Class names should not be forwarded through the web component, they should apply to the custom element itself. As for a full-widht button, the best would be to apply utility classes like `<post-button class="w-100">` (https://design-system.post.ch/?path=/docs/eb78afcb-ce92-4990-94b6-6536d5ec6af4--docs#full-width).

> 2. How do we deal with buttons that are used as `a` tags or that are using custom components like `Link` in NextJS? Do we just not support them through this component as for example loading will not be necessary for links?

The web component can't support `<Link>` in React or any comparable concept since it's meant to be framework agnostic. In the react case, I don't think it's possible to define what element the component should render, in other cases it might be configurable. When it's not, it might be possible to use the HTML/CSS version that just needs to add classes to the `<Link>` component. If even that's not possible, then the project using this component has to find another way to style it according to the Design Guidelines.

Meierschlumpf:
Okay not sure if the way I did it now is how you like it, if not can you let me know. I think in general the pull request would now be ready for it's first review

packages/components/src/components/post-button/post-button.tsx
```diff
@@ -0,0 +1,243 @@
...
+          popoverTargetAction={this.popovertargetaction}
+          type={this.type}
+          value={this.value}
+        >
+          {startSection}
+          <span class={this.iconOnly ? 'visually-hidden' : undefined}>
+            <slot />
+          </span>
+          {endSection}
+        </button>
```
gfellerph, line 206: Generally, it would be nice to treat the `post-button` as the button instead of nesting a button inside. I expect this to simplify things because attributes added to the custom element don't need to be passed down to the button. What we need to make sure then is, that the custom button plays nice with all the attributes that can be added to real buttons.

For forms, for example, this means adding form association (https://www.raymondcamden.com/2023/05/24/adding-form-participation-support-to-web-components) which gives access to the form the button is part of and enables us to react to `type="reset"` to clear the form. This would not work in the current implementation, because of the shadow boundary (https://codepen.io/tuelsch/pen/NWZvEQw, only the reset light deletes the text in the input field). The reset shadow button has no form association, but is inside the shadow dom.

packages/components/src/components/post-button/post-button.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+  /**
+   * Defines the variant of the component.
+   */
+  @Prop() readonly variant: string = 'secondary';
+
+  /**
+   * Defines the size of the component.
+   */
+  @Prop() readonly size: null | 'sm' | 'rg' | 'md' | 'lg' = 'md';
```
gfellerph, line 30: The new button uses words instead of abbreviations for naming the sizes (https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations-%26-Components-Next-Level?node-id=1913-7309&t=OfKC9fyWzpVtdkZF-4). We should implement this here as well:
- Small
- Medium
- Large
- Extra large (extra-large)

packages/documentation/src/stories/components/button/web-components/post-button.stories.ts
```diff
@@ -0,0 +1,309 @@
...
+      'coral-bright',
+      'olive',
+      'olive-bright',
+      'purple',
+      'purple-bright',
+      'aubergine',
+      'aubergine-bright',
+    ],
+  },
+};
```
gfellerph, line 285: Accent colors are deprecated and should be removed here.

gfellerph (changes requested):
@Meierschlumpf this is a first review on the basic architecture of the component. It's very nice to have a base to iterate on.

PR #3196, opened 2024-06-28
feat(tooltip)!: add post-tooltip-trigger

gfellerph:
In order to get rid of mutation observers on attributes, a new element to wrap the trigger was added for the tooltip. This enables easier lifecycle management and more precise event listeners with less overhead.

- [ ] Update with new delay property


packages/components/src/components/post-tooltip-trigger/post-tooltip-trigger.tsx
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    // Add tooltip to aria-describedby
+    const describedBy = this.trigger.getAttribute('aria-describedby');
+    if (!describedBy?.includes(this.for)) {
+      const newDescribedBy = describedBy ? `${describedBy} ${this.for}` : this.for;
+      this.trigger.setAttribute('aria-describedby', newDescribedBy);
+    }
+
+    this.host.addEventListener('pointerover', this.localInterestHandler);
```
alizedebray, line 47: Why not adding the event listeners on the trigger instead of the host?

packages/components/src/components/post-tooltip-trigger/post-tooltip-trigger.tsx
```diff
@@ -0,0 +1,70 @@
...
+    this.host.addEventListener('pointerout', this.localInterestLostHandler);
+    this.host.addEventListener('focusin', this.localInterestHandler);
+    this.host.addEventListener('focusout', this.localInterestLostHandler);
+    this.host.addEventListener('long-press', this.localInterestHandler);
+  }
+
+  private get tooltip(): HTMLPostTooltipElement | null {
+    const ref = document.getElementById(this.for);
+    if (ref && ref.tagName === 'POST-TOOLTIP') {
```
alizedebray, line 56:
```suggestion
    if (ref && ref.localName=== 'post-tooltip') {
```

packages/components/src/components/post-tooltip-trigger/post-tooltip-trigger.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  constructor() {
+    this.localInterestHandler = this.interestHandler.bind(this);
+    this.localInterestLostHandler = this.interestLostHandler.bind(this);
+  }
+
+  componentDidLoad() {
+    this.host.setAttribute('data-version', version);
+
+    if (this.host?.children.length > 0 && this.host.children[0].nodeType === 1) {
```
alizedebray, line 29:
```suggestion
    if (this.host?.children.length > 0 && this.host.children[0].nodeType === Node.ELEMENT_NODE) {
```

packages/components/src/components/post-tooltip-trigger/post-tooltip-trigger.tsx
gfellerph: Just in case the trigger gets replaced or re-rendered by react/angular in which case the event bindings would be lost. If the trigger is re-rendered, the lifecycle hooks will trigger again.

packages/components/src/components/post-tooltip-trigger/post-tooltip-trigger.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    this.host.addEventListener('pointerover', this.localInterestHandler);
+    this.host.addEventListener('pointerout', this.localInterestLostHandler);
+    this.host.addEventListener('focusin', this.localInterestHandler);
+    this.host.addEventListener('focusout', this.localInterestLostHandler);
+    this.host.addEventListener('long-press', this.localInterestHandler);
+  }
+
+  private get tooltip(): HTMLPostTooltipElement | null {
+    const ref = document.getElementById(this.for);
```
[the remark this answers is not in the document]
gfellerph: That's necessary to deal with the possibility that the tooltip element gets replaced/re-rendered by react/angular.

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -181,7 +62,7 @@ export class PostTooltip {
 
   /**
    * Toggle tooltip display
-   * @param target An element with [data-tooltip-target="id"] where the tooltip should be shown
+   * @param target An element where the tooltip should be shown
    * @param force Pass true to always show or false to always hide
```
alizedebray, line 66:
```suggestion
   * @param [force] Pass true to always show or false to always hide
```

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -106,65 +25,27 @@ export class PostTooltip {
...
-   * Wheter or not to display a little pointer arrow
+   * Choose a tooltip animation
    */
-  @Prop() readonly arrow?: boolean = true;
+  @Prop() readonly animation?: 'pop-in' | null = 'pop-in';
+
+  /**
+   * Indicates the open state of the tooltip
+   */
+  @Prop({ reflect: true, mutable: true }) open = false;
```
alizedebray, line 35: Props validation is missing, they should also be added to the changeset.

packages/components/src/index.html
alizedebray, line 1: These changes can probably be reverted.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -55,12 +55,17 @@ export class PostPopovercontainer {
...
+   * Defines the placement of the popover-container according to the floating-ui options available at https://floating-ui.com/docs/computePosition#placement.
+   * popover-containers are automatically flipped to the opposite side if there is not enough available space and are shifted
    * towards the viewport if they would overlap edge boundaries.
    */
   @Prop() readonly placement?: Placement = 'top';
 
+  /**
+   * Animation style
+   */
+  @Prop() readonly animation?: 'pop-in' = null;
```
alizedebray, line 67:
```suggestion
  @Prop() readonly animation?: 'pop-in';
```

OR 

```suggestion
  @Prop() readonly animation: 'pop-in' | null = null;
```

packages/documentation/src/stories/components/tooltip/tooltip.stories.ts
```diff
@@ -62,11 +62,17 @@ const meta: MetaComponent = {
     placement: {
       name: 'Placement',
     },
-    arrow: {
-      name: 'Arrow',
+    animation: {
```
alizedebray, line 65: Why did you override the animation arg?
It would be better to use the definition derived from the JSDoc of the component.

alizedebray (changes requested):

gfellerph:
Closing in favour of a larger effort to refactor tooltips with https://github.com/swisspost/design-system/issues/4170

PR #3205, opened 2024-07-02
feat(components, components-angular): update the post-collapsible col…

alizedebray:
…lapsed property


packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -29,17 +29,19 @@ export class PostCollapsible {
...
-  validateCollapsed(newValue = this.collapsed) {
+  collapsedChange() {
     checkEmptyOrType(
-      newValue,
+      this.collapsed,
       'boolean',
       'The `collapsed` property of the `post-collapsible` must be a boolean.',
     );
+
+    void this.toggle(!this.collapsed);
```
oliverschuerch, line 44: You call `collapsedChange()` in `componentDidLoad()`, probably to run the `checkEmptyOrType` functionality.
But this will call the current line of code as well right from the beginning.
Is this intentionally? Or do I miss something?

packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -69,8 +67,9 @@ export class PostCollapsible {
   async toggle(open = !this.isOpen): Promise<boolean> {
     if (open === this.isOpen) return open;
 
-    this.isOpen = !this.isOpen;
-    if (this.isLoaded) this.postToggle.emit(this.isOpen);
+    this.isOpen = open;
```
oliverschuerch, line 70: Is this.isOpen still needed?

oliverschuerch (changes requested):

packages/components/src/components/post-collapsible/post-collapsible.tsx
alizedebray: Yes it is intentional, if the collapsible has the prop `collpased="true"` we want it to be initially closed, to do so the `toggle` method has to be called.

packages/components/src/components/post-collapsible/post-collapsible.tsx
alizedebray: Yes, this is still needed. The component initially renders the host without any styles (i.e., in the 'opened' state), even when it loads with the `collapsed="true"` prop. In this case, `this.isOpen === !this.collapsed` is only true after the toggle method has been called once in `componentDidLoad`.

oliverschuerch (approved):

PR #3209, opened 2024-07-02
feat(components): add a post-collapsible-trigger

alizedebray:


packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
gfellerph, line 1: The only semantically valid child or slotted element is a single button (it still makes sense to slot it because there are many types of buttons and you'd not want to rebuild the trigger as a button. With this in mind, it would be possible to simplify the code somewhat, because clicks on buttons bubble up through the shadow dom and buttons are focusable anyways.

- No need for keydown or pointerdown, just listen for click events on the host (https://codepen.io/tuelsch/pen/zYVYgEd?editors=1111)
- No need to check if it's focusable
- Check if the slotted element is a button, otherwise warning (on componentDidLoad)
- Set all aria attributes on the slotted button

gfellerph (changes requested):

packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
alizedebray: I updated the PR to make it mandatory to use a `<button>` inside the trigger. Do you think we should allow the `[role="button"]` as well?
I also attached the click listener directly to the trigger as I need to query the element to set the ARIA attributes anyways. 

Concerning slotted buttons, I am not sure this is a real use case in the end. The component itself does not have a shadow root thus in most cases both the button and the collapsible will be in the light DOM. And if we implement the post-collapsible-trigger into another component, we will most probably also add the button and the collapsible ourselves so they will both be in the shadow DOM. What do you think?
Anyways, if that's okay with you I would rather merge this PR without this feature and potentially handle slotted buttons in the [accordion PR](https://github.com/swisspost/design-system/pull/3215) since it contains the necessary helpers already.

gfellerph: I think for the moment we can go without `role=button` support. It's a valid use case, but pretty rare (hopefully). There are many other downsides to not using propper buttons (e.g. role buttons are not focusable by default), therefore we should not "advertise" this use case too much.

As for the slotted button, I think I missed that point in the review, or got confused by the two PRs. I'll review with that in mind again, thanks for the heads up.

packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
```diff
@@ -0,0 +1,55 @@
...
+    // Add collapsible id to aria-controls
+    if (this.trigger) {
+      this.trigger.setAttribute('aria-controls', this.for);
+
+      const isOpen = !this.collapsible?.collapsed;
+      if (isOpen !== undefined) this.trigger.setAttribute('aria-expanded', `${isOpen}`);
+    }
+  }
+
+  componentWillLoad() {
```
gfellerph, line 32: In case the button renders slower than the trigger.

```suggestion
  componentDidLoad() {
```

packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
```diff
@@ -0,0 +1,55 @@
...
+    this.host.setAttribute('data-version', version);
+    this.trigger = this.host.querySelector('button');
+
+    if (!this.trigger) throw new Error('The post-collapsible-trigger must contain a button.');
```
gfellerph, line 36: Not sure if we just want to write `console.error("")` in this case and not break the whole JS code. If there is no button, the trigger won't work and people will hopefully read the errors in the console.

packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    this.trigger.addEventListener('click', () => this.toggleCollapsible());
+    this.setAriaAttributes();
+  }
```
gfellerph, line 40: Since the click event bubbles, it would be safer to register the event on the trigger. The button could be re-rendered and the listener lost. If the trigger re-renders, the listener will get attached again by this lifecycle function.

packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+
+  componentWillLoad() {
+    this.host.setAttribute('data-version', version);
+    this.trigger = this.host.querySelector('button');
+
+    if (!this.trigger) throw new Error('The post-collapsible-trigger must contain a button.');
+
+    this.trigger.addEventListener('click', () => this.toggleCollapsible());
```
gfellerph, line 38: We should also add an event listener on the linked collapsible to catch collapse events triggered via JS on the collapsible itself and update the aria attributes on the trigger button accordingly.

If these events bubble, it would be a good idea to register a global event listener on the document or root to prevent having to watch the collapsible element. The correct trigger can be identified based on the id on the event target.

gfellerph (changes requested):

packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
alizedebray: As discussed, I used a mutation observer instead.

gfellerph (approved):
Awesome work

200w

PR #3215, opened 2024-07-04
chore(components): use the post-collapsible-trigger within the post-accordion-item

alizedebray:


packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.scss
gfellerph, line 1: I don't think this is necessary, buttons are not block level elements by default. I'd just not ship any styles at all.

.changeset/plenty-experts-admire.md
alizedebray, line 1: Remove file

.changeset/thin-eyes-join.md
alizedebray, line 1: Remove file

oliverschuerch (commented):
@alizedebray did you do something within this PR, which is was not implemented in the v7 patch and the cherry-picked PR for v8?
Because if the answer is "YES", you should add a changeset to this PR ;)

packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
```diff
@@ -1,6 +1,6 @@
-import { Component, Element, Listen, Method, Prop, Watch } from '@stencil/core';
-import { version } from '@root/package.json';
-import { checkNonEmpty, checkType, debounce } from '@/utils';
+import { Component, Element, Method, Prop, Watch } from '@stencil/core';
+import { version } from 'typescript';
```
oliverschuerch, line 2: Nope!
```suggestion
import { version } from '@root/package.json';
```

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #3228, opened 2024-07-10
feat(styles): update select disabled styles

davidritter-dotcom:


alizedebray (changes requested):
Can you update the styles to match [Figma](https://www.figma.com/design/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?node-id=7802-33963&m=dev)? When the selection is disabled:
- the label and hint should be `--post-gray-60`
- the chevron icon should be `--post-gray-40` 
- the text should be `--post-gray-60` and not strikethrough

davidritter-dotcom:
> Can you update the styles to match [Figma](https://www.figma.com/design/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?node-id=7802-33963&m=dev)? When the selection is disabled:
> 
> * the label and hint should be `--post-gray-60`
> * the chevron icon should be `--post-gray-40`
> * the text should be `--post-gray-60` and not strikethrough

The Design in Figma is already updated and the text is now strikethrough also in the select.

packages/styles/src/variables/components/_forms.scss
```diff
@@ -37,7 +37,7 @@ $input-line-height-lg: type.$line-height-copy !default;
 
 $input-bg: color.$white !default;
 $input-disabled-bg: rgba(var(--post-contrast-color-inverted-rgb), 0.4) !default;
-$input-disabled-color: rgba(var(--post-contrast-color-rgb), 0.6); // Design System only
+$input-disabled-color: color.$gray-40; // Design System only
```
alizedebray, line 40:
```suggestion
$input-disabled-color: var(--post-gray-60); // Design System only
```

packages/styles/src/variables/components/_forms.scss
```diff
@@ -37,7 +37,7 @@ $input-line-height-lg: type.$line-height-copy !default;
...
 $input-disabled-border-color: color.$gray-40; // Design System only
```
alizedebray, line 41:
```suggestion
$input-disabled-border-color: var(--post-gray-40); // Design System only
```

alizedebray (approved):
If you declare the variables like so:
```
$input-disabled-color: var(--post-gray-60);
$input-disabled-border-color: var(--post-gray-40);
```
Then the other files can remain unchanged, still using `forms.$input-disabled-color` and `forms.$input-disabled-border-color`

The only place where you will have to declare the gray manually is here, because you cannot use a var() in a background URL: 
```
$form-select-disabled-color: forms.$input-disabled-color;
$form-select-indicator-disabled: url('#{icons.get-colored-svg-url('2113', $form-select-disabled-color)}') !default;
```

This approach helps with consistency as you only define the gray twice: once in the form variables and once in the select background.

davidritter-dotcom:
> If you declare the variables like so:
> 
> ```
> $input-disabled-color: var(--post-gray-60);
> $input-disabled-border-color: var(--post-gray-40);
> ```
> 
> Then the other files can remain unchanged, still using `forms.$input-disabled-color` and `forms.$input-disabled-border-color`
> 
> The only place where you will have to declare the gray manually is here, because you cannot use a var() in a background URL:
> 
> ```
> $form-select-disabled-color: forms.$input-disabled-color;
> $form-select-indicator-disabled: url('#{icons.get-colored-svg-url('2113', $form-select-disabled-color)}') !default;
> ```
> 
> This approach helps with consistency as you only define the gray twice: once in the form variables and once in the select background.

The svg doesn't work if there is a css variable behind the sass variable, so I just put a color sass variable for the background URL.

PR #3245, opened 2024-07-10
feat(components): add "delayed" property to the post-tooltip

alizedebray:


alizedebray:
@oliverschuerch @gfellerph I am wondering if it wouldn't be better to create a `post-title` component instead of adding this property. It could have a simpler design (no arrow, no configurable background) but also have specific logic (i.e. if the text is fully visible in the trigger then do not show the title). What do you think? 

Gmail has a similar component they use on all icons: 

https://github.com/user-attachments/assets/2fc29e16-7219-4f09-b4f2-ad9caa29b01d

gfellerph:
@alizedebray let's discuss this in the pattern meeting. I'm not sure if it should be the same component and people can recreate the behaviour with attributes or a dedicated one (naming would need to be discussed as well, `post-title` could also mean heading).

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -1,9 +1,12 @@
-import { Component, Element, h, Host, Method, Prop } from '@stencil/core';
+import { Component, Element, h, Host, Method, Prop, Watch } from '@stencil/core';
 import { Placement } from '@floating-ui/dom';
 import { version } from '@root/package.json';
 import isFocusable from 'ally.js/is/focusable';
 import 'long-press-event';
 import { getAttributeObserver } from '@/utils/attribute-observer';
+import { checkEmptyOrType, timeout } from '@/utils';
+
+const OPEN_DELAY = 650; // matches HTML title delay
```
oliverschuerch, line 9: :heart:

oliverschuerch (approved):

PR #3256, opened 2024-07-11
feat(workflows): add a build-tokens workflow

oliverschuerch:
How will a workflow run look?
You can see that [here](https://github.com/swisspost/design-system/actions/workflows/build-tokens.yaml).

### DIFF
--- a/.github/workflows/build-tokens.yaml
+++ b/.github/workflows/build-tokens.yaml
@@ -30,8 +30,8 @@ jobs:
             const fs = require('fs')
             const path = require('path')
 
-            const inputfileNames = fs.readdirSync('packages/tokens/tokensstudio-generated')
-            const inputFiles = inputfileNames.map(fileName => ({
+            const inputFileNames = fs.readdirSync('packages/tokens/tokensstudio-generated')
+            const inputFiles = inputFileNames.map(fileName => ({
               type: path.extname(fileName),
               name: fileName,
               content: fs.readFileSync(`packages/tokens/tokensstudio-generated/${fileName}`, 'utf8')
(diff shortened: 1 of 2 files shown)


.github/workflows/build-tokens.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Create Summary
+        id: summary
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const fs = require('fs')
+            const path = require('path')
+
+            const inputfileNames = fs.readdirSync('packages/tokens/tokensstudio-generated')
+            const inputFiles = inputfileNames.map(fileName => ({
```
alizedebray, line 34:
```suggestion
            const inputFiles = inputFileNames.map(fileName => ({
```

.github/workflows/build-tokens.yaml
```diff
@@ -0,0 +1,71 @@
...
+              'channel.scss',
+              'theme.scss',
+              'components.scss',
+            ]
+            const outputFileNames = fs.readdirSync('packages/tokens/dist')
+            const outputFiles = outputFileNames.map(fileName => ({
+              type: path.extname(fileName),
+              name: fileName,
+              content: fs.readFileSync(`packages/tokens/dist/${fileName}`, 'utf8')
+            })).sort((a, b) => (outputOrder.indexOf(a.name) === -1 ? 1000 : outputOrder.indexOf(a.name)) - (outputOrder.indexOf(b.name) === -1 ? 1000 : outputOrder.indexOf(b.name)))
```
alizedebray, line 54:
```suggestion
            })).sort((a, b) => (outputOrder.includes(a.name) ? outputOrder.indexOf(a.name) : 1000) - (outputOrder.includes(b.name) ? outputOrder.indexOf(b.name) : 1000))
```

.github/workflows/build-tokens.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+            return `# Build Tokens
```
alizedebray, line 56:
```suggestion
            return `# Token Build
```

.github/workflows/build-tokens.yaml
```diff
@@ -0,0 +1,71 @@
+name: Build Tokens
+on:
+  pull_request:
+    types: [opened, synchronize, edited, reopened]
```
alizedebray, line 4: Is this necessary to run the workflow when the PR id edited as well?

alizedebray (approved):
👍

.github/workflows/build-tokens.yaml
oliverschuerch: I'm not even sure if it is necessary at all.
What do you think?

alizedebray: The changeset magic force pushes to the changeset PR instead of doing a regular push; therefore, the synchronize trigger does not activate. Fortunately, changeset magic also updates the PR description, which allowed us to use the edited trigger as a workaround to ensure the documentation preview remains up-to-date after a force push on this PR.

If the build token summary needs to be updated with any change on the changeset PR, then you can leave the triggers as-is. Otherwise, you can fully remove the types.

PR #3262, opened 2024-07-12
feat(tokens): add `Components/Interactives` and `Components/Notifications` tokens

oliverschuerch:
To see if the tokens can be built, we need to merge the following PR first:
- [x] #3256


packages/tokens/tokensstudio-generated/tokens.json
```diff
@@ -571,626 +596,2619 @@
...
+            },
+            "type": "border",
+            "description": "To be used for everything disabled that has a natural border"
           }
         }
       }
     }
   },
-  "Device/Tablet": {
+  "Mode/Light": {
```
alizedebray, line 623: Shouldn't this be lower-cased?

oliverschuerch: Doesn't matter, because it will be outputted lower-cased anyway ;)

gfellerph (approved):

alizedebray (approved):

PR #3343, opened 2024-07-24
fix(styles): invalid CSS selector

gfellerph:


packages/styles/src/components/alert.scss
```diff
@@ -76,10 +76,11 @@
...
         flex: 1 0 0;
         width: 100%;
+      }
 
-        &:first-child {
-          margin-inline-start: 0;
-        }
+      > .btn:first-child,
+      > ::slotted(.btn:first-child) {
+        margin-inline-start: 0;
```
oliverschuerch, line 83: I can't see, that this is doing anything in the browser, but never the less, at least it's not producing wrong css anymore.

oliverschuerch (approved):

PR #3346, opened 2024-07-24
feat: add device tokens

gfellerph:
Added device token mappings, intended to be used in output files.


gfellerph:
Will merge token prs to one

PR #3349, opened 2024-07-24
feat: add semantic layer token machinery

gfellerph:


packages/styles/package.json
```diff
@@ -44,6 +44,7 @@
   },
   "dependencies": {
     "@popperjs/core": "2.11.8",
+    "@swisspost/design-system-tokens": "workspace:*",
```
oliverschuerch, line 47: Should we implement the correct version here instead of using an asterisk?

packages/styles/src/placeholders/_modes.scss
```diff
@@ -0,0 +1,10 @@
+@use '@swisspost/design-system-tokens/mode' as mode;
+@use './../mixins/tokens';
+
+%post-light {
```
oliverschuerch, line 4: What about using the name `%mode-light` instead?

packages/styles/src/tokens/_device.scss
```diff
@@ -0,0 +1,19 @@
+@use '@swisspost/design-system-tokens/device' as device;
+
+@use './core';
+@use './../mixins/tokens';
+@use './../mixins/media';
+
+:root {
+  @include media.max(600px) {
+    @include tokens.from(device.$post-mobile);
+  }
```
oliverschuerch, line 10: These are the default values and may not be included in a media query. Are you sure this works?

packages/styles/src/tokens/_external.scss
```diff
@@ -0,0 +1,8 @@
+@use '@swisspost/design-system-tokens/channel' as channel;
+
+@use './core';
+@use './../mixins/tokens';
+
+:root {
```
oliverschuerch, line 6: To separate these semantic tokens from the rest, we could put them in a `:scope` selector instead of `:root`.
> When used at the root level of a stylesheet, :scope is equivalent to :root, which in a regular HTML document matches the <html> element.

Source: [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/:scope)

packages/styles/src/tokens/_modes.scss
oliverschuerch, line 1: Will we be able to use https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark if we implement mode tokens like this?

oliverschuerch (changes requested):

packages/styles/src/tokens/_device.scss
gfellerph: By being specific, there's reduced complexity in the devtools because the default tokens don't have to be overwritten. The way the queries are set up, max, between, min, every possible screen size has been covered.

packages/styles/package.json
gfellerph: This is how workspace dependencies get installed by pnpm when using the `--workspace` flag. I'm wondering if we should change all other workspace dependencies to that.

packages/styles/src/placeholders/_modes.scss
gfellerph: Good idea, I think I'll call it `%color-mode-light` which would be even closer to the selector.

packages/styles/src/mixins/_tokens.scss
```diff
@@ -0,0 +1,12 @@
...
+@use 'sass:map';
+
+@mixin from($list) {
+  @each $key, $value in $list {
+    --#{$key}: #{$value};
+  }
+}
+
+@function component($name) {
+  @return #{map.get(components.$post-interactives, $name)};
```
oliverschuerch, line 11: This is fine for testing, but it will not work as soon as you like to get other component tokens (not from the hardcoded map `$post-interactives`).

packages/styles/src/tokens/_modes.scss
gfellerph: Yes, but not without workarounds. I would say we wait until it's broadly adopted and the improve the token export script to better support this use case. It would save us a bunch of tokens in the end.

packages/styles/src/mixins/_tokens.scss
gfellerph: Yes, I think we need to either flatten the components into one map or export component-button.scss with button only tokens from the tokens package.

packages/styles/src/tokens/_external.scss
gfellerph: While I don't see an immediate benefit, this could be nice in combination with `@scope` in the future. Not sure though if there is an actual use case. As long as we don't need to use this, I would avoid implementing it as not to create unexpected and hard to fix issues when suddenly someone uses `@scope` without thinking about the tokens.

packages/styles/src/mixins/_tokens.scss
gfellerph: @oliverschuerch I've removed the component token implementation again since we need to make changes to how component tokens are rendered.

packages/documentation/.storybook/styles/preview.scss
```diff
@@ -1,6 +1,6 @@
 // importing the complete styles package scss
 // TODO: replace intranet.scss with index.scss as soon as the issues in the portal.scss are fixed
```
alizedebray, line 2:
```suggestion
```

alizedebray (approved):
This is neat, I look forward to working with these tokens!

oliverschuerch (approved):

packages/styles/src/tokens/_device.scss
```diff
@@ -0,0 +1,19 @@
+@use './temp/device' as device;
+
+@use './core';
+@use './../mixins/tokens';
+@use './../mixins/media';
+
+:root {
+  @include media.max(600px) {
```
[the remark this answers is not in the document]
gfellerph: Yes, we were talking about this. We'll update this once they are (also, there might be two more).

PR #3358, opened 2024-07-30
fix(components): remove scss module import

alizedebray:


gfellerph:
I somehow doubt that both solutions (the `:export` and the new `hostContext` patch) are the right approach to solve the underlying problem (which I don't really understand from context here). Web components should be built context independent and if the context has an influence on how the component is rendered, the context should be responsible of doing the customization while the web component just offers the possibility to customize what's needed through options.

TLDR; no component should change based on context, the context should change the component.

alizedebray:
> 1. As far as I understand, the solution provided in this PR is doing the same thing as before (with some small updates), just with another technic. @alizedebray please correct me if I'm wrong.

@oliverschuerch Exact. `:export` cannot be used with the 'dist-custom-elements' output target so I created this PR as a workaround to fix the bug while keeping your initial logic.

alizedebray:
> I somehow doubt that both solutions (the `:export` and the new `hostContext` patch) are the right approach to solve the underlying problem (which I don't really understand from context here). 

@gfellerph I completely agree and we should no longer have this issue  if we implement the component tokens right. This PR fixes to build in the meantime, giving us time to come with a better solution.

oliverschuerch (approved):
Please add a changeset

PR #3366, opened 2024-07-31
fix(styles): realign radios / checkboxes labels

davidritter-dotcom:


packages/styles/src/components/form-check.scss
```diff
@@ -11,6 +11,7 @@
 @use '../mixins/utilities' as utility-mx;
 
 .form-check {
+  align-items: center;
```
gfellerph, line 14: This won't work if the label text breaks on two lines. After a style update, the checkbox is now 22px high and the label 24px. This means that the checkbox needs a margin-top of 1px to re-center it with the label.

gfellerph (changes requested):

packages/styles/src/components/form-check.scss
schaertim: I'm not quite sure if i understood the comment correctly. I just added the top-margin to the checkbox/radio and the alignment looks to be correct now.

.changeset/fair-kids-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+realigned the checkbox and the radio button with the lable
```
gfellerph, line 5:
```suggestion
Realigned the checkbox and the radio button with the label.
```

gfellerph (approved):

PR #3376, opened 2024-08-05
feat(components): add the post-avatar component

alizedebray:


gfellerph (changes requested):
Before hitting the button: discuss naming to align with other components naming scheme

oliverschuerch:
Naming suggestions for this component:

- post-profile
- post-profile-card
- post-account-info
- post-avatar-info
- ...

PR #3396, opened 2024-08-13
fix(styles): Repeating icons in validation messages

schaertim:


packages/styles/src/components/form-validation.scss
```diff
@@ -18,7 +18,7 @@
 
 .invalid-feedback,
 .valid-feedback {
-  display: none;
+  display: flex;
```
gfellerph, line 21: This change will show validation feedback in certain situations when it's rendered in the DOM but the control has not been validated yet. See line https://github.com/swisspost/design-system/pull/3396/files#diff-1c6c2dabdd404cc629b0659cde023035d93529fe69b3798056ef1b04ef8dc8c4L92 where the control gets `display: inline-flex` if there's a sibling with `is-valid` or `is-invalid` only.

```suggestion
  display: none;
```

packages/styles/src/components/form-validation.scss
```diff
@@ -31,9 +31,9 @@
 
   &::before {
     content: '';
+    flex-shrink: 0;
```
gfellerph, line 34: While `flex-shrink` generally solves the issue, I think it would be a good idea to add `mask-repeat: no-repeat;` as well for stability. Also, the icon should be aligned at the top, so `align-self: flex-start;`.

```suggestion
    flex-shrink: 0;
    mask-repeat: no-repeat;
    align-self: flex-start;
```

gfellerph (changes requested):

gfellerph:
@schaertim this is a fix we're releasing on v8 (I changed the base branch). After you merge it, can you cherry pick the commit to current main and open another PR?

gfellerph (approved):

PR #3406, opened 2024-08-14
chore(components): update the component's package output target

alizedebray:
In the end this task does not solve any bug with importing web components individually, it is just a nice-to-have simplification.


oliverschuerch (approved):
Does this have any effect on the components-angular or components-react package?

PR #3408, opened 2024-08-14
chore: create header/footer data structure proposal

gfellerph:
Proposed data structure for the header and footer: https://github.com/swisspost/design-system/pull/3408/files


gfellerph:
Missing points:
- [ ] Error message when track'n'trace does not find shipping number

gfellerph:
@gfellerph move to a discussion

PR #3427, opened 2024-08-20
fix(components): Unexpected behavior of multilevel accordions

schaertim:


packages/components/src/components/post-accordion/post-accordion.tsx
```diff
@@ -50,23 +50,24 @@ export class PostAccordion {
 
   @Listen('postToggle')
   collapseToggleHandler(event: CustomEvent<boolean>) {
-    if ((event.target as HTMLElement).localName === 'post-accordion-item') {
-      event.stopPropagation();
+    const toggledItem = event.target as HTMLElement;
```
alizedebray, line 53: Here you must still verify that the event.target is a `post-accordion-item`. 

The `postToggle` is a generic event name that could be used for other components, and we don't want to close accordion items if a popover is toggle for example.

alizedebray (changes requested):

alizedebray (approved):

PR #3441, opened 2024-08-20
feat(docs): add toolbar to switch theme, channel, and mode

alizedebray:


packages/documentation/.storybook/styles/preview.scss
```diff
@@ -199,3 +199,9 @@
     right: 0;
   }
 }
+
+.custom-story-container {
+  display: block;
+  margin: -39px -29px;
```
oliverschuerch, line 205: I would suggest to set the negative margin to the same values as the posive padding. Or if you prefer this solution, add `border-radius-top-left` and `border-radius-top-right` on the custom-story-container to prevent the strange look like in the following screenshot (open it in its full resolution and inspect the top borders to see what I mean). 
image

```suggestion
  margin: -40px -30px;
```

packages/documentation/.storybook/toolbar/index.ts
```diff
@@ -0,0 +1,77 @@
...
+  // mode is added through the `data-color-mode` attribute, we also set a dark background color when necessary
+  const mode = context.globals.mode || 'light';
+  const classes = {
+    'custom-story-container': true,
+    'bg-dark': mode === 'dark',
+  };
+
+  return html`
+    <${unsafeStatic(storyContainerComponentName)}
+      class=${classMap(classes)}
```
oliverschuerch, line 71:
```suggestion
  const classes = {
    'bg-dark': mode === 'dark',
  };

  return html`
    <${unsafeStatic(storyContainerComponentName)}
      class="custom-story-container ${classMap(classes)}"
```

image

packages/documentation/.storybook/toolbar/index.ts
```diff
@@ -0,0 +1,77 @@
...
+};
+
+/**
+ * Decorator to apply the toolbar selection to all stories.
+ * Stories are then rendered in a shadow dom were the expected styles are applied
+ */
+export const applyToolbarSelection = (story: StoryFn, context: object) => {
+  // theme and channel require importing a specific CSS file, this is handled by specific web-components
+  const theme = context.globals.theme || 'post';
+  const channel = context.globals.channel || 'external';
```
oliverschuerch, line 59: I would like to have the `globalTypes` properties `description` and `defaultValue` being implemented, so users may see the description through the `title` attribute and you would not need to set the default values here.
https://storybook.js.org/docs/essentials/toolbars-and-globals#global-types-and-the-toolbar-annotation

And also it would be nice to use some icons, but only if we find some that match:
https://storybook.js.org/docs/faq#what-icons-are-available-for-my-toolbar-or-my-addon

packages/documentation/.storybook/toolbar/story-container-base.ts
```diff
@@ -0,0 +1,26 @@
...
+    return this.story(this.context, this.context.args);
+  }
+}
+
+/**
+ * Transforms a style input into a CSSResult output adapted to web components
+ */
+export function useStyles(styles: any) {
+  if (typeof styles !== 'string') return unsafeCSS('');
+  return unsafeCSS(styles.replace(/:root/g, ':host'));
```
oliverschuerch, line 25: Did you try to change the `:root` selector to `:scope` in the tokens core.scss file?
I thought, this was a topic we discussed last week?

Ignore this if it just did not work ;)

packages/documentation/.storybook/toolbar/story-container-post-external.ts
oliverschuerch, line 1: When I look at this file (and also the story-container-post-internal), I come to the conclusion, that we will end up with `post-internal`, `post-external`, `cargo-internal`, `cargo-external`, and many more.
Do we really need to change the tagname in the toolbar/index.ts file, or woule it be enough to replace the used css?

oliverschuerch (changes requested):

packages/documentation/.storybook/toolbar/story-container-base.ts
alizedebray: I added a comment about that but it was pending... :sweat_smile:
Here it is: https://github.com/swisspost/design-system/pull/3441#discussion_r1727173846

packages/documentation/.storybook/toolbar/index.ts
alizedebray: I couldn't find any icon that fit, if you see some that make sense just tell me.

oliverschuerch (approved):

PR #3466, opened 2024-08-27
fix(docs): removed typos in href paths

schaertim:


.changeset/yellow-yaks-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': major
```
oliverschuerch, line 2: This is not a major update for the docs package!
It's just a typo, but not something that had to be fixed in the actual internet-header package 😉 

```suggestion
'@swisspost/design-system-documentation': patch
```

oliverschuerch (changes requested):

oliverschuerch (approved):
Nice, thanks!

PR #3481, opened 2024-08-29
fix: release workflow

oliverschuerch:
Not sure if it will work this time. But I don't see a way to test it somewhere else than on the live system.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -25,7 +25,18 @@ jobs:
       # This will fail the build if something in the publish setup is not correct
       # before changeset magic is starting to run
       - name: Test publish step
-        run: pnpm -r publish --dry-run
+        id: publish
+        run: |
+          branch=$(git rev-parse --symbolic-full-name --abbrev-ref HEAD);
+
+          if [ $branch != 'main' ];
+            then
+              echo "command=pnpm changeset publish --tag latest-${branch:8:2}" >> "$GITHUB_OUTPUT";
+              pnpm -r publish --publish-branch $branch --tag latest-${branch:8:2} --dry-run;
+            else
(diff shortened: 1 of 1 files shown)


oliverschuerch:
@gfellerph I'm not able to test this from my post computer, because of a timeout.
Can you try it from your machine?

.github/workflows/release.yaml
```diff
@@ -25,7 +25,18 @@ jobs:
...
       - name: Test publish step
-        run: pnpm -r publish --dry-run
+        id: publish
+        run: |
+          branch=$(git rev-parse --symbolic-full-name --abbrev-ref HEAD);
+
+          if [ $branch != 'main' ];
+            then
+              echo "command=pnpm changeset publish --tag latest-${branch:8:2}" >> "$GITHUB_OUTPUT";
+              pnpm -r publish --publish-branch $branch --tag latest-${branch:8:2} --dry-run;
```
gfellerph, line 35: Are we not on the branch that we want to publish? If current branch and ${branch} don't match, pnpm will ask a question and the workflow will hang.

```suggestion
              pnpm -r publish --tag latest-${branch:8:2} --dry-run;
```

gfellerph (changes requested):

.github/workflows/release.yaml
oliverschuerch: The problem here is, that the current $branch is not the "main" branch. This causes the publish task to ask the question: "you are not on the main branch, do you really want to publish?" or something like that. with this option, the question is omitted.

oliverschuerch:
This is not needed, because changeset publish should setting the tag specified in the `pre.json` file within the `.changeset` folder.
Currently defined tag: `next`

PR #3499, opened 2024-09-05
fix(docs): stepper and docs examples not working

schaertim:


gfellerph (approved):
Works smoothly again. Nice fix!

PR #3502, opened 2024-09-05
feat(styles): add tokens get function

oliverschuerch:


packages/styles/src/functions/_tokens.scss
```diff
@@ -0,0 +1,41 @@
+@use 'sass:map';
+@use 'sass:string';
+
+/**
+ * @function get($keyName, $m: $map)
+ * Gets token-key, normalizes and resolved it with the $map or de one specified in the second parameter.
```
alizedebray, line 6:
```suggestion
 * Gets token-key, normalizes and resolved it with the $map or the one specified in the second parameter.
```

packages/styles/src/functions/_tokens.scss
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @param {string} $keyName - The token key to get (with or without the "post" prefix).
+ * @param {map} $m - The map to get the token key from.
+ *                   The $map variable can be specified through the "with" keyword in the import statement:
+ *                   @use '../functions/tokens' as tokens with ($map: namespace.$my-default-map);
```
alizedebray, line 11:
```suggestion
 *                   @use '../functions/tokens' with ($map: namespace.$my-default-map);
```

gfellerph (approved):

packages/styles/src/functions/_tokens.scss
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @returns {any} $value of $map[$key]
+ *
+ * @example
+ * @use '../tokens/components' as componentTokens;
+ * @use '../functions/tokens' as tokens with ($map: componentTokens.$post-body);
```
alizedebray, line 17:
```suggestion
 * @use '../functions/tokens' with ($map: componentTokens.$post-body);
```

alizedebray (approved):

packages/styles/src/functions/_tokens.scss
oliverschuerch: Nope. This must stay, as we want to call `tokens.get()`

packages/styles/src/functions/_tokens.scss
oliverschuerch: Same here

packages/styles/src/functions/_tokens.scss
alizedebray: Yes, but the file is already called "tokens"

alizedebray (changes requested):
I get this error when using "with" in two files that end up in the same index file:
image

gfellerph:
> But maybe we could have only one map for all component tokens, what do you think?

Dang, I had so much fun. If we use one map, we loose the discoverability of all the tokens that belong to one component. What we could do instead is using a convention for the map name, so you'd only had to provide the name of the component without `post-` prefix, e.g. `tokens.get('h1', 'font-family')` and the function would do `map.get(post-h1, post-h1-font-family)`. Would that work with our current naming conventions for component tokens?

oliverschuerch:
> > But maybe we could have only one map for all component tokens, what do you think?
> 
> Dang, I had so much fun. If we use one map, we loose the discoverability of all the tokens that belong to one component. What we could do instead is using a convention for the map name, so you'd only had to provide the name of the component without `post-` prefix, e.g. `tokens.get('h1', 'font-family')` and the function would do `map.get(post-h1, post-h1-font-family)`. Would that work with our current naming conventions for component tokens?

This will not work easily also, because we have 3 different files where the map could come from (elements, components, utilities).
Threrefore the map need to be namespaced, e.g. `map.get(ns.$post-headings, 'key')`

oliverschuerch:
> I get this error when using "with" in two files that end up in the same index file: image
> 
> But maybe we could have only one map for all component tokens, what do you think?

What happens, when you only build the styles package? Do you have the same problem there?
I'm asking, because if it's just a configuration problem (in vite), we could maybe fix it there.

alizedebray:
> What happens, when you only build the styles package? Do you have the same problem there?

I get the same issue: 
image

alizedebray:
> What we could do instead is using a convention for the map name

@gfellerph as far as I know, Sass doesn’t allow retrieving variables using a dynamic name like `token-maps.$post-#{$component-name}`. However, to improve discoverability, we could organize our tokens with nested maps. This way, all tokens for a component are grouped within a single map inside a root token map, like so:
```scss
$tokens: (
  'h1': (
    'post-h1-font-family': var(--post-sem-font-family-default),
  ),
);

map.get($tokens, h1, post-h1-font-family)
```

> This will as well not work without any problem, because we have 3 different files where the map could come from (elements, components, utilities).
@oliverschuerch is it necessary to have 3 separate files? If so, we could also create 3 files for the token functions: `element-tokens.get`, `component-tokens.get`, and `utility-tokens.get`, which could all be simplified to `tokens.get` when imported using the `as` keyword. What do you think?

oliverschuerch:
@alizedebray and @gfellerph 
I found another solution, how to set the used map.

```scss
 // here you could also load elements and/or utilities
@use '../tokens/components' as components;
@use '../functions/tokens' as tokens;
 
// set the default map to use or always use the third function parameter to specify a different map.
tokens.$default-map: components.$post-body;
 
// lets say you have two keys: 'post-selector-font-family' and 'post-selector-font-size' in the same map.
// then use it like so:
selector {
  // tokens.get($prefix, $key-name, $map)
  // $prefix can be used with or without "post-"
  font-family: tokens.get('post-selector', 'font-family');
  font-family: tokens.get('selector', 'font-family');
  font-family: tokens.get('selector-font', 'family');
  font-size: tokens.get('selector-font', 'size');
  font-family: tokens.get('selector', 'font-family', $my-custom-map);  // use a different map than the default one
}
```

I've pushed this solution. Try it out if you like ;)

oliverschuerch:
> > What we could do instead is using a convention for the map name
> 
> @gfellerph as far as I know, Sass doesn’t allow retrieving variables using a dynamic name like `token-maps.$post-#{$component-name}`. However, to improve discoverability, we could organize our tokens with nested maps. This way, all tokens for a component are grouped within a single map inside a root token map, like so:
> 
> ```css-scss
> $tokens: (
>   'h1': (
>     'post-h1-font-family': var(--post-sem-font-family-default),
>   ),
> );
> 
> map.get($tokens, h1, post-h1-font-family)
> ```
> 
> > This will as well not work without any problem, because we have 3 different files where the map could come from (elements, components, utilities).
> 
> @oliverschuerch is it necessary to have 3 separate files? If so, we could also create 3 files for the token functions: `element-tokens.get`, `component-tokens.get`, and `utility-tokens.get`, which could all be simplified to `tokens.get` when imported using the `as` keyword. What do you think?

This would be a possible solution, but as soon as we add another file (like 'elements', 'components' or 'utilities') we would need to add more functions.

alizedebray:
> This would be a possible solution, but as soon as we add another file (like 'elements', 'components' or 'utilities') we would need to add more functions.

Sure, though this will probably not happen happen too often.
The upside is that it will save us from having to write the map for every file.

oliverschuerch:
Depends on #3506

alizedebray (approved):

PR #3506, opened 2024-09-06
fix: e2e tests

oliverschuerch:


gfellerph (approved):
Thanks for looking into it @oliverschuerch!

PR #3509, opened 2024-09-09
chore(documentation): reset tabs story-container selectors

oliverschuerch:
We resetted the story-container implementation, because it caused a lot of problems in stories, where we make usage of `document`, etc.

depends on #3512


oliverschuerch:
Will be merged with #3512

Issue #3437, opened 2024-08-20
[element]: Bullet list


Cian77:
### Description

**Open issues:**

### Design
- [x]  Integrated new typography
- [ ]  Spacing is defined
- [x] Documentation final

https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations-%26-Components-Next-Level?node-id=21-170

### Tokens
- [x] Tokens for this component are ready

#### DEVs proposal

Base styles: `body`
Token set name: `Elements/ListBullet`
- `post-list-bullet-margin-block` (padding top/bottom on `List simple / Bullets`)
- `post-list-bullet-item-gap-block` (gap between `listItem 1` and `listItem 2`)
- `post-list-bullet-item-icon-gap-inline` (gap between `bulletIcon` and `bulletLabel`)
- `post-list-bullet-item-icon-size` (width of `bulletIcon`)
- `post-list-bullet-item-icon-text-padding-block` (padding of `bulletLabel`)
- `post-list-bullet-item-icon-bullet-size` (width/height of `Bullet`)

Code example:
```scss
ul {
  --post-body-line-height: 1.5em;
  --post-list-bullet-margin-block: 1rem;
  --post-list-bullet-item-gap-block: 8px;
  --post-list-bullet-item-icon-size: 32px;
  --post-list-bullet-item-icon-gap-inline: 8px;
  --post-list-bullet-item-icon-text-padding-block: 1px;
  --post-list-bullet-item-icon-bullet-size: 8px;

  margin-block-start: var(--post-list-bullet-margin-block);
  padding-inline-start: calc(
    var(--post-list-bullet-item-icon-size) +
      var(--post-list-bullet-item-icon-gap-inline)
  );
  list-style: none;

  li {
    position: relative;
    padding-block: var(--post-list-bullet-item-icon-text-padding-block);

    &:not(:first-child) {
      margin-block-start: var(--post-list-bullet-item-gap-block);
    }

    &::before {
      box-sizing: content-box;
      display: block;
      content: "";
      position: absolute;
      right: 100%;
      margin-block-start: calc(
        var(--post-body-line-height) * 0.5 -
          var(--post-list-bullet-item-icon-bullet-size) * 0.5
      );
      margin-inline-end: calc(
        var(--post-list-bullet-item-icon-size) * 0.5 -
          var(--post-list-bullet-item-icon-bullet-size) * 0.5 +
          var(--post-list-bullet-item-icon-gap-inline)
      );
[...]

PR #3511, opened 2024-09-09
feat: component check list

bashirkarimi:


packages/styles/src/components/list.scss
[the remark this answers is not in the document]
bashirkarimi: The `bottom` and `left` which have `0` value are reseting the `reboot` and can not be changed. The rest are changed accordingly.

oliverschuerch:
@bashirkarimi create a :after element, place it the same way as the :before element.
Set the background-color of it through the icon and define the icon as a mask-image (there is a mixin for that).

packages/styles/src/components/list.scss
```diff
@@ -0,0 +1,54 @@
...
+    position: relative;
+    padding-inline-start: calc(
+      #{tokens.get('post-list-checks', 'spacing-gap-inline-text'),
+      } + #{$sizing-icon}
+    );
+
+    &::before {
+      content: '';
+      position: absolute;
+      left: 0;
```
gfellerph, line 27: The check is not centrally aligned with the first line of text of the list item
image

packages/styles/src/components/list.scss
```diff
@@ -0,0 +1,54 @@
+@use './../themes/bootstrap/core' as *;
```
gfellerph, line 1: We're going to remove bootstrap, this import can't stay here. Please replace every reference to bootstrap with internal functionality.

packages/styles/src/components/list.scss
```diff
@@ -0,0 +1,54 @@
...
+      background-repeat: no-repeat;
+      background-position-x: 50%;
+      background-position-y: center;
+      background-color: tokens.get('post-list-checks', 'color-icon-bg');
+      border-radius: 50%;
+      background-size: calc(#{$sizing-icon} - #{$padding-icon} * 2);
+      background-image: url('#{icons.get-colored-svg-url("2105", "white")}'); // 'white' should be changed with variable
+    }
+
+    @include media-breakpoint-down(sm) {
```
gfellerph, line 40: Please replace this bootstrap mixin with it's counterpart from https://github.com/swisspost/design-system/blob/main/packages/styles/src/mixins/_media.scss

packages/styles/src/components/list.scss
bashirkarimi: It is removed.

packages/styles/src/components/list.scss
bashirkarimi: It is replaced.

packages/styles/src/components/list.scss
bashirkarimi: The size of icon is fixed.

gfellerph (changes requested):
On mobile, the checklist is now misaligned.

image

And in high contrast mode, the round checks disappear:
image

Can you add a rule for high contrast mode like

```CSS
    background-color: CanvasText;
    color: Canvas;
}
```

bashirkarimi:
> On mobile, the checklist is now misaligned.
> 
> image
> And in high contrast mode, the round checks disappear: image
> 
> Can you add a rule for high contrast mode like
> 
> ```css
> @include utilities.high-contrast-mode() {
>     background-color: CanvasText;
>     color: Canvas;
> }
> ```



> On mobile, the checklist is now misaligned.
> 
> image
> And in high contrast mode, the round checks disappear: image
> 
> Can you add a rule for high contrast mode like
> 
> ```css
> @include utilities.high-contrast-mode() {
>     background-color: CanvasText;
>     color: Canvas;
> }
> ```

bashirkarimi:
> On mobile, the checklist is now misaligned.
> 
> image
> And in high contrast mode, the round checks disappear: image
> 
> Can you add a rule for high contrast mode like
> 
> ```css
> @include utilities.high-contrast-mode() {
>     background-color: CanvasText;
>     color: Canvas;
> }
> ```

The high contrast mode is added.
The Tokens for icon size does not change on breakpoints, currently. Therefore, the size of icon is the same for all screen sizes and it overflow to bottom on mobile.
I asked Alessio about tokens changing on breakpoints and here is his reply: 

> The font-size is atm device dependent, meaning it pulls different numbers from the semantical level based on what screensice people are on. Not sure if it will just stay that way or if they want to switch it up once we got figma enterprise as we then can have 40 different breakpoints compared to 4 right now. Therefore as it is now: Desktop ≥ 1024, Tablet 600-1023, Mobile <600

gfellerph:
There should be a way to always align the checkmark on the first line of content, no matter the size of the checkmark or the text.

gfellerph (changes requested):
There should be a way to always align the checkmark on the first line of content, no matter the size of the checkmark or the text.

alizedebray:
Waiting for the bullet list styles to be implemented with [#3437](https://github.com/swisspost/design-system/issues/3437)

Issue #3412, opened 2024-08-19
[bug]: Counter increment on the stepper


gfellerph:
**Describe the bug**
Stepper increments happen on `::before` elements for the stepper. This creates wrong numbers in certain cases, when navigating back to previous steps.

**To Reproduce**
Design System Version used: 7.1
Reproducible in an internal Angular application, write @gfellerph or Pfeuti Andreas for more infos.

**Expected behavior**
Step increments are always correct.

**Screenshots**
![image](https://github.com/user-attachments/assets/95121509-32ce-4d87-b520-7d46c1798a6a)

**Additional context**
Proposed fix: move `counter-increment` from the `::before` element to the `li` element. Fix works on the internal application.

schaertim:
@gfellerph is there a link to the page where the behavior can be observed?

PR #3519, opened 2024-09-10
[bug]: Counter increment on the stepper #3412 (v8)

schaertim:


gfellerph (approved):
@schaertim please check the PR title, it does not have the format "changetype: description". That's why the action is failing.

PR #3521, opened 2024-09-10
1305 remove deprecated line height variables

veyaromain:


packages/documentation/src/stories/utilities/text/text-line-height.sample.scss
```diff
@@ -1,5 +1,5 @@
 @use '@swisspost/design-system-styles/core' as post;
 
 .my-element {
-  line-height: post.$line-height-tiny;
+  line-height: post.$line-height-regular;
```
alizedebray, line 4: `$line-height-regular`  is also part of the variables you removed. 
You can replace it by any `$line-height-*` variable that will remain in [_type.scss](https://github.com/swisspost/design-system/pull/3521/files#diff-902e14d896af3207e36ba58534b8c42295d698d54348a49abfde86b4b83c9063) after your changes.

alizedebray (approved):
Congratulations on your first PR for the Design System 👏

PR #3525, opened 2024-09-10
chore: reset release v7 workflow and set tag "latest-v7" to all publishable package.json files

oliverschuerch:


gfellerph (commented):
@oliverschuerch, the styles v7 are currently published under `7.4.2`. If we release v7 with a tag, say `@swisspost/design-system-styles@7.4.3@latest-v7`. Will somebody installing with `npm i @swisspost/design-system-styles@7` get `7.4.2` or `7.4.3@latest-v7`?

What happens on `npm install` when the styles are installed as `{ "@swisspost/design-system-styles": "^7.4.2" }`? Do they get `7.4.3@latest-v7`?

Maybe we should do some testing before implementing this.

oliverschuerch:
> @oliverschuerch, the styles v7 are currently published under `7.4.2`. If we release v7 with a tag, say `@swisspost/design-system-styles@7.4.3@latest-v7`. Will somebody installing with `npm i @swisspost/design-system-styles@7` get `7.4.2` or `7.4.3@latest-v7`?
> 
> What happens on `npm install` when the styles are installed as `{ "@swisspost/design-system-styles": "^7.4.2" }`? Do they get `7.4.3@latest-v7`?
> 
> Maybe we should do some testing before implementing this.

As far as I know:

- A taged version can always be installed without the tag, by just using a semver (like @7, @7.4, etc.).
- A tag set on a specific semver is always optional to use.
- A tag is unic, which means, you will never end up having to semvers using the same tag.

So:

- If you run `npm i package-name@7`, you will always get the latest version with the major version `7`.
- You could also use `npm i package-name@latest-v7`, but no one will probably ever do that.
- If you run `npm i package-name@latest` you will always get the version which contains this tag.

I don't think this should be a problem. But we can for sure test it ;)

alizedebray (approved):

gfellerph:
@oliverschuerch confirmed the behavior you described with an internal test. Tagged packages installed with in-the-middle-versions will install according to their version disregarding the tag and will not override the higher versioned latest tag.

gfellerph (approved):

PR #3528, opened 2024-09-10
feat(docs): add toolbar to switch theme, channel, and mode

alizedebray:


oliverschuerch:
I thought, we decided to integrate the attributes on the preview body element? They are still on the `.sbdocs-preview` element.
@alizedebray can you confirm or decline?

alizedebray:
> I thought, we decided to integrate the attributes on the preview body element? They are still on the `.sbdocs-preview` element. @alizedebray can you confirm or decline?

@oliverschuerch the post-internal/port-external styles are applied to the whole preview document, only the `data-color-mode` attribute is set on the `.sbdocs-preview` to avoid having to fix the dark theme for the preview: 
image

oliverschuerch (approved):

PR #3530, opened 2024-09-10
chore: use a namespace for sites on netlify

gfellerph:
@oliverschuerch I think it would be a good idea to namespace netlify pages to the swisspost company. `design-system-version-x` is living in the global netlify namespace and could be any design system in the world.

To complete this task, we should copy netlify pages and create `swisspost-...` prefixed versions of all old version sites. If I understand the workflow correctly, with these updates and the new prefixed pages on netlify, everything should work again. Correct me if I'm wrong.

- [ ] Copy existing old version sites on netlify and create `swisspost-...` namespaced ones

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -120,7 +120,7 @@ jobs:
               old: {
                 version: release.oldVersion,
                 major: oldMajor,
-                siteUrl: `design-system-version-${oldMajor}`,
+                siteUrl: `swisspost-design-system-version-${oldMajor}`,
                 dependencies: {}
               },
               new: {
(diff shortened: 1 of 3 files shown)


oliverschuerch:
The following steps would be necessary if you like to prefix all of the netlify pages:

- [ ] Create the new pages on netlify for v6, v7, v8 v9 (I don't know why `design-system-version-9` already exists, but anyway).
- [ ] Update the `versions.json` on the `main` branch, to make it future proof.
The version switcher on design-system.post.ch is broken since several weeks now, because the `release/v8` `versions.json` file already specifies the prefixed URLs, but there are no existing netlify pages for those URLs!
- [x] Update the `release.yaml` file, so the old version is specified with the `swisspost` prefix.

oliverschuerch (approved):

oliverschuerch:
Fixing the branch `release/v8` should be done with #3663

PR #3534, opened 2024-09-11
chore: foundations typography heading h1   h6

bashirkarimi:


oliverschuerch:
Depends on #3543

packages/styles/src/elements/heading.scss
```diff
@@ -0,0 +1,62 @@
+@use '../functions/tokens' as tokens;
+@use '../tokens/elements' as elements;
```
gfellerph, line 2: With `@use`, the file name automatically becomes the namespace. Overriding the namespace name with `as X` is only necessary when the filename is not optimal or two identically named files are being imported.

```suggestion
@use '../functions/tokens';
@use '../tokens/elements';
```

.changeset/eleven-keys-work.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': patch
+'@swisspost/design-system-styles': patch
+---
+
+Updated the style of Headings with design token
```
gfellerph, line 6:
```suggestion
Updated the style of Headings (h1-h6) using design tokens.
```

.changeset/eleven-keys-work.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': patch
+'@swisspost/design-system-styles': patch
+---
+
+Updated the style of Headings (h1-h6) using design tokens.
```
alizedebray, line 6: I suggest not to mention the tokens in the change log, as they are part of our internal technology that users don't necessarily need to be aware of:
```suggestion
Updated the style of headings (h1-h6).
```

alizedebray (approved):

gfellerph (approved):

PR #3539, opened 2024-09-11
feat: allow to include the navigation bar in a single line header

yopemu:
Allow the option to include the navigation bar in a single line as a condensed header. Instead of 2 lines only one (the top one) will contains the logo, title and navigation bar, as well as the settings. As discussed with Debray Alize and Gfeller Philipp.

Team NES
Langhard Matthias
Pedroso Yohandris


yopemu:
Hi @gfellerph as suggested, here is the direct PR

gfellerph:
Resolution pattern discussion: thumbs up

packages/intranet-header-workspace/projects/intranet-header/src/lib/swisspost-intranet-header.component.html
```diff
@@ -107,32 +115,48 @@ <h1 class="m-0 font-weight-bold">{{ siteTitle }}</h1>
       </div>
     </div>
     <nav
-      *ngIf="hasNavbar === true"
+      *ngIf="hasNavbar === true && !condenseHeader"
       class="row justify-content-center top-navigation m-0 d-md-flex"
+      aria-label="navigation"
```
gfellerph, line 120: This causes screenreaders to announce "navigation navigation" as the `<nav>` already prompts a (localized) announcement of navigation. When adding this statically, there is also no chance to localize it.

```suggestion
```

packages/intranet-header-workspace/projects/intranet-header/src/lib/swisspost-intranet-header.component.html
```diff
@@ -2,32 +2,40 @@
   <div class="nav-down">
     <div class="intranet-header">
       <div class="d-flex align-items-center flex-nowrap">
-        <div class="bg-logo d-sm-block">
+        <div id="logo" class="bg-logo d-sm-block">
```
gfellerph, line 5: Why was this id added? There is a chance of this id already being in use by a project, which could cause hard to find errors. If this id is necessary, we should prefix it with `post-` or an even more specific prefix. Otherwise, please remove it.

*Same goes for the other instances where ids were added.*

packages/intranet-header-workspace/projects/intranet-header/src/lib/swisspost-intranet-header.component.html
```diff
@@ -2,32 +2,40 @@
...
-                <button class="btn pr-4" type="submit">
-                  <span><em class="pi pi-2069"></em></span>
-                </button>
-              </div>
-            </div>
-          </form>
+
+        <nav
+          *ngIf="condenseHeader && hasNavbar"
+          aria-label="navigation"
```
gfellerph, line 20: This causes screenreaders to announce "navigation navigation" as the `<nav>` already prompts a (localized) announcement of navigation. When adding this statically, there is also no chance to localize it.

```suggestion
```

packages/intranet-header-workspace/projects/intranet-header/src/lib/swisspost-intranet-header.component.ts
```diff
@@ -333,6 +345,23 @@ export class SwissPostIntranetHeaderComponent implements OnInit, OnChanges, Afte
...
+      this.titleElement,
+      this.optionHeaderContentElement,
+      this.profileMenuElement,
+      this.intranetSearchElement,
+    ];
+    const totalWidth = elements.reduce((sum, element) => sum + (element?.scrollWidth || 0), 0);
+
+    return document.documentElement.scrollWidth - totalWidth;
+  }
+
```
gfellerph, line 364: Nice!

packages/styles/src/components/intranet-header/_top-navigation.scss
gfellerph, line 1: There's a one pixel gap around the logo, possibly caused by the nav items that are 2px bigger than the logo.

image

gfellerph (changes requested):
Thanks for doing the extensive rework necessary for providing this option and also thinking about documentation. The PR generally looks good, there are some things however that need to be fixed before we can merge it.

gfellerph:
Notice for @swisspost/design-system-maintainers, this PR needs to go to main and then needs a cherry pick to `release/v8`.

yopemu:
@gfellerph Thank you for the review and your feedback. I will have a look at your suggestions.

packages/intranet-header-workspace/projects/intranet-header/src/lib/swisspost-intranet-header.component.html
yopemu: I see. I only added it afterwards because sonar failed when it wasn't there. I will remove it again.

packages/intranet-header-workspace/projects/intranet-header/src/lib/swisspost-intranet-header.component.html
yopemu: Yes, the IDs are needed to identify the containers and then calculate the available space. This identifies them uniquely. I see the point with the other projects and will add the prefix “post-”.

riscie:
Any news on this one @gfellerph? Is the ball in our court?

gfellerph:
Superseded by #4084 because this change here was based on main, where the Demo App is already removed and the Intranet Header will be removed soon (to be replaced by the new web component header).

PR #3545, opened 2024-09-16
feat: add CSS reset style

bashirkarimi:


packages/styles/src/elements/reset.scss
```diff
@@ -0,0 +1,122 @@
...
+  margin: 0;
+}
+
+/*
+  Typographic tweaks!
+  - Add accessible line-height
+  - Improve text rendering
+*/
+body {
+  line-height: 1.5;
```
oliverschuerch, line 25: I think we can remove this, as we will define it anyway ;)
Everything else looks good!

oliverschuerch (changes requested):

packages/styles/src/_core.scss
```diff
@@ -3,3 +3,4 @@
 @forward './mixins';
 
 @import './themes/bootstrap/core';
+@import './elements';
```
oliverschuerch, line 6: I'd say, this is definitively the wrong place to integrate the elements files.
And please do not use `@import` statements (it's deprecated and we only use it to import bootstrap files, because they can not be used in a different way). Instead, use `@use` or `@forward`.

In my opinion, we need to import it here:
- `styles/src/base.scss`
- `styles/src/index.scss`
- `styles/src/intranet.scss`
- `styles/src/post-external.scss`
- `styles/src/post-internal.scss`

oliverschuerch (changes requested):

packages/styles/src/_core.scss
bashirkarimi: The `elements` style is moved to mentioned places.
The `elements` style is moved to `styles/src/components/base` for now and it should be moved/removed when the `/root` and `.reboot` are cleaned-up.

packages/styles/src/elements/reset.scss
bashirkarimi: It is removed.

packages/styles/src/_core.scss
bashirkarimi: The `elements` is moved to `styles/src/basic` instead of `base`.

oliverschuerch (approved):

PR #3557, opened 2024-09-18
feat(styles): create spacing utilities from tokens

alizedebray:
I recommend reviewing this pull request one commit at a time:

- The ***first commit*** (`feat(styles): create spacing utilities from tokens`) is the actual implementation of the new margin, padding, and gap utilities from the token map.

- The ***second commit*** (`chore: apply new spacing utilities`) is purely a refactor, where the old spacing utilities are replaced with the newly created ones.


gfellerph (approved):
I've added #3639 to document the breaking change.

PR #3592, opened 2024-09-26
fix(component): fix an issue related to conflicting pointer and focus events hiding the tooltip unexpectedly

alionazherdetska:


packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -186,10 +195,22 @@ export class PostTooltip {
...
-  async show(target: HTMLElement) {
+  async show(target: HTMLElement, triggeredByFocus = false) {
     if (this.delayed) await timeout(OPEN_DELAY);
+
+    // Determine if the tooltip was opened by a focus event
+    this.wasOpenedByFocus = triggeredByFocus;
+
+    // Disable pointer events if triggered by focus, otherwise enable them
+    if (this.wasOpenedByFocus) {
+      this.host.style.pointerEvents = 'none';
```
gfellerph, line 209: There is an accessibility issue with this. If the tooltip was opened by focus, you can no longer interact with it with a pointer. The tooltip behaves differently from a pointer-opened tooltip.

Pointer opened:
- Can click on it, tooltip does not disappear
- Can select/copy text

Focus opened:
- Closes on click
- Cannot select text

It would be better to unregister the pointerout event that would close the tooltip (or something along those lines) in this case.

gfellerph (changes requested):

gfellerph:
@alionazherdetska thanks for tackling this issue. I think the direction is generally the right one.

gfellerph (approved):

PR #3596, opened 2024-09-27
chore(docs): document `noFlyout` property in Internet Header

alionazherdetska:


gfellerph (changes requested):
Good job! Can you please add a changeset for the internet-header package stating something like "Added and documented the possibility to create plain links in the main navigation by adding `noFlyout: true` to the config. The `flyout` property is now optional and can be omitted.".

gfellerph (approved):

PR #3616, opened 2024-10-02
feat(styles): component form footer

leagrdv:


packages/styles/src/components/form-footer.scss
```diff
@@ -0,0 +1,36 @@
...
+    @include utility-mx.responsive-actions();
+    gap: tokens.get('post-form-footer', 'post-form-footer-gap');
+
+    @include media-breakpoint-up(md) {
+      margin-left: auto;
+    }
+  }
+
+  @include media-breakpoint-up(md) {
+    padding-top: tokens.get('post-form-footer', 'post-form-footer-padding-top-m');
```
alizedebray, line 33: The tokens are CSS custom properties that are automatically updated for all breakpoints, you do not have to handle the responsiveness manually in the component styles.

packages/styles/src/components/form-footer.scss
```diff
@@ -0,0 +1,36 @@
...
+@use '../mixins/utilities' as utility-mx;
+@use '../tokens/temp-form-footer' as components;
+@use '../functions/tokens' as tokens;
+
+tokens.$default-map: components.$post-form-footer;
+
+.form-footer {
+  border-top: tokens.get('post-form-footer', 'post-form-footer-border-top-width') solid
+    tokens.get('post-form-footer', 'post-form-footer-border-top-color');
+  padding-top: tokens.get('post-form-footer', 'post-form-footer-padding-top-s');
```
alizedebray, line 17: We recently decided to move to inline/block properties instead of top/bottom/right/left.
```suggestion
  padding-block-start: tokens.get('post-form-footer', 'post-form-footer-padding-top-s');
```

packages/styles/src/mixins/_utilities.scss
```diff
@@ -176,3 +178,12 @@
     }
   }
 }
+
+@mixin responsive-actions {
```
alizedebray, line 182: Can you move the [margin-left](https://github.com/swisspost/design-system/pull/3616/files#diff-2f8b0d15fcfb213824d105a0ac65edbd2aefe77692061fcf8d1585f1c0773b64R28) and [flex-direction](https://github.com/swisspost/design-system/pull/3616/files#diff-2f8b0d15fcfb213824d105a0ac65edbd2aefe77692061fcf8d1585f1c0773b64R34) to the mixin as well so that we do not have media queries in the component styles at all?

packages/documentation/src/stories/components/forms/form-footer/form-footer.stories.ts
```diff
@@ -0,0 +1,56 @@
...
+    },
+  },
+  args: {
+    primaryButton:
+      '<button class="btn btn-primary">Send<post-icon aria-hidden="true" name="3020"></post-icon></button>',
+    secondaryButton: '<button class="btn btn-secondary">Cancel</button>',
+    tertiaryButton:
+      '<button class="btn btn-tertiary"><post-icon aria-hidden="true" name="3024"></post-icon>Back</button>',
+  },
+  argTypes: {
```
alizedebray, line 25: We usually add a category and a description for all story args as we have [here](https://github.com/swisspost/design-system/blob/main/packages/documentation/src/stories/components/badge/badge.stories.ts#L29). If there is only one category you can name it "General".

alizedebray (changes requested):

packages/styles/src/mixins/_utilities.scss
leagrdv: I moved the margin-left to the mixin, but the flex-direction: row is outside of those two action buttons on the right

alizedebray: You can refactor the mixin and apply it to the parent, I could imagine it being used in multiple components like modal, [notifications](https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations-%26-Components-Next-Level?node-id=85-321&node-type=canvas&t=ia7Kw9j7TcqUAhOm-0), and [accordions](https://www.figma.com/design/xZ0IW0MJO0vnFicmrHiKaY/Components-Post?node-id=42-358&node-type=canvas&m=dev).

This way the &-right-actions` selector would be part of the mixin and the classes would be consistent for all components. What do you think?

leagrdv: You're right, it makes more sense that way :)

packages/styles/src/components/form-footer.scss
```diff
@@ -1,32 +1,18 @@
...
-  justify-content: space-between;
-  gap: tokens.get('form-footer', 'gap');
-
-  &-right-actions {
-    @include utility-mx.responsive-actions();
-    gap: tokens.get('form-footer', 'gap');
-  }
+  @include utility-mx.responsive-actions();
+  border-block-start: tokens.get('form-footer-border-block-start-width')
+    tokens.get('form-footer-border-start-color');
```
alizedebray, line 10:
```suggestion
  border-block-start-width: tokens.get('form-footer-border-block-start-width');
  border-block-start-color: tokens.get('form-footer-border-start-color');
```

packages/documentation/src/stories/components/forms/form-footer/form-footer.stories.ts
```diff
@@ -0,0 +1,70 @@
...
+  parameters: {
+    badges: [],
+    design: {
+      type: 'figma',
+      url: 'https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations-%26-Components-Next-Level?node-id=1498-28215',
+    },
+  },
+  args: FooterArgs,
+  argTypes: {
+    primaryButton: {
```
alizedebray, line 28: Can you separate this control into three: 
- showPrimary button
- primaryButtonType
- primaryButtonText 

Same for the other buttons, it will prevent people from testing with invalid HTML.

packages/documentation/src/stories/components/forms/form-footer/form-footer.stories.ts
```diff
@@ -0,0 +1,70 @@
...
+};
+
+export default meta;
+
+type Story = StoryObj;
+
+export function render(args: Args) {
+  return html`
+    <div class="form-footer">
+      <div class="form-footer-primary-actions">
```
alizedebray, line 62: Ca you hide this div if there is no button inside please?

alizedebray (changes requested):

packages/documentation/src/stories/components/forms/form-footer/form-footer.stories.ts
leagrdv: I added the showXButton and xButtonText for each of them :) Do you think adding the xButtonType is necessary? As even the name suggests its type already, I don't know if we want users to be able to change the type of the button?

alizedebray: You're right, it's good like this!

packages/documentation/src/stories/components/forms/form-footer/form-footer.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export function render(args: Args) {
+  return html`
+    <div class="form-footer">
+      ${args.showPrimaryButton || args.showSecondaryButton
+        ? html`
+            <div class="form-footer-primary-actions">
+              ${args.showPrimaryButton
+                ? html`<button class="btn btn-primary">
+                    ${args.primaryButtonText}<post-icon aria-hidden="true" name="3020"></post-icon>
```
alizedebray, line 99:
```suggestion
                    ${args.primaryButtonText}<post-icon aria-hidden="true" name=${args.primaryButtonIcon}></post-icon>
```

packages/documentation/src/stories/components/forms/form-footer/form-footer.stories.ts
```diff
@@ -0,0 +1,117 @@
...
+      description: 'Show or hide the primary button (last one on the right)',
+      control: { type: 'boolean' },
+      table: {
+        category: 'Primary button',
+      },
+    },
+    primaryButtonText: {
+      name: 'Primary button text',
+      description: 'Text to display on the primary button',
+      control: { type: 'text' },
```
alizedebray, line 39: Can you hide the controls for button X when showXButton is false, like so:
```suggestion
      control: { type: 'text' },
      if: {
        arg: 'showPrimaryButton',
      },
```

alizedebray (approved):

PR #3619, opened 2024-10-02
fix(components): fixed accessibility issue with tooltips

schaertim:


oliverschuerch (changes requested):
I don't think to move the role attribute into the shadowdom is a good solution.
And it does for sure not fix the mentioned error, but bypass it, by simply hiding the semantic tooltip from a screenreader.
Instead a tooltip should have a `aria-label`, `aria-labelledby` or a `title` attribute, which specifies its name.
See: [Learn how to name tooltip elements](https://dequeuniversity.com/rules/axe/4.9/aria-tooltip-name)

Lighthouse original message:
image

oliverschuerch (approved):

PR #3623, opened 2024-10-03
chore(styles): remove deprecated text placeholders

leagrdv:


packages/styles/tests/placeholders/text.test.scss
```diff
@@ -1,8 +1,6 @@
 @use 'src/placeholders/text';
 
 .list-adjustment {
```
leagrdv, line 3: Should this also be removed? It isn't used anywhere in the project.

alizedebray: These `*.test.scss` files are tests files, the tests are run on every PR and you can also run them manually with the `pnpm run styles:test` command. 
This class is not exported and its name does not really matter. It is here to make sure the `src/placeholders/text` file can be imported correctly and that it contains all expected placeholder.

alizedebray (approved):

PR #3624, opened 2024-10-03
chore(demo): removed demo app and workflows

schaertim:


oliverschuerch (changes requested):
❤️❤️❤️❤️❤️
 
Also remove `.github/actions/deploy-to-netilfy`

oliverschuerch (approved):

PR #3638, opened 2024-10-03
feat(documentation): reorganized accessibility folder and added form grouping controls section

myrta2302:


packages/documentation/src/stories/foundation/accessibility/regulation.docs.mdx
```diff
@@ -0,0 +1,17 @@
+import { Meta, Source, Canvas } from '@storybook/blocks';
+import * as RegulationStories from './regulation.stories';
+
+<Meta of={RegulationStories} />
+
+# Accessibility
+
+<figure>
+  <blockquote class="blockquote" lang="de">
+    <p>Die Postdienste müssen so angeboten werden, dass Menschen mit Behinderungen
```
oliverschuerch, line 10: Can we translate this to english? For example with the post translater or with deepl or some other online service.

packages/documentation/src/stories/foundation/accessibility/regulation.docs.mdx
```diff
@@ -0,0 +1,17 @@
...
+sie in qualitativer, quantitativer und wirtschaftlicher Hinsicht unter vergleichbaren
+Bedingungen wie Menschen ohne Behinderungen beanspruchen können.</p>
+  </blockquote>
+  <figcaption class="blockquote-footer">
+    <cite>§14.7 <a href="https://www.fedlex.admin.ch/eli/cc/2012/585/de#art_14">Swiss postal law</a></cite>
```
oliverschuerch, line 15: So the language can be choosen by the page itself, depending on the browser language.
```suggestion
    <cite>§14.7 <a href="https://www.fedlex.admin.ch/eli/cc/2012/585">Swiss postal law</a></cite>
```

oliverschuerch (changes requested):
Also it would be nice to have the "Regulations" page at the beginning of the newly created subpages. So "Accessibility/Regulations" comes first and the others do not need an order.

To do so, you can open the `documentation/.storybook/preview.ts` and add the following:
image

oliverschuerch (approved):

PR #3659, opened 2024-10-04
feat(components): create the post-breadcrumb-item

alionazherdetska:


packages/styles/src/components/breadcrumb.scss
```diff
@@ -18,24 +18,3 @@
   font-weight: type.$font-weight-normal;
   white-space: nowrap;
 }
-
-.breadcrumb-item {
```
alizedebray, line 22: Removing these styles is a breaking change. Could you write a changeset to documented this?

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.scss
```diff
@@ -0,0 +1,47 @@
+@use 'sass:map';
+@use '@swisspost/design-system-styles/mixins/media';
+@use '@swisspost/design-system-styles/functions/tokens';
+@use '@swisspost/design-system-styles/tokens/components';
+@use '@swisspost/design-system-styles/mixins/utilities';
+
+tokens.$default-map: components.$post-breadcrumb;
+
+:host {
+  display: block;
```
alizedebray, line 10:
```suggestion
  display: inline-block;
```

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.scss
```diff
@@ -0,0 +1,47 @@
...
+}
+
+a {
+  display: inline-flex;
+  align-items: center;
+  padding-block: tokens.get('post-breadcrumb', 'padding-block-text');
```
alizedebray, line 16: Can you update the styles to use the tokens function the new way?
```suggestion
  padding-block: tokens.get('breadcrumb-padding-block-text');
```

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.tsx
```diff
@@ -0,0 +1,41 @@
...
+  styleUrl: 'post-breadcrumb-item.scss',
+  shadow: true,
+})
+export class PostBreadcrumbItem {
+  @Element() host: HTMLPostBreadcrumbItemElement;
+
+  /**
+   * The URL to which the breadcrumb item will link.
+   */
+  @Prop() readonly url!: string;
```
alizedebray, line 20: Are we sure the URL should be mandatory?

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  @Watch('url')
+  validateUrl() {
+    checkEmptyOrUrl(this.url, 'The "url" property is invalid');
```
alizedebray, line 24: If the URL is mandatory:

```suggestion
    checkUrl(this.url, 'The "url" property is invalid');
```

alizedebray (changes requested):

.changeset/proud-cheetahs-act.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-components': major
+'@swisspost/design-system-styles': major
+---
+
+Removed the .breadcrumb-item class, which previously handled styling for breadcrumb items. Introduced a new BreadcrumbItemComponent that should be used in place of the .breadcrumb-item class.
```
alizedebray, line 6:
```suggestion
Removed the `.breadcrumb-item` class, which previously handled styling for breadcrumb items. Introduced a new `post-breadcrumb-item` that should be used in place of the `.breadcrumb-item` class.
```

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.scss
```diff
@@ -0,0 +1,44 @@
...
+  display: inline-flex;
+  align-items: center;
+  padding-block: tokens.get('breadcrumb-padding-block-text');
+  gap: tokens.get('breadcrumb-gap-inline-inner');
+  color: tokens.get('breadcrumb-enabled-fg');
+  text-decoration: none;
+
+  &:hover {
+    color: tokens.get('breadcrumb-hover-fg');
+    text-decoration: underline;
```
alizedebray, line 23: This text decoration should probably be tokenized.

packages/components/src/utils/property-checkers/check-url.ts
```diff
@@ -2,7 +2,7 @@ export function checkUrl(value: unknown, error: string) {
   if (typeof value !== 'string' && !(value instanceof URL)) throw new Error(error);
 
   try {
-    new URL(value);
+    URL.canParse(value, window.location.origin);
```
alizedebray, line 5: This always returns `true`. In your component it would be better to use `checkType` instead of `checkUrl`.

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.tsx
```diff
@@ -0,0 +1,49 @@
...
+  @Element() host: HTMLPostBreadcrumbItemElement;
+
+  /**
+   * The optional URL to which the breadcrumb item will link.
+   */
+  @Prop() url?: string | URL;
+
+  @Watch('url')
+  validateUrl() {
+    checkUrl(this.url, 'The "url" property of the post-breadcrumb-item is invalid');
```
alizedebray, line 23: Since the URL is optional, you should use an `emptyOr` validator like so:
```suggestion
    checkEmptyOrUrl(this.url, 'The "url" property of the post-breadcrumb-item is invalid');
```

alizedebray (changes requested):

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.tsx
```diff
@@ -0,0 +1,61 @@
...
+  @Prop() url?: string | URL;
+
+  private validUrl?: string;
+
+  @Watch('url')
+  validateUrl() {
+    try {
+      this.validUrl = this.constructUrl(this.url);
+    } catch (error) {
+      console.error(error);
```
alizedebray, line 28: Can you just check if the error is logged twice and if so just remove this line?

alizedebray (approved):

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.tsx
alionazherdetska: Will do.

PR #3667, opened 2024-10-07
fix: remove margin reset

gfellerph:
Resetting the margin on all elements breaks some default styles like the native dialog placement.


packages/styles/src/elements/reset.scss
```diff
@@ -9,13 +9,6 @@
   box-sizing: border-box;
 }
 
-/*
-  Remove default margin
-*/
-* {
-  margin: 0;
-}
```
oliverschuerch, line 17: But to simply remove these lines will not be the solution imo, because other elements like the `body` normaliy need to have the margin resetted.

What do you think will be the best solution here:
- [ ] Reset the margin per component (where needed)?
- [ ] Reset the margin globally in the reset.scss but only for some specific elements?
- [ ] Reset the margin globally in the reset.scss for all elements and revert that for elements, where it's needed (for example with the `margin: revert` statemant)?

oliverschuerch (changes requested):

packages/styles/src/elements/reset.scss
gfellerph: True, I don't really like the `margin: revert` approach, it makes it more complicated for the components that depend on standard margins. Since the goal of the reset is to basically smooth over browser differences, I'd put it to the element/component.

oliverschuerch (approved):

PR #3688, opened 2024-10-09
chore: create token PR workflow

alizedebray:


.github/workflows/create-token-pr.yaml
```diff
@@ -0,0 +1,58 @@
...
+    steps:
+      # Checkout the token branch which was pushed
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      # Get the name of the token branch which was pushed and the corresponding PR branch
+      - name: Get Branch
+        id: branch-names
+        run: |
+          echo "token-branch=tokens/$(echo ${GITHUB_REF##*/})"    >> $GITHUB_OUTPUT
```
oliverschuerch, line 26: Wouldn't that be the same as `${{ github.ref_name }}` (which by the way is provided globally in a workflow)?
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

With this, you may be able to reduce on step.

.github/workflows/create-token-pr.yaml
```diff
@@ -0,0 +1,58 @@
...
+        env:
+          PR_BRANCH: ${{ steps.branch-names.outputs.pr-branch }}
+
+      # if the branch already exits, update the PR
+      - name: Update Branch
+        if: steps.git-branch.outputs.pr_branch != ''
+        run: |
+          git checkout $PR_BRANCH
+          git merge $TOKEN_BRANCH -X theirs --no-edit
+          git push
```
oliverschuerch, line 44: Would be nice, if the `Swiss Post Bot` could be used as the author. See `release.yaml` -> `Commit Changes and Push Branch` how this can be done.

.github/workflows/create-token-pr.yaml
```diff
@@ -0,0 +1,58 @@
...
+      # if the branch does not exit, create the PR
+      - name: Create PR
+        if: steps.git-branch.outputs.pr_branch == ''
+        run: |
+          git checkout -b $PR_BRANCH $TOKEN_BRANCH
+          gh pr create --title "chore(tokens): :art: update tokens"
+        env:
+          TOKEN_BRANCH: ${{ steps.branch-names.outputs.token-branch }}
+          PR_BRANCH: ${{ steps.branch-names.outputs.pr-branch }}
+          GITHUB_TOKEN: ${{ secrets.SWISSPOSTDEVS_ACCESS_TOKEN }}
```
oliverschuerch, line 58: Would be nice to use the `actions/checkout@v4` action to checkout this branch (since you already using it at the begining).
For exampe, see `release.yaml` -> `update-changeset-branch`.
Also, if you once defined the token (GITHUB_TOKEN) with said action, it will be used through all following git actions in the workflow.

oliverschuerch (changes requested):

.github/workflows/create-token-pr.yaml
alizedebray: I don't think I can use EndBug/add-and-commit@v9 here since I am not creating a new commit but a merge commit.

.github/workflows/create-token-pr.yaml
alizedebray: I don't think I can use `actions/checkout@v4` to create a new branch either...

oliverschuerch: You're already using it on line19/20 ^^, but yeah, you would need to split this step up in at least to independant steps.

oliverschuerch: As mentioned, as long as it works, I'm fine with it.

oliverschuerch (approved):

.github/workflows/create-token-pr.yaml
alizedebray: I added the GITHUB_TOKEN which should do the same.

PR #3704, opened 2024-10-09
fix(styles): reworked get-function for tokens

schaertim:


packages/documentation/.storybook/styles/components/typo.scss
```diff
@@ -30,7 +30,7 @@ body {
     }
 
     .docs-story > :first-child {
-      font-family: tokens.get('post-body', 'font-family');
+      font-family: tokens.get('post-body-font-family');
```
alizedebray, line 33: This PR would be a good opportunity to remove the "post-" prefix, don't you think?
```suggestion
      font-family: tokens.get('body-font-family');
```

alizedebray (approved):

alizedebray:
Please notify all developers as soon as this PR is merged so that they can update their PR's.

gfellerph (changes requested):
@schaertim please address the linting error

packages/documentation/.storybook/styles/components/typo.scss
schaertim: Right now the post- prefix is optional. By removing it do you mean just removing it from the method calls using it right now?

gfellerph (approved):

packages/documentation/.storybook/styles/components/typo.scss
oliverschuerch: @alizedebray This has been implemented as you suggested it now 😉

PR #3709, opened 2024-10-10
fix(styles): compile element output to CSS

gfellerph:
All files without leading underscore are now compiled on styles:build. This will also fix future issues in this direction but requires stricter usage of the sass partial file naming convention with leading underscores.


packages/styles/gulpfile.js
```diff
@@ -136,7 +136,7 @@ gulp.task('sass', () => {
  */
 gulp.task('build-components', () => {
   return gulp
-    .src('./src/components/*.scss')
+    .src('./src/**/*.scss')
```
alizedebray, line 139: @gfellerph It looks like the `sass` and `build-components` tasks are now the same, no?

gfellerph: True, it can be removed.

alizedebray (approved):

PR #3711, opened 2024-10-10
patch(intranet-header): add alt text to user profile image

leagrdv:


packages/intranet-header-workspace/projects/intranet-header/src/lib/swisspost-intranet-header.component.html
```diff
@@ -60,7 +60,7 @@ <h1 class="m-0 font-weight-bold">{{ siteTitle }}</h1>
               id="profileMenu"
               ngbDropdownToggle
             >
-              <img [src]="avatarUrl" class="profile-image align-self-center" alt="" />
+              <img [src]="avatarUrl" class="profile-image align-self-center" alt="User's avatar" />
```
alizedebray, line 63: This has to be translated, you can use the `localization` object for that: https://github.com/swisspost/design-system/blob/2f8ea64c0748a9ba1e0983e3c421eb816b4f58a8/packages/intranet-header-workspace/projects/intranet-header/src/lib/swisspost-intranet-header.component.ts#L53

alizedebray (changes requested):

alizedebray (approved):

PR #3714, opened 2024-10-10
feat(tokens): added token coverage script

schaertim:


package.json
```diff
@@ -81,7 +81,8 @@
     "primeng:start": "pnpm --filter design-system-styles-primeng-workspace start",
     "primeng:build": "pnpm --filter design-system-styles-primeng-workspace build",
     "primeng:lint": "pnpm --filter design-system-styles-primeng-workspace lint",
-    "primeng:lint:fix": "pnpm --filter design-system-styles-primeng-workspace lint:fix"
+    "primeng:lint:fix": "pnpm --filter design-system-styles-primeng-workspace lint:fix",
+    "check-tokens": "node ./utilities/token-checker.cjs"
```
alizedebray, line 85:
```suggestion
    "tokens:usage": "node ./utilities/token-checker.cjs"
```

alizedebray (changes requested):
Very nice! 

There is just an issue with the "post-utility-*" tokens, that are reported as not used even though they are (not with the `tokens.get` function though). Maybe we can just remove them from the report.

Also, it would be nice to be able to filter with a component name, don't you think?

schaertim:
> Very nice!
> 
> There is just an issue with the "post-utility-*" tokens, that are reported as not used even though they are (not with the `tokens.get` function though). Maybe we can just remove them from the report.
> 
> Also, it would be nice to be able to filter with a component name, don't you think?

I changed it that the utility tokens are not checked anymore. Right now it is possible to filter by certain token maps like this: ´tokens:usage --maps post-button post-switch´ for example. Is that what you meant by filtering by component name?

alizedebray:
> Right now it is possible to filter by certain token maps like this: ´tokens:usage --maps post-button post-switch´ for example. Is that what you meant by filtering by component name?

Yes, perfect!

alizedebray (approved):

PR #3719, opened 2024-10-10
chore: fix homepage link

gfellerph:
This sets the homepage field in the package.json of some packages to our actual, nice URL.


oliverschuerch (approved):
What about the other packages?
Shouldn't be this link set/updated there as well?
For example in the components/package.json

gfellerph:
> What about the other packages? Shouldn't be this link set/updated there as well? For example in the components/package.json

I'll create a follow up ticket to align all package.json files on this and other fields.

PR #3734, opened 2024-10-11
fix(styles): mixed decls for chip

oliverschuerch:


.changeset/brave-coins-grin.md
gfellerph, line 1: pls remove

gfellerph (approved):

PR #3735, opened 2024-10-11
fix(styles): mixed decls for close

oliverschuerch:


.changeset/breezy-vans-cough.md
gfellerph, line 1: I don't think this is necessary. This is an internal change only, the component should look the same as before.

gfellerph (approved):

.changeset/breezy-vans-cough.md
oliverschuerch: Removed it

PR #3740, opened 2024-10-11
feat(documentation,styles): update list group component to v2 and add list links, list documents and list switches

leagrdv:


oliverschuerch (commented):
@leagrdv Where is the name `list-group` coming from?

packages/styles/src/components/list-group.scss
```diff
@@ -1,35 +1,21 @@
...
 
 .list-group-item-action {
-  @include transition(
-    color animation.$transition-time-simple animation.$transition-easing-default,
-    background-color animation.$transition-time-simple animation.$transition-easing-default
-  );
-
   &:hover,
   &:focus {
     .list-group-item-right-svg {
```
oliverschuerch, line 21: Would be nice if we could omit the class here: `> post-icon:last-child {...}`

packages/styles/src/components/list-group.scss
```diff
@@ -38,125 +24,87 @@
...
-    background-color: transparent;
-
-    &:hover,
-    &:focus {
-      background-color: list-group.$list-group-hover-bg-inverted;
-    }
-  }
 }
 
 .list-group-item {
```
oliverschuerch, line 36: If possible, omit this class.
```suggestion
> li {
```

packages/styles/src/components/list-group.scss
```diff
@@ -38,125 +24,87 @@
...
+    &::before {
+      content: '';
+      position: absolute;
+      width: 100%;
+      height: 1px;
+      border-block-end: tokens.get('list-item-border-width') tokens.get('list-item-border-style')
+        tokens.get('list-item-enabled-border');
+      bottom: 0;
+      left: 0;
     }
```
oliverschuerch, line 61:
```suggestion
    &::after {
      display: block;
      content: '';
      position: absolute;
      inset: auto 0 0;
      border-block-end: tokens.get('list-item-border-width') tokens.get('list-item-border-style')
        tokens.get('list-item-enabled-border');
    }
```

packages/styles/src/components/list-group.scss
```diff
@@ -38,125 +24,87 @@
...
-
-    &-transparent {
-      padding-right: map.get(spacing.$post-sizes, 'mini');
-      padding-left: map.get(spacing.$post-sizes, 'regular');
+  border: unset;
+  padding: 0;
+  background: transparent;
+
+  > a,
+  .form-check {
```
oliverschuerch, line 42: Would be nice, if we could extract these styles and provide them without the need to be nested in the `.list-group > li`. This would allow us using them also on other places (for example in the <post-list> web-component).

```suggestion
  .list-group-link,
  .list-group-document,
  .list-group-switch {
```

oliverschuerch: Also adopt the styles defined below for these elements.

packages/styles/src/components/list-group.scss
```diff
@@ -38,125 +24,87 @@
...
-    border-top: list-group.$list-group-border-width solid
-      rgba(var(--post-contrast-color-inverted-rgb), 0.2);
+  a,
+  div.form-check {
+    @include utilities-mx.focus-style(
+      null,
+      tokens.get('list-item-focus-outline-offset'),
+      tokens.get('list-item-focus-outline-width'),
+      tokens.get('list-item-focus-outline-color'),
+      tokens.get('list-item-focus-outline-style')
```
oliverschuerch, line 107:
```suggestion
      $offset: tokens.get('list-item-focus-outline-offset'),
      $width: tokens.get('list-item-focus-outline-width'),
      $color: tokens.get('list-item-focus-outline-color'),
      $style: tokens.get('list-item-focus-outline-style')
```

oliverschuerch (changes requested):
@leagrdv Sorry for adding so many change requests. But I think, because of the way we want to use the elements styles here, we need to update the whole thing a little bit.

**Do not hesitate to contact me if you have any questions about one of my comments 😉**

.changeset/dirty-mayflies-taste.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': minor
+'@swisspost/design-system-styles': minor
```
oliverschuerch, line 3:
```suggestion
'@swisspost/design-system-styles': major
```

oliverschuerch (changes requested):

packages/styles/src/components/list-group.scss
leagrdv: I targeted only the post-icon of the list-group-link instead of using :last-child

packages/styles/src/components/list-group.scss
```diff
@@ -1,162 +1,103 @@
...
+        margin-left: auto;
       }
     }
-  }
 
-  &-no-border {
-    border: 0;
-  }
+    .list-group-switch {
+      justify-content: space-between;
```
oliverschuerch, line 73: We need to reset the `margin-bottom` here, which is defined on the `.form-check` class.

```suggestion
      justify-content: space-between;
      margin-bottom: 0;
```

oliverschuerch (changes requested):

packages/styles/src/components/list-group.scss
```diff
@@ -1,162 +1,103 @@
...
-    &-big {
-      padding: map.get(spacing.$post-sizes, 'small-large') map.get(spacing.$post-sizes, 'mini')
-        map.get(spacing.$post-sizes, 'small-large') map.get(spacing.$post-sizes, 'regular');
-    }
-  }
-
-  &-svg {
+  > post-icon {
     flex: 0 0 auto;
+    margin-left: auto;
```
oliverschuerch, line 69:
```suggestion
    
    &:last-child {
      margin-left: auto;
    }
```

oliverschuerch (changes requested):

packages/styles/src/components/list-group.scss
```diff
@@ -1,162 +1,106 @@
...
-    &:hover,
-    &:focus {
-      background-color: list-group.$list-group-hover-bg-inverted;
-    }
-  }
+  @include utilities-mx.focus-style(
+    $offset: tokens.get('list-item-focus-outline-offset'),
+    $width: tokens.get('list-item-focus-outline-width'),
+    $color: tokens.get('list-item-focus-outline-color'),
+    $style: tokens.get('list-item-focus-outline-style')
```
oliverschuerch, line 57: I think you will need to remove these tokens, as they do not exist on `main` anymore.
The general focus tokens are now used by the focus mixins itself.

oliverschuerch (approved):

PR #3751, opened 2024-10-14
chore(styles): updated borders utility

myrta2302:


alizedebray (commented):
@myrta2302 I let you update the PR, just request my review again when it's ready. Thank you!

packages/styles/src/utilities/_variables.scss
```diff
@@ -31,6 +31,144 @@ $position-values: (
...
+    property: border,
+    class: border,
+    values: (
+      null: var(--post-border-width) var(--post-border-style)
+        rgba(var(--post-border-color-base), var(--post-border-opacity)),
+      0: 0,
+    ),
+  ),
+  'border-top': (
+    property: border-top,
```
alizedebray, line 45: Please always use the `block` and `inline` properties instead of the `top`, `bottom`, `right`, and `left` properties:
```suggestion
    property: border-block-start,
```

packages/styles/src/utilities/_variables.scss
```diff
@@ -31,6 +31,144 @@ $position-values: (
...
+  'border-start': (
+    property: border-left,
+    class: border-start,
+    values: (
+      null: var(--post-border-width) var(--post-border-style) var(--post-border-color),
+      0: 0,
+    ),
+  ),
+  'border-width': (
+    responsive: true,
```
alizedebray, line 75: The border utilities probably do not need to be responsive, there is no real use case for this and this adds a lot of classes.
```suggestion
```

packages/styles/src/utilities/_variables.scss
```diff
@@ -31,6 +31,144 @@ $position-values: (
 );
 
 $utilities: (
+  'border': (
+    responsive: true,
+    property: border,
+    class: border,
+    values: (
+      null: var(--post-border-width) var(--post-border-style)
+        rgba(var(--post-border-color-base), var(--post-border-opacity)),
```
alizedebray, line 40: This is good! Please use the same logic for the `border-top`,  `border-bottom`,  `border-start`, and  `border-end` utilities

packages/styles/src/utilities/_variables.scss
```diff
@@ -31,6 +31,144 @@ $position-values: (
...
+      25: 0.25,
+      50: 0.5,
+      75: 0.75,
+      100: 1,
+    ),
+  ),
+  'rounded': (
+    property: border-radius,
+    class: rounded,
+    values: (
```
alizedebray, line 104: This kind of border utilities should get their values from the tokens, using the `from-tokens` function.

packages/styles/src/tokens/_post-theme.scss
```diff
@@ -5,4 +5,27 @@
...
+  --post-border-color-info: 13, 202, 240;
+  --post-border-color-light: 245, 245, 245;
+  --post-border-color-dark: 33, 37, 41;
+  --post-border-color-white: 245, 245, 245;
+  --post-border-radius-sm: 0.25rem;
+  --post-border-radius: 0.375rem;
+  --post-border-radius-lg: 0.5rem;
+  --post-border-radius-xl: 1rem;
+  --post-border-radius-xl: 2rem;
+  --post-border-radius-pill: 50rem;
```
alizedebray, line 30: Here you only have need the variables for the base border utility, i.e. the `.border` class. The other utilities will use values from the utility tokens.
```suggestion
  --post-border-width-base: 1px;
  --post-border-color-base: 204, 204, 204;
  --post-border-opacity-base: 1;
  --post-border-style-base: solid;
```

alizedebray (changes requested):

PR #3752, opened 2024-10-14
feat(Styles): updated float utility

myrta2302:


packages/documentation/src/stories/utilities/float/float.stories.ts
```diff
@@ -0,0 +1,82 @@
...
+    },
+  },
+  args: {
+    floatPosition: '',
+    floatBreakPoint: '',
+  },
+  render: (args: Args) => {
+    const float = args.floatPosition !== '' ? `float` : '';
+    const floatPosition = args.floatPosition !== '' ? `-${args.floatPosition}` : '';
+    const floatBreakPoint = args.floatBreakPoint !== '' ? `-${args.floatBreakPoint}` : '';
```
oliverschuerch, line 49:
```suggestion
    const classMembers = ['float', args.floatBreakPoint ?? 'xs', args.floatPosition].filter(m => m);
    const floatClass = classMembers.length === 3 ? classMembers.filter(m => m !== 'xs').join('-') : nothing;
```

This gathers the possible class parts (`float-{breakpoint}-{position}`) and passes them as a `-` separated string to `floatClass` if `args.floatPosition` is defined. The breakpoint `xs` will always be filtered out and if `args.floatPosition` is undefined, the `floatClass` will be defined as `nothing` (which is the lit way to set a `undefined` or `null` value).

packages/documentation/src/stories/utilities/float/float.stories.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    return html`
+      <div class="float-example">
+        <div class="${float}${floatBreakPoint}${floatPosition}">Sample Text</div>
```
oliverschuerch, line 53:
```suggestion
        <div class="${floatClass}">Sample Text</div>
```
If `floatClass` is `nothing`, the whole `class` attribute will be omitted. Otherwhise, the class will be visible in the code panel and can be in the format of `float-{position}` or `float-{breakpoint}-{position}`.

packages/documentation/src/stories/utilities/float/float.stories.ts
oliverschuerch, line 1: If one or more change requests are unclear or you need further explanations, just call me 😉

packages/documentation/src/stories/utilities/float/float.snapshot.stories.ts
oliverschuerch, line 1: Please visit (for example) [badge.snapshot.stories.ts](https://github.com/swisspost/design-system/blob/main/packages/documentation/src/stories/components/badge/badge.snapshot.stories.ts) to see how we set up the snapshot pages.

With the Snapshot pages we try to create the following:
- Show a component variant for every possible `args` combination.
- Show all these component variants on a light and on a dark background.

The `bombArgs` function helps us to produce the component variants.
For your component a snapshot story could look like this:

```tsx
import type { Args, StoryContext, StoryObj } from '@storybook/web-components';
import meta from './float.stories';
import { html } from 'lit';
import { bombArgs } from '@/utils';

const { id, ...metaWithoutId } = meta;

export default {
  ...metaWithoutId,
  title: 'Snapshots',
};

type Story = StoryObj;

export const FloatUtility: Story = {
    return html`
      <div class="d-flex flex-wrap gap-4 align-items-start">
        ${['bg-white', 'bg-dark'].map(
          bg => html`
            <div class="${bg} p-16">
              ${(meta.argTypes?.floatBreakPoint?.options ?? []).map(breakpoint => {
                return html`
                  <h2>Breakpoint ${breakpoint}</h2>
                  <div>
                    ${bombArgs({
                      floatPosition: meta.argTypes?.floatPosition?.options ?? [],
                    })
                      .map((args: Args) => meta.render?.({ ...context.args, ...args }, context))}
                  </div>
                `;
              })}
            </div>
          `,
        )}
      </div>
    `;
  },
};
```

oliverschuerch (changes requested):

packages/documentation/src/stories/utilities/float/float.stories.ts
myrta2302: Great :)

packages/documentation/src/stories/utilities/float/float.stories.ts
```diff
@@ -0,0 +1,56 @@
...
+    const classMembers = ['float', args.floatBreakPoint ?? 'xs', args.floatPosition].filter(m => m);
+    const floatClass =
+      classMembers.length === 3 ? classMembers.filter(m => m !== 'xs').join('-') : nothing;
+
+    return html`
+      <div class="${floatClass}">Sample Text</div>
+      <div class="clearfix"></div>
+    `;
+  },
+};
```
oliverschuerch, line 56: Because you have only one story, you could define all of the used properties (`args`, `argTypes`, `render`) also directly in the meta object.
The difference is, if you define something in the `meta` object, it will be used (but also can be overriden) by all the stories you create here.

In case you're going to move those props to `main` (I'll leave the decision to you), you can create the story with everything on it by only writing `export const Float: Story = {};`. Also you would only need to import the `meta` object into your snapshot stories file and use the `argTypes` options and also the `render` function from there.

packages/documentation/src/stories/utilities/float/float.stories.ts
```diff
@@ -0,0 +1,56 @@
...
+    floatBreakPoint: null,
+  },
+  render: (args: Args) => {
+    const classMembers = ['float', args.floatBreakPoint ?? 'xs', args.floatPosition].filter(m => m);
+    const floatClass =
+      classMembers.length === 3 ? classMembers.filter(m => m !== 'xs').join('-') : nothing;
+
+    return html`
+      <div class="${floatClass}">Sample Text</div>
+      <div class="clearfix"></div>
```
oliverschuerch, line 53: 1. The element `div.clearfix` should be a wrapper around the floated element(s), because it dynamicaly adds an `::after` pseudo element, which then makes sure, that the wrapper element itself gets sized correctly (because floating elements do not communicate there height to the parent element, because they are rendered different than other elements).
2. The same element should not be a visible part of the story. Please add it in a storybook decorator.
https://storybook.js.org/docs/writing-stories/decorators

packages/documentation/src/stories/utilities/float/float.stories.ts
```diff
@@ -0,0 +1,56 @@
...
+    floatPosition: null,
+    floatBreakPoint: null,
+  },
+  render: (args: Args) => {
+    const classMembers = ['float', args.floatBreakPoint ?? 'xs', args.floatPosition].filter(m => m);
+    const floatClass =
+      classMembers.length === 3 ? classMembers.filter(m => m !== 'xs').join('-') : nothing;
+
+    return html`
+      <div class="${floatClass}">Sample Text</div>
```
oliverschuerch, line 52:
```suggestion
      <div class="${floatClass}">Sample Text (${args.floatPosition})</div>
```
This would be helpful for the snapshots.

oliverschuerch (changes requested):
Good work so far 😄 Just some more thoughts.

packages/components/src/components/post-list-item/readme.md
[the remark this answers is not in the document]
myrta2302: omg! I dont have a clue how this happened... :(

packages/documentation/src/stories/utilities/float/float.stories.ts
myrta2302: Thanx i ll do so! Alize also told me the same thing this morning for another PR :)

packages/documentation/src/stories/utilities/float/float.docs.mdx
```diff
@@ -0,0 +1,36 @@
...
+<div className="alert alert-info">
+  The `float-{breakpoint}-{position}` classes provide control over floating elements at specific
+  breakpoints, enabling you to float elements or remove floating based on the screen size.
+</div>
+
+> - `float-sm-{start|end|none}`
+> - `float-md-{start|end|none}`
+> - `float-lg-{start|end|none}`
+> - `float-xl-{start|end|none}`
+> - `float-xxl-{start|end|none}`
```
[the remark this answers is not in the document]
myrta2302: If i understand correctly, for the moment we just integrate the bootstrap breakpoints to the ds as they are. I think the classes must work now.

packages/documentation/src/stories/utilities/float/float.stories.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    return html`
+      <div class="${floatClass}">
+        Sample Text ${args.floatPosition ? '(float-' + args.floatPosition + ')' : ''}
+      </div>
+    `;
+  },
+  decorators: [
+    (story: StoryFn, context: StoryContext) => {
+      const storyTemplate = html`<div class="clearfix float-example">
```
oliverschuerch, line 52: Please remove this obsolete class
```suggestion
      const storyTemplate = html`<div class="clearfix">
```

packages/documentation/src/stories/utilities/float/float.docs.mdx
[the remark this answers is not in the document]
oliverschuerch: Was my fault, probably the viewport was just to small when I checked it.

packages/documentation/src/stories/utilities/float/float.stories.ts
```diff
@@ -0,0 +1,64 @@
...
+    floatBreakPoint: null,
+  },
+  render: (args: Args) => {
+    const classMembers = ['float', args.floatBreakPoint ?? 'xs', args.floatPosition].filter(m => m);
+    const floatClass =
+      classMembers.length === 3 ? classMembers.filter(m => m !== 'xs').join('-') : nothing;
+
+    return html`
+      <div class="${floatClass}">
+        Sample Text ${args.floatPosition ? '(float-' + args.floatPosition + ')' : ''}
```
oliverschuerch, line 46: Sorry for adding a CR for this again, but it would be helpful to also know on which breakpoint the content should be floated.
```suggestion
        Content ${floatClass ? '(' + floatClass + ')' : ''}
```

packages/documentation/cypress/snapshots/utilities/float.snapshot.ts
```diff
@@ -0,0 +1,7 @@
+describe('Float', () => {
+  it('float', () => {
+    cy.visit('/iframe.html?id=snapshots--float');
+    cy.get('.float-example', { timeout: 30000 }).should('be.visible');
+    cy.percySnapshot('Float', { widths: [320, 1440] });
```
oliverschuerch, line 5:
```suggestion
    cy.percySnapshot('Float', { widths: [320, 400, 600, 780, 1024, 1280, 1440] });
```

oliverschuerch (changes requested):

packages/documentation/src/stories/utilities/float/float.snapshot.stories.ts
```diff
@@ -0,0 +1,46 @@
...
+};
+
+type Story = StoryObj;
+
+export const Float: Story = {
+  render: (_args: Args, context: StoryContext) => {
+    return html`
+      <div class="snapshot-example">
+        <h1>Snapshot with Float</h1>
+        ${['bg-white', 'bg-dark'].map(
```
oliverschuerch, line 20: Testing a floating utility on different colored backgrounds, does not really bring any advantage. But since I told you to do so, I'm ok if you keep it this way 😉

oliverschuerch (approved):

PR #3762, opened 2024-10-15
chore(intranet-header): fix dropdown nav-link documentation

alizedebray:


oliverschuerch (approved):
@alizedebray Is this something we need to cherry pick for v8 also?
If so, please create a ticket!

PR #3785, opened 2024-10-17
fix(components): switch stencil hydrated flag from `class` to `attribute`

oliverschuerch:
This changes how the hydrated flag is add to a web-component.
The class `hydrated` is no longer important/used and therefore also not causing the component to show up, when added statically or just before the component has been hydrated.

Developers which copied the components dom strucure from the code panel of our docs pages (for example for the components `tooltip` and `placeholder`) should remove the class `hydrated` out of there HTML, even though it's not mandatory, because the class will just not have any impact anymore.
On the other hand, if the class was used in some e2e tests of the project, it may be necessary to update those.

### DIFF
--- a/packages/components/stencil.config.ts
+++ b/packages/components/stencil.config.ts
@@ -10,6 +10,10 @@ export const config: Config = {
   buildDist: true,
   sourceMap: false,
   validatePrimaryPackageOutputTarget: true,
+  hydratedFlag: {
+    name: 'data-hydrated',
+    selector: 'attribute',
+  },
   outputTargets: [
     {
       type: 'dist',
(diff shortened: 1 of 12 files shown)


packages/components/stencil.config.ts
```diff
@@ -10,6 +10,10 @@ export const config: Config = {
   buildDist: true,
   sourceMap: false,
   validatePrimaryPackageOutputTarget: true,
+  hydratedFlag: {
+    name: 'data-hydrated',
```
gfellerph, line 14: We could think about using our prefix here as well to make it clear where that comes from. Browsers are following the convention that dashed attribute names are user-land (except for aria-*) so it should be safe to use.

```suggestion
    name: 'post-hydrated',
```

gfellerph (approved):

packages/components/stencil.config.ts
oliverschuerch: If we do this, we should align it on the main also.
Do you want me to do so?

oliverschuerch: Won't do, because the flag is not set by us but by the stencil logic implemented in the components.

PR #3802, opened 2024-10-21
feat(components): create post-language-option

alizedebray:


packages/components/src/components/post-language-option/post-language-option.tsx
```diff
@@ -0,0 +1,147 @@
...
+      value,
+      'boolean',
+      'The "active" property of the post-language-option component must be a boolean value.',
+    );
+  }
+
+  /**
+   * The full name of the language. For example, "Deutsch".
+   */
+  @Prop() name: string;
```
gfellerph, line 57: The name is a required prop, even if not active. It will be read when choosing options, thanks to the language code in the appropriate language.

packages/components/src/components/post-language-option/post-language-option.tsx
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  render() {
+    const lang = this.code.toLowerCase();
+
+    return (
+      <Host data-version={version} role="listitem">
+        {this.url ? (
+          <a
+            aria-current="page"
```
gfellerph, line 125: This should only be there if this option actually is the currently selected option.

packages/components/src/components/post-language-option/post-language-option.tsx
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * Selects the language option programmatically.
+   */
+  @Method()
+  async select() {
+    this.active = true;
+  }
+
+  private handleClick() {
+    this.active = true;
```
gfellerph, line 110: I don't think we should set active here. We don't know if the page actually changed language or something went wrong. It should be up to the project to update after the switch.

packages/components/src/components/post-language-option/post-language-option.tsx
```diff
@@ -0,0 +1,147 @@
...
+   * An event emitted when the language option is clicked. The payload is the ISO 639 code of the language.
+   */
+  @Event() postClick: EventEmitter<string>;
+
+  /**
+   * Selects the language option programmatically.
+   */
+  @Method()
+  async select() {
+    this.active = true;
```
gfellerph, line 106: It could be good to emit the language change event here.

packages/components/src/components/post-language-option/post-language-option.tsx
```diff
@@ -0,0 +1,147 @@
...
+      throw new Error(
+        'The "name" property of the post-language-option component is required when the full language name is not displayed.',
+      );
+    }
+  }
+
+  /**
+   * An event emitted when the language option is clicked. The payload is the ISO 639 code of the language.
+   */
+  @Event() postClick: EventEmitter<string>;
```
gfellerph, line 99: Should we rename this to postLanguageChange? This would align with the current event.

gfellerph (changes requested):

packages/components/src/components/post-language-option/post-language-option.tsx
alizedebray: I renamed it to `postChange` to be even more aligned.

packages/components/src/index.html
gfellerph, line 1: Please reset changes to this file

packages/documentation/src/stories/components/header/language-option/language-option.stories.ts
```diff
@@ -0,0 +1,67 @@
...
+// RENDERERS
+function renderLanguageOption(args: Partial<HTMLPostLanguageOptionElement>) {
+  return html`<post-language-option
+    code=${args.code}
+    name=${args.name ? args.name : nothing}
+    active=${args.active ? args.active : nothing}
+    url=${args.url ? args.url : nothing}
+    onClick=${args.url ? nothing : 'e => changeLangTo(e.detail)'}
+  >
+    EN
```
gfellerph, line 54: This is statically set, it should take the value from the args.

```suggestion
    { args.innerHtml }
```

gfellerph (approved):

PR #3814, opened 2024-10-24
feat(styles): set initial component visibility to hidden to prevent flickering

myrta2302:


packages/styles/src/utilities/_not-defined.scss
```diff
@@ -0,0 +1,25 @@
...
+post-rating,
+post-tab-header,
+post-tab-panel,
+post-tabs,
+post-tag,
+post-tooltip {
+  &:not(:defined) {
+    visibility: hidden;
+  }
+}
```
gfellerph, line 25: This will create a way smaller selector without the repeated `:not(:defined)` statement.

```suggestion
:where(
  post-list,
  post-list-item,
  post-accordion,
  post-accordion-item,
  post-alert,
  post-card-control,
  post-collapsible,
  post-icon,
  post-logo,
  post-popover,
  post-popovercontainer,
  post-rating,
  post-tab-header,
  post-tab-panel,
  post-tabs,
  post-tag,
  post-tooltip
):not(:defined) {
  visibility: hidden;
}
```

gfellerph (changes requested):

gfellerph (changes requested):
Document this workflow in the contribution guidelines for the components package.

gfellerph (changes requested):
@myrta2302 can you also add this file to the utilities index file so it's automatically loaded?

myrta2302:
I see the styles being applied in Storybook examples, but not at the Component playground.

gfellerph:
> I see the styles being applied in Storybook examples, but not at the Component playground.

@myrta2302 the component playground probably does not load all the styles necessary. You can check the index.html in the components package src folder to see what's exactly being loaded.

gfellerph (approved):

PR #3820, opened 2024-10-24
chore(styles): added focus styles for chips

schaertim:


packages/styles/src/components/chip.scss
alizedebray, line 1: Please remove the focus styles on [line 115](https://github.com/swisspost/design-system/blob/42a27ef859c94060422c529b1ae26fa8aa8a273e/packages/styles/src/components/chip.scss#L115) as it duplicates the focus you just added.

PR #3835, opened 2024-10-28
feat(documentation): accessibility

myrta2302:


packages/documentation/src/stories/foundation/accessibility/forms/forms-reference-labels.sample.html
```diff
@@ -0,0 +1,15 @@
+<form>
+    <div role="group" aria-labelledby="complex-label">
+      <span id="complex-label">Select your favorite fruits:</span>
```
alizedebray, line 3: Accessibility requires all texts to be inside `<p>` tags: 
```suggestion
      <p id="complex-label">Select your favorite fruits:</p>
```

packages/documentation/src/stories/foundation/accessibility/regulation/regulation.stories.ts
```diff
@@ -2,7 +2,7 @@ import { StoryObj } from '@storybook/web-components';
 import { MetaExtended } from '@root/types';
 
 const meta: MetaExtended = {
-  id: '759a427c-7884-474e-b6b7-de32d3acf5df',
+  id: 'ea0e885a-ab89-492a-8905-afbed7ad9548',
```
alizedebray, line 5: Why did you change the id here? Most of the time ids shouldn't be changes as it also changes the page URL.

alizedebray (changes requested):

packages/documentation/src/stories/foundation/accessibility/regulation/regulation.stories.ts
myrta2302: i got an error about duplicate ids, but it was coming from a different file after all.

alizedebray (approved):

PR #3836, opened 2024-10-28
feat(styles): enabled media mixins with string params

gfellerph:
This allows to call the media mixins with keys from the breakpoint list directly, e.g. `@include media.max(lg) { }`.


oliverschuerch:
Easy fix:
image

oliverschuerch (approved):
Isn't this a feature we should have a changeset for?
At least for the users using our scss files.

oliverschuerch (changes requested):
Fix the lint issues and add a changset (if necessary)

gfellerph:
> Isn't this a feature we should have a changeset for? At least for the users using our scss files.

This was not documented and not available in v9.

oliverschuerch (approved):

PR #3845, opened 2024-10-28
feat(components): add edge gap property to popovercontainer

gfellerph:
This feature allows to specify how much space the floating-ui library should account for when positioning overlays near the edge of the page. Set to 0 if popovers should be allowed to touch the edge of the window.

No changeset because this component is not documented.

This is needed for the megadropdown in the header component.

### DIFF
--- a/packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
+++ b/packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
@@ -60,6 +60,11 @@ export class PostPopovercontainer {
   @Prop() readonly placement?: Placement = 'top';
 
   /**
+   * Gap between the edge of the page and the popover
+   */
+  @Prop() readonly edgeGap?: number = 8;
+
+  /**
    * Wheter or not to display a little pointer arrow
    */
   @Prop() readonly arrow?: boolean = false;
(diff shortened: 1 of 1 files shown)


packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -59,6 +59,11 @@ export class PostPopovercontainer {
    */
   @Prop() readonly placement?: Placement = 'top';
 
+  /**
+   * Gap between the edge of the page and the popover
+   */
+  @Prop() readonly edgeGap?: number = 8;
```
oliverschuerch, line 65: Should we really set the default value to something else than `0`?
It will possibly change the default behaviour for some projects.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -159,15 +165,15 @@ export class PostPopovercontainer {
       size({
         apply({ availableWidth, elements }) {
           Object.assign(elements.floating.style, {
-            maxWidth: `${availableWidth - 16}px`,
+            maxWidth: `${availableWidth - gap * 2}px`,
           });
         },
       }),
-      offset(this.arrow ? 12 : 8), // 4px outside of element to account for focus outline + ~arrow size
+      offset(this.arrow ? gap + 4 : gap), // 4px outside of element to account for focus outline + ~arrow size
```
oliverschuerch, line 172: Sad this `4` needs to be hardcoded here 😢

oliverschuerch (approved):

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
gfellerph: This is the value that was used before, so anything else would mean that it changes for projects that don't specify. It was hardcoded before.

PR #3849, opened 2024-10-29
feat(components): button

leagrdv:


packages/styles/src/components/button.scss
```diff
@@ -95,84 +79,70 @@
...
-  }
 
-  @include color-mx.on-dark-background() {
-    .pi {
-      filter: invert(0);
+    @include utilities.not-disabled-hover() {
+      @include button-mx.button-variant-def('hover', $color);
+    }
+
+    @include color-mx.on-dark-background {
```
gfellerph, line 101: This old mixin should not really be used anymore, at least not where we have tokens. For example for the border, we just need to define

```CSS
button {
  border-color: tokens.get('button-primary-enabled-stroke');
}
```
 and the dark mode tokens will take care of mapping to the correct color in each mode.

packages/styles/src/components/button.scss
```diff
@@ -95,84 +79,70 @@
...
-    color: var(--post-contrast-color);
-
-    &:disabled {
-      color: color.$white-alpha-60;
-      border-color: color.$white-alpha-60;
+    .pi {
+      filter: invert(0);
     }
   }
 }
```
gfellerph, line 127: This is the older method of inverting icons. As these icons are no longer available, we also don't need to support inverting them. This can safely be removed.

packages/styles/src/mixins/_button.scss
```diff
@@ -34,14 +35,11 @@
...
-  $button-height: map.get(button.$btn-height-map, $size);
-  $padding-y: ($button-height - $icon-size) * 0.5 - button.$input-btn-focus-width;
-  min-width: $button-height;
-  padding-right: $padding-y;
-  padding-left: $padding-y;
+@mixin button-variant-def($type, $color, $inverted: false) {
+  $inverted-suffix: if($inverted, '-inverted', '');
+  border-color: tokens.get('button-#{$color}#{$inverted-suffix}-#{$type}-stroke');
+  background-color: tokens.get('button-#{$color}-#{$type}-bg');
+  color: tokens.get('button-#{$color}-#{$type}-fg');
```
gfellerph, line 42: The inverted tokens should not be used (Alessio will remove them). The schemes token layer will handle the correct mapping in light and dark mode.

gfellerph (changes requested):
image
Buttons... buttons everywhere

.changeset/fifty-students-call.md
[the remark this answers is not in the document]
leagrdv: Done! I've also added that we removed `.btn-animated`

packages/styles/src/variables/components/_button.scss
gfellerph, line 1: This already cleans up a lot of variables. Nice work! I've created a follow up ticket for having a closer look at the variables in here because some of them are cross-linked (https://github.com/swisspost/design-system/issues/3870). This can be tackled in a later milestone.

.changeset/fifty-students-call.md
gfellerph, line 1: It would be nice to have multiple, separate changesets:

styles: minor -> Updated button component to v2
styles: major -> Removed regular button size
styles: major -> Removed the `btn-animated` class. It no longer has any effect and can be removed from existing buttons.

I think this makes the changelog a little more readable since every change is addressed in it's own bullet point.

gfellerph (approved):
Thanks for prioritizing this!

PR #3859, opened 2024-10-31
feat: add styles and docs for ul elements

oliverschuerch:


packages/components/src/components.d.ts
gfellerph, line 1: Can you remove this file from the PR? It's being updated in a dedicated PR.

packages/components/src/components/post-popovercontainer/readme.md
gfellerph, line 1: Same here, does not belong in this one.

packages/documentation/src/stories/foundation/list/list-reset.sample.scss
```diff
@@ -0,0 +1,5 @@
+@use '@swisspost/design-system-styles/core' as post;
+
+ul.my-list {
+  @include post.list-reset();
+}
```
gfellerph, line 5: Including the whole styles library is not necessary for using the reset and the mixin name is `reset-list`.

```suggestion
@use '@swisspost/design-system-styles/mixins/utilities' as post;

ul.my-list {
  @include post.reset-list();
}
```

packages/documentation/src/stories/foundation/list/list.docs.mdx
```diff
@@ -24,6 +25,16 @@ This only applies to immediate child list items, meaning you will need to add th
 
 <Canvas of={ListStories.UnstyledList} sourceState="shown" />
 
+#### Sass mixin
+
+If you wan't to use the class above, you can still reset the list styles with the following mixin:
```
gfellerph, line 30:
```suggestion
If you want to use the class above, you can still reset the list styles with the following mixin:
```

packages/styles/src/elements/bullet-list.scss
```diff
@@ -0,0 +1,58 @@
...
+    #{tokens.get('list-bullet-item-icon-size')} + #{tokens.get('list-bullet-item-icon-gap-inline')}
+  );
+
+  > li {
+    &::marker {
+      line-height: 1;
+    }
+  }
+
+  &.bullet {
```
gfellerph, line 27: Can we prefix the class name with the component name: `.list-bullet`? Could imagine naming conflicts with this class name.

(Would obviously have to be updated throughout the PR).

```suggestion
  &.list-bullet {
```

packages/styles/src/elements/_index.scss
```diff
@@ -2,6 +2,7 @@
 
 @use 'body';
 @use 'anchor';
+@use 'bullet-list';
```
gfellerph, line 5: Could be named `list-bullet` as well just to place it with all the other list types.

```suggestion
@use 'list-bullet';
```

gfellerph (changes requested):
Solid implementation!

packages/documentation/src/stories/foundation/list/list-reset.sample.scss
oliverschuerch: Fixed the mixin name.

The core file only contains `variables`, `mixins` and `functions`. This means, that nothing is added to the CSS output, except what is actually used. With this in mind, it does not make sense, to add more specific import statements imo.
What do you think?

gfellerph (approved):

PR #3882, opened 2024-11-05
feat(styles): create blockquote component

myrta2302:


packages/styles/src/components/blockquote.scss
```diff
@@ -0,0 +1,34 @@
...
+
+  .blockquote {
+    margin: 0;
+    padding: 0;
+    border: none;
+  }
+}
+
+.blockquote-footer {
+  @extend %body;
```
alizedebray, line 32: This is not necessary since the figcaption inherits its font styles from the body anyways:
```suggestion
```

packages/styles/src/placeholders/_blockquote.scss
alizedebray, line 1: Now that that blockquote has its own tokens, the placeholders are no longer needed.

packages/styles/src/components/blockquote.scss
```diff
@@ -0,0 +1,34 @@
+@use '../tokens/components';
+@use '../functions/tokens';
+@use './../placeholders/blockquote' as blockquote-ph;
+
+tokens.$default-map: components.$post-quote;
+
+.blockquote {
```
alizedebray, line 7: I am just wondering if the blockquote is a component or an element since the `.blockquote` class should only be used on `<blockquote>` elements.... 
I am adding this as a topic for the devs roundtable.

alizedebray (changes requested):

packages/styles/src/components/blockquote.scss
```diff
@@ -0,0 +1,38 @@
...
+    border-style: none;
+    border-color: transparent;
+  }
+}
+
+figcaption.blockquote-footer {
+  margin-top: tokens.get('post-blockquote-gap-inline');
+
+  cite {
+    font-style: normal;
```
alizedebray, line 36:
```suggestion
    font-style: inherit;
```

alizedebray (changes requested):

.changeset/twenty-items-drum.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': minor
+'@swisspost/design-system-styles': minor
+---
+
+implement blockquote component
```
alizedebray, line 6: The blockquote is not a new component, you updated the existing one therefore this is a patch.

```suggestion
---
'@swisspost/design-system-documentation': patch
'@swisspost/design-system-styles': patch
---

Updated the blockquote styles to match the new Post design.
```

alizedebray (commented):
In the previous version, the quotes character was different depending on the language: https://design-system.post.ch/?path=/docs/a4c7d5fa-9f8d-4a95-820b-ae4ce9ab4f03--docs.

This is not the case anymore. Is that intentional?

myrta2302:
> I checked out your branch and tested the blockquotes locally on Firfox and I still cannot see the quote character changing depending on the language.

Hey! So i notice the following:

1. Localhost Styles:Playground -> Language dependent quotes work in Edge/Chrome and Firefox. 
2. Localhost Documentation -> Language dependent quotes work only in Edge/Chrome
3. Branch Preview -> Language dependent quotes work only in Edge/Chrome.
4. (v8.5) Published Documentation -> Language dependent quotes work partially in Edge/Chrome and Firefox. They change only from " to « (missing 'de' version)

alizedebray (commented):
@myrta2302 This styles are missing, this is why the quotes aren't correct anymore: https://github.com/swisspost/design-system/blob/release/v8/packages/styles/src/components/type.scss#L103

alizedebray (approved):

PR #3902, opened 2024-11-06
chore(docs): refactored `div` element to `p` for hint text in input, textarea, and select

alionazherdetska:


oliverschuerch (commented):
Please add a changeset, because this is relevant for our users.

PR #3903, opened 2024-11-07
chore(docs): refactored `div` to `p` for hint text in input, text area and select

alionazherdetska:


oliverschuerch:
I just requested the preview page to be recategorized on https://sitereview.bluecoat.com/, so we're able to see how it looks on v7.

PR #3927, opened 2024-11-08
fix(styles): fix nested rules warning in scss files for reset-list utility

leagrdv:
…ility


gfellerph (approved):

gfellerph (changes requested):
@leagrdv can you fix the code smell? There seems to be a duplicate class style block for `.stepper` (according to sonar).

gfellerph (approved):

PR #3850, opened 2024-10-29
feat(styles): add palettes

alizedebray:
To ease the review, the changes are grouped into 3 commits: 

- feat(styles): add palette styles
- feat(docs): add palette stories
- chore: add palette tests

PR #3945, opened 2024-11-08
feat: add default and overscroll backgrounds

gfellerph:


packages/styles/src/elements/body.scss
```diff
@@ -1,17 +1,28 @@
...
+}
+
 body {
   font-family: tokens.get('body-font-family');
   font-size: tokens.get('body-font-size');
   font-weight: tokens.get('body-font-weight');
   line-height: tokens.get('body-line-height');
   letter-spacing: tokens.get('body-letter-spacing');
   color: tokens.get('body-color');
+  background-color: tokens.get('palettes-color-default-bg', palettes.$post-palettes);
```
alizedebray, line 27: The background color of the body is already set in the palettes PR: https://github.com/swisspost/design-system/pull/3850/files#diff-8c14f2102052982c3be175ed844b74d595f43df77e4c22707011eecdd5b4ad2dR5

```suggestion
```

alizedebray: In the end, I think I had you create this PR for nothing... 🙈 I thought you would need the palettes to fetch the Cargo green for the HTML element.

packages/styles/src/elements/body.scss
```diff
@@ -1,12 +1,22 @@
 @use 'sass:map';
+@use '../variables/color';
 @use '../functions/tokens';
 @use '../tokens/elements';
+@use '../tokens/palettes';
```
gfellerph, line 5:
```suggestion
```

packages/styles/src/elements/body.scss
gfellerph: There's still the background of the `<html>` element that changes, so not entirely for nothing.

alizedebray (approved):

gfellerph:
@alizedebray retargeting to the palettes PR created a bunch of issues. Maybe better to just go to main?

PR #3947, opened 2024-11-08
chore(styles): remove CSS icons

alizedebray:


packages/styles/src/functions/_icons.scss
```diff
@@ -61,9 +61,9 @@
 
   @if (map.has-key(icons.$svg-icon-map, #{$icon-name})) {
     $svg-url: map.get(icons.$svg-icon-map, #{$icon-name});
-  } @else if(map.has-key(icons.$svg-unofficial-icon-map, #{$icon-name})) {
+  } @else if(map.has-key(icons.$next-icons, #{$icon-name})) {
```
gfellerph, line 64: What are next-icons? Seems like a name that will age badly (what will the next next-icons be called? Next-next-icons?).

gfellerph (approved):
Awesome! That's a huge change for a bright future.

gfellerph (changes requested):
I'll approve as soon as unit and e2e tests are green ^.-

packages/styles/src/functions/_icons.scss
alizedebray: These are PPNL icons, I suppose they will come from the icons package in a near future but I can rename the map. What name would you choose?

gfellerph: @alizedebray sync with @oliverschuerch, I think he called the set UI icons.

oliverschuerch: Rouven defined them as the „UI“ Icons, because they are exclusively made for the UI.
Since each of these icons lifes in a seperate file, contains up to 6 size variants, which are nested in `<defs>` and `<symbol>` elements and used through multiple `<use>` tags combined with dark container query magic defined in the `<post-icon>` component, I‘d not expect it’s even possible to implement them correctly in SCSS.

@alizedebray If you really want to see how far you can get, I can send you some of the output files already. Or I can open a draft PR (with out the `<post-icon>` being ready), so you can export them by your self or start the icons preview page, where you’ll see how they work.

alizedebray: @oliverschuerch we are not talking about the same icon set. What I named `$next-icons` are not the responsive icons but the 3 notification icons used for the PPNL form validation: 
image

I'll rename the map to `$notification-icons`.

gfellerph (approved):

PR #3954, opened 2024-11-11
chore: update tokens workflow

alizedebray:
Updated the tokens PR workflow to:
- have only one merging branch to prevent overriding changes from the current month with changes from the past month,
- stop merging the main branch into the token branch, merge it into the merging branch instead


.github/workflows/create-token-pr.yaml
```diff
@@ -60,6 +52,7 @@ jobs:
         if: steps.pr-branch.outputs.exists == 'true'
         run: |
           git checkout ${{ steps.pr-branch.outputs.name }}
+          git merge origin/main -X ours --no-edit
           git merge ${{ github.ref_name }} -X theirs --no-edit
```
alizedebray, line 56: @oliverschuerch I updated the workflow to work as follow: 
1. Check if the merging branch exist:
    - If **it does not exists** ⇨ create the merging branch from the tokens branch
    - if **it exists** ⇨ update the merging branch with both the main and the token branches (in case of conflicts accept changes from the tokens branch because it "is always king", thus the `-X theirs`)
2. Check if the PR exists:
    - If **it does not exists** ⇨ create it
    - If **it exists** ⇨ we are good to go

oliverschuerch (approved):

PR #3965, opened 2024-11-12
feat(styles): implement v2 styles for checkbox, radio-button & switch using tokens

schaertim:
Checkbox: https://preview-3965--swisspost-design-system-next.netlify.app/?path=/docs/e6ecc86f-d148-413b-b796-614a89da54be--docs
Radio button: https://preview-3965--swisspost-design-system-next.netlify.app/?path=/docs/151242aa-a074-4a55-a81c-db597c83cdad--docs


schaertim:
E2E test for the card-control currently fail. The issue is that some tests use a fixture file which needs to load the styles file and the fonts. The fonts however are not able to be loaded due to proxy problems. After discussing this with philipp we decided to postpone finding a solution for this so the PR can be reviewed and merged.

Some of the spacing styles of the radio-button and the checkbox within fieldsets is not looking like the design right now. This is due to some styling having to be done in the fieldset styles instead of the checkbox/radio styles. A ticket to do this has been created.

packages/styles/src/elements/reset.scss
```diff
@@ -145,3 +145,9 @@ legend {
 fieldset {
   border: 0;
 }
+
+input,
+select,
+textarea {
+  margin: 0;
+}
```
gfellerph, line 153: Can you move this to the component styles? We probably don't want to remove margins on elements that we don't style.

packages/styles/src/components/_form-check.scss
```diff
@@ -0,0 +1,76 @@
...
+  > label {
+    transition: color #{animation.$transition-base-timing};
+
+    @include utility-mx.high-contrast-mode {
+      transition: none;
+    }
+  }
+}
+
+.form-check-input {
```
gfellerph, line 40: We're telling everybody that this class can safely be removed. It should not be used here.

gfellerph (changes requested):

packages/styles/src/elements/reset.scss
schaertim: Previously the margin reset was part of the bootstrap reboot file. Because of that it can be implemented in our reset file instead of having it on a component level.

gfellerph (changes requested):
Unfortunately, this change breaks storybook controls. I think not all selectors are scoped to `.form-check`. 

image

packages/styles/src/components/checkbox.scss
```diff
@@ -0,0 +1,120 @@
...
+
+$checkbox-icon-checked: url("data:image/svg+xml,<svg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'><path d='M7.51712 12.1332L4.30712 8.69325C4.02712 8.39325 4.03712 7.91325 4.34712 7.63325C4.64712 7.35325 5.12712 7.36325 5.40712 7.67325L7.31712 9.72325L10.7171 4.57325C10.9471 4.22325 11.4071 4.13325 11.7571 4.36325C12.1071 4.59325 12.1971 5.05325 11.9671 5.40325L7.51712 12.1332Z'/></svg>");
+$checkbox-icon-indeterminate: url("data:image/svg+xml,<svg width='16' height='16' viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'><path d='M11.21 8.5H4.75C4.34 8.5 4 8.16 4 7.75C4 7.34 4.34 7 4.75 7H11.21C11.62 7 11.96 7.34 11.96 7.75C11.96 8.16 11.62 8.5 11.21 8.5Z'/></svg>");
+
+:where(input[type='checkbox']) {
+  @include utility-mx.focus-style();
+}
+
+input[type='checkbox'] {
+  appearance: none;
```
gfellerph, line 20: Checkbox styles are leaking into other parts of the document, that is not good. If we want a default look for checkboxes and radio buttons, these files need to go to the elements folder and concern themselves with the very basics and the checkbox component can build up from there. This would need to go in another ticket, not for PPNL Components.

image

```suggestion
:where(.form-check > input[type='checkbox']) {
  @include utility-mx.focus-style();
}

.form-check > input[type='checkbox'] {
  appearance: none;
```

packages/styles/src/components/radio-button.scss
```diff
@@ -0,0 +1,117 @@
...
+@use '../tokens/components';
+@use '../functions/tokens';
+
+tokens.$default-map: components.$post-radio-button;
+
+:where(input[type='radio']) {
+  @include utility-mx.focus-style();
+}
+
+input[type='radio'] {
```
gfellerph, line 16:
```suggestion
:where(.form-check > input[type='radio']) {
  @include utility-mx.focus-style();
}

.form-check > input[type='radio'] {
```

gfellerph (changes requested):

schaertim:
The switch alignment is not correct at the moment. A follow up ticket was created to fix this: #4039

gfellerph (approved):
party-time

Good job @schaertim!

PR #3976, opened 2024-11-12
chore(styles): remove rg and xxl breakpoints

alizedebray:
Removing the `rg` and `xxl` breakpoints to prepare for the tokenized grid system which has only 5 breakpoints.


gfellerph (approved):
@alizedebray can you create a ticket to document this breaking change with a changeset and migration documentation?

PR #3978, opened 2024-11-13
feat(component): form select

leagrdv:


.changeset/loud-peaches-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': major
+---
+
+Removed size variants for form select.
```
gfellerph, line 5:
```suggestion
Removed size variants for form select. The sizing classes `.form-select-sm`, `.form-select-rg`, `.form-select-md` and `.form-select-lg` no longer have any effect and can be removed safely.
```

packages/styles/src/components/form-select.scss
gfellerph, line 1: image

The padding-inline-end should be big enough to account for validation icons as well. This prevents long option text from flowing behind the icons.

gfellerph (changes requested):

PR #3982, opened 2024-11-13
chore(styles): tokenize containers

alizedebray:


packages/styles/src/components/grid.scss
```diff
@@ -4,53 +4,16 @@
 
 @use './../themes/bootstrap/core' as *;
 @use './../themes/bootstrap/grid' as bg;
```
gfellerph, line 6: Will these be removed in a separate PR?

gfellerph (changes requested):
I found 169 files on bitbucket and 33 files on GitHub using "col-rg-" on internal projects. Can you add a major changeset describing the change and how to map it to the new class name?

gfellerph (approved):

PR #3984, opened 2024-11-13
fix: removed duplicate class definition

jarosmpost:


schaertim:
This issue has been fixed on main. I created two PR's to fix it on v7 and v8.

Issue #3585, opened 2024-09-25
Check if we can sign our packages using the `--provenance` flag


gfellerph:
What kind of setup would be necessary to use this feature? Gather requirements or suggest an implementation according to https://github.blog/security/supply-chain-security/introducing-npm-package-provenance/.

PR #3986, opened 2024-11-13
chore: publish package with provenance

Framoroso:
fix: #3585


gfellerph:
@Framoroso we're waiting with this until the end of the frozen zone.

oliverschuerch (commented):
According to the [npm docs](https://docs.npmjs.com/generating-provenance-statements#using-third-party-package-publishing-tools), this seems to be the easiest way to publish all of our packages with provenance.

@Framoroso Thanks for the PR!
@gfellerph Let's try it out!

PR #3988, opened 2024-11-14
chore(documentation,styles): add position utilities

leagrdv:


oliverschuerch (commented):
Please add a changeset

packages/documentation/public/images/browser-bg-top.png
oliverschuerch, line 1: It would be nice, if you could remove `google.com` and instead use `post.ch` or keep the address bar empty.

packages/documentation/src/stories/utilities/position/position.snapshot.stories.ts
```diff
@@ -0,0 +1,62 @@
...
+            ${bombArgs({
+              x: ['start-0', 'start-50', 'start-100', 'end-0', 'end-50', 'end-100'],
+              y: ['top-0', 'top-50', 'top-100', 'bottom-0', 'bottom-50', 'bottom-100'],
+            }).map(
+              args =>
+                html`
+                  <p class="mt-12">
+                    ${(args.x as string).split('-')[0]}: ${(args.x as string).split('-')[1]}% /
+                    ${(args.y as string).split('-')[0]}: ${(args.y as string).split('-')[1]}% /
+                    yellow: translate-middle / blue: no translate-middle
```
oliverschuerch, line 49: I suggest to insert the legend text only once after the `Position arrangement` heading.

packages/documentation/src/stories/utilities/position/position.snapshot.stories.ts
```diff
@@ -0,0 +1,62 @@
...
+                    <div class="bg-dark"></div>
+                    <div class="bg-yellow position-${position} top-0 start-50">I'm ${position}</div>
+                    <div class="bg-dark"></div>
+                    <div class="bg-dark"></div>
+                    <div class="bg-dark"></div>
+                  </div>
+                </div>
+              `;
+            })}
+            <h3>Position arrangement</h3>
```
oliverschuerch, line 39: Are these examples for position absolute? If so, add the info after the heading.

packages/documentation/src/stories/utilities/position/position.snapshot.stories.ts
oliverschuerch, line 1: I don't know why, but it seems, that the bottom positioning is not working (or is at least not visualized) as expected:
image

oliverschuerch (changes requested):
Wow, the docs page looks great now!

packages/documentation/src/stories/utilities/position/position.snapshot.stories.ts
leagrdv: The bottom positioning actually is working, bottom: 0 actually positions the bottom of the square to the bottom of the container. So if you only look at the blue square, you see that bottom: 0 positions it at the bottom, and bottom: 50% positions the bottom of the square to the middle of the container. The yellow square actually adds a `translate: (-50%, -50%)` which only makes sense on top and left positioned items, so I removed them on elements that were positioned either from the bottom or the right of the container.

oliverschuerch: Did you see that bootstrap also implemented `translate-middle-x` and `translate-middle-y` classes?
Do we have them as well?

https://getbootstrap.com/docs/5.3/utilities/position/#center-elements

oliverschuerch (changes requested):

packages/documentation/src/stories/utilities/position/position.snapshot.stories.ts
leagrdv: We have them as well and I had forgotten to document them. I updated the snapshots accordingly and separated the position arrangement and translate middle.

oliverschuerch (approved):

PR #3993, opened 2024-11-14
feat(docs): add cargo theme

alizedebray:


packages/styles/src/post-external.scss
oliverschuerch, line 1: This update changes the output, because post-tokens-external are placed in a CSS-layer, while the tokens-files define their content un-layered. Is this the intention?
Also why should the `post-tokens-{idk|edk}.scss` file even exist, when we don't make use of it?

Same counts also for `post-internal.scss`.

oliverschuerch (changes requested):
There is a very heavy flickering when I change from `post` to `cargo` theme.
Would it maybe be possible to load the cargo theme first, add it to the page and only when this happend remove the post styles?

packages/styles/src/post-external.scss
alizedebray: Yes, this is intentional. I check with @gfellerph and the token must be in a layer only in the token entry files. 

Concerning internal entry files, we are not using them at the moment but our users might when they will be developing internal applications. But we can discuss this point during the round-table!

oliverschuerch (approved):

PR #3998, opened 2024-11-15
feat(icons): update source storage folder for v1 icons and add v2 source icons

oliverschuerch:


oliverschuerch:
will redo after massive source branch updates

PR #4012, opened 2024-11-18
feat(styles): updated sizing utility (no tokens) 

myrta2302:


packages/components/cypress/e2e/logo.cy.ts
```diff
@@ -49,14 +49,14 @@ describe('logo', () => {
...
-        });
-    });
+    // it('should not fit the parent height', () => {
+    //   cy.get('@logo')
+    //     .parent()
+    //     .invoke('innerHeight')
+    //     .then(parentHeight => {
+    //       cy.get('@logo').invoke('outerHeight').should('not.eq', parentHeight);
+    //     });
+    // });
```
oliverschuerch, line 59: Why is this commented out?
If we don't need this this anymore, please delete it. We can always restore it from the history if necessary.

packages/migrations/src/utils/constants.ts
```diff
@@ -1,6 +1,8 @@
 export const breakpoints = [ 'xs', 'sm', 'rg', 'md', 'lg', 'xl', 'xxl' ];
 
 const bootstrapSizes = [ '0', '1', '2', '3', '4', '5', 'auto' ];
+
```
oliverschuerch, line 4:
```suggestion
```

oliverschuerch (changes requested):

packages/components/cypress/e2e/logo.cy.ts
myrta2302: yes, this was by mistake. I ll revert the file.

oliverschuerch (changes requested):

packages/migrations/src/migrations/post/custom-select/index.ts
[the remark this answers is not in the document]
myrta2302: This is just updating the mw-100 to min-w-100. Saving the file does automatically the reformat.

gfellerph (changes requested):

packages/styles/src/utilities/_variables.scss
```diff
@@ -28,6 +29,126 @@ $position-values: (
...
+    values: $post-combined-sizes,
+  ),
+  'min-width': (
+    responsive: true,
+    property: min-width,
+    class: min-w,
+    values: $post-combined-sizes,
+  ),
+  'min-height': (
+    responsive: false,
```
oliverschuerch, line 63: Should this also be `true`?

oliverschuerch (changes requested):

oliverschuerch (approved):

gfellerph (approved):

PR #4019, opened 2024-11-19
chore: e2e workflow

oliverschuerch:


oliverschuerch:
this approach does not work

PR #4030, opened 2024-11-20
feat(styles): Added post-linkarea component (3830)

veyaromain:


veyaromain:
@alizedebray do you now why i get this code smell? This `h` import is used by `<slot>`

alizedebray:
> @alizedebray do you now why i get this code smell? This `h` import is used by `<slot>`

@veyaromain this is a false positive, I just accepted the issue on SonarCloud

packages/components/src/components/post-linkarea/post-linkarea.scss
```diff
@@ -0,0 +1,4 @@
+:host {
+  display: block;
```
alizedebray, line 2: Since you're not sure whether the element inside `post-linkarea` will be an inline or block element, it's better to set the wrapper to `display: contents`. This makes the `post-linkarea` behave as if it were replaced by its content in the DOM. Its logic remains intact, but it no longer acts as a box. Here’s an article that explains this well: https://la-cascade.io/articles/comment-fonctionne-css-display-content.
```suggestion
  display: contents;
```

packages/documentation/src/stories/components/linkarea/linkarea.stories.ts
```diff
@@ -0,0 +1,93 @@
...
+      control: {
+        type: 'boolean',
+      },
+      table: {
+        category: 'General',
+      },
+    },
+    anchorDefaultLink: {
+      name: 'Default link',
+      description: 'This is the link that will be used if no specific link is set.',
```
alizedebray, line 39:
```suggestion
      name: 'First link URL',
      description: 'This is the URL used for the first link in the card.',
```

packages/documentation/src/stories/components/linkarea/linkarea.stories.ts
```diff
@@ -0,0 +1,93 @@
...
+        type: 'text',
+      },
+      table: {
+        category: 'Links',
+      },
+    },
+    anchorSepcifiedLink: {
+      name: 'Specified link',
+      description:
+        'This is the link that will be used if you manually assign one using the `data-link` option.',
```
alizedebray, line 50:
```suggestion
      name: 'Second link URL',
      description: 'This is the URL used for the second link in the card.',
```

packages/documentation/src/stories/components/linkarea/linkarea.stories.ts
```diff
@@ -0,0 +1,93 @@
...
+  },
+  args: {
+    dataLink: false,
+    anchorDefaultLink: '#',
+    anchorSepcifiedLink: '#',
+  },
+  argTypes: {
+    dataLink: {
+      name: 'Custom selector',
+      description: 'Allows you to use a different link instead of the default first one.',
```
alizedebray, line 29:
```suggestion
      description: 'If `false`, clicking the card redirects to the first link. If `true`, a `data-link` attribute is added to the second link, which is used instead, overriding the default behavior.',
```

packages/documentation/src/stories/components/linkarea/linkarea.stories.ts
```diff
@@ -0,0 +1,93 @@
...
+  return html`
+    <post-linkarea>
+      <div class="card">
+        <div class="card-body">
+          <h5 class="card-title">Titulum</h5>
+
+          <p class="card-text">Contentus momentus vero siteos et accusam iretea et justo.</p>
+
+          <a class="card-link" href="${args.anchorDefaultLink}">Ligilo teksto</a>
+          <a class="card-link" href="${args.anchorSepcifiedLink}" ${spread(props)}>Pli da ligo</a>
```
alizedebray, line 75: Here you can use `nothing`, when a attribute receives `nothing` then it is not rendered at all: 
```suggestion
          <a class="card-link" href="${args.anchorSepcifiedLink}" data-link=${args.dataLink ? '' : nothing}>Pli da ligo</a>
```
Then you no longer need the `createSecondAnchorProps` function.

packages/documentation/src/stories/components/linkarea/linkarea.stories.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (args.dataLink) {
+    return { 'data-link': '' };
+  }
+  return {};
+}
+
+export const Default: StoryObj<HTMLPostLinkareaElement> = {};
+
+export const InitiallySpecified: StoryObj<HTMLPostLinkareaElement> = {
+  render: args => html` ${renderLinkarea({ ...args, dataLink: true })} `,
```
alizedebray, line 92: You can simply override the `args` like so:
```suggestion
  args: {
    dataLink: true,
  }
```

packages/components/src/components/post-linkarea/post-linkarea.tsx
```diff
@@ -0,0 +1,33 @@
+import { Component, Element, h, Host } from '@stencil/core';
+import { version } from '@root/package.json';
+
+@Component({
+  tag: 'post-linkarea',
+  styleUrl: 'post-linkarea.scss',
+  shadow: true,
```
alizedebray, line 7: You do not need to use a shadow DOM for this component as it does not render additional elements (only the host and its content) and the styles are minimal. 
```suggestion
```

alizedebray (changes requested):
Your code is very clean, well done! 👍
I've added a few change requests. If you need more details about anything, feel free to ask.

alizedebray (approved):

PR #4053, opened 2024-11-22
fix(components): delayed tooltip shown even after blur event

leagrdv:


packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -201,6 +203,9 @@ export class PostTooltip {
   async show(target: HTMLElement, triggeredByFocus = false) {
     if (this.delayed) await timeout(OPEN_DELAY);
 
+    // If focus or pointer event is not on the button anymore, don't show the tooltip
+    if (globalCurrentTarget !== target) return;
```
gfellerph, line 207: Do we need to reset the timeout here? I noticed a strange behaviour when continuously hovering and blurring.

Tried to record the behaviour I mean. When I hover and blur the trigger a couple of times but always leave before the timeout, I expect to never see the tooltip. Currently, it sometimes appears really quick even if I think I did not hover for the full timeout.

https://github.com/user-attachments/assets/b1774266-5aa8-4aca-8856-7160b3ad55d2

gfellerph (changes requested):

gfellerph (approved):

PR #4059, opened 2024-11-25
fix(documentations): fix nested rules warning in scss files for tiles

leagrdv:


gfellerph (approved):
Why does that fix the warning though?

leagrdv:
> Why does that fix the warning though?

@gfellerph Because the size `rg` does not exist anymore in the breakpoint, so it was not rendering the media queries, but only just the content inside directly which resulted in the width and height being under nested elements.

PR #4069, opened 2024-11-25
fix(components): use base tag href if defined for post-icon

leagrdv:


gfellerph (approved):
@leagrdv can you cherry pick this one to v8 as well?

PR #4070, opened 2024-11-25
fix(components): card-control display on light bg

leagrdv:


leagrdv:
Discussed with Alizé, closed as won't do as finding a fix seems too complicated for something that will be deprecated anyway.

PR #4085, opened 2024-11-27
chore(documentation): update styles import files with theme and channel

leagrdv:


packages/documentation/src/stories/components/app-store-badge/app-store-badge.mdx
oliverschuerch, line 1: I don't think this belongs here, please reset.

packages/documentation/.storybook/addons/styles-switcher/StylesSwitcher.tsx
```diff
@@ -94,6 +96,49 @@ function StylesSwitcher() {
...
+    if (!stylesCodeBlocks) return;
+
+    const t = currentTheme.toLowerCase();
+    const c = currentChannel.toLowerCase();
+    const domain = "'@swisspost/design-system-styles/";
+
+    stylesCodeBlocks.forEach(stylesCodeBlock => {
+      let sourceArray = stylesCodeBlock.querySelectorAll('.token.string');
+      if (sourceArray.length) {
+        sourceArray.forEach((s, i) => {
```
oliverschuerch, line 112: This three lines could be written as one.
Try it out if you like.

`Array.from(stylesCodeBlock.querySelectorAll('.token.string')).forEach((s, i) => {`

packages/documentation/.storybook/addons/styles-switcher/StylesSwitcher.tsx
```diff
@@ -94,6 +96,49 @@ function StylesSwitcher() {
...
 
+  /**
+   * Sets the design system styles import SCSS file to the correct theme and channel file
+   */
+  useEffect(() => {
+    if (!stylesCodeBlocks) return;
+
+    const t = currentTheme.toLowerCase();
+    const c = currentChannel.toLowerCase();
+    const domain = "'@swisspost/design-system-styles/";
```
oliverschuerch, line 107: This does not look like a domin to me.
`packageName` would maybe match better.

packages/documentation/.storybook/addons/styles-switcher/StylesSwitcher.tsx
```diff
@@ -94,6 +96,49 @@ function StylesSwitcher() {
...
+      let sourceArray = stylesCodeBlock.querySelectorAll('.token.string');
+      if (sourceArray.length) {
+        sourceArray.forEach((s, i) => {
+          let source = s.innerHTML;
+          // Remove the domain from the source to make sure we don't override it
+          source = source.replace(domain, '');
+
+          // Check if one of the themes or channels are in the scss path
+          let theme = THEMES.find(t => source.indexOf(t.toLowerCase()) > -1);
+          let channel = CHANNELS.find(c => source.indexOf(c.toLowerCase()) > -1);
```
oliverschuerch, line 119: The variable names `t` and `c` have already been used in the parent scope. Even thought JS allows you to write your code like this, I do not recommend it.

packages/documentation/.storybook/addons/styles-switcher/StylesSwitcher.tsx
```diff
@@ -94,6 +96,49 @@ function StylesSwitcher() {
...
+
+          const updateTheme = theme && theme.toLowerCase() !== t;
+          const updateChannel = channel && channel.toLowerCase() !== c;
```
oliverschuerch, line 122: Would that work too?
```suggestion
          const updateTheme = theme?.toLowerCase() !== t;
          const updateChannel = channel?.toLowerCase() !== c;
```

packages/documentation/.storybook/addons/styles-switcher/StylesSwitcher.tsx
```diff
@@ -94,6 +96,49 @@ function StylesSwitcher() {
...
+            if (updateTheme) {
+              source = source.replace((theme as string).toLowerCase(), t);
+            }
+
+            if (updateChannel) {
+              source = source.replace((channel as string).toLowerCase(), c);
+            }
+
+            const newSource = domain + source;
+            sourceArray[i].innerHTML = newSource;
```
oliverschuerch, line 135: Isn't `sourceArray[i].innerHTML` the same as `source`?

packages/documentation/.storybook/addons/styles-switcher/StylesSwitcher.tsx
oliverschuerch, line 1: I know there are a lot of CRs and I had even more, but I think it makes sense clearify some things in a talk. So call me, whenever you see my comments ;)

oliverschuerch (changes requested):

packages/documentation/.storybook/addons/styles-switcher/StylesSwitcher.tsx
leagrdv: I changed it to remove the if (sourceArray.length) but I kept the const sourceArray as I use it later in the code

packages/documentation/.storybook/addons/styles-switcher/StylesSwitcher.tsx
leagrdv: no, that does not work because it tries to apply toLowerCase() to undefined so it breaks

packages/documentation/.storybook/addons/styles-switcher/StylesSwitcher.tsx
leagrdv: yes, pretty much but if I put source = newSource, that only updates the source variable and not the actual innerHTML of my element.

oliverschuerch (approved):

PR #4095, opened 2024-11-28
chore(styles, components): remove strikethrough on disabled elements

leagrdv:


.changeset/lovely-seas-call.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-components': minor
+'@swisspost/design-system-styles': minor
```
gfellerph, line 3: It's not a new feature, so I'd propose to change it to patch.

```suggestion
'@swisspost/design-system-components': patch
'@swisspost/design-system-styles': patch
```

gfellerph (approved):

PR #4097, opened 2024-11-28
feat(styles): add list mixins

oliverschuerch:


packages/documentation/.storybook/styles/preview.scss
```diff
@@ -128,6 +128,13 @@ $monospace: SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier
     }
   }
 
+  [data-color-scheme='dark'] {
+    code {
```
leagrdv, line 132: I'm not sure this is useful as the color-scheme is set on the storybook preview elements, and not the text content of the stories.

packages/documentation/src/stories/foundations/typography/list/list.snapshot.stories.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export const Lists: Story = {
+  render: () => {
+    return schemes(
+      () => html`
+        <h1>Lists</h1>
+
+        <h2>Bullet</h2>
+        <div class="row">
+          <div class="col-6">
```
leagrdv, line 24: `col-md-6` might be better here to have full width columns on mobile

packages/styles/src/mixins/_utilities.scss
```diff
@@ -8,10 +8,10 @@
 @use '../variables/commons';
 @use '../variables/breakpoints';
 
+@use './list' as list-mx;
+
 @mixin reset-list() {
-  margin-block: 0;
-  padding-inline-start: 0;
-  list-style: none;
+  @include list-mx.list-unstyled();
```
leagrdv, line 14: reset-list is being used on the stepper and the layout seems broken now.
I'm not sure reset-list() was doing the same as list-unstyled.

leagrdv (changes requested):

packages/documentation/src/stories/foundations/typography/list/list.snapshot.stories.ts
```diff
@@ -0,0 +1,124 @@
...
+            <code>&lt;ul&gt;</code>
+            ${renderList({ el: 'ul' })} ${renderList({ el: 'ul', classList: 'list-revert' })}
```
alizedebray, line 26: This list has a `class="undefined"`.

packages/documentation/src/stories/raw-components/list/list.snapshot.stories.ts
```diff
@@ -50,40 +50,3 @@ export const PostList: Story = {
...
-            item.
-          </li>
-          <li>
-            Nested list:
-            <ol>
-              <li>This is a nested list</li>
-              <li>It is further indented, depending on the depth of nesting.</li>
-              <li>Nested lists numbers are independent form the numbers of their parents.</li>
-            </ol>
-            After nested list item
```
alizedebray, line 72: It would be good to have a text after the nested lists in the snapshots as we found a bug with this specific use case when developing the ordered lists.

alizedebray (changes requested):

packages/documentation/.storybook/styles/preview.scss
oliverschuerch: This is useful, if you use `code` elements in your snapshot stories, because the `schemes` wrapper function uses also the `data-color-scheme` attribute.

packages/styles/src/mixins/_utilities.scss
oliverschuerch: I've tryed to fix the stepper component, but after some attempts, I resetted the reset-list mixin to its original state, so we don't mess up anything already using it.

alizedebray (approved):

packages/documentation/src/stories/foundations/typography/list/list.snapshot.stories.ts
```diff
@@ -0,0 +1,132 @@
...
+
+          <div class="col-6">
+            <code>&lt;ul class="list-bullet"&gt;</code>
+            ${renderList({ el: 'ul', classList: 'list-bullet' })}
+            ${renderList({ el: 'ul', classList: 'list-bullet list-revert' })}
+            ${renderList({ el: 'ul', classList: 'list-bullet list-unstyled' })}
+          </div>
+
+          <div class="col-6">
+            <code>&lt;ul class="list-bullet list-inline"&gt;</code>
```
leagrdv, line 56: Shouldn't `list-bullet list-inline` have bullets? 
Also putting the lists on two columns make it hard to see the list-inline as there is not enough space for the list item to be on the same line.

oliverschuerch: No, a `.list-bullet .list-inline` should not have any bullets.
Can you explain your second point again? I don't really get it.

leagrdv (approved):

PR #4110, opened 2024-11-28
fix(styles): ol styles without using grid

leagrdv:


oliverschuerch (changes requested):
Please add a changeset

packages/styles/src/mixins/_utilities.scss
```diff
@@ -8,6 +8,8 @@
 @use '../variables/commons';
 @use '../variables/breakpoints';
 
+@use './list' as list-mx;
+
```
oliverschuerch, line 12: Why is that necessary?

oliverschuerch (changes requested):

oliverschuerch (approved):
Good job!

PR #4114, opened 2024-11-29
chore(styles): quick win styles build improvement

gfellerph:
Using `sass-embedded` should improve performance of our sass builds a bit and it's very easy to implement. Stayed with version 1.78.0 due to additional deprecation warnings when updating to 1.81.0.


packages/documentation/vite.config.js
```diff
@@ -23,4 +23,11 @@ export default {
     ],
     exclude: ['@swisspost/design-system-styles'],
   },
+  css: {
```
gfellerph, line 26:
```suggestion
  // https://www.oddbird.net/2024/08/14/sass-compiler/
  css: {
```

oliverschuerch (approved):
Can't sass-embedded also be used in the components package?

PR #4116, opened 2024-11-29
chore(styles): move sizing utilities to utilities index

gfellerph:
This enables more comprehensive utilities import because sizing utilities were part of the components and not the utilities import.


packages/styles/src/utilities/_index.scss
oliverschuerch, line 1: Shouldn't we move the imported file as well?
I mean from the `components` folder to the `utilities` folder?

oliverschuerch (changes requested):

PR #4119, opened 2024-11-29
chore(styles): fixes playground

gfellerph:
Adds more specific token selects and fixes the preview, also adds compatibility for the `sass-embedded` package.

### DIFF
--- a/packages/styles/index.html
+++ b/packages/styles/index.html
@@ -12,6 +12,7 @@
     <link id="theme" rel="stylesheet" href="src/tokens/_post-theme.scss" />
     <link rel="stylesheet" href="src/tokens/_device.scss" />
     <link rel="stylesheet" href="src/tokens/_components.scss" />
+    <link rel="stylesheet" href="src/tokens/_schemes.scss" />
 
     <!-- Styles -->
     <link rel="stylesheet" href="src/layout/index.scss" />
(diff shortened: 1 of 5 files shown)


packages/styles/index.html
```diff
@@ -4,33 +4,47 @@
...
-      id="tokens-select"
-      value="external"
-      class="mb-4"
-      onchange="document.getElementById('tokens').setAttribute('href', `/src/post-tokens-${this.value}.scss`)"
-    >
-      <option value="external">Post external</option>
-      <option value="internal">Post internal</option>
-    </select>
-    <hr />
+  <body class="palette-default">
```
alizedebray, line 23: You don't need to set a palette on the body: 
```suggestion
  <body>
```

packages/styles/src/palettes/_mixins.scss
```diff
@@ -13,6 +13,7 @@
 @mixin palettes($theme) {
   body {
     background-color: var(--post-current-palette-bg) !important;
+    color: var(--post-current-palette-fg) !important;
```
alizedebray, line 16: The `--post-current-palette-fg` is for the `.palette-text` utility which allows to have blue headings on cargo green. The body should not use it.

Feel free to rename the custom property if you think it's unclear!

```suggestion
```

PR #4136, opened 2024-12-03
feat: add the ability to use our icons from within SCSS

oliverschuerch:
I've used one icon from the current `Post` IconSet (icon: 1000) and one from the new `UI` IconSet (icon accessibility).

Both have been tested with the new `post-icon` mixin and the `post-icon` web-component.
The icons have been implemented into and were served by the test project itself.

**Windows 10**
- Chrome 80, 90, 100, 110, 120, 130, 131
- Edge 80, 100, 120, 131
- Firefox 80, 100, 120, 131
- Opera 80, 100, 115

**Android**
- Chrome 9, 10, 12, 14

**Mac OS**
- Safari 12, 13, 14, 15, 16, 17, 18
- Chrome 80
- Firefox 80
- Opera 80

**iOS**
- Safari 13, 16, 18

**iPadOS**
- Safari 12, 14, 16, 18


gfellerph (approved):

packages/icons/src/constants.ts
```diff
@@ -19,7 +19,35 @@ export const ICON_V1_TEMPLATE = `<svg xmlns="http://www.w3.org/2000/svg">
...
+const ICON_V2_CONTAINER_STYLES = ICON_V2_SIZES.map((size, i) => {
+  const nextSize = ICON_V2_SIZES[i + 1];
+  const min = i !== 0 && `(min-width: ${size}px)`;
+  const max = i < ICON_V2_SIZES.length - 1 && `(max-width: ${nextSize - 1}px)`;
+
+  return `@container pi ${[min, max].filter(Boolean).join(' and ')} {
+    g {
+      --pis-${size}: block;
+    }
+  }`;
```
gfellerph, line 31: Can you add a fallback for browsers that don't support container queries?

gfellerph (changes requested):

packages/icons/src/constants.ts
oliverschuerch: Not necessary anymore, because I switched to use media queries. They are working great, and are supported since almost a decade. Please have a look at the tested browsers at the top of this PR.

packages/icons/src/constants.ts
```diff
@@ -9,24 +9,43 @@ export const ID_UNWANTED_PARTS = ['shape'];
...
 
+const ICON_V2_TEMPLATE_STYLES = `<style>
+  ${ICON_V2_SIZES.map((size, i) => {
+    const isFirst = size === ICON_V2_SIZES[0];
+    const isLast = size === ICON_V2_SIZES[ICON_V2_SIZES.length - 1];
+    const nextSize = ICON_V2_SIZES[i + 1];
+    const min = !isFirst && `(min-width: ${size}px)`;
+    const max = !isLast && `(max-width: ${nextSize - 1}px)`;
+
+    return `@media ${[min, max].filter(Boolean).join(' and ')} {
```
gfellerph, line 27:
```suggestion
    // Concisely filter out null, undefined or false values, https://michaeluloth.com/javascript-filter-boolean/
    return `@media ${[min, max].filter(Boolean).join(' and ')} {
```

packages/icons/src/constants.ts
```diff
@@ -9,24 +9,43 @@ export const ID_UNWANTED_PARTS = ['shape'];
...
+      g {
+        --${ID_SYMBOL_PREFIX}${size}: block;
+      }
+    }`;
+  }).join('\n\n')}
```
gfellerph, line 32: This function puzzled me quite a bit and just for a mental exercise I added a suggestion for making it a little bit more concise. While I like the shorthand boolean filter, this version does not need it. If you decide to not implement it, consider subtracting only 0.01 px from the max-width media query to catch edge cases where subpixel rendering would end up between two queries and render the mobile (?) icon unexpectedly.

```suggestion
const ICON_V2_TEMPLATE_STYLES = `<style>
  ${ICON_V2_SIZES.map((size, i) => {
    const query = [];
    if (i > 0) {
      query.push(`(min-width: ${ICON_V2_SIZES[i]}px)`);
    }
    if (i < ICON_V2_SIZES.length - 1) {
      query.push(`(max-width: ${ICON_V2_SIZES[i + 1] - 0.01}px)`);
    }

    return `@media ${query.join(' and ')} {
      g {
        --${ID_SYMBOL_PREFIX}${size}: block;
      }
    }`;
  }).join('\n\n')}
```

gfellerph (approved):

PR #4137, opened 2024-12-03
fix(components): update post-list component

alizedebray:
- Reflected the `horizontal` property on the host to allow updating the orientation with `document.querySelector('post-list).horizontal = true;`
- Reduced the selectors specificity to allow flexible styling
- Gave a default value to all CSS custom properties


oliverschuerch (changes requested):
This PR needs a changeset imo. At least, because of the specificity change.

oliverschuerch (approved):

PR #4148, opened 2024-12-04
feat(component): add main navigation styles

alizedebray:
This PR is only about styling the main navigation according to the design. 
The behavior will come with another PR.


packages/components/src/index.html
gfellerph, line 1: Can you update this structure in the [documentation/components/header](https://preview-4148--swisspost-design-system-next.netlify.app/?path=/docs/header--docs) file as well?

gfellerph (changes requested):

gfellerph (approved):

PR #3775, opened 2024-10-16
fix: check props on componentDidLoad

gfellerph:
Delays the prop checks until the last lifecycle hook to give frameworks time to render their props before the check runs.

PR #4151, opened 2024-12-04
fix: check props on componentDidLoad #3775 (cherry-pick v7)

schaertim:


schaertim:
No need for review yet, will have to fix build first

alizedebray (approved):

PR #4177, opened 2024-12-09
feat(components): mega dropdown component

leagrdv:


oliverschuerch (changes requested):
Could you intergrate the markup for the post-megadropdown into the storybook file located under src/stories/components/header/components/header.markup.ts?

And since this is an internal component it would be good to have:
- [ ] A docs page located under "Raw Components"
- [ ] End-to-end tests

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
oliverschuerch, line 1: How can I navigate from one level of the megadropdown to a lower one and back?
image

packages/components/src/components/post-megadropdown/post-megadropdown.scss
```diff
@@ -1,71 +1,161 @@
...
+    }
   }
 
   post-list {
-    display: flex;
-    flex-direction: column;
-    gap: 0.5rem;
+    display: grid;
+    grid-row: span 2;
+    grid-template-rows: subgrid !important;
```
oliverschuerch, line 93: Is the `!important` really necessary?

packages/components/src/components/post-megadropdown-trigger/post-megadropdown-trigger.tsx
```diff
@@ -1,29 +1,78 @@
...
   }
 
   render() {
     return (
-      <Host slot="post-megadropdown-trigger">
-        <button aria-expanded={this.toggled} onClick={() => this.handleClick()}>
-          <slot></slot>
-        </button>
+      <Host data-version={version} tab-index="-1">
+        <slot></slot>
```
oliverschuerch, line 75: Since we need a `button` element here, wouldn't it be possible to implement the button directly into the component and use a prop to let the users define the label of it?
This would not be an accessibility problem imo, because it's set up as a non-shadow component.

packages/components/src/components/post-megadropdown-trigger/post-megadropdown-trigger.tsx
```diff
@@ -1,29 +1,78 @@
...
+
+  private get megadropdown(): HTMLPostMegadropdownElement | null {
+    const ref = this.root.getElementById(this.for);
+    return ref && ref.localName === 'post-megadropdown'
+      ? (ref as HTMLPostMegadropdownElement)
+      : null;
+  }
+
+  private handleToggle() {
+    const megadropdown = this.megadropdown;
```
oliverschuerch, line 47: Can we remove this variable and instead just use `this.megadropdown` below?

packages/components/src/components/post-megadropdown-trigger/post-megadropdown-trigger.tsx
```diff
@@ -1,29 +1,78 @@
...
+   * Watch for changes to the `for` property to validate its type and ensure it is a string.
+   * @param forValue - The new value of the `for` property.
+   */
+  @Watch('for')
+  validateControlFor(forValue = this.for) {
+    checkType(forValue, 'string', 'The "for" property is required and should be a string.');
+  }
+
+  private get megadropdown(): HTMLPostMegadropdownElement | null {
+    const ref = this.root.getElementById(this.for);
```
oliverschuerch, line 40: You're looking for an ID in the light dom here, right? So I don't see the point, why we need to have this `root` variable at all? `document.getElementById` or `document.querySelector` would do the same.

oliverschuerch (changes requested):

oliverschuerch:
> Could you intergrate the markup for the post-megadropdown into the storybook file located under src/stories/components/header/components/header.markup.ts?
> 
> And since this is an internal component it would be good to have:
> 
> * [ ]  A docs page located under "Raw Components"
> * [ ]  End-to-end tests

This can be done in another PR.

packages/documentation/src/stories/components/header/components/header.markup.ts
```diff
@@ -67,14 +73,19 @@ export default html`<post-header>
...
-          <post-closebutton>Schliessen</post-closebutton>
         </post-megadropdown>
       </post-list-item>
       <post-list-item>
-        <post-megadropdown-trigger> Pakete </post-megadropdown-trigger>
-        <post-megadropdown>
-          <button slot="back-button"><- Zurück</button>
-          <h2><a href="">Pakete title</a></h2>
+        <post-megadropdown-trigger for="pakete">
+          <button class="btn btn-link">Pakete</button>
```
alizedebray, line 80: Please remove the button classes for all `post-megadropdown-trigger`:
```suggestion
          <button>Pakete</button>
```

packages/components/src/components/post-megadropdown/post-megadropdown.scss
```diff
@@ -1,71 +1,161 @@
...
+}
+
+@keyframes slide-out {
+  from {
+    transform: translateX(0%);
+  }
+  to {
+    transform: translateX(100%);
+  }
+}
```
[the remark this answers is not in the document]
leagrdv: it doesn't seem to be that simple, putting reverse on animation-direction does not do the animation as the timeline is considered finished already.

packages/components/src/components/post-megadropdown-trigger/post-megadropdown-trigger.tsx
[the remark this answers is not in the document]
leagrdv: It's there. You don't see it in the PR because it was already created before

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
```diff
@@ -1,56 +1,113 @@
...
-   * @param element HTMLElement
-   * @returns boolean
+   * Toggles the dropdown visibility based on its current state.
    */
   @Method()
-  async show(element: HTMLElement) {
-    return this.popoverRef.show(element);
+  async toggle(target: HTMLElement) {
+    this.isVisible ? await this.hide() : await this.show(target);
   }
```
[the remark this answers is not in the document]
leagrdv: I removed the async/await on the hide method but in order to be a method that can be called from outside, it needs to return a Promise, hence the async/await

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
```diff
@@ -1,56 +1,113 @@
...
+    }
   }
 
   private handleBackButtonClick() {
-    this.hide();
+    this.animationClass = 'slide-out';
+    forceUpdate(this);
+    setTimeout(() => {
+      this.hide();
+    }, 350);
```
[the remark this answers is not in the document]
leagrdv: 350 ms is the time it takes to make the slide out animation.

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
leagrdv: The top level elements are one after the other on mobile (left screenshot), you click on one label, it opens the subnavigation (right screenshot) with its children. To go back to the top level elements, you click on "zurück"

packages/components/src/components/post-megadropdown-trigger/post-megadropdown-trigger.tsx
leagrdv: I've move it there but I kept the slot for the label instead of adding the prop. What do you think?

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
```diff
@@ -1,56 +1,113 @@
...
+   * This state is internally managed to track whether the dropdown is open (`true`) or closed (`false`),
+   * and updates automatically when the dropdown is toggled.
+   */
+  @State() isVisible: boolean = false;
+
+  /**
+   * Emits when the dropdown is shown or hidden.
+   * The event payload is a boolean: `true` when the dropdown was opened, `false` when it was closed.
+   **/
+  @Event() toggleMegadropdown: EventEmitter<boolean>;
```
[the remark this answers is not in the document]
leagrdv: I checked with Philipp, I'm changing it to postToggleMegadropdown for now

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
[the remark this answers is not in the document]
oliverschuerch: Then we should add an animation end event listener. And if we change the animation duration in the future, we will not run in a problem, because we forgot to update this timeout.
https://developer.mozilla.org/en-US/docs/Web/API/Element/animationend_event

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
oliverschuerch: 😄 I know how it should work, but I was not able to test that behaviour on your component.
Before I check the branch out again and try it, can you confirm this behaviour is built in your component?

packages/components/src/components/post-megadropdown-trigger/post-megadropdown-trigger.tsx
oliverschuerch: I can't see this change? I mean, there is no button wrapper around the slot.

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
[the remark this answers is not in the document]
oliverschuerch: You can resolve this CR, as soon as it's implemented.

oliverschuerch (changes requested):

packages/components/src/components/post-megadropdown-trigger/post-megadropdown-trigger.tsx
leagrdv: If you click on the file from the CR, you don't see it but if you go to the list of files, you should be able to see it.

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
leagrdv: Sorry, I hadn't understood the question 😆 but yes, it does work as intended.

oliverschuerch (approved):

packages/components/src/components/post-mainnavigation/post-mainnavigation.scss
```diff
@@ -11,18 +11,20 @@ $nav-height: var(--post-core-dimension-56);
...
   post-list-item {
-    a {
+    > a {
       text-decoration: none;
       border-radius: 0;
     }
 
-    button {
+    > button,
+    post-megadropdown-trigger button {
```
alizedebray, line 20: I think this is not necessary, there will only be buttons in the post-megadropdown-trigger. 
```suggestion
    post-megadropdown-trigger button {
```

PR #4190, opened 2024-12-09
feat(components): add composable footer

oliverschuerch:


packages/components/src/components/post-footer/post-footer.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  connectedCallback() {
+    window.addEventListener('postBreakpoint:name', this.breakpointChange.bind(this));
+  }
+
+  private breakpointChange(e: CustomEvent) {
+    this.isMobile = e.detail === 'mobile';
+  }
+
+  render() {
```
gfellerph, line 37: You could simplify the rendering logic a bit by outsourcing the individual structures to functions like

```tsx
private function renderMobileFooter() { return (<div>Accordion footer</div>); }
private function renderDesktopFooter() { return (<div>Desktop footer</div>); }

render() { return (<stuff /> { this.isMobile ? this.renderMobileFooter() : this.renderDesktopFooter() } <otherstuff/>); }
```

gfellerph (approved):

PR #4200, opened 2024-12-10
chore(styles, documentation): update styles of btn-link

leagrdv:


packages/styles/src/components/button.scss
```diff
@@ -105,8 +108,11 @@ tokens.$default-map: components.$post-button;
...
 .btn-link {
-  padding-inline-start: 0;
-  padding-inline-end: 0;
+  text-decoration: underline;
+  min-height: 0;
+  font-size: inherit;
+  font-weight: inherit;
+  padding: 0;
+  border-radius: tokens.get('focus-border-radius', helpers.$post-focus);
 }
```
oliverschuerch, line 118: Should we use the link tokens here (which are defined in the file `elements/anchor.scss`)?
What do you think?

oliverschuerch (changes requested):

packages/styles/src/components/button.scss
leagrdv: yes, good idea! I used the ones defined there for text-decoration, radius and hover color :)

packages/styles/src/components/button.scss
```diff
@@ -105,8 +108,15 @@ tokens.$default-map: components.$post-button;
...
-// Tertiary with no padding (overrides the padding defined by the sizing classes above)
 .btn-link {
-  padding-inline-start: 0;
-  padding-inline-end: 0;
+  text-decoration: tokens.get('link-decoration', elements.$post-link);
+  border-radius: tokens.get('link-border-radius', elements.$post-link);
+  min-height: 0;
+  font-size: inherit;
+  font-weight: inherit;
+  padding: 0;
```
oliverschuerch, line 117:
```suggestion
  display: inline;
  text-decoration: tokens.get('link-decoration', elements.$post-link);
  border: 0 none;
  border-radius: tokens.get('link-border-radius', elements.$post-link);
  min-height: 0;
  font-size: inherit;
  font-weight: inherit;
  padding: 0;
```
Even with this changes, the button has still a different height than an anchor element:
image

To fix this, we could add `display: inline-block` to the anchor styles, but I'm not sure if this would produce unwanted side-effects.
image

So I'd say, we go with the solution suggested above.

oliverschuerch (changes requested):

packages/styles/src/components/button.scss
leagrdv: I commited your suggestion but then it doesn't look good if there is an icon, so I just removed the display: inline and it seems to work better that way.
Screenshot 2024-12-12 100905
Screenshot 2024-12-12 100939

oliverschuerch (approved):

PR #4211, opened 2024-12-11
fix(component): fix tab isolation in nested `post-tabs` component

alionazherdetska:


gfellerph (approved):
@alionazherdetska can you cherry pick to v7 and v8?

PR #4241, opened 2024-12-12
chore(components): update slide animations

alizedebray:


gfellerph (approved):

oliverschuerch (changes requested):
When clicking on the back to top button, the button disappears immediately and the page scrolls up, then the button gets visible again, before it finally moves above the viewport.
https://github.com/user-attachments/assets/3ba5f7c2-c2b6-4e19-9f22-a36a4caa02d0

Tested in Chrome & FF

alizedebray:
> When clicking on the back to top button, the button disappears immediately and the page scrolls up, then the button gets visible again, before it finally moves above the viewport. https://github.com/user-attachments/assets/3ba5f7c2-c2b6-4e19-9f22-a36a4caa02d0
> 
> Tested on Chrome & FF

@oliverschuerch the behavior did not change compared to the released back-to-top button.
This "bug" is because of the padding that storybooks adds to the full-screen preview, but in real application there will be no padding and the button will behave as expected. You can test it by removing the padding on the `.sb-main-padded` element. 
(The [header's preview](https://preview-4241--swisspost-design-system-next.netlify.app/?path=/story/header--default&full=true) doesn't have a padding so we should be able to do the same for the back-to-top preview)

PR #4258, opened 2024-12-13
fix(documentation): move internet-header, footer and breadcrumbs to raw-components and new header to modules

oliverschuerch:


packages/documentation/src/stories/modules/header/language-switch/language-switch.stories.ts
```diff
@@ -4,7 +4,7 @@ import { MetaComponent } from '@root/types';
 
 const meta: MetaComponent<HTMLPostLanguageSwitchElement> = {
   id: 'decbb10c-2b39-4f47-b67d-337d8111a3ae',
-  title: 'Components/Header/Language Switch',
+  title: 'Modules/Header/Language Switch',
```
alizedebray, line 7: The language switch is not a module:
```suggestion
  title: 'Components/Language Switch',
```

gfellerph (approved):

packages/documentation/src/stories/modules/header/header.stories.ts
```diff
@@ -4,7 +4,7 @@ import HeaderMarkup from './components/header.markup';
 
 const meta: MetaComponent = {
   id: 'header',
-  title: 'Raw Components/Header',
+  title: 'Modules/Header/Header',
```
alizedebray, line 7:
```suggestion
  title: 'Modules/Header',
```

alizedebray (changes requested):
Can you make sure the "Modules" section is initially collapsed?

packages/documentation/src/stories/modules/header/language-switch/language-switch.stories.ts
oliverschuerch: I know, but I thought you want to have it at the same place, where the header docs are. At least this was the case before.

alizedebray (approved):

PR #4260, opened 2024-12-13
fix(components): update style of language switch and fix keyboard nav

leagrdv:


alizedebray (commented):
The focus ring is still blue on desktop and the gap between the language options is no correct on mobile (should be 8px).

PR #4262, opened 2024-12-13
chore(components): header hcm

myrta2302:


alizedebray (approved):
Nice! Can you also add a border to separate the mobile menu from the page bellow?
image

packages/components/src/components/post-closebutton/post-closebutton.scss
```diff
@@ -0,0 +1,7 @@
+@use '@swisspost/design-system-styles/mixins/utilities';
+
+post-closebutton button post-icon {
```
alizedebray, line 3:
```suggestion
post-closebutton button {
```

packages/components/src/components/post-header/post-header.scss
```diff
@@ -12,6 +13,9 @@
   --global-header-minimal-height: 24px;
   --main-header-height: 56px;
   --header-height: calc(var(--global-header-height) + var(--main-header-height));
+  @include utilities.high-contrast-mode() {
+    display: block;
```
alizedebray, line 17: There would be an issue with the stickiness if you display the header as a block. Can you add the border to the local-header instead?

packages/components/src/components/post-closebutton/post-closebutton.scss
```diff
@@ -0,0 +1,7 @@
...
+  @include utilities.high-contrast-mode() {
+    color: LinkText !important;
```
alizedebray, line 5: The close button is a button not a link, you should therefore use ButtonFace and ButtonText so that HCM users easily understand it. Can you please update the whole PR taken this into account?

alizedebray (changes requested):

alizedebray (commented):
The preview does not seem to be up-to-date.

alizedebray (changes requested):
Can you try and reduce the amount of custom HCM styles and use the default HCM instead? 
You might need to remove any `color: inherit` set for example on the buttons, if that is the case it means that the color should be tokenized.

packages/styles/src/components/header/index.scss
```diff
@@ -64,4 +64,9 @@ post-header {
       background: var(--post-core-color-sandgrey-100);
     }
   }
+  button {
+    @include utilities.high-contrast-mode() {
+      color: ButtonText !important;
+    }
+  }
```
alizedebray, line 71: This does not seem necessary and overrides the color on hover so it can probably be removed: 
```suggestion
```

packages/styles/src/mixins/_button.scss
```diff
@@ -17,10 +17,14 @@ tokens.$default-map: components.$post-button;
...
   background: none;
-  color: inherit;
+  color: tokens.get('button-secondary-enabled-fg', components.$post-button);
   font: inherit;
   user-select: none;
   appearance: button; // for input
+
+  @include utilities.high-contrast-mode() {
+    background-color: ButtonFace !important;
+  }
```
alizedebray, line 27: The background-color is missing on hover: 
```suggestion
  @include utilities.high-contrast-mode() {
    background-color: ButtonFace !important;
    
    &:hover {
      background-color: Highlight !important;
    }
  }
```

packages/components/src/components/post-header/post-header.scss
```diff
@@ -42,6 +43,11 @@
   .local-sub {
     display: none;
   }
+  .navigation {
```
alizedebray, line 46: This does not seem to be applied:
image

alizedebray (changes requested):
Nice!

packages/components/src/components/post-togglebutton/post-togglebutton.scss
```diff
@@ -6,7 +6,13 @@
   cursor: pointer;
   outline-offset: tokens.get('focus-outline-offset', helpers.$post-focus) !important;
   outline: tokens.get('focus-outline-color', helpers.$post-focus) none
-  tokens.get('focus-outline-width', helpers.$post-focus) !important;
+    tokens.get('focus-outline-width', helpers.$post-focus) !important;
+
+  @include utilities.high-contrast-mode() {
```
alizedebray, line 11: The hover styles are missing for the toggle button. But actually, this only happens in the header as the `post-togglebutton` should otherwise be used with a `.btn` class. 

Can you move this styles to the `packages/styles/src/components/header/index.scss` file?

alizedebray (changes requested):
There is no border at the bottom of the megadropdown: 
image

packages/styles/src/components/header/index.scss
```diff
@@ -64,4 +64,12 @@ post-header {
       background: var(--post-core-color-sandgrey-100);
     }
   }
+
+  post-togglebutton {
+    @include utilities.high-contrast-mode() {
+      background-color: ButtonFace !important;
+      border-color: ButtonBorder !important;
+      color: ButtonText !important;
```
alizedebray, line 72: Please also add the hover styles.

alizedebray (changes requested):

alizedebray (approved):
👍

PR #4301, opened 2024-12-16
fix(components): add border top main navigation on mobile

leagrdv:


packages/components/src/components/post-header/post-header.scss
```diff
@@ -172,6 +172,7 @@ slot[name='post-logo'] {
   ::slotted(post-mainnavigation) {
     background-color: var(--post-core-color-sandgrey-002);
     gap: var(--post-core-dimension-32);
+    border-top: 1px solid var(--post-core-color-sandgrey-012);
```
gfellerph, line 175: Would be nice to have the border during the animation, right now it looks a bit wild and having the border there as part of the local header (or a fixed element) might help visually separating things during the slide up and down.

gfellerph (changes requested):

gfellerph (approved):

PR #4306, opened 2024-12-16
fix(styles): fix focus indicator visibility for links in accordion items inside the footer on mobile

alionazherdetska:
…ems on mobile


alizedebray (changes requested):
If this issue happens with the footer, it could also happen for anyone using the post-collapsible. I would rather fix the post-collapsible overflow than adding specific code for the footer. The hidden overflow is actually only necessary when the collapsible is closed and during the transition but it can be removed when it is open.

alizedebray (changes requested):
Can you rather update the `expand` animation so that we do not have the same issue with another component in the future?

alizedebray (approved):

PR #4311, opened 2024-12-16
fix(styles): HCM on segmented buttons

leagrdv:


packages/styles/src/components/segmented-button.scss
```diff
@@ -173,6 +181,10 @@ $post-segmented-button-max-count: 8;
       &:hover {
         input {
           border-color: tokens.get('button-segmented-hover-border');
+
+          @include utilities.high-contrast-mode() {
+            border-color: Highlight;
```
alizedebray, line 186: I cannot see this border-color on Firefox.

packages/styles/src/components/segmented-button.scss
```diff
@@ -91,6 +91,10 @@ $post-segmented-button-max-count: 8;
             tokens.get('button-segmented-selected-border');
           border-radius: inherit;
           transition: transform 0.4s cubic-bezier(0.25, 1.4, 0.5, 0.9);
+
+          @include utilities.high-contrast-mode() {
+            background-color: SelectedItem;
```
alizedebray, line 96: Just in case: 
```suggestion
            background-color: SelectedItem;
            color: SelectedItemText;
```

.changeset/stale-months-dance.md
alizedebray, line 1: 👍

alizedebray (changes requested):

packages/styles/src/components/segmented-button.scss
alizedebray: Borders are still missing when hovering over inactive button.

alizedebray (changes requested):

packages/styles/src/components/segmented-button.scss
leagrdv: I updated from color Highlight to SelectedItem as in Firefox, the "Highlight" color is a half transparent blue so it was hard to see. Here's what I get on my side with the new color: 
Screenshot 2025-02-11 151351

alizedebray (approved):

PR #4316, opened 2024-12-17
fix(styles): added hover color to ´post-accordion´ in HCM

schaertim:


.changeset/violet-geckos-shave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Added hover color to ´post-accordion´ in high contrast mode.
```
alizedebray, line 5:
```suggestion
Added hover color to `post-accordion` in high contrast mode.
```

alizedebray (approved):

PR #4319, opened 2024-12-17
fix(styles): HCM on form validation icons

leagrdv:


packages/styles/src/functions/_forms.scss
```diff
@@ -25,3 +30,39 @@
...
+  @return $state;
+}
+
+@function encode-color($color) {
+  @return utilities.replace($color, '#', '%23');
+}
+
+@function valid-icon($color) {
+  $encoded-color: encode-color($color);
+  $icon: utilities.replace(forms.$valid-icon, '%23107800', $encoded-color);
```
alizedebray, line 60: There is a function you can used for that in packages/styles/src/functions/_icons.scss. It is called `add-fill-color`.

alizedebray (changes requested):
The multiple select has a chevron icon that should not be there: 
image

packages/styles/src/functions/_forms.scss
leagrdv: The thing is my variable is not a svg but a `url('...')`. I tried removing the url() wrapper but I get an error in the console telling me it's invalid and missing a `;` and I don't get why. And if I try to use it as it in the fill function, it does not work.

alizedebray: Ah ok, then no worries I simply wrote that comment in case you didn't know

alizedebray (approved):

PR #4321, opened 2024-12-17
fix(styles): HCM card control checkbox text color

leagrdv:


oliverschuerch (approved):
Please, add a changeset!

PR #4325, opened 2024-12-18
fix(styles): fixed visibility of the arrow-icon in HCM

alionazherdetska:


oliverschuerch (changes requested):
Please add a changeset

packages/styles/src/components/form-select.scss
oliverschuerch, line 1: In dark hc-mode the arrows contrast to the background is really bad.
Default:
image
Hovered:
image

oliverschuerch (changes requested):

packages/styles/src/components/form-select.scss
oliverschuerch: I also recognized, that the text color of the select element does not change on hover (without hcm activated).
This could be a token issue, or we just did not update/use the hover-fg token there.
Please check and fix or contact @Vandapanda to fix it in the token system.

oliverschuerch (changes requested):

packages/styles/src/components/form-select.scss
alionazherdetska: image
Sandra confirmed that this is the intended behavior.

oliverschuerch (approved):

PR #4334, opened 2024-12-18
fix(styles): added missing hcm styles for radio-button and checkbox

schaertim:


alizedebray (approved):

alizedebray (changes requested):
The labels are not showing in Firefox.

packages/styles/src/components/checkbox.scss
```diff
@@ -61,6 +62,10 @@ $checkbox-icon-indeterminate: url("data:image/svg+xml,<svg width='16' height='16
 
     &:checked,
     &:indeterminate {
+      @include utility-mx.high-contrast-mode() {
+        background-color: SelectedItem;
+      }
+
       &::after {
         background-color: currentColor;
```
alizedebray, line 70: User can select any color for HCM, the indicator should use the `SelectedItemText` color to ensure it is always visible.
Same for the radio buttons.

alizedebray (changes requested):

alizedebray (approved):

PR #4337, opened 2024-12-18
chore(components): update the post-menu to be usable inside other com…

alizedebray:
…ponents

### DIFF
--- a/packages/components/src/utils/get-focusable-children.ts
+++ b/packages/components/src/utils/get-focusable-children.ts
@@ -23,6 +23,6 @@ const focusDisablingSelector = `:where(${[
   'details:not([open]) > *:not(details > summary:first-of-type) *',
 ].join(',')})`;
 
-export const isFocusable = (element: Element) => {
-  return element?.matches(focusableSelector) && !element?.matches(focusDisablingSelector);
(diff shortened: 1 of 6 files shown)


packages/components/src/utils/get-focusable-children.ts
```diff
@@ -23,6 +23,6 @@ const focusDisablingSelector = `:where(${[
   'details:not([open]) > *:not(details > summary:first-of-type) *',
 ].join(',')})`;
 
-export const isFocusable = (element: Element) => {
-  return element?.matches(focusableSelector) && !element?.matches(focusDisablingSelector);
+export const getFocusableChildren = (element: Element): NodeListOf<HTMLElement> => {
+  return element.querySelectorAll(`& > ${focusableSelector}:not(${focusDisablingSelector})`);
```
oliverschuerch, line 27: I think there is a problem in the `focusableSelector`, because `tabindex` with any negative value will make an element unfocusable.
Something like `:not([tabindex^="-"])` could maybe work.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

oliverschuerch (changes requested):
Not sure anymore, if we decided to add changesets for raw components to or not?

oliverschuerch (approved):

PR #4338, opened 2024-12-18
fix(styles): fixed HCM hover and focus styles for `post-avatar` component in anchor and button contexts

alionazherdetska:


alizedebray (changes requested):
Can you also update the border-color of the avatar when it is interactive?
Since there is only 1 to 2 characters, the difference in the text color is not very visible.

alizedebray (changes requested):
Please keep the default HCM colors as much as possible: 
- when in an anchor, the avatar has a LinkText color and border-color automatically set, it does not have to change color on hover
- when in a button, the  avatar has a ButtonText color and border-color changing to HighlightText on hover, also automatically set

packages/styles/src/components/avatar.scss
```diff
@@ -14,14 +15,27 @@ button {
...
+        border-color: Highlight !important;
+      }
 
       &::after {
         background-color: tokens.get('avatar-image-hover-opacity');
       }
     }
   }
+
+  @include utilities.focus-style;
```
alizedebray, line 40: 👍

packages/styles/src/components/avatar.scss
```diff
@@ -18,10 +19,24 @@ button {
...
+}
+
+button {
+  &:where(:has(post-avatar:only-child)):hover {
+    @include utilities.high-contrast-mode() {
+      post-avatar {
+        border-color: Highlight !important;
+        --post-avatar-fg: Highlight !important;
```
alizedebray, line 38:
```suggestion
        --post-avatar-bg: Highlight !important;
```

alizedebray (approved):

PR #4343, opened 2024-12-19
chore(styles, documentation): remove usage of deprecated utility sizing classes

leagrdv:


packages/components/cypress/fixtures/post-popover.test.html
```diff
@@ -4,6 +4,10 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Document</title>
+    <link
+      href="https://next.design-system.post.ch/styles/post-external.css"
```
oliverschuerch, line 8:
```suggestion
      href="../../node_modules/@swisspost/design-system-styles/post-external.css"
```

That's how we did it for the `post-card-control.form-association.test.html` fixture.
This way, we ensure, we're always using the current version of the file, not the one which was released with the latest release (maybe already a week ago).

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #4351, opened 2024-12-20
feat(documentation): separate icon sets and implement pagination

oliverschuerch:
Depends on https://github.com/swisspost/design-system/pull/4136
Merge #4136 first to main, then update this one and merge it afterwards.


packages/documentation/src/stories/foundations/icons/search/search-icons.styles.scss
```diff
@@ -4,113 +4,87 @@
...
   }
 
-  .icon--info {
-    flex: 1 0 auto;
-    display: flex;
-    flex-direction: column;
+  .result-list {
+    @include post.list-unstyled();
+    display: grid;
+    grid-template-columns: repeat(2, 120px);
```
leagrdv, line 40: In desktop, the space around the list elements is great as it makes the layout look more light but I feel like in mobile, it makes the list really long because it's on two columns (so 48 rows) but there is still a lot of space available around.

packages/documentation/src/stories/foundations/icons/search/search-icons.styles.scss
```diff
@@ -4,113 +4,87 @@
...
-      height: 60px;
+    @include post.min('lg') {
+      grid-template-columns: repeat(6, 140px);
+    }
+
+    @include post.min('xl') {
+      grid-template-columns: repeat(8, 140px);
+    }
+
+    .icon {
```
leagrdv, line 61: Icon width should either be a bit bigger or there should be a word break on the icon title because long names break out of the container.
Screenshot 2025-01-14 165041

leagrdv (changes requested):

packages/documentation/src/stories/foundations/icons/search/search-icons.styles.scss
oliverschuerch: The columns now use fractional values, so the columns always span across the whole available width.

packages/documentation/src/stories/foundations/icons/search/search-icons.styles.scss
oliverschuerch: The words will now be wrapped to the new line if necessary (with a dash, where the word break happens).

leagrdv (commented):
Approved

gfellerph (changes requested):
image

Seems like the new icons are rendered too bold.

Tested on Chrome and Edge, Firefox looks ok.

Firefox:
image

leagrdv (approved):

gfellerph (approved):

PR #4358, opened 2024-12-24
chore(components): remove usage of utility classes in post-header

leagrdv:


.changeset/breezy-poets-shout.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Removed usage of utility classes in `post-header` component.
```
alizedebray, line 5:
```suggestion
Removed the dependency on utility classes in `post-header` component. The header can now be used without importing CSS utility classes.
```

alizedebray (approved):

PR #4364, opened 2024-12-30
chore(components): update overflow layout header in tablet and mobile

leagrdv:


alizedebray (changes requested):
Can you add a `max-width` to the title so that it wraps as well? 
Screenshot 2024-12-31 at 14 57 03

And on bigger screens, I would probably only wrap the title not the buttons: 
Screenshot 2024-12-31 at 14 59 58

alizedebray (approved):

PR #4367, opened 2024-12-30
fix(components, documentation): z-index on header + fake content on header preview

leagrdv:
…eader preview


packages/documentation/src/utils/fake-content.ts
```diff
@@ -0,0 +1,7 @@
+import { html } from 'lit-html';
+
+export function fakeContent(paragraphs = 6) {
+  return html`<div class="container">
+    ${Array.from({ length: paragraphs }).map(() => html`<p class="fake-content"></p>`)}
```
alizedebray, line 5:
```suggestion
    ${Array.from({ length: paragraphs }).map(() => html`<p aria-hidden="true" class="fake-content"></p>`)}
```

alizedebray (changes requested):
Can you add a max-height to the header story so that it is scrollable?

alizedebray (commented):
The header's behavior is not correct in the story: it is sticky when you start scrolling but disappears afterwards. Can you please investigate this? 

The behavior is correct on the full screen preview.

alizedebray (approved):

PR #4368, opened 2024-12-31
fix(styles): update valid/invalid icons in form fields

leagrdv:


packages/styles/src/variables/components/_forms.scss
alizedebray, line 1: Isn't it possible to use the icons from `packages/styles/src/variables/_icons.scss`?

leagrdv: I didn't see they were defined there as well, PR has been updated with the changes!

PR #4369, opened 2024-12-31
chore(components): update crypto to nanoid

leagrdv:


packages/components/src/components/post-banner/post-banner.tsx
```diff
@@ -29,7 +30,7 @@ import { BANNER_TYPES, BannerType } from './banner-types';
 export class PostBanner {
   @Element() host: HTMLPostBannerElement;
 
-  @State() bannerId = crypto.randomUUID();
+  @State() bannerId = `p${nanoid(6)}`;
```
alizedebray, line 33: Why are you using `6`? Is it still safe to avoid collisions?

leagrdv: The ticket said to set it with at least 6 characters and I checked for the safety, it'd take 37k IDs for a 1% chance of collision.

.changeset/lucky-grapes-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Used nanoid instead of crypto to generate random IDs.
```
alizedebray, line 5:
```suggestion
Reduced the length of random IDs generated in the components; they are now generated using the [nanoid library](https://github.com/ai/nanoid) instead of the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API).
```

alizedebray (approved):

PR #4376, opened 2025-01-03
feat(components): update language switch behavior

myrta2302:


packages/components/src/components/post-language-switch/post-language-switch.tsx
```diff
@@ -111,11 +111,10 @@ export class PostLanguageSwitch {
```
leagrdv, line 111: Could you update the link to Figma for the component? I don't know if I put the wrong one in the first place or it changed.

packages/components/src/components/post-language-switch/post-language-switch.tsx
leagrdv, line 1: The keyboard navigation for the menu version is not working as expected.

packages/components/src/components/post-language-switch/post-language-switch.scss
```diff
@@ -8,12 +8,22 @@
...
 :host {
-  display: block;
+  display: flex;
 
   @include media.max(lg) {
     display: flex !important;
     gap: 0.5rem;
   }
+
+  > span {
```
leagrdv, line 18: The selected language should have a small border-radius.

packages/components/src/components/post-language-switch/post-language-switch.scss
leagrdv, line 1: There is a gap between the language options in mobile that are missing.

packages/components/src/components/post-language-switch/post-language-switch.tsx
```diff
@@ -125,15 +124,16 @@ export class PostLanguageSwitch {
...
+          <button class="post-language-switch-trigger" aria-label={this.description}>
             {this.activeLang.toUpperCase()}
             <post-icon aria-hidden="true" name="chevrondown"></post-icon>
           </button>
         </post-menu-trigger>
-        <post-menu id={this.menuId} class="post-language-switch-dropdown-container">
+        <post-menu
+          id={this.menuId}
+          class="post-language-switch-dropdown-container"
+          aria-label={this.caption}
```
leagrdv, line 135: Why did you move the caption from the trigger button to here?

packages/components/src/index.html
```diff
@@ -35,15 +35,15 @@
 
       <!-- Language switch -->
       <post-language-switch
-        caption="Caption"
-        description="Description"
+        caption="Change the language"
+        description="The currently selected language is English."
```
leagrdv, line 39: Description should be dynamic based on the selected language.

packages/components/src/components/post-language-switch/post-language-switch.scss
myrta2302: ok. Its also in the non selected elements so added it there too.

packages/components/src/components/post-language-switch/post-language-switch.tsx
myrta2302: There was a change to the foundation components. I ll create a separate issue for this.

packages/components/src/components/post-language-switch/post-language-switch.tsx
myrta2302: I see it working. Have not change something relevant to that.

packages/components/src/components/post-language-switch/post-language-switch.scss
myrta2302: I set the gap to 0.5rem for horizontal version and to no gap for dropdown version.

packages/components/src/components/post-language-switch/post-language-switch.tsx
myrta2302: This was suggested by Alize based on the post.ch current langugage switcher.

packages/components/src/index.html
myrta2302: The dynamic translation of the caption/description strings is to be implemented by the projects (usually with .json translation files).

packages/components/src/components/post-language-switch/post-language-switch.tsx
leagrdv: It's still not working on my side. When the dropdown is opened, you need to press on the arrow up to go down to the next selection, and the arrow down to go to the previous selection. In the header preview, it seems to be the opposite (which is the correct behaviour).
Also the selection should loop (pressing the arrow down on the last item should go back to the first one).

leagrdv (changes requested):
The keyboard navigation still isn't working. When the dropdown is opened, you need to press on the arrow up to go down to the next selection, and the arrow down to go to the previous selection which should be the opposite. In the header preview, it seems to be the opposite (which is the correct behaviour) but on click on the dropdown trigger, the focus is not on the first element (which it should be).

packages/components/src/components/post-language-switch/post-language-switch.tsx
```diff
@@ -62,18 +62,24 @@ export class PostLanguageSwitch {
...
   private updateChildrenVariant() {
     this.host.querySelectorAll('post-language-option').forEach(el => {
-      el.setAttribute('variant', this.variant);
+      const isActive = el.getAttribute('active') === 'false';
+      if (this.variant === 'menu' && isActive) {
+        el.setAttribute('variant', this.variant);
+      } else {
+        el.removeAttribute('variant');
+        if (this.variant !== 'menu') el.setAttribute('variant', this.variant);
+      }
```
leagrdv, line 82:
```suggestion
      if (this.variant === 'list' || (this.variant === 'menu' && isActive)) {
        el.setAttribute('variant', this.variant);
      } else {
        el.removeAttribute('variant');
      }
```

leagrdv: Also, could you write a short comment on top of that code explaining you're doing that for accessibility reasons and why?

leagrdv (changes requested):

leagrdv (commented):
Approved

packages/components/src/components/post-language-option/post-language-option.scss
```diff
@@ -20,30 +20,32 @@ post-language-option {
...
     border-radius: 2px;
     @include utilities-mx.focus-style;
   }
+
+  &:has([aria-current='true']),
+  &:has([aria-current='page']) {
+    :where([variant='menu']) & {
+      display: none;
+    }
+  }
```
alizedebray, line 32: If the aria-current option is hidden for both the menu and list variant, then it is more readable to add the `display: none` only once.

```suggestion
  &:has([aria-current]):not([aria-current='false']) {
    display: none;
  }
```

packages/components/src/index.html
alizedebray, line 1: Can you add this changes to the header stories as well?

packages/components/src/utils/get-focusable-children.ts
```diff
@@ -22,6 +22,7 @@ const focusDisablingSelector = `:where(${[
   'details:not([open]) > *:not(details > summary:first-of-type)',
   'details:not([open]) > *:not(details > summary:first-of-type) *',
   '[tabindex^="-"]',
+  '[role=menuitem]:has([aria-current="true"]) *',
```
alizedebray, line 25: This function is an helper therefore it should remain generic. Can you update the function to check if the element is visible or not using the `getComputedStyles` function for example?

packages/components/src/components/post-language-switch/post-language-switch.tsx
```diff
@@ -111,11 +110,10 @@ export class PostLanguageSwitch {
...
     return (
-      <Host
-        data-version={version}
-        aria-label={`${this.caption}, ${this.description}`}
-        aria-role="list"
-      >
+      <Host data-version={version} aria-role="list">
+        <span aria-label={`${this.caption}, ${this.description}`}>
+          {this.activeLang.toUpperCase()}
+        </span>
```
alizedebray, line 116: This behavior does not match with the behavior on post.ch, the caption should at least be on the host.

alizedebray (changes requested):

packages/components/src/components/post-language-option/post-language-option.scss
```diff
@@ -20,30 +20,29 @@ post-language-option {
...
-      background-color: #050400;
-      color: #fff;
-
-      &:hover {
-        background-color: #504f4b;
-      }
-    }
+  &:has([aria-current='true']),
+  &:has([aria-current='page']) {
+    display: none;
```
alizedebray, line 45: This is not necessary anymore as it duplicates line 27 to 29

packages/components/src/components/post-menu/post-menu.tsx
```diff
@@ -198,6 +198,8 @@ export class PostMenu {
         .filter(el => el.getAttribute('role') === 'menuitem')
         // For each menu item, get any focusable children (e.g., buttons, links)
         .flatMap(el => Array.from(getFocusableChildren(el)))
+        // Filter elements whose parent is hidden
+        .filter(el => window.getComputedStyle(el.parentElement).display !== 'none')
```
alizedebray, line 202: It would be better to filter these elements within the `getFocusableChildren` helper as they are not actually focusable.

packages/components/src/components/post-language-switch/post-language-switch.tsx
```diff
@@ -111,11 +110,8 @@ export class PostLanguageSwitch {
 
   private renderList() {
     return (
-      <Host
-        data-version={version}
-        aria-label={`${this.caption}, ${this.description}`}
-        aria-role="list"
-      >
+      <Host data-version={version} aria-role="list" aria-label={this.caption}>
+        <span aria-label={this.description}>{this.activeLang.toUpperCase()}</span>
```
alizedebray, line 114:
```suggestion
        <span aria-label={this.description} role="listitem">{this.activeLang.toUpperCase()}</span>
```

alizedebray (changes requested):

packages/components/src/utils/get-focusable-children.ts
```diff
@@ -25,5 +25,14 @@ const focusDisablingSelector = `:where(${[
...
+  const focusableChildren = element.querySelectorAll(
+    `${focusableSelector}:not(${focusDisablingSelector})`,
+  );
+
+  const visibleFocusableChildren = Array.from(focusableChildren).filter(child => {
+    const style = window.getComputedStyle(child.parentElement);
+    return style.display !== 'none';
+  });
+
+  return visibleFocusableChildren as unknown as NodeListOf<HTMLElement>;
```
alizedebray, line 37: This typing is incorrect, you're turning the node list into an array on line 32 so the return type is an array of HTML elements (`HTMLElement[]`)

packages/components/src/utils/get-focusable-children.ts
```diff
@@ -25,5 +25,14 @@ const focusDisablingSelector = `:where(${[
...
 
 export function getFocusableChildren(element: Element): NodeListOf<HTMLElement> {
-  return element.querySelectorAll(`& > ${focusableSelector}:not(${focusDisablingSelector})`);
+  const focusableChildren = element.querySelectorAll(
+    `${focusableSelector}:not(${focusDisablingSelector})`,
+  );
+
+  const visibleFocusableChildren = Array.from(focusableChildren).filter(child => {
+    const style = window.getComputedStyle(child.parentElement);
+    return style.display !== 'none';
```
alizedebray, line 34:
```suggestion
    return style.display !== 'none' && style.visibility !== 'hidden';
```

packages/components/src/utils/get-focusable-children.ts
```diff
@@ -25,5 +25,14 @@ const focusDisablingSelector = `:where(${[
 ].join(',')})`;
```
alizedebray, line 25: To make sure all hidden cases are taken into account:
```suggestion
  '[hidden]:not([hidden="false"])',
].join(',')})`;
```

alizedebray (changes requested):

packages/components/src/utils/get-focusable-children.ts
```diff
@@ -22,8 +22,18 @@ const focusDisablingSelector = `:where(${[
...
+  const focusableChildren = element.querySelectorAll(
+    `${focusableSelector}:not(${focusDisablingSelector})`,
+  );
+
+  const visibleFocusableChildren = Array.from(focusableChildren).filter(child => {
+    const style = window.getComputedStyle(child.parentElement);
+    return style.display !== 'none' && style.visibility !== 'hidden';
+  });
+
+  return visibleFocusableChildren as HTMLElement[];
```
alizedebray, line 38: With `as` you are forcing the type, this is not a good practice and should be avoided as much as possible:
```suggestion
  return visibleFocusableChildren;
```

alizedebray (approved):

alizedebray (approved):

leagrdv (approved):

PR #4378, opened 2025-01-03
feat(styles): implement gutter classes

myrta2302:


packages/documentation/cypress/snapshots/utilities/gutters.snapshot.ts
```diff
@@ -0,0 +1,7 @@
+describe('Gutters', () => {
+  it('Gutters', () => {
+    cy.visit('/iframe.html?id=snapshots--gutters');
+    cy.get('.snapshot', { timeout: 30000 }).should('be.visible');
```
alizedebray, line 4: The `cy.get` will fail because the is no element with a `.snapshot` class on the snapshot page: https://preview-4378--swisspost-design-system-next.netlify.app/iframe.html?id=snapshots--gutters&viewMode=story...

alizedebray (changes requested):

PR #4379, opened 2025-01-03
fix(focus): fix focus order in `post-megadropdown` component

alionazherdetska:


packages/components/src/components/post-megadropdown/post-megadropdown.tsx
```diff
@@ -88,6 +95,14 @@ export class PostMegadropdown {
     }
   }
 
+  private getPostListItems(): HTMLElement[] {
```
leagrdv, line 98: It might be that the first focusable element of the mega dropdown is not the first post-list-item but the post-list title as it could be a link as well.

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
```diff
@@ -88,6 +95,14 @@ export class PostMegadropdown {
...
+    const postListItems = Array.from(this.host.querySelectorAll('post-list-item'));
+    const focusableChildren = postListItems.flatMap(el => Array.from(getFocusableChildren(el)));
+
+    return focusableChildren;
```
leagrdv, line 102:
```suggestion

    return postListItems.flatMap(el => Array.from(getFocusableChildren(el)));
```

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
```diff
@@ -98,16 +113,16 @@ export class PostMegadropdown {
           ref={el => (this.popoverRef = el)}
         >
           <div class="megadropdown" onFocusout={e => this.handleFocusout(e)}>
+            <slot name="megadropdown-title"></slot>
```
leagrdv, line 116: This works in desktop but not in mobile, as it puts the back button after the elements.

leagrdv (changes requested):

leagrdv (changes requested):
I have a weird glitch on the top bar when I go back and forth between main navigation and mega dropdown, could you investigate on that?
https://github.com/user-attachments/assets/ce38f04c-71f5-4380-aed9-219c267b55d4

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
```diff
@@ -166,6 +172,16 @@ export class PostMegadropdown {
     }
   }
 
+  private handleFocusOutside = (event: FocusEvent) => {
```
leagrdv, line 175: Detection of focus outside of the megadropdown doesn't work in the header module. When using tab and opening the "Briefe" megadropdown, it does close correctly after the close button focus, but when doing the same for the "Pakete" element, then the megadropdown never closes when focus is outside.

leagrdv (changes requested):

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
leagrdv: Discussed with Alona, it's because there is no more focusable element after.

leagrdv (commented):
Approved

leagrdv (approved):

PR #4391, opened 2025-01-07
chore(components): fixed typo in `post-accordion-item`

schaertim:


schaertim:
Not sure if a changeset is needed here...

leagrdv:
> Not sure if a changeset is needed here...

Because of the typo, I'm guessing the heading level was not working before so it should be a patch on the components package.

leagrdv (changes requested):

leagrdv (commented):
Approved

oliverschuerch (approved):

PR #4395, opened 2025-01-07
fix(components): update mobile menu behaviour

leagrdv:


packages/components/src/components/post-header/post-header.tsx
```diff
@@ -57,6 +86,47 @@ export class PostHeader {
...
+
+    this.firstFocusableEl = focusableChildren[0];
+    this.lastFocusableEl = focusableChildren[focusableChildren.length - 1];
+  }
+
+  private keyboardHandler(e: KeyboardEvent) {
+    if (e.key === 'Tab') {
+      if (e.shiftKey && document.activeElement === this.firstFocusableEl) {
+        // If back tab (Tab + Shift) and first element is focused, focus goes to the last element of the megadropdown
+        e.preventDefault();
```
gfellerph, line 120: This should only be done if the menu is open, otherwise it's impossible to go back to the logo and skiplinks:

image

Pressing [shift] + [tab] in this situation locks focus to the menu button.

gfellerph (changes requested):

gfellerph (approved):

PR #4396, opened 2025-01-08
chore(header): changed path matching for`active-route` prop to also work for relative paths

schaertim:


packages/internet-header/src/services/route.service.spec.ts
```diff
@@ -50,46 +51,47 @@ describe('route.service.ts', () => {
...
 
   describe('compareRoutes', () => {
-    const post = new URL('https://post.ch');
-    const letters = new URL('https://post.ch/briefe');
-    const deep = new URL('https://post.ch/briefe/inland');
-    const search = new URL('https://post.ch/briefe?q=search');
-    const hash = new URL('https://post.ch/briefe#hash');
-    const nope = new URL('https://post.de/briefe');
-    const upper = new URL('https://post.ch/Briefe');
+    const urls = {
```
leagrdv, line 54: It would be nice to have some tests for relative path URLs as well

packages/internet-header/src/services/route.service.spec.ts
```diff
@@ -50,46 +50,61 @@ describe('route.service.ts', () => {
...
       // @ts-expect-error
       expect(compareRoutes(null, undefined, 'auto')).toBe(0);
       // @ts-expect-error
-      expect(compareRoutes(post, nope, null)).toBe(0);
+      expect(compareRoutes(urls.post, urls.nope, null)).toBe(0);
+    });
+
+    it('handles relative and absolute paths the same', () => {
+      const relativeUrl = new URL('/briefe', 'https://post.ch');
+      const faultyRelativeUrl = new URL('/briefee', 'https://post.ch');
```
leagrdv, line 98: You're not actually testing with a relative path here, you're creating an absolute path before actually comparing the routes

leagrdv (changes requested):

packages/internet-header/src/services/route.service.spec.ts
schaertim: If I want to test the compareRoutes function, we need to use the URL constructor since it only accepts URL objects, not strings. While this means we need a valid base path when constructing test URLs, this doesn't limit our testing capability - relative URLs are transformed into absolute ones in the markActiveRoute function anyway using `new URL(activeRouteProp, window.location.href)`. The `isRelativeReference` flag in compareRoutes then ensures proper relative path comparison by ignoring origins when set to true.

leagrdv:
Approved

oliverschuerch (changes requested):
Please ask @gfellerph  if he want us to implement this change in v8, v9 or in both versions?

PR #4400, opened 2025-01-09
chore(components): set header z-index

myrta2302:


myrta2302:
This only ensures that the header is stacked on top of other elements - including the popover (trigger). The popovercontainer issue still remains (since it belongs to the #toplayer and is stacked above everything). This probably needs to be resolved with setting the post-header as a boundary via the floating ui API.

PR #4403, opened 2025-01-09
feat(documentation): styling shadowdom parts

myrta2302:


packages/documentation/src/stories/guidelines/shadowdom-parts/shadowdom-parts.docs.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+# Styling Shadowdom Parts
+
+## Parts
+
+<div className="lead">
+  The `part` attribute allows a shadow tree element to be exposed to its parent DOM.
+</div>
+
+To style an HTML shadow element, first set the `part` attribute on the element and assign it a name. Then, use the `::part` selector in your CSS to apply styles, referencing the name you specified in the `part` attribute. This allows you to style specific parts of a shadow DOM element from outside the shadow tree.
```
oliverschuerch, line 18: This sounds like an explanation for us. Our users will not be able to add a `part` attribute to any shadowdom element in our components. I would like to rephrase this text and explain how to check in our documentation if there are `part` attributes available in a component and how to add custom styles for it.
Maybe also explain the boundaries (which elements can be styled with the `part` attribute and which not, e.g. it only allows to style the element which contains the `part` attribute, not its children).

> [Pseudo-classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes) (such as ::part(label):hover) can be appended to the ::part() selector, but [structural pseudo-classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes#tree-structural_pseudo-classes) that match based on tree information (such as :empty), rather than local element information (such as :last-child), cannot be appended.

> Additional pseudo-elements, such as ::before, can be appended to the ::part() selector, but additional ::part() element can't be appended. For example, ::part(confirm-button)::part(active) never matches anything, i.e, it is not the same as ::part(confirm-button active). This is because doing so would expose more structural information than is intended.

Source: https://developer.mozilla.org/en-US/docs/Web/CSS/::part

oliverschuerch (changes requested):

packages/documentation/src/stories/guidelines/shadowdom-parts/shadowdom-parts.docs.mdx
myrta2302: Updated. I had totally missunderstood the POV.

oliverschuerch (approved):

PR #4421, opened 2025-01-13
fix(component): remove `post-popovercontainer` from `post-megadropdown`

alionazherdetska:


gfellerph (changes requested):
image

I don't see any icons in the preview, possibly related to another issue since the base URL seems to be wrong. Can you update from main to make sure before investigating the issue?

gfellerph (approved):

Issue #3278, opened 2024-07-16
[component]: Popover


Cian77:
### Open issues:
**Design:**
- [ ] New font
- [ ] Spacing

**Tokens:**
- [x] Text

### Design

https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations-%26-Components-Next-Level?node-id=21-173

### Tokens

- [ ] Tokens for this component are ready

### Tasks

```[tasklist]
### 💻 Tasks
- [ ] Define more generic tokens
- [ ] Update styles with tokens
- [ ] Documentation in Storybook
- [ ] End to end tests
- [ ] Visual regression tests
- [ ] Design review
```

gfellerph:
There is an existing PR, but the tokens need re-specification for this component (some of them are way too specific) and the component styles in the existing PR require a re-work.

gfellerph:
@Vandapanda content does not need to be tokenised, just padding, border-radius, etc.

PR #4436, opened 2025-01-14
chore(components): added optional safe space to `post-popovercontainer`

schaertim:


packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -69,13 +62,30 @@ export class PostPopovercontainer {
...
+    if (!this.safeSpace || !this.host.matches(':where(:popover-open, .popover-open)')) return;
+
+    this.host.style.setProperty('--safe-space-cursor-x', `${event.clientX}px`);
+    this.host.style.setProperty('--safe-space-cursor-y', `${event.clientY}px`);
+  };
+
   componentDidLoad() {
     this.host.setAttribute('popover', '');
     this.host.addEventListener('beforetoggle', this.handleToggle.bind(this));
+    window.addEventListener('mousemove', this.mouseTrackingHandler);
```
gfellerph, line 81: Only add this listener if there's a `safe-space="triangle"` requested, otherwise it's there for nothing. Ideally the event listener is being added on open and removed on close.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -69,13 +62,30 @@ export class PostPopovercontainer {
    */
   @Prop() readonly arrow?: boolean = false;
 
+  /**
+   * Enables a safespace through which the cursor can be moved without the popover being disabled
+   */
+  @Prop() readonly safeSpace?: 'triangle' | 'trapezoid';
```
gfellerph, line 68: If you use [reflection](https://stenciljs.com/docs/properties#reflect-properties-values-to-attributes-reflect), you can use the prop as a selector in CSS and don't have to write a data-attribute for it, duplicating the setting.

```suggestion
  @Prop({ reflect: true }) readonly safeSpace?: 'triangle' | 'trapezoid';
```

packages/components/src/components/post-popovercontainer/post-popovercontainer.scss
```diff
@@ -41,6 +41,39 @@
...
+  &[data-safe-space]::after {
+    content: '';
+    position: fixed;
+    inset: 0;
+    z-index: calc(#{commons.$zindex-popover} - 1);
+    pointer-events: auto;
+    background: transparent;
+  }
+
+  &[data-safe-space='triangle']::after {
```
gfellerph, line 54: See comment about reflection.

```suggestion
  &[safe-space='triangle']::after {
```

packages/components/src/components/post-popovercontainer/post-popovercontainer.scss
```diff
@@ -41,6 +41,39 @@
...
+    clip-path: polygon(
+      var(--safe-space-cursor-x, 0) var(--safe-space-cursor-y, 0),
+      var(--safe-space-popover-x-start, var(--safe-space-popover-x, 0))
+        var(--safe-space-popover-y, var(--safe-space-popover-y-start, 0)),
+      var(--safe-space-popover-x-end, var(--safe-space-popover-x, 0))
+        var(--safe-space-popover-y, var(--safe-space-popover-y-end, 0))
+    );
+  }
+
+  &[data-safe-space='trapezoid']::after {
```
gfellerph, line 64:
```suggestion
  &[safe-space='trapezoid']::after {
```

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -169,51 +209,103 @@ export class PostPopovercontainer {
...
+      '--safe-space-trigger-y',
+      '--safe-space-trigger-x-start',
+      '--safe-space-trigger-x-end',
+      '--safe-space-trigger-y-start',
+      '--safe-space-trigger-y-end',
+    ];
+
+    propertiesToClear.forEach(prop => {
+      this.host.style.removeProperty(prop);
+    });
```
gfellerph, line 283: Why is this necessary?

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -169,51 +209,103 @@ export class PostPopovercontainer {
...
+      this.host.style.setProperty('--safe-space-trigger-x', `${posData.trigger.x}px`);
+      this.host.style.setProperty('--safe-space-trigger-y-start', `${posData.trigger.yStart}px`);
+      this.host.style.setProperty('--safe-space-trigger-y-end', `${posData.trigger.yEnd}px`);
     }
   }
 
   render() {
     return (
-      <Host data-version={version}>
+      <Host data-version={version} data-safe-space={this.safeSpace}>
```
gfellerph, line 308: See the comment about reflection on line 68.

```suggestion
      <Host data-version={version}>
```

gfellerph (changes requested):
Nice work!

packages/components/src/components/post-popovercontainer/post-popovercontainer.scss
```diff
@@ -41,6 +41,39 @@
...
   overflow: visible;
 
+  // Safe space overlay styles
+  &[data-safe-space]::after {
+    content: '';
+    position: fixed;
+    inset: 0;
+    z-index: calc(#{commons.$zindex-popover} - 1);
+    pointer-events: auto;
+    background: transparent;
```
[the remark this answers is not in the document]
schaertim: The background is not necessary but the pointer-events is since its not working without it.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
schaertim: Not all positions use the same set of properties. When the position changes and stale values are not cleared they can interfere with the new placement.

gfellerph (approved):
I'll approve to unblock, but please also get Oli's review on this @schaertim.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
oliverschuerch: I understand the issue and I'm fine with it.
The `propertiesToClear` map could be declared outside of this function (e.g. as private static map directly in the class or as a static variable outside of the class). Just implement it the same way as the `staticSide` map.

packages/components/src/components/post-popovercontainer/post-popovercontainer.scss
```diff
@@ -20,49 +20,83 @@
...
   padding: 0;
+  // Keeps the little arrow visible
+  overflow: visible;
+
+  > div {
+    box-shadow: elevation.$elevation-300;
+    width: max-content;
+    color: var(--post-current-palette-fg);
+    background-color: var(--post-current-palette-bg);
+    border: 2px solid transparent; // Keeping the default border for HCM
```
oliverschuerch, line 37: Sad we don't have any variables/tokens for that. 😞

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -70,16 +85,35 @@ export class PostPopovercontainer {
...
+   * Enables a safespace through which the cursor can be moved without the popover being disabled
+   */
+  @Prop({ reflect: true }) readonly safeSpace?: 'triangle' | 'trapezoid';
+
+  /**
+   * Updates cursor position for safe space feature when popover is open.
+   * Sets CSS custom properties for dynamic styling of safe area.
+   * @param event MouseEvent with cursor position
+   */
+  private mouseTrackingHandler = (event: MouseEvent) => {
```
oliverschuerch, line 102: Please use the same function definition styles as used below.
```suggestion
  private mouseTrackingHandler(event: MouseEvent) {
```

oliverschuerch (changes requested):

packages/components/src/components/post-popovercontainer/post-popovercontainer.scss
schaertim: There is a popover ticket in "ready for token spec" ([#3278](https://github.com/swisspost/design-system/issues/3278)). This can hopefully be done there.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -70,16 +85,35 @@ export class PostPopovercontainer {
...
+   */
+  @Prop({ reflect: true }) readonly safeSpace?: 'triangle' | 'trapezoid';
+
+  /**
+   * Updates cursor position for safe space feature when popover is open.
+   * Sets CSS custom properties for dynamic styling of safe area.
+   * @param event MouseEvent with cursor position
+   */
+  private mouseTrackingHandler(event: MouseEvent) {
+    if (!this.safeSpace || !this.host.matches(':where(:popover-open, .popover-open)')) return;
```
oliverschuerch, line 103: What do you think couldn't we use this guard in the `handleToggle` function already and just not bind the event listener? This would avoid a lot of unnecessary event bindings/removes.

```js
private handleToggle(e: ToggleEvent) {
    this.toggleTimeoutId = window.setTimeout(() => (this.toggleTimeoutId = null), 10);
    const trackMouse = this.safeSpace && this.host.matches(':where(:popover-open, .popover-open)');
    const isOpen = e.newState === 'open';

    if (trackMouse) {
      if (isOpen) {
        window.addEventListener('mousemove', this.mouseTrackingHandler);
        this.startAutoupdates();
      } else {
        window.removeEventListener('mousemove', this.mouseTrackingHandler);
        if (typeof this.clearAutoUpdate === 'function') this.clearAutoUpdate();
      }
    }
    this.postToggle.emit(isOpen);
	@@ -153,15 +189,40 @@ export class PostPopovercontainer {
  }
```

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #4452, opened 2025-01-16
chore(documentation): add docs for responsive design

leagrdv:


oliverschuerch (changes requested):
Would be nice if we could integrate the browser icons as svg files, like all the others.

oliverschuerch (approved):

PR #4453, opened 2025-01-16
chore(icons): automatically add a link to the health page on icon update

myrta2302:


myrta2302:
Not sure if this needs a changeset.

.github/actions/preview/message/update/action.yaml
```diff
@@ -11,6 +11,9 @@ inputs:
   preview-url:
     description: The preview url to add in the comment.
     required: true
+  issue-name:
+    description: The PR title to check if it involves an 'icons update'.
+    required: true
```
oliverschuerch, line 16: This won't work, if you do not add the `issue-name` property in all of our workflows, who use this action.

oliverschuerch (changes requested):
Let's discuss the whole topic again ;)

oliverschuerch (approved):

PR #4467, opened 2025-01-20
feat: update font v9

gfellerph:
Updates to the new font.


packages/styles/src/variables/_type.scss
```diff
@@ -58,9 +58,9 @@ $line-height-sm: 1;
 $line-height-copy: 1.5;
 $line-height-heading: 1.2;
 
-$font-weight-light: 300 !default; // Design System -> Frutiger Neue Light
-$font-weight-normal: 400 !default; // Design System -> Frutiger Neue Roman
-$font-weight-bold: 700 !default; // Design System -> Frutiger Neue Bold
+$font-weight-light: 300 !default;
+$font-weight-normal: 400 !default;
+$font-weight-bold: 700 !default;
```
oliverschuerch, line 63: What about `$font-weight-black`?
And since we remove the light font variant, we should also remove the variable `$font-weight-light`.

oliverschuerch (changes requested):
Why do we need to add the new font files in the styles and the docs packages, if we're always loading/using them from the cdn?

gfellerph:
> Why do we need to add the new font files in the styles and the docs packages, if we're always loading/using them from the cdn?

That was by mistake, removed them again.

oliverschuerch (commented):
Why all these changes in the jest files?

oliverschuerch (approved):

gfellerph:
> Why all these changes in the jest files?

I removed the forwarded options variable that is being checked in the components tests. This is no longer working, therefore I removed the check in the tests.

PR #4468, opened 2025-01-20
feat: update font v8

gfellerph:
Update fonts for v8.

### DIFF
--- a/packages/documentation/src/stories/foundation/typography/typography-standalone.sample.css
+++ b/packages/documentation/src/stories/foundation/typography/typography-standalone.sample.css
@@ -1,14 +1,6 @@
 @font-face {
   font-display: swap;
   font-family: 'Swiss Post Sans';
-  font-weight: 900;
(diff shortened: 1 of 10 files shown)


packages/icons/.gitignore
oliverschuerch, line 1: This should probably not be part of this PR.

packages/documentation/src/stories/foundation/typography/typography-standalone.sample.css
oliverschuerch, line 1: Remove visualisation of font version `black` (font-weight: 900)

packages/internet-header/package.json
oliverschuerch, line 1: Why this needs to be part of this PR?

packages/styles/src/components/fonts.scss
oliverschuerch, line 1: Remove the integration of the font variant `black` (on v8 only)

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #4451, opened 2025-01-16
chore: update eslint to v9 with flat config files

oliverschuerch:

PR #4482, opened 2025-01-21
feat(eslint): create ESLint Plugin

alizedebray:


packages/eslint-plugin/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@swisspost/design-system-eslint-plugin",
```
gfellerph, line 2: I would propose to call the package 
- `@swisspost/design-system-eslint-migrations` 
- `@swisspost/design-system-eslint-plugin-migrations`
- `@swisspost/eslint-plugin-design-system-migrations`

This would better communicate the purpose of the package and prevent naming conflicts if for example we decide to have a default ruleset eslint plugin shared with other projects. What do you think?

packages/eslint-plugin/src/rules/no-deprecated-btn-rg.ts
```diff
@@ -0,0 +1,38 @@
...
+        'The "btn-rg" class is deprecated. Please remove it or replace it with "btn-sm".',
+    },
+    type: 'suggestion',
+    schema: [],
+  },
+  defaultOptions: [],
+  create(context) {
+    return {
+      tag(node: TemplateNode) {
+        if (node.name === 'button') {
```
gfellerph, line 26: Button can be put on multiple tag names.

```suggestion
        if (['button', 'input', 'a'].includes(node.name)) {
```

packages/eslint-plugin/src/rules/no-deprecated-btn-rg.ts
```diff
@@ -0,0 +1,38 @@
...
+          const $node = node.toCheerio();
+          if ($node.hasClass('btn-rg')) {
+            context.report({
```
gfellerph, line 29: Is it possible to auto-fix instead of just report?

packages/eslint-plugin/tsconfig.eslint.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    // ensure that nobody can accidentally use this config for a build
```
gfellerph, line 4: https://github.com/swisspost/design-system/pull/4451#pullrequestreview-2567397264

gfellerph (changes requested):
Really cool work!

packages/eslint-plugin/package.json
alizedebray: We build the package for migration purposes but it is now ready to export any eslint config or plugin. This is the biggest advantage of the ESLint flat config!
Therefore, I would simply call the package `@swisspost/design-system-eslint`.

Then we could rethink the name of the plugin itself: https://github.com/swisspost/design-system/blob/077bb70019ab0d427543eec41d9cd6ed3020a7fa/packages/eslint-plugin/src/index.ts#L13
But do you think it's an issue to have the rules installed in a project at all times? For example, the rule I created notifies you when you use the `.btn-rg` class, this might be useful even outside of a migration phase, don't you think?

packages/eslint-plugin/src/rules/no-deprecated-btn-rg.ts
alizedebray: Sure! But for this rule I am unsure what to do: replace the class with `.btn-sm`? remove the class altogether? I therefore removed the fixer to let project decide case by case. I can add it back if you have a preferred solution.

gfellerph: I think it would be useful to have fixes for most of the things. Even if it does not map 100%, if we can handle 80% of the cases automatically, that will be a huge boost compared to have to do 100% of the changes manually.

For this specific case, `btn-sm` seems to be closer to what `btn-rg` was (40px on Desktop), therefore I'd probably use this instead of removing it, which would fall back to medium.

packages/eslint/CONTRIBUTING.md
```diff
@@ -0,0 +1,121 @@
...
+Please note that these guidelines extend the [general contribution guidelines](../../CONTRIBUTING.md) from the root of the repository.
+
+## Getting Started
+
+### Build the Plugin
+
+To build the ESLint plugin for production, run:
+
+```bash
+pnpm rules:build
```
oliverschuerch, line 14:
```suggestion
pnpm eslint:build
```

packages/eslint/CONTRIBUTING.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+### Run Unit Tests
+
+To run the unit tests for the ESLint plugin, use the following command:
+
+```bash
+pnpm rules:test
```
oliverschuerch, line 22:
```suggestion
pnpm eslint:test
```

packages/eslint/src/configs/template.ts
```diff
@@ -0,0 +1,23 @@
...
+  name: 'post/template',
+  files: ['**/*.html'],
+  languageOptions: {
+    parser: templateParser,
+  },
+  plugins: {
+    'post/template': plugin,
+  },
+  rules: templateRules,
+});
```
oliverschuerch, line 23: It could be a good idea to split this a little bit up and implement different things in different config objects.
At least this is done by most the flat configs I saw (for example from typescript-eslint).

Almost all of them implemented at least two objects, one for the base config, containing `languageOptions`, `plugins` and other configurations and one to add the `files` filter and the `rules`.

```js
export default (plugin: FlatConfig.Plugin): FlatConfig.Config => ({
  name: 'post/design-system/base',
  languageOptions: {
    parser: templateParser,
  },
  plugins: {
    'post/template': plugin,
  },
},
{
  name: 'post/design-system/recommended',
  files: ['**/*.html'],
  rules: templateRules,
});
```

Let me know, what you think about this idea.

packages/eslint/CONTRIBUTING.md
```diff
@@ -0,0 +1,121 @@
...
+- **name**: The name of the rule (e.g. `no-foo`).
+- **meta**: An object containing all necessary information about the rule.
+- **defaultOptions**: Optional default values for any configuration options for the rule.
+- **create**: The function where you implement the rule logic. A detailed guide on writing custom rules is available in the ESLint Custom Rule Tutorial.
+
+_More details about the structure can be found in the [ESLint documentation](https://eslint.org/docs/latest/extend/custom-rules#rule-structure)._
+
+The **meta** object also contains additional details about the rule, including its category.
+The category determines which ESLint configuration the rule belongs to.
+Currently, there is only the `template` category, which is used for HTML-related rules and is included in the `plugin:@design-system/template` configuration.
```
oliverschuerch, line 68: Is the mentioned plugin name still correct?

oliverschuerch (changes requested):
Do we want projects to include the eslint configs defined in this project into there own eslint.config.js?

If "yes", we should make sure the configs can be imported like so (not sure if this is already the case):
```js
import postEslint from '@swisspost/design-system-eslint';

export defaults [
  ...postEslint.configs.recommended
];
```

packages/eslint-plugin/package.json
oliverschuerch: @alizedebray Most of the packages I saw, are names like this `@namespace/eslint-plugin`.
But I really support to not use `migration` in it's name!

packages/eslint/src/configs/template.ts
alizedebray: Yes, this is the idea. I only implemented the template config for now but we can definitely add more.

What differs from typescript-eslint is that we will have multiple parsers: one for HTML, one for TypeScript and one for Sass. Then, the config object cannot really be split as the template parser only applies to HTML files and HTML files can only be parsed by the template parser.

alizedebray: angular-eslint is a more appropriate example for us as they use different parsers as well. Here are their configs: https://github.com/angular-eslint/angular-eslint/tree/main/packages/angular-eslint/src/configs

alizedebray: Do you mean using a `ConfigArray`? 
If so we can definitely align with other plugins, do you know what are the benefits of the array over the object? It seems more difficult to override a rule with an array...

packages/eslint-plugin/package.json
alizedebray: The `@namespace/eslint-plugin` packages are usually used for classic config, the flat config packages are for example `typescript-eslint` and `angular-eslint`. We could use `swisspost-eslint`, but I have a preference for `@swisspost/design-system-eslint` to align with our naming convention.

alizedebray:
> Do we want projects to include the eslint configs defined in this project into there own eslint.config.js?

It depends on the outcome of [this discussion](https://github.com/swisspost/design-system/pull/4482#discussion_r1926788587).

> If "yes", we shoule make sure the configs can be imported like so (not sure if this is already the case):

Currently project can import the config as explained here in the [README.md](https://github.com/swisspost/design-system/blob/0d1fce9aad3bf53376243297926bba54ef26aff4/packages/eslint/README.md#configuration) (there is no spread operator because the package currently exports a config not a config array).

packages/eslint-plugin/package.json
alizedebray: @gfellerph grouping the rules to a specific migration might not be necessary as the rules will simply differs from one version of the package to the other: `@swisspost/design-system-eslint@8` exports rules to migrate from v7 to v8, but `@swisspost/design-system-eslint@9` exports rule to migrate from v8 to v9.

oliverschuerch: @alizedebray This would mean, we would remove existing rules, as soon as we start working on a next version of this package!?

The more I think about it, the more this makes sense!
Because else, we would still provide rules to migrate from v10 to v11, which have been created to migrate from v8 to v9.
And this, would not make any sense and could even produce wrong output.

packages/eslint/src/configs/template.ts
oliverschuerch: Since every `ConfigObject` ends up in a `ConfigArray` anyway and acts as an override for the previous `ConfigObject`, this should be save and is (as explained) exaclty what `typescript-eslint`, `angular-eslint` and other packages are doing already. Also splitting up the `files` filter and the `parser` config is not an issue, as following `ConfigObjects` can always override this parameter.

oliverschuerch (changes requested):
@alizedebray please let me know if you want to merge this and do the changes discussed [here](https://github.com/swisspost/design-system/discussions/4525) in a different pr, so my review does not block you.

gfellerph (approved):

package.json
```diff
@@ -87,6 +87,13 @@
     "migrations:build": "pnpm --filter design-system-migrations build",
     "migrations:lint": "pnpm --filter design-system-migrations lint",
     "migrations:lint:fix": "pnpm --filter design-system-migrations lint:fix",
+    "eslint:play": "pnpm --filter design-system-eslint play",
+    "eslint:start": "pnpm --filter design-system-eslint start",
+    "eslint:build": "pnpm --filter design-system-eslint build",
+    "eslint:lint": "pnpm --filter design-system-eslint lint",
+    "eslint:lint:fix": "pnpm --filter design-system-eslint lint:fix",
+    "eslint:test": "pnpm --filter design-system-eslint test",
+    "eslint:test:watch": "pnpm --filter design-system-eslint test:watch",
```
oliverschuerch, line 96: If these are unit tests, they should be renamed to `eslint:unit` and `eslint:unit:watch` here and `unit` and `unit:watch` in the package.json of the eslint package. Otherwise, they will not being executed within our `unit-tests.yaml` workflow on github.

packages/eslint/package.json
```diff
@@ -0,0 +1,69 @@
...
+  "scripts": {
+    "clean": "rimraf dist",
+    "build": "pnpm clean && tsc -p tsconfig.build.json",
+    "start": "pnpm build --watch",
+    "inspect": "wait-on dist && eslint --c eslint.play.js --inspect-config",
+    "play": "npm-run-all --parallel --print-label start inspect",
+    "lint": "eslint",
+    "lint:fix": "eslint --fix",
+    "test": "jest",
+    "test:watch": "jest --watchAll"
```
oliverschuerch, line 33: Rename this to `unit` and `unit:watch`.

oliverschuerch (dismissed):
Awesome!
Thanks @alizedebray, for the great work!

package.json
oliverschuerch, line 96: These scripts should be renamed to `eslint:unit` and `eslint:unit:watch` to align them with all the other packages.

packages/eslint/package.json
oliverschuerch, line 33:
```suggestion
    "unit": "jest",
    "unit:watch": "jest --watchAll"
```
Otherwise, they will not being executed within our unit-tests.yaml workflow on github.

oliverschuerch (changes requested):

gfellerph (changes requested):

packages/eslint/docs/rules/template/migrations/no-deprecated-btn-rg.md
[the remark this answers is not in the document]
alizedebray: This is only in our GitHub, similar to how we have readme.md for all components, and the link to this docs folder is in the README.md.

oliverschuerch (approved):

packages/eslint/docs/rules/template/migrations/no-deprecated-btn-rg.md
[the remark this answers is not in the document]
gfellerph: Would be really cool if this could be displayed somehow when you encounter this specific error in your IDE. But I think that's another topic for another time ^^

gfellerph (approved):
Good work!

PR #4498, opened 2025-01-23
chore: run clean before each build for all packages

gfellerph:
This reduces locked file errors when switching branches.


packages/components-angular/package.json
```diff
@@ -6,7 +6,8 @@
     "build": "ng build components",
```
oliverschuerch, line 6:
```suggestion
    "build": "pnpm clean && ng build components",
```

packages/nextjs-integration/package.json
```diff
@@ -6,7 +6,8 @@
     "dev": "next dev",
     "build": "next build",
```
oliverschuerch, line 7:
```suggestion
    "build": "pnpm clean && next build",
```

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #4522, opened 2025-01-24
fix(styles): make stepper always start at 1

myrta2302:


alizedebray (commented):
This change was already merged a few months back: https://github.com/swisspost/design-system/pull/3513/files
Can you please check where/why it was removed?

alizedebray (approved):

PR #4523, opened 2025-01-24
fix: remove yellow background on html

gfellerph:
Looks good on a mac, not such a good idea for everything else.


leagrdv (commented):
Approved

oliverschuerch (approved):
Thank god 🤣 
By the way, there is a meta tag for doing what you want:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color

PR #4531, opened 2025-01-27
chore(documentation, styles): clean up font weight light

leagrdv:


packages/documentation/src/stories/misc/migration-guide/migrationv8-9.component.ts
```diff
@@ -154,6 +154,30 @@ export class MigrationV89Component extends LitElement {
...
+                    <span class="tag tag-sm tag-danger">breaking</span>
+                  </p>
+                  <ul>
+                    <li><code>.fw-light</code></li>
+                    <li><code>.light</code></li>
+                  </ul>
+                </li>
+                <li class="mb-16">
+                  <p>
+                    Removed the following display SCSS variables:
```
oliverschuerch, line 170: It would be nice to tell people what to use instead. For example, we could recommend to use our headings instead and/or our line-height utilities.

packages/styles/src/utilities/temp/_legacy.scss
oliverschuerch, line 1: FYI: I've added a ticket to remove this entire file #4555

oliverschuerch (changes requested):

packages/documentation/src/stories/misc/migration-guide/migrationv8-9.component.ts
leagrdv: These SCSS variables were just there to override bootstrap's default variables, it was simply defining of weight of `300`, so I'm not sure it's relevant to talk about alternatives here. And after checking on bitbucket, no projects are using those variables.

oliverschuerch: Ah sorry, I thought this was all about the display headings (specified by bootstrap), which we probably gonna remove completely in the future.
https://getbootstrap.com/docs/4.0/content/typography/#display-headings

oliverschuerch (approved):

PR #4532, opened 2025-01-27
fix(component): fixed header navigation issue for screens <1025px

alionazherdetska:


leagrdv (changes requested):
When opened, the mega dropdown should not have a shadow on the top
Screenshot 2025-01-27 144613

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -164,7 +164,10 @@ export class PostHeader {
           </div>
         </div>
         <div
-          class={'title-header ' + (this.mobileMenuExtended ? 'title-header-mobile-extended' : '')}
+          class={
+            'title-header ' +
+            (this.mobileMenuExtended ? 'title-header-mobile-extended' : 'title-header-shadow')
```
leagrdv, line 169: In the SCSS, you can simply do &:not('.title-header-mobile-extended') to add the box-shadow instead of adding another class here

leagrdv (changes requested):

leagrdv (commented):
Just needs a changeset, then it's approved!

gfellerph (approved):

leagrdv (approved):

PR #4534, opened 2025-01-27
fix(tokens): add custom transform in the scss output config, to avoid unitless zero values

oliverschuerch:
This is necessary, so these values can also be used within a css `calc()` function:
> Note: Because `number-token`s are always interpreted as `number`s or `integer`s, "unitless 0" `length`s aren’t supported in calc(). That is, width: calc(0 + 5px); is invalid, even though both width: 0; and width: 5px; are valid.
> Source: https://drafts.csswg.org/css-values-3/#calc-type-checking


packages/tokens/_build/style-dictionary.js
```diff
@@ -18,6 +18,30 @@ StyleDictionary.registerFileHeader({
...
+ * Defines a custom StyleDictionary transform.
+ *
+ * swisspost/scss-no-unitless-zero-values:
+ * Used to keep pixel unit for zero values.
+ * This is necessary, so these values can also be used within a css `calc()` function:
+ * > Note: Because <number-token>s are always interpreted as <number>s or <integer>s, "unitless 0" <length>s aren’t supported in calc(). That is, width: calc(0 + 5px); is invalid, even though both width: 0; and width: 5px; are valid.
+ * > Source: https://drafts.csswg.org/css-values-3/#calc-type-checking
+ */
+
+StyleDictionary.registerTransform({
```
leagrdv, line 32: Some CSS properties would work with a `0` value but not with `0px` (e.g., flex order, opacity, flex grow, etc.). I haven't seen any tokens for those elements, I don't think we will have tokens for flex values, but we might for the opacity for instead. Should that function take that into account?

packages/tokens/_build/style-dictionary.js
```diff
@@ -18,6 +18,38 @@ StyleDictionary.registerFileHeader({
...
+  name: 'swisspost/scss-no-unitless-zero-values',
+  type: 'value',
+  filter: token => {
+    const usesDtcg = token.$type && token.$value;
+    const transformType = NO_UNITLESS_ZERO_VALUE_TOKEN_TYPES.includes(
+      usesDtcg ? token.$type : token.type,
```
leagrdv, line 38: Do `token.type` and `token.value` ever have values? From what I've checked, it's always undefined.

alizedebray (approved):

PR #4540, opened 2025-01-29
fix(setup): use rimraf instead of rm to fix clean command

schaertim:


packages/components-angular/package.json
```diff
@@ -7,7 +7,7 @@
     "lint": "ng lint",
     "e2e": "ng e2e --watch=false",
     "e2e:watch": "ng e2e",
-    "clean": "rm -rf dist"
+    "clean": "rimraf -rf dist"
```
alizedebray, line 10:
```suggestion
    "clean": "rimraf dist"
```

alizedebray (approved):

PR #4574, opened 2025-01-31
feat(utility): added eventGuard utility for custom Events

alionazherdetska:


leagrdv (commented):
I tested putting an accordion within another accordion and closing a child accordion item closes the parent's one as well. Should the event guard take that into account as well?

leagrdv (approved):
Approved

leagrdv (commented):
Approved

.changeset/twelve-taxis-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': minor
+---
+
+Implemented eventGuard to ensure postToggle events are only handled when originating from the expected element.
```
oliverschuerch, line 5:
```suggestion
Implemented an event-guard utility to ensure our custom events are only handled when originating from the expected element.
```

packages/components/src/utils/event-guard.ts
```diff
@@ -0,0 +1,15 @@
+export function eventGuard(
+  event: CustomEvent,
+  callback: () => void,
+  options: { targetLocalName?: string; delegatorSelector?: string } = {}
```
oliverschuerch, line 4:
```suggestion
  options: { targetLocalName!: string; delegatorSelector?: string } = {}
```
The guard does not make any sence without the targetLocalName being specified.
Since this is the case, the options object is always needed, why I would suggest to set it as the second parameter. Because it will be better readable, when you use the utility if you first specify the event, then the options object, then the callback, which will take multiple lines in most cases.

packages/components/src/utils/event-guard.ts
```diff
@@ -0,0 +1,15 @@
...
+): void {
+  const target = event.target as HTMLElement | null;
+
+  if (!target) return;
+
+  if (target.localName === options.targetLocalName) {
+    if (!options.delegatorSelector || target.closest(options.delegatorSelector)) {
```
oliverschuerch, line 11: On this line you're testing if `options.delegatorSelector` is undefined OR `target.closest(options.delegatorSelector)` is NOT undefined.

Instead you should test if `options.delegatorSelector` is undefined OR `target.closest(options.delegatorSelector)` is `this.host`.

Or maybe a little bit more human readable:
To become relevant for a `post-accordion` component, a `postToggle` event must be emitted by an element with the tag name `post-accordion-item` and `this.host` (which means `this.post-accordion`), needs to be the closest ancestor of the `post-accordion-item`, which emitted said event.

So, checking if the target has an ancestor with the tag name `post-accordion` is not enough, because yes it has one even thought, it might not be the same element as `this.host`.

Therefore, the correct syntax would be:

```js
if (!delegatorSelector || this.host === event.target.closest(delegatorSelector)) {
  callback();
}
```
This but also means, you need to to be able to access `this.host` from within the guard.
That can be done by sending it as a parameter, or if you simply bind the `this` context of the component to the eventGuard utility before you call the function, with `eventGuard.bind(this)({ options... })`

```js
eventGuard.bind(this)(event, function() {
  // do something
}, { targetLocalName: 'post-accordion-item', delegatorSelector: 'post-accordion' });
```

oliverschuerch (changes requested):
We have much more components with custom events. Imo all of them should use the eventGuard so, we'll never have an issue in the future.

oliverschuerch (commented):
Will not approve this, as long as there is no useful preview link...

packages/components/src/components/post-accordion-item/post-accordion-item.tsx
```diff
@@ -55,15 +56,17 @@ export class PostAccordionItem {
...
+  // Capture to make sure the "collapsed" property is updated before the event is consumed
   @Listen('postToggle', { capture: true })
   onCollapseToggle(event: CustomEvent<boolean>): void {
-    if (
-      event.target === this.host &&
-      (event.target as HTMLElement).localName === 'post-accordion-item'
-    ) {
-      this.collapsed = !event.detail;
-    }
+    eventGuard.call(
```
oliverschuerch, line 62:
```suggestion
    eventGuard(
```

packages/components/src/components/post-accordion/post-accordion.tsx
```diff
@@ -52,28 +53,30 @@ export class PostAccordion {
...
-
-    if (closestParentAccordion === this.host && toggledItem.localName === 'post-accordion-item') {
-      const toggledAccordionItem = event.target as HTMLPostAccordionItemElement;
-      const isClosing = this.expandedItems.has(toggledAccordionItem);
-
-      if (isClosing) {
-        this.expandedItems.delete(toggledAccordionItem);
-      } else {
-        this.expandedItems.add(toggledAccordionItem);
+    eventGuard.call(
```
oliverschuerch, line 56:
```suggestion
    eventGuard(
```

packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
```diff
@@ -43,9 +43,12 @@ export class PostCollapsibleTrigger {
...
-      if (!this.trigger || !e.target.isEqualNode(this.collapsible)) return;
-      this.trigger.setAttribute('aria-expanded', `${e.detail}`);
+    this.root.addEventListener('postToggle', (e: CustomEvent) => {
+      eventGuard.call(this, e, { targetLocalName: 'post-collapsible' }, () => {
+        if (this.trigger) {
+          this.trigger.setAttribute('aria-expanded', `${e.detail}`);
+        }
+      });
     });
   }
```
oliverschuerch, line 53: Switch hook from componentWillLoad to connectedCallback.
Make sure, you remove the registered eventlistener in the disconnectedCallback.
And by the way, arrow/inline functions, can not be used if you properly want to remove an eventlistener later.
Instead, pass a private function defined in the class it self as the callback function parameter of the addEventListener.

```ts
class bla {
  connectedCalllback() {
    this.root = getRoot(this.host);

    this.root.addEventListener('postToggle', this.toggleHandler);
  }

  disconnectedCallback() {
    this.root.removeEventListener('postToggle', this.toggleHandler);
  }
  
  private toggleHandler(e: CustomEvent) {
     ...
  }
}
```

packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
```diff
@@ -43,9 +43,12 @@ export class PostCollapsibleTrigger {
   componentWillLoad() {
     this.root = getRoot(this.host);
 
-    this.root.addEventListener('postToggle', (e: PostCollapsibleCustomEvent<boolean>) => {
-      if (!this.trigger || !e.target.isEqualNode(this.collapsible)) return;
-      this.trigger.setAttribute('aria-expanded', `${e.detail}`);
+    this.root.addEventListener('postToggle', (e: CustomEvent) => {
+      eventGuard.call(this, e, { targetLocalName: 'post-collapsible' }, () => {
```
oliverschuerch, line 47:
```suggestion
      eventGuard(this, e, { targetLocalName: 'post-collapsible' }, () => {
```

packages/components/src/components/post-language-switch/post-language-switch.tsx
```diff
@@ -59,24 +60,34 @@ export class PostLanguageSwitch {
...
   handlePostChange(event: CustomEvent<string>) {
-    this.activeLang = event.detail;
-
-    // Update the active state in the children post-language-option components
-    this.languageOptions.forEach(lang => {
-      if (lang.code && lang.code === this.activeLang) {
-        lang.setAttribute('active', '');
-      } else {
-        lang.removeAttribute('active');
+    eventGuard.call(
```
oliverschuerch, line 68:
```suggestion
    eventGuard(
```

packages/components/src/components/post-megadropdown-trigger/post-megadropdown-trigger.tsx
```diff
@@ -70,22 +71,29 @@ export class PostMegadropdownTrigger {
     this.validateControlFor();
 
     // Check if the mega dropdown attached to the trigger is expanded or not
-    document.addEventListener('postToggleMegadropdown', (event: CustomEvent) => {
-      if ((event.target as HTMLPostMegadropdownElement).id === this.for) {
-        this.ariaExpanded = event.detail.isVisible;
+    document.addEventListener('postToggleMegadropdown', (event: CustomEvent<{ isVisible: boolean; focusParent: boolean }>) => {
+      eventGuard.call(
```
oliverschuerch, line 75:
```suggestion
      eventGuard(
```

packages/components/src/components/post-menu/post-menu.tsx
```diff
@@ -71,20 +72,27 @@ export class PostMenu {
...
-      requestAnimationFrame(() => {
-        if (this.isVisible) {
-          this.lastFocusedElement = this.root.activeElement as HTMLElement;
-          const menuItems = this.getSlottedItems();
-          if (menuItems.length > 0) {
-            (menuItems[0] as HTMLElement).focus();
-          }
-        } else if (this.lastFocusedElement) {
-          this.lastFocusedElement.focus();
+      eventGuard.call(
```
oliverschuerch, line 75:
```suggestion
      eventGuard(
```

packages/components/src/components/post-accordion-item/post-accordion-item.tsx
```diff
@@ -55,15 +56,17 @@ export class PostAccordionItem {
...
+      this,
+      event,
+      { targetLocalName: 'post-accordion-item', delegatorSelector: 'post-accordion-item' },
```
oliverschuerch, line 65: Are you sure, we need to check the delegatorSelector here? And if so, does the selector need to be 'post-accordion-item'?
This would mean, the event must be triggered from a `post-accordion-item` and its closest parent which listens to the event, must be a `post-accordion-item` as well.

packages/components/src/components/post-accordion/post-accordion.tsx
```diff
@@ -52,28 +53,30 @@ export class PostAccordion {
   collapseToggleHandler(event: CustomEvent<boolean>) {
     event.stopPropagation();
```
oliverschuerch, line 54: With the event guard in place, I would expect, this is not needed anymore.
Not 100% sure.

packages/components/src/components/post-language-switch/post-language-switch.tsx
```diff
@@ -59,24 +60,34 @@ export class PostLanguageSwitch {
...
+      this,
+      event,
+      { targetLocalName: 'post-language-option' },
```
oliverschuerch, line 71: I think here, we should specify the delegatorSelector with the value `post-language-swtich`.

packages/components/src/components/post-menu/post-menu.tsx
```diff
@@ -71,20 +72,27 @@ export class PostMenu {
...
+        this,
+        event,
+        { targetLocalName: 'post-popovercontainer' },
```
oliverschuerch, line 78: If the post-popovercontainer here always is a child of the post-menu, we should use the delegatorSelector with the value `post-menu`.

oliverschuerch (changes requested):

packages/components/src/components/post-menu/post-menu.tsx
oliverschuerch, line 1: Make sure you do not only add the event listener, but also remove it in the disconnectedCallback!

packages/components/src/components/post-megadropdown-trigger/post-megadropdown-trigger.tsx
oliverschuerch, line 1: Make sure you do not only add the event listener, but also remove it in the disconnectedCallback!

packages/components/src/components/post-accordion-item/post-accordion-item.tsx
```diff
@@ -55,15 +56,17 @@ export class PostAccordionItem {
...
-    if (
-      event.target === this.host &&
-      (event.target as HTMLElement).localName === 'post-accordion-item'
-    ) {
-      this.collapsed = !event.detail;
-    }
+    eventGuard(
+      this.host,
+      event,
+      { targetLocalName: 'post-accordion-item', delegatorSelector: 'post-collapsible' },
```
oliverschuerch, line 65: This still seems to be wrong imo.
Is the `post-accordion-item` component really listening to a `postToggle` event, emitted by a `post-accordion-item` element, where its host element (`this.host`) equals an element which is selectable with `post-collapsible`?

In my opinion, we're listening to a `postToggle` event, emitted by a `post-collapsible` element, while its closest parent should be a `post-accordion-item` element that matches `this.host`.

oliverschuerch (changes requested):
Also check if the following components need a eventGuard implementation:
- post-header
- post-megadropdown
- post-language-switch

alionazherdetska:
> Also check if the following components need a eventGuard implementation:
> 
> * post-header
> * post-megadropdown
> * post-language-switch

I’ve implemented eventGuard in the `post-header` component.
For the `post-megadropdown` component, there’s currently no use case that requires it.
In the `post-language-switch`, eventGuard is already correctly implemented.

leagrdv (approved):

PR #4577, opened 2025-02-03
feat(styles): foundation interactive card

leagrdv:


packages/styles/src/components/card.scss
```diff
@@ -10,26 +10,79 @@
...
+@use './../tokens/components';
 
-tokens.$default-map: utilities.$post-spacing;
+tokens.$default-map: components.$post-cards;
 
 .card-group {
   margin-top: commons.$border-width;
   margin-left: commons.$border-width;
-  box-shadow: commons.$box-shadow-lg;
+  box-shadow: tokens.get('card-interactive-elevation');
```
alizedebray, line 24: This shadow look a little strange because of a 1px margin on the cards, can you update that please?

Also, if the group contains only non-interactive cards it should probably not have a shadow.

packages/styles/src/components/card.scss
```diff
@@ -10,26 +10,79 @@
...
+  // Add a border if card has no interactive element
+  &:not(.interactive-card) {
+    border: 1px solid tokens.get('card-non-interactive-stroke');
+  }
+
+  .card-header:first-child,
+  .list-group:first-child,
+  .card-img-top {
+    border-top-left-radius: tokens.get('card-border-radius');
+    border-top-right-radius: tokens.get('card-border-radius');
```
alizedebray, line 77: Can you replace all `top`, `bottom`, `right`, and `left` properties with their `start`, `end` equivalent? 

```suggestion
    border-start-start-radius: tokens.get('card-border-radius');
    border-start-end-radius: tokens.get('card-border-radius');
```

packages/styles/src/components/card.scss
```diff
@@ -10,26 +10,79 @@
...
   color: card.$card-color;
+  border-radius: tokens.get('card-border-radius');
+  background-color: tokens.get('card-enabled-bg1');
+
+  .palette-alternate & {
+    background-color: tokens.get('card-enabled-bg2');
+  }
+
+  // Add an elevation only if card is interactive
+  &.interactive-card {
```
alizedebray, line 39: It seems that the interactive cards are the same as cards wrapped in a post-linkarea: https://next.design-system.post.ch/?path=/docs/1d52b794-768b-464e-90eb-4fd15774aa90--docs

As the post-linkarea is more flexible and potentially better for accessibility, I would drop the `.interactive-card` class and set the styles on the `post-likarea .card`.

packages/documentation/src/stories/components/teaser/teaser.stories.ts
alizedebray, line 1: It isn't part of your ticket but since its a quick change, can you please update the meta to use `Card Teaser` instead of `Teaser Card` so that they are grouped with the cards in the side navigation? 😇

alizedebray (changes requested):

alizedebray:
Question to the team: should we hide buttons and links that aren't in interactive cards to prevent misaligned UI?

packages/documentation/src/stories/components/card/card.docs.mdx
alizedebray, line 1: Can you please adapt the docs so that it follows the structure of the component documentation pages? 
Meaning, the interactive example is at the top followed by the controls table.

alizedebray (changes requested):
Can you please check these points? 
- The text color does not adapt when you hover the interactive card, you can see it in the [post-linkarea documentation](https://preview-4577--swisspost-design-system-next.netlify.app/?path=/docs/1d52b794-768b-464e-90eb-4fd15774aa90--docs)
- The cursor should be a pointer on the whole link area
- The examples in the documentation are no longer correct: 
  - The Basic Content example should be wrapped in a post-linkarea when interactive and shouldn't have a button when non-interactive
  - The Custom Content example should be interactive since it contains buttons
- Please add a changeset for the new styles

packages/documentation/src/stories/components/card/card.docs.mdx
leagrdv: Comments have been adressed.
Regarding interactive cards with multiple links: as stated in figma, if a card has more than one interactive element, it should not be an interactive card. I added a banner explaining that in the docs as well

PR #4588, opened 2025-02-04
chore(styles, documentation): remove some utility sizes + clean up

leagrdv:


packages/styles/src/variables/_sizing.scss
alizedebray, line 1: The `$post-px-sizes` map should be removed otherwise we will have to update it anytime there is a change in the utility tokens and there is a risk that they are not in sync anymore.

alizedebray (changes requested):

PR #4608, opened 2025-02-05
feat(styles)!: rename entry points according to new appearance

gfellerph:
Updated CSS and SCSS entry file names to better represent the styles contained. There is now the choice between compact and default appearance instead of internal and external styles. This change better represents the resulting look of the components.

- post-external.[s]css -> post-default.[s]css
- post-internal.[s]css -> post-compact.[s]css
- post-tokens-external.[s]css -> post-tokens-default.[s]css
- post-tokens-internal.[s]css -> post-tokens-compact.[s]css
- cargo-external.[s]css -> cargo-default.[s]css
- cargo-internal.[s]css -> cargo-compact.[s]css
- cargo-tokens-external.[s]css -> cargo-tokens-default.[s]css
- cargo-tokens-internal.[s]css -> cargo-tokens-compact.[s]css


packages/styles/src/tokens/_device-compact.scss
```diff
@@ -0,0 +1,15 @@
...
+@use './../mixins/media';
+
+:root {
+  @include media.max(600px) {
+    @include tokens.from(device.$post-mobile);
+  }
+
+  @include media.min(600px) {
+    @include tokens.from(device.$post-tablet);
+  }
```
leagrdv, line 14: How come the compact appearance does not use the desktop tokens?

pnpm-lock.yaml
leagrdv, line 1: Why are there changes on that file?

leagrdv (changes requested):

.changeset/slimy-items-happen.md
```diff
@@ -0,0 +1,14 @@
+---
+'@swisspost/design-system-styles': major
+---
+
+Updated CSS and SCSS entry file names to better represent the styles contained. There is now the choice between compact and default appearance instead of internal and external styles. This change better represents the resulting look of the components.
+
+- post-external.[s]css -> post-default.[s]css
```
alizedebray, line 7: Didn't we define that we would simply use `post.[s]css` for the entry file?

packages/styles/src/tokens/_appearance-compact.scss
alizedebray, line 1: Do we still need these files? I understood we would completely drop this layer in favor of simply using tablet tokens on desktop for the application.

gfellerph: This is a temporary measure to test if we can really do that. We still need to rename the tokens and the tokens package and apply the measure to the token system in order for it to work in figma. Otherwise we'd just have appearance: compact styles in code.

packages/styles/src/tokens/_device-compact.scss
gfellerph: This is a temporary measure to test if this really holds true. We still have to rename the tokens and adapt the system to work like this from token studio.

pnpm-lock.yaml
gfellerph: Updated the branch from main and they're gone^^. Possibly some system differences that produced different lockfile content.

.changeset/slimy-items-happen.md
gfellerph: I tried to apply this naming pattern, but I found it leads to more confusion than it saves on typing or cleanlyness. For example in the imports:

```scss
@layer design-system {
  @include meta.load-css('./tokens/schemes');
  @include meta.load-css('./tokens/device-default');
  @include meta.load-css('./tokens/appearance-default');
  @include meta.load-css('./tokens/post-theme');
}
```

If the device and appearance imports are not named "default", it looks like we're importing all device and appearance tokens and not only the default set. As a consequence of this, also the main import files have the default postfix. What do you think?

alizedebray: That make sense

alizedebray (approved):

leagrdv (approved):

PR #4614, opened 2025-02-05
fix(component): set --main-header-height for `post-header` on component load

alionazherdetska:


packages/components/src/components/post-header/post-header.tsx
```diff
@@ -193,9 +198,6 @@ export class PostHeader {
       this.mobileMenuAnimation.finish(); // no animation
     }
 
-    const mhh = this.host.shadowRoot.querySelector('.local-header')?.clientHeight;
```
alizedebray, line 196: This should remain in the `handleResize` as well otherwise the value is not updated when you go from desktop to mobile and vice-versa.

alizedebray (approved):

PR #4616, opened 2025-02-05
fix(components): add base to URL type check to allow for relative URLs

gfellerph:


oliverschuerch (approved):
If this fixes an issue, we should add a changeset!

PR #4623, opened 2025-02-06
fix(component): ensure proper stacking order to prevent megadropdown from being hidden behind content

alionazherdetska:
…navigation for the header


alionazherdetska:
Closing the issue as it was resolved in the latest version by setting z-indexes on the `post-header` and its components.

alionazherdetska:
Closing the issue as it was resolved in the latest version by setting z-indexes on the `post-header` and its components.

PR #4625, opened 2025-02-06
chore(header): impoved focus behaviour of megadropdown

schaertim:


.changeset/blue-eyes-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Improved keyboard navigation for `post-megadropdown` by auto-focusing first element when opened with Enter key and returning focus to trigger when closed.
```
alizedebray, line 5:
```suggestion
Improved keyboard navigation for the `post-megadropdown` by focusing on the first element when it is opened with the Enter key and returning the focus to the trigger when it is closed.
```

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
```diff
@@ -75,6 +77,9 @@ export class PostMegadropdown {
     PostMegadropdown.activeDropdown = this;
     this.postToggleMegadropdown.emit(this.isVisible);
     this.addOutsideClickListener();
+    this.triggerElement = document.querySelector(
```
alizedebray, line 80: In theory there could be multiple triggers. 
Can you rather listen to the `postToggleMegadropdown` in the trigger that opened the megadropdown?

alizedebray (changes requested):

PR #4631, opened 2025-02-07
chore(documentations, styles): remove legacy classes file

leagrdv:


packages/documentation/src/stories/components/select-custom/select-custom-floating.sample.html
```diff
@@ -5,7 +5,7 @@
     aria-haspopup="listbox"
     (keydown)="setFocus($event)"
     aria-labelledby="customSelectLabel"
-    class="form-select custom-select text-start no-toggle-arrow"
```
alizedebray, line 8: This class is often used in projects, we should probably keep it.

alizedebray (changes requested):

packages/documentation/src/stories/components/select-custom/select-custom-floating.sample.html
leagrdv: I put it back and moved it to the `form-select.scss` file

alizedebray (approved):

PR #4633, opened 2025-02-07
chore(styles): setting initial height for the `post-header` and its components

alionazherdetska:


gfellerph (changes requested):
I think this is a duplicate PR and can be closed. Is that correct @alionazherdetska?

alionazherdetska:
The issue was already addressed in this PR #4769

PR #4638, opened 2025-02-10
chore(components): align prop validation messages

myrta2302:


packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -135,11 +135,7 @@ export class PostTooltip {
 
   @Watch('delayed')
   validateDelayed() {
-    checkEmptyOrType(
-      this.delayed,
-      'boolean',
-      'The post-tooltip "delayed" property should be a boolean.',
-    );
+    checkEmptyOrType(this, 'delayed', 'boolean');
```
oliverschuerch, line 138: Instead of passing `this` as an argument, we could use the `checkEmptyOrType.call()` function to bind the context and pass all the needed properties.
This should be discussed with the team, so we can find a common way to do this.
Because Alona is implementing the `eventGuard` function which has pretty much the same issue and it would be nice to implement context binding the same way everywhere it's needed.

oliverschuerch (changes requested):

packages/components/src/components/post-tooltip/post-tooltip.tsx
alizedebray: I created a separate ticket or that [#4703](https://github.com/swisspost/design-system/issues/4703)

oliverschuerch (approved):

PR #4650, opened 2025-02-11
fix(styles): adjusted the composable header height to match the design specification

alionazherdetska:


gfellerph:
@alionazherdetska merge main into this branch and re-check after core token removal

oliverschuerch (changes requested):
Don't know if this is related, but there is again a gap between the global and the local-header and the local-header is not scrolling.
image

packages/components/src/components/post-header/post-header.scss
```diff
@@ -96,7 +96,15 @@
...
   transition: box-shadow 0.2s ease-in-out;
-  border-block-end: 1px solid #e1e0dc;
+  
+  &::after {
+    content: '';
+    position: absolute;
+    width: 100%;
+    height: 1px;
+    background-color: #e1e0dc;
+    bottom: -1px;
```
gfellerph, line 106:
```suggestion
    bottom: -1px;
    z-index: 1;
```

gfellerph (changes requested):
There is also a border top on the `.megadropdown-container` that can be removed.

oliverschuerch (approved):

PR #4652, opened 2025-02-11
feat(component): implemented breakpoints utlity in the `post-header` component

alionazherdetska:


packages/components/src/components/post-header/post-header.tsx
```diff
@@ -186,35 +194,8 @@ export class PostHeader {
...
-      newDevice = 'mobile';
-    }
-
-    // Close any open mobile menu
-    if (newDevice === 'desktop' && this.mobileMenuExtended) {
-      this.toggleMobileMenu();
-      this.mobileMenuAnimation.finish(); // no animation
-    }
-
-    this.updateLocalHeaderHeight();
```
leagrdv, line 207: The local header height is no longer updated on resize

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -186,35 +194,8 @@ export class PostHeader {
...
-    } else if (width >= 600) {
-      newDevice = 'tablet';
-    } else {
-      newDevice = 'mobile';
-    }
-
-    // Close any open mobile menu
-    if (newDevice === 'desktop' && this.mobileMenuExtended) {
-      this.toggleMobileMenu();
-      this.mobileMenuAnimation.finish(); // no animation
```
leagrdv, line 204: Same here, in case the mobile menu is expanded on mobile, it should be closed when switching to desktop.

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
```diff
@@ -41,6 +39,11 @@ export class PostMegadropdown {
     if (PostMegadropdown.activeDropdown === this) {
       PostMegadropdown.activeDropdown = null;
     }
+    window.removeEventListener('postBreakpoint:name', this.breakpointChange.bind(this));
+  }
+
+  private breakpointChange(e: CustomEvent) {
```
leagrdv, line 45: This is not called on resize, which breaks the function `keyboardHandler`.

packages/components/src/components.d.ts
leagrdv, line 1: Could you add a changeset to your PR?

leagrdv (changes requested):

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -189,41 +211,16 @@ export class PostHeader {
...
-    if (newDevice !== previousDevice) {
-      this.device = newDevice;
-
-      this.postUpdateDevice.emit(this.device);
-      window.requestAnimationFrame(() => {
-        this.switchLanguageSwitchMode();
-      });
-    }
+    this.device = breakpoint.get('name');
+    this.switchLanguageSwitchMode();
```
leagrdv, line 223: `this.switchLanguageSwitchMode()` should only be called when breakpoint changes cause it only switches between desktop and tablet/mobile, so you can remove it from here. But `this.updateLocalHeaderHeight()` should be called here as it needs to update the value on every resize.

leagrdv (changes requested):

leagrdv (changes requested):
The calls seem to be made at the right time now, but on scroll the header height is not reduced anymore in desktop

oliverschuerch (changes requested):

alionazherdetska:
After discussion in the dev roundtable, we've decided to close this PR and create a new ticket: https://github.com/orgs/swisspost/projects/3/views/12?pane=issue&itemId=108261388&issue=swisspost%7Cdesign-system%7C5357.

**Reasons:**
This PR has been on hold for an extended period
There are now several merge conflicts that would be time-consuming to resolve
It would be more efficient to implement this from scratch with the current codebase

PR #4654, opened 2025-02-12
chore(styles): update utility mixin

leagrdv:


.changeset/great-houses-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Updated the utilities API mixin to shorten output CSS as well as removed unnecessary code regarding pseudo elements in that same mixin.
```
alizedebray, line 5: Might be too technical for a changeset as users do not know about the utility API and mixin, only about the utility classes. 
```suggestion
Merged utility classes that were setting the same properties, which reduces the size of the CSS output.
```

packages/styles/src/utilities/_mixins.scss
```diff
@@ -39,34 +42,36 @@
     $state: if(map.has-key($utility, state), map.get($utility, state), ());
 
     $infix: if(
-      $property-class == '' and string.slice($infix, 1, 1) == '-',
+      $class-list == '' and string.slice($infix, 1, 1) == '-',
```
alizedebray, line 45: Is this correct? Since `$class-list` is now a list I would imagine it is never equal to `''`

alizedebray (changes requested):

packages/documentation/src/stories/utilities/opacity/opacity.stories.ts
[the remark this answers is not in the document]
leagrdv: They're just small issues I had noticed when reworking the utilities mixin, and it seemed trival to make a PR just for that

PR #4657, opened 2025-02-12
chore(components, styles): remove core tokens from scss files

leagrdv:


.changeset/fuzzy-files-lie.md
alizedebray, line 1: Not sure a changeset is needed here since this is only a technical change but the styles, components, and documentation remain the same in the end.

alizedebray (approved):

PR #4661, opened 2025-02-12
fix(component): refactor `post-menu` and `post-menu-trigger` for correct aria-expanded attribute handling

alionazherdetska:


leagrdv (changes requested):
There is a weird behavior when using tab to navigate, if I select one of the elements of the menu button and press `Enter`, another menu appears in the page.


https://github.com/user-attachments/assets/75dacdf2-062b-4bce-ae5b-05d509914422

leagrdv (commented):
Approved

packages/components/src/components/post-menu-trigger/post-menu-trigger.tsx
oliverschuerch, line 1: Would it work to define `this.root = getRoot(this.host)` inside a `connectedCallback()`?
Because define it in `componentDidLoad()` means the component has been already rendered once before we set `this.root` to its value.

And also, why is `this.root` flagged as optional (on line 28)?
At least initialize it with `null` and add `null` as one of the possible types.

OR

You can implement it the same way you implemented the menu getter:
```ts
private get root(): Document | ShadowRoot | null {
  return getRoot(this.host);
}
```

packages/components/src/components/post-menu/post-menu.tsx
oliverschuerch, line 1: Same is true for `this.root` here as well.
It's not optional, but can be of type `Document | ShadowRoot | null`.

However you implement `root` in the `post-menu-trigger`, do it the same way here as well!

packages/components/src/components/post-menu/post-menu.tsx
```diff
@@ -81,10 +93,11 @@ export class PostMenu {
    */
   @Method()
   async toggle(target: HTMLElement) {
-    if (this.isVisible) {
-      await this.hide();
+
+    if (this.popoverRef) {
+      await this.popoverRef.toggle(target);
```
oliverschuerch, line 98: Why is the keyword `await` needed here? You're not doing anything else afterwards.

packages/components/cypress/e2e/menu-button.cy.ts
oliverschuerch, line 1: Add the following test cases:
- [ ] Test if `@menu` is visible after toggeling
- [ ] Add two triggers and two menus, connect toggleA to menuA and toggleB to menuB, then:
  Test if triggerA toggles only menuA and triggerB toggles only menuB
- [ ] Add two triggers and two menus, connect toggleA and toggleB to menuA and none to menuB, then:
  Test if triggerA toggles menuA and triggerB toggles menuA as well, make sure menuB is not toggled

oliverschuerch (changes requested):

packages/components/cypress/e2e/menu-button.cy.ts
alionazherdetska: Created a separate ticket to add more tests: [#4735](https://github.com/swisspost/design-system/issues/4735)

oliverschuerch (approved):

leagrdv (approved):

PR #4663, opened 2025-02-13
chore(components, styles): move header css variables to styles package

leagrdv:


packages/styles/src/components/header/index.scss
```diff
@@ -4,6 +4,28 @@
 @use 'mixins' as *;
 
 post-header {
+  --global-header-height: 72px;
+  --global-header-minimal-height: 24px;
```
alizedebray, line 8: Don't know if this one is used, I think this was renamed to `--global-header-reduced-height`. Also, it should probably be set to `--global-header-reduced-height: --global-header-height;` on mobile... 
Can you please check this out and clean the variables?

alizedebray (changes requested):

packages/styles/src/components/header/index.scss
```diff
@@ -5,6 +5,28 @@
 @use 'mixins' as *;
 
 post-header {
+  --global-header-height: 72px;
+  --local-header-height: 112px;
+  --local-header-min-height: 56px;
+  --header-height: calc(var(--global-header-height) + var(--local-header-height));
+  --global-header-reduced-height: var(--global-header-height);
```
alizedebray, line 12: Can you place these properties either in the `media.min(lg)` or the `media.max(lg)` mixins for better readability? Currently it is not clear that the `--global-header-height: 72px;` is only true on large screens whereas the `--global-header-reduced-height: var(--global-header-height);` is only true on small screens. 

I would only keep the `--header-height` outside of the mixins since its calculated.

alizedebray (approved):

PR #4666, opened 2025-02-13
chore(components): add tests for the post-mainnavigation

alizedebray:


alizedebray:
> https://github.com/user-attachments/assets/b9938a00-1309-4f51-94a7-dcac174debf3

That is strange, I cannot reproduce this and the test for that seems to be passing... What browser are you using? Is it also happening on full screen? There is a Mutation Observer so adding links client side should be an issue.

alizedebray:
> Check if test is actually broken or just flaky.

Tests were broken because of changes on the main branch, I updated them.

oliverschuerch:
> > Main.Navigation.Scroll.Behaviour.mp4
> 
> That is strange, I cannot reproduce this and the test for that seems to be passing... What browser are you using? Is it also happening on full screen? There is a Mutation Observer so adding links client side should not be an issue.

I used chrome 132 on Win11

oliverschuerch (approved):

alizedebray:
> > > Main.Navigation.Scroll.Behaviour.mp4
> > 
> > 
> > That is strange, I cannot reproduce this and the test for that seems to be passing... What browser are you using? Is it also happening on full screen? There is a Mutation Observer so adding links client side should not be an issue.
> 
> I used chrome 132 on Win11

I opened a ticket to investigate the issue:  #4742

PR #4674, opened 2025-02-14
fix(components): make it possible to use multiple language switches in the post header

oliverschuerch:


packages/components/src/components/post-language-switch/post-language-switch.tsx
```diff
@@ -56,28 +56,45 @@ export class PostLanguageSwitch {
     );
   }
 
+  /**
+   * Variant that determines the rendering of the language switch either as a list (used on mobile in the header) or a dropdown (used on desktop in the header)
+   */
```
alizedebray, line 61: Please update this description.

packages/components/src/components/post-language-option/post-language-option.tsx
```diff
@@ -56,6 +56,11 @@ export class PostLanguageOption {
    */
   @Prop() variant?: SwitchVariant | null;
 
+  /**
+   * The type of the post-language-switch parent (dynamically set by the parent)
+   */
```
alizedebray, line 61: Can you make the description more specific so that we can easily differentiate the `variant` prop form the `type` prop?

packages/documentation/src/stories/components/language-switch/language-switch.stories.ts
```diff
@@ -26,6 +27,12 @@ const meta: MetaComponent<HTMLPostLanguageSwitchElement> = {
         category: 'content',
       },
     },
+    type: {
+      description: 'View type of the language switch.',
+      table: {
+        category: 'content',
+      },
+    },
```
alizedebray, line 35: This override is unnecessary, this information is generated from the tsdocs in the components package.
```suggestion
```

alizedebray (changes requested):

alizedebray (approved):

alizedebray (changes requested):
Is that what we want? 
image

alizedebray (approved):

oliverschuerch:
> Is that what we want? image
> 
> Also, is the `type` property only setting styles, if so why not using a class instead?

No it's not, that's still a issue I have to fix.

Adding a modifier class to a shadow root component seems to be the wrong approch to me. I'll go with the prop for now, since we don't know if the prop needs to change logic, etc. as well in the future.

PR #4680, opened 2025-02-14
feat(styles): add gulp task to update _not-defined.scss with built components

myrta2302:
…mponents


packages/styles/gulpfile.js
```diff
@@ -160,6 +160,48 @@ gulp.task(
...
+    }
+
+    const content = data;
+    const componentNames = content
+      .match(/export \{ (\w+) \} from/g)
+      .map(match => match.replace(/export \{ (\w+) \} from/, '$1'));
+
+    const kebabCaseNames = componentNames.map(name =>
+      name.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase(),
+    );
```
oliverschuerch, line 182:
```suggestion
    ```suggestion
    const kebabCaseNames = Array.from(data.matchAll(/export \{ (\w+) \} from/g), (m) => m[1].replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase());
```

Since you're not using the vars `data` and `componentNames` anymore later in code, you can combine everything to this single line. And also avoid implementing the regex twice.

packages/styles/gulpfile.js
```diff
@@ -160,6 +160,48 @@ gulp.task(
...
+    const componentNamesContent = `/*
+  Initial visibility of the components is set to hidden to prevent 'flickering' effect due to stencil js/scss delay.
+*/
+
+:where(
+    ${kebabCaseNames.join(',\n    ')}
+):not(:defined) {
+  visibility: hidden;
+}
+    `;
```
oliverschuerch, line 193: Would be nice, if we could extract this code and save it (for example) as a `_not-defined.template.scss`.
We can then load it in this gulp task and update it with theg given kebabCaseNames, but keep the format and have it checked in to git. While the output file would not be checked in (because it's getting generated).

oliverschuerch (changes requested):
Since we're generating this file now, we should add it in the /styles/.gitignore file.

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #4687, opened 2025-02-17
chore(workflow): create a GitHub workflow for dry running publish process

alionazherdetska:


oliverschuerch (changes requested):
Should this workflow just show if we can publish or not?
What if it fails, shouldn't it stop some of our workflows from doing something, like stopping the publish workflows from publishing the docs or the packages?
If so, we should make it an action and integrate it into the other workflows, or combine it at least so the other workflows only run if this finishes successfully.

gfellerph:
> Should this workflow just show if we can publish or not? What if it fails, shouldn't it stop some of our workflows from doing something, like stopping the publish workflows from publishing the docs or the packages? If so, we should make it an action and integrate it into the other workflows, or combine it at least so the other workflows only run if this finishes successfully.

This should just be a regular check on release PRs. If it's green, we should be able to merge and release, if it's red, fix the issue before we can release. Very simple and straight forward. No dependencies to anything, just a simple check.

.github/workflows/dry-run-publish.yaml
```diff
@@ -0,0 +1,23 @@
+name: Dry Run Publish
+on:
+  pull_request:
+    branches:
+      - changeset-release/main
+      - changeset-release/release/v*
+
```
gfellerph, line 7: @alionazherdetska can you add [workflow permissions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token) to this workflow (see GitHub alert). Read permissions should be enough to run this task.

.github/workflows/dry-run-publish.yaml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  dry-run-publish:
+    name: Publish Verification
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v4
```
gfellerph, line 14: Can you reference this action by hash instead of version?

gfellerph (changes requested):

gfellerph (approved):

oliverschuerch (approved):

PR #4691, opened 2025-02-17
chore(component): stop using the window object in the post-linkarea

alizedebray:


packages/components/src/components/post-linkarea/post-linkarea.tsx
```diff
@@ -1,30 +1,49 @@
 import { Component, Element, h, Host } from '@stencil/core';
 import { version } from '@root/package.json';
 
+const INTERACTIVE_ELEMENTS = ['a'].join(',');
```
alizedebray, line 4: I added this constant so that we can refactor the component in the future to also work with buttons (this will be needed for interactive cards). But for now it only accepts anchor elements as it did before.

packages/components/src/components/post-linkarea/post-linkarea.tsx
oliverschuerch, line 1: The logic should also take into account whether the user presses the “Ctrl” key (open in a new tab) or the “Shift” key (open in a new window) when clicking on the element.

oliverschuerch (changes requested):

packages/components/src/components/post-linkarea/post-linkarea.tsx
```diff
@@ -1,30 +1,49 @@
...
 
-  private handleClick = () => {
-    const customSlottedLink: HTMLAnchorElement = this.hostElement.querySelector('a[data-link]');
+  private delegateClick() {
+    // get any child with a data-link attribute and check that it is interactive
+    const elementWithDataLink = this.host.querySelector('[data-link]');
+    if (!this.isInteractive(elementWithDataLink)) {
+      throw new Error(
+        `The \`data-link\` attribute must be used on a interactive element inside the \`post-linkarea\` component. Possible elements are: ${INTERACTIVE_ELEMENTS}`,
+      );
```
oliverschuerch, line 22: I feel like we should not throw an error if there is at least one interactive element, even though it may not contain the [data-link] attribute. Instead I would log a warning and only throw an error if we really can't find any interactive element. Because then, and only then, the component can not work as expected.

oliverschuerch (changes requested):

PR #4696, opened 2025-02-17
chore(component): stop using the window object in the post-linkarea

alizedebray:


oliverschuerch (approved):
Works like a charme, thanks!

PR #4697, opened 2025-02-17
feat(components): make the post-icon component ssr ready

oliverschuerch:


alizedebray (dismissed):

packages/components/src/components/post-icon/post-icon.tsx
```diff
@@ -113,6 +115,16 @@ export class PostIcon {
 
   // Construct the icon url from different possible sources
   private getUrl() {
+    const file = `${this.name}.svg`;
+
+    if (this.isSSR) {
+      if (this.base) {
+        return `/${this.base}/${file}`.replaceAll(/\/\//g, '/');
```
oliverschuerch, line 122: Don‘t merge it like that, this.base could be an absolute url and then the first '/' in the return value would make the it an invalid url!

gfellerph (approved):

PR #4701, opened 2025-02-18
fix(components): fix the active attribute of the post-language-option

alizedebray:


gfellerph (approved):
Thanks for the quick fix, I think we should update tests accordingly as well, but can be done in another ticket.

PR #4702, opened 2025-02-19
chore(styles): clean up button variables

schaertim:


oliverschuerch (changes requested):
If there are variables, who have been removed completely, we need to add information about them in the migration guide.

oliverschuerch (approved):

PR #4704, opened 2025-02-19
fix(component): prevent slide-down animation for already open `post-megadropdown` component when resizing from mobile/tablet to desktop

alionazherdetska:


gfellerph:
Wait for https://github.com/swisspost/design-system/pull/4723

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
```diff
@@ -104,9 +107,18 @@ export class PostMegadropdown {
...
+        (event: CustomEvent<DEVICE_SIZE>) => {
+          const newDevice = event.detail;
+          if (this.device !== newDevice) {
+            this.device = newDevice;
+            if (newDevice === 'desktop' && this.isVisible) {
+              this.animationClass = null;
+            }
+          }
+        },
+      );
```
oliverschuerch, line 121: Just no!
Adding an eventlistener on the parent component is just not how web-components should communicate with each other, it's creating way too much and unnecessary dependencies!

Use the breakpoints utility to know if there is a breakpoint change or on which breakpoint you currently are.
Also, the utility already is optimize to only emit an event if there is a change, making this.device !== newDevice obsolete.

oliverschuerch (changes requested):

gfellerph (approved):

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
```diff
@@ -1,25 +1,33 @@
...
+  @State() device: string = breakpoint.get('name');
 
   @Element() host: HTMLPostMegadropdownElement;
 
   /** Tracks the currently active dropdown instance. */
   private static activeDropdown: PostMegadropdown | null = null;
 
+  private breakpointChange(e: CustomEvent) {
+    const newDevice = e.detail; 
+    if (this.device !== newDevice) {
```
oliverschuerch, line 23: Is this check even necessary? I think the postBreakpoint:name event should only trigger, if there really is a change. Can you test that and remove the line here, if it's unnecessary?

And also to save an object property in a new var/const is unnecessary. Instead of `const newDevice = e.detail` and then using `newDevice`, you can just use `e.detail` and save some Bits in memory.

oliverschuerch (approved):

PR #4705, opened 2025-02-19
chore(migrations): remove package

alizedebray:


oliverschuerch (approved):
Good to see that this package can be removed finally!
Where do we need to inform our users, that this package will no longer be maintained?

PR #4712, opened 2025-02-20
chore(components): enable header overflow-y on tablet & mobile

myrta2302:


packages/components/src/components/post-header/post-header.scss
```diff
@@ -205,6 +205,7 @@
     height: calc(100vh - var(--header-height));
     display: flex;
     flex-direction: column;
+    overflow-y: auto;
```
gfellerph, line 208: This also introduces the possibility to scroll the first level navigation when the second one is open:

image

When the second level navigation is open, the first level navigation should not be scrollable.

gfellerph (changes requested):

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -234,10 +241,15 @@ export class PostHeader {
...
-    const navigationClasses = ['navigation'];
+    let navigationClasses = ['navigation'];
     if (this.mobileMenuExtended) {
       navigationClasses.push('extended');
     }
+    if (this.megadropdownOpen) {
+      navigationClasses = navigationClasses.filter(className => className !== 'scroll-y');
+    } else {
+      navigationClasses.push('scroll-y');
+    }
```
gfellerph, line 252: `navigationClasses` is defined each time the function is called. The string `scroll-y` will never be part of the array because the value is not persisted anywhere.

```suggestion
    if (!this.megadropdownOpen) {
      navigationClasses.push('scroll-y');
    }
```

gfellerph (changes requested):

gfellerph (approved):

PR #4733, opened 2025-02-20
feat(styles): add new `.section` css component

oliverschuerch:


packages/styles/src/layout/index.scss
```diff
@@ -1,2 +1,3 @@
+@forward './sections';
```
gfellerph, line 1: I'd call it singular "section.scss" instead of plural.

```suggestion
@forward './section';
```

gfellerph (approved):

PR #4741, opened 2025-02-21
feat(styles): add bleeding classes to the `.section` component

oliverschuerch:


packages/styles/src/layout/_section.scss
gfellerph, line 1: With that file name, no individual CSS will be generated to selectively import these styles. Maybe we should think about the layout folder and if we can offer selective imports on this one or not. But that's a different issue.

gfellerph (approved):

packages/styles/src/layout/_section.scss
oliverschuerch: I know. But once again, I just aligned it with the existing files.

PR #4747, opened 2025-02-21
fix(components): update from `vh` to `dvh` in `post-header` to take iOS devices into account

leagrdv:


.changeset/nervous-rice-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Changed from `vh` to device viewport height (`dvh`) unit on post-header to take into account the iOS devices height change.
```
oliverschuerch, line 5:
```suggestion
Changed from `vh` to dynamic viewport height (`dvh`) unit on post-header to take into account the mobile devices height change.
```

oliverschuerch (changes requested):

PR #4759, opened 2025-02-24
fix(styles): fix text alignment for `post-language-option` component to match the design

alionazherdetska:


oliverschuerch (commented):
I don't thinks this is implemented as wanted.
In Figma we have a padding-left of 16px and a padding-right of 12px (which is weird also, because it should have 16px on both sides imo).
But if we do that, the dropdown gets more narrow. Maybe adding a min-width to the post-language-option element would help.
I've added the pattern-discussion label to discuss this with designers.

gfellerph:
@oliverschuerch is the width of the dropdown not aligned with the width of the trigger? Not sure if that makes sense, but that seems to be the explanation for the dropdown width.

oliverschuerch:
> @oliverschuerch is the width of the dropdown not aligned with the width of the trigger? Not sure if that makes sense, but that seems to be the explanation for the dropdown width.

This would not make any sense, both the trigger and the dropdown width should be based on the widths of the ISO country codes included. "IT" (italian) would be much less wide than for example "MX" (mexican spanisch) and if "IT" is the current selected language, the trigger could become less wide than the dropdown should be.
That's why I asked @alionazherdetska to add a min-width to the dropdown, but not align trigger and dropdown widths at all.
According to her, it is implemented the same way in Figma anyway.

packages/components/src/components/post-language-option/post-language-option.scss
```diff
@@ -48,18 +48,18 @@ post-language-option {
   }
 
   &:where([variant='menu']) {
-    width: 100%;
+    min-width: 73px;
```
oliverschuerch, line 51: I know this is the exact width specified in Figma. But please can we ask the designers if we can use an even number like 80px, because:
1. It feels odd to use such a random number
2. The 73px are based on the trigger width in Figma which is using "DE" as the selected language. But with a different language, the trigger would have a different width for sure.

oliverschuerch (changes requested):

PR #4767, opened 2025-02-24
fix(components,styles): `post-header` margins and global buttons shrink on scroll

leagrdv:


packages/components/src/components/post-header/post-header.scss
```diff
@@ -59,6 +59,13 @@
   display: flex;
   align-items: center;
   gap: 24px;
+
+  @include media.min(lg) {
+    // targets the right links and language switch to hide them on scroll
+    &:nth-child(2) {
+      transform: scale(calc(35 / max(var(--header-scroll-top), 35)));
+    }
+  }
```
gfellerph, line 68: This is a good idea, but it leaves us with the issue that the controls are still focusable. Earlier, I was thinking that we should not make them focusable by hiding them with `display: none`, but that leaves us with the issue that we don't know what to do with these controls when they are only partially visible.

I think a better idea would be to only apply the sticky behaviour when the focus is not inside the global navigation. When the page is scrolled, header is sticky, and the user navigates to the language switch via keyboard, we should show the header in full to make sure everything is properly visible. I think that could be done through CSS alone, making it more reliable.

```css
.global-nav:focus-within {
  // Reset all sticky styles
}

// Or the opposite
.global-nav:not(:focus-within) {
  // Sticky styles
}
```

gfellerph (changes requested):

packages/components/src/components/post-header/post-header.scss
```diff
@@ -14,6 +14,25 @@
...
+  --global-controls-top: calc(min((var(--header-scroll-top) / 4), 20px) * -1);
+
+  // If focus is within the header (except the logo and the post-mainnavigation elements), the header should have be shown fully
+  &:has(.global-sub > *:not(.logo):focus-within),
+  &:has(.local-sub:focus-within) {
+    --global-header-top: 0;
+    --local-header-top: var(--global-header-height);
+    --logo-height: var(--global-header-height);
+    --global-controls-top: 0;
+  }
```
oliverschuerch, line 35: I think we moved these variables to the styles package now?
Can you check that and implement your update there instead of adding these variables back to the component styles?

oliverschuerch (changes requested):

oliverschuerch (approved):

packages/components/src/components/post-header/post-header.scss
```diff
@@ -14,6 +14,25 @@
...
+  --global-header-top: calc(
+    (var(--global-header-height) - var(--global-header-reduced-height)) * -1
+  );
+  --local-header-top: calc(
+    -1 * (var(--local-header-height) - var(--main-navigation-height)) + var(--global-header-reduced-height)
+  );
+  --logo-height: calc(var(--global-header-height) - var(--header-scroll-top));
+  --global-controls-top: calc(min((var(--header-scroll-top) / 4), 20px) * -1);
+
+  // If focus is within the header (except the logo and the post-mainnavigation elements), the header should have be shown fully
```
gfellerph, line 28:
```suggestion
  // If focus is within the header (except the logo and the post-mainnavigation elements), the header should be shown fully
```

.changeset/many-cherries-destroy.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-components': patch
+'@swisspost/design-system-styles': patch
+---
+
+Adjusted paddings and hid top global header buttons on scroll in the `post-header`.
```
gfellerph, line 6: I don't think this appropriately describes what's happening here. How about:

```suggestion
Improved header accessibility by ensuring that elements which are hidden on scroll get visible again if they receive focus.
```

gfellerph (approved):

PR #4777, opened 2025-02-25
chore(components): fixed the scrolling behaviour in `post-mainnavigation` to show all the nav items correctly

schaertim:


alizedebray (commented):
Can you add e2e tests for the resize observer please? It can be done in a follow up ticket if needed.

PR #4783, opened 2025-02-25
fix(styles, components): update the icon sizes in the `post-header` and `post-language-switch` components to match the design specs  

oliverschuerch:


pnpm-lock.yaml
leagrdv, line 1: Could you remove that file?

leagrdv (changes requested):

leagrdv (approved):

gfellerph (approved):

PR #4785, opened 2025-02-25
chore(styles): apply header list styles to buttons

myrta2302:


packages/styles/src/components/header/index.scss
```diff
@@ -84,7 +89,16 @@ post-header {
...
+  > ul.list-inline li button {
+    @include button-mx.reset-button;
+
+    &:hover {
+      color: tokens.get('link-hover-fg');
+    }
+  }
+
   a,
+  > ul.list-inline li button,
```
alizedebray, line 101: Does this selector need to be this specific?

.changeset/nervous-glasses-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Updated header list styles to also be applied to buttons.
```
alizedebray, line 5:
```suggestion
Updated header button styles to match the header anchor styles.
```

alizedebray (changes requested):

packages/styles/src/components/header/index.scss
myrta2302: It was targeting also the language switch buttons.

leagrdv:
Approved on my side

packages/styles/src/components/header/index.scss
alizedebray: Ar you sure? The selector bellow is `&:not(post-mainnavigation *, post-language-switch *)` so it should exclude the language switch buttons.

myrta2302: You are right! It is needed only in the previous selector.   

```
button:not(post-mainnavigation *, post-language-switch *) {
    @include button-mx.reset-button;
```

alizedebray (approved):

leagrdv (approved):

PR #4826, opened 2025-02-28
revert: reset the netlify config to point to the next netlify page again

oliverschuerch:
Since netlify unblocked our next url again, we can revert the netlify.config.json back to what we had before we switched to the temp url.


leagrdv:
Approved

gfellerph (approved):

PR #4828, opened 2025-02-28
feat(documentation): deep link to specific component configuration

myrta2302:
This PR goal is to add to the storybook 'docs' mode the ability to link from a component page to another one with specific args set. (e.g. button-secondary). With the current PR changes the args are applied to the component via the use of a global decorator which takes the URL params and updates the args. (e.g.  this link will show the secondary button: http://localhost:9000/?path=/docs/eb78afcb-ce92-4990-94b6-6536d5ec6af4--docs&args=variant:btn-secondary). 

A 'Copy variation' button is also added, to copy the current configuration link to the clipboard.


packages/documentation/.storybook/preview.ts
```diff
@@ -81,6 +83,10 @@ const preview: Preview = {
             title: 'View full screen',
             onClick: openFullScreenDemo,
           },
+          {
+            title: 'Copy variation',
```
gfellerph, line 87: I think this is a bit clearer, variation could also mean "copy the code for this variation". It could also be "Copy link to current config", but it starts to get a bit long.

```suggestion
            title: 'Copy link',
```

packages/documentation/.storybook/preview.ts
```diff
@@ -109,5 +115,5 @@ const preview: Preview = {
     },
   },
 };
-
+//just to wake it up
```
gfellerph, line 118: What does this mean?

packages/documentation/.storybook/preview.ts
```diff
@@ -81,6 +83,10 @@ const preview: Preview = {
             title: 'View full screen',
             onClick: openFullScreenDemo,
           },
+          {
+            title: 'Copy variation',
+            onClick: copyStoryConfigUrl,
```
gfellerph, line 88: Is there the possibility to shortly display a different title to get feedback on the interaction?

-> Initially: "Copy link"
-> On successfully copying a link: "Copied!"
-> After 3 seconds: "Copy link" again

packages/documentation/.storybook/helpers/with-url-params.ts
```diff
@@ -0,0 +1,74 @@
...
+    const observer = new MutationObserver((_, obs) => {
+      const scrollTarget = document.querySelector(elementId)?.closest('.docs-story');
+      if (scrollTarget) {
+        setTimeout(() => {
+          scrollTarget.scrollIntoView({ behavior: 'smooth' });
+        }, 700);
+        obs.disconnect();
+      }
+    });
+    observer.observe(document, { childList: true, subtree: true });
```
gfellerph, line 32: This is a really expensive junk of JS to replace something that is really easily done with hash refs (`https://web.site#id-of-the-element-to-scroll-to`). Since there is a known issue with this, we could just append a hash ref to the copied link and wait for the other issue to resolve itself. What do you think?

gfellerph (changes requested):

packages/documentation/.storybook/helpers/with-url-params.ts
myrta2302: I agree, i will update it.

packages/documentation/.storybook/preview.ts
```diff
@@ -105,5 +111,5 @@ const preview: Preview = {
     },
   },
 };
-
+//just to wake it up
```
gfellerph, line 114:
```suggestion
```

gfellerph (approved):
Awesome quality of life improvement!

PR #4829, opened 2025-02-28
fix(components): fix post-header mobile menu height if scrollParent is not document.body

oliverschuerch:
Depends on #4823 

- This pr should fix the mobile menu height in our post-header docs page.
- It fixes also the mobile menu second level position, if the user opens the menu first level, scrolls down and then opens the second level.
- If the user opens the mobile menu first and second level, then closes the menu with the toggle button, the second level should slideUp together with the first level, before it gets closed. Before, we had the first level sliding up and the second level sliding to the right at the same time, which looked weird.


packages/components/src/components/post-header/post-header.tsx
```diff
@@ -307,15 +305,22 @@ export class PostHeader {
...
+
+    if (this.megadropdownOpen) {
+      navigationClasses.push('megadropdown-open');
     }
 
     return (
-      <div class={navigationClasses.join(' ')}>
+      <div
+        class={navigationClasses.join(' ')}
+        style={{ '--header-navigation-scroll-top': `${mobileMenuScrollTop}px` }}
```
alizedebray, line 322: This will be updated only when the `renderNavigation` is called so it would have an incorrect value if you just scroll the page.

alizedebray (changes requested):
It would be good to add tests to ensure there is no regression with future PRs.

packages/components/src/components/post-header/post-header.tsx
oliverschuerch: This is intentionally, because the custom-property is only needed, if the mobile menu is open, the user scrolled down the main navigation level and then opens the second navigation level.

alizedebray: Ok! This could be renamed to something like `--header-navigation-initial-inset`, just to make it clear that it the value is ot dynamic in opposition to `--header-scroll-top` that is.

Issue #4860, opened 2025-03-03
Local header padding


alizedebray:
The local header padding is different whether the title wraps or not. 
In the currently released version the local header therefore has an incorrect height on mobile and tablet due to this padding.

This unexpected height is causing a layout shift. 

https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR?node-id=11565-26175&m=dev#1156166949

alizedebray:
**Resolution:** use padding defined for a one-line title in Figma.

PR #4861, opened 2025-03-03
chore(components): add e2e for the header layout shift

alizedebray:


alizedebray:
On hold until the header is fixed with #4860 and #4862

leagrdv:
@alizedebray Tickets #4860 and #4862 have both been closed if you want to continue working on that ticket. Since it's in draft mode, I'll remove myself from the review in the meantime :)

packages/components/cypress/fixtures/post-mainnavigation.test.html
```diff
@@ -111,17 +143,29 @@ <h3><a href="/schritt-für-schritt">Schritt für Schritt</a></h3>
               <h2 slot="megadropdown-title">Pakete title</h2>
               <post-list>
                 <h3>Pakete senden</h3>
-                <post-list-item><a href="/sch">Pakete Schweiz</a></post-list-item>
-                <post-list-item><a href="/kl">Kleinwaren Ausland</a></post-list-item>
-                <post-list-item><a href="">Waren Ausland</a></post-list-item>
-                <post-list-item><a href="">Express und Kurier</a></post-list-item>
+                <post-list-item slot="post-list-item"
```
leagrdv, line 146: Is it necessary to define the slot name for each `post-list-item`? the `post-list-item` only has one default slot.

packages/components/cypress/fixtures/post-header.test.html
leagrdv, line 1: This file seem almost identical to the `post-mainnavigation.test.html` one, minus the extra 20 links on that other one. Is there a way we could only have one file, or to at least make some parts of them common?

leagrdv (changes requested):

packages/components/cypress/fixtures/post-header.test.html
alizedebray: I removed the files and used the examples from the docs.

leagrdv (approved):

PR #4863, opened 2025-03-04
feat(styles): Added position boostrap classes in Design System

veyaromain:


leagrdv (changes requested):

alizedebray:
These classes seem redundant since we already have position utility classes: https://next.design-system.post.ch/?path=/docs/803a58e8-c734-4ad7-80a8-62da1bb29d4b--docs

The only added property is the `z-index`, can we simply create utility classes with fixed z-indexes to place something above the page content or above the header? (we already have relative z-index utility classes: #5741)

gfellerph (commented):
Dev roundtable resolution: implement z-index utilities for named z-indices and provide automigrations for previous uses of the combined utility classes to migrate them to composable utility classes.

Issue #4848, opened 2025-03-03
Expose real header height on the `:root`


gfellerph:
**Is your feature request related to a problem? Please describe it.**
For in-page navigation, hash links are being used (`<a href="#about">Go to section about</a>`). This automatically scrolls the page to the element with the linked id. Since the header is sticky, it always hides the linked to element below and the navigation feels broken (it scrolls too far down). Setting `scroll-padding-top` (https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-padding-top) resolves this issue.

**Describe the solution you'd like**
The page needs to know how high the header is (amount of pixels, not amount of cookies). The `<post-header>` component should expose it's height on the `:root` as a CSS variable so other implementations can refer to it and react to it. This value should be updated as the header changes height due to page scrolling or mobile/desktop browser resizes. As all pages profit from the `scroll-padding-top`, we should consider setting this globally.

### Tasks

- [x] Expose actual header height (of `<post-header>`) on the `:root` as a CSS variable (currently `--header-expanded-height`)
- [x] Expose visual header height on the `:root` as a CSS variable (currently `--header-height`)
- [x] Set `scroll-padding-top` by default if the header is on the page

PR #4879, opened 2025-03-04
chore(components): update back to top button to adapt to sticky header

myrta2302:


packages/documentation/src/stories/components/back-to-top/back-to-top.stories.ts
```diff
@@ -25,7 +25,7 @@ const meta: MetaComponent = {
       language="en"
     ></swisspost-internet-header>
     ${fakeContent(17)}
-    <post-back-to-top />
+    <post-back-to-top label="Back to top button" />
```
leagrdv, line 28: Not sure if it's because the previous swisspost-internet-header is not sticky, but on the preview, the back to top button is visible because it's not being hidden by the header. I think in all cases, it'd be best that the back to top button is not visible at all before we start scrolling.

leagrdv (changes requested):

packages/components/cypress/e2e/back-to-top.cy.ts
```diff
@@ -55,7 +55,7 @@ describe('Back-to-top', () => {
   describe('Accessibility', () => {
     it('Has no detectable a11y violations on load for all variants', () => {
       cy.getSnapshots('post-back-to-top');
-      cy.checkA11y('#root-inner');
+      cy.checkA11y('post-back-to-top');
```
oliverschuerch, line 58: In all the other component tests, we're using `#root-inner` as the selector to be checked. Is this a problem here or why did you change that?

packages/components/src/components/post-back-to-top/post-back-to-top.tsx
oliverschuerch, line 1: Maybe a docs issue, but since this is the only thing I can test, it seems to be buggy:
https://github.com/user-attachments/assets/bce6e8de-4334-48d5-b616-a29639418c49

packages/components/src/components/post-back-to-top/post-back-to-top.tsx
```diff
@@ -60,17 +60,43 @@ export class PostBackToTop {
   componentDidLoad() {
     window.addEventListener('scroll', this.handleScroll, false);
 
-    this.translateY = window
-      .getComputedStyle(this.host)
-      .getPropertyValue('--post-floating-button-translate-y');
+    // Get the back-to-top button top postiion
```
oliverschuerch, line 63: Please move the following calculations and sub functions in a dedicated class member (private function).
Also, are we 100% sure, we need that much JS to solve the issue. I would prefer to use as much CSS as possible to do the same thing and only use JS, if we have no other possibility.

oliverschuerch (changes requested):

packages/components/cypress/e2e/back-to-top.cy.ts
myrta2302: The test was not passing because of one of the header components. Because the #root-inner in this case includes the header.

packages/components/src/components/post-back-to-top/post-back-to-top.tsx
myrta2302: This is because when the post-header is present it is supposed to be sticky.

myrta2302: Also the PR takes as a fact that the -post-header-height is set to the root. So to actually test it the related PR https://github.com/swisspost/design-system/issues/4848 should be merged first.

oliverschuerch (approved):
Wow, now it looks very nice!

leagrdv (approved):

PR #4889, opened 2025-03-05
chore: Create web_component.md

gfellerph:
Adds an issue template for describing new web components to be added to the system.

### DIFF
--- a/.github/ISSUE_TEMPLATE/web_component.md
+++ b/.github/ISSUE_TEMPLATE/web_component.md
@@ -1,0 +1,36 @@
+---
+name: Web Component
+about: Parent issue for adding sub-issues related to this component.
+title: ''
+labels: '📦 components'
+assignees: ''
+type: 'Task'
+---
+
+### 📝 Description
+<!-- Detailed description of the component and a link to the design  -->
+
+### ♿ Accessibility
+<!-- List accessibility considerations such as ARIA attributes, focus handling, and color contrast -->
+
+### ⚙️ Requirements
+<!-- List technical requirements like responsive behaviour, animations, interactions, ... -->
+
+### 🧪 Testing
+<!-- Describe how the component should be tested -->
+
+### 💻 Proposed implementation
+<!-- If available, propose an implementation or hints that help with the implementation -->
+
+### 🔗 References
+<!-- If available reference to existing implementations in other Design Systems -->
+
+### 📃 Tasks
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/web_component.md
```diff
@@ -0,0 +1,36 @@
+---
+name: Web Component
+about: Parent issue for adding sub-issues related to this component.
```
leagrdv, line 3: The web component is already a sub-issue of the base component, right?

.github/ISSUE_TEMPLATE/web_component.md
```diff
@@ -0,0 +1,36 @@
...
+title: ''
```
leagrdv, line 4: Same thing here, maybe a `Component name [Web Component]`?

.github/ISSUE_TEMPLATE/web_component.md
gfellerph, line 3:
```suggestion
name: Component/Web Component
about: For creating a web component implementation.
```

.github/ISSUE_TEMPLATE/web_component.md
gfellerph: I'm not sure about this. For the bug template for example, the default title is "[bug]: ". When creating issues, it's very easy to overlook that the title needs to be set as well. As it already has some content, the field is validated and the issue is created with just the title "[bug]: ". If the title is empty from the start, it's impossible to forget to put one.

Also, since labels and issue type are already set, I was hoping this would be enough to distinguish the different types of tickets and we would not need an additional distinction in the title. What do you think?

.github/ISSUE_TEMPLATE/web_component.md
gfellerph: Yeah... copy & paste error ^^

.github/ISSUE_TEMPLATE/web_component.md
leagrdv: I think visually, it does help to have the `[type]:` before name of the component, as it's currently the case for `[bug]` and `[component]` tickets now. I think Sandra had also made a comment last time we discussed it that it helped her as well. Maybe it's something we could discuss tomorrow with the team, to have their opinion?

leagrdv (approved):

PR #4891, opened 2025-03-06
chore(components): review and update prop validation across component library

myrta2302:
PR Changes

1) Replaced thrown errors with console errors
2) Updated checkType function to handle NaN in case of type number
3) Removed boolean checks
4) Updated logic to:

- a) if required: if checkNonEmpty false then checkType
- b) if not required: keep checkEmptyOr.


packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -60,6 +60,21 @@ export class PostAvatar {
...
+
+  @Watch('userid')
+  validateUserid() {
+    checkNonEmpty(this, 'userid', undefined, 'warning');
+  }
+
+  @Watch('email')
+  validateEmail() {
+    checkNonEmpty(this, 'email', undefined, 'warning');
+  }
```
alizedebray, line 76: These properties are declared with the `?:` notation which means they are optional and therefore can be empty. To validate them, you can use the `checkEmptyOrType()` validator instead.

packages/components/src/components/post-breadcrumb/post-breadcrumb.tsx
```diff
@@ -29,6 +29,7 @@ export class PostBreadcrumb {
 
   @Watch('homeUrl')
   validateUrl() {
+    checkNonEmpty(this, 'homeUrl');
```
alizedebray, line 32: The `homeUrl` is mandatory here so the `checkNonEmpty()` validator is appropriate. However, to make sur the property is correctly marked as mandatory in the docs it should be declared with the `!:` notation on line 16.

packages/components/src/components/post-card-control/post-card-control.tsx
```diff
@@ -148,11 +148,31 @@ export class PostCardControl {
...
+  @Watch('value')
+  validateControlValue() {
+    checkNonEmpty(this, 'value', undefined, this.type === 'radio' ? 'error' : 'warning');
+  }
+
+  @Watch('description')
+  validateControlDescription() {
+    checkNonEmpty(this, 'description', undefined, 'warning');
+  }
+
```
alizedebray, line 165: Can you check all the properties you validated to know if they are actually mandatory or not, and adapt either the validator or the declaration? 

To know if a property is mandatory you have to check how it is used: 
- if it is used within `if` conditions then it is probably optional, declared with `?:` and validated with a `checkEmptyOr` validator,
- if it is used without a condition then it is mandatory, declared with `!:` and validated with the `checkNonEmpty()` validator.

alizedebray (changes requested):

packages/components/src/components.d.ts
alizedebray, line 1: To ensure we keep on defining the properties that way you could enable the `strictPropertyInitialization` in the tsconfig file of the components project.

packages/components/src/components/post-accordion/post-accordion.tsx
```diff
@@ -25,33 +25,33 @@ export class PostAccordion {
...
-      'headingLevel',
-      HEADING_LEVELS,
-      'The `heading-level` property of the `post-accordion` must be a number between 1 and 6.',
-    );
-
-    this.accordionItems.forEach(item => {
-      item.setAttribute('heading-level', String(newValue));
-    });
+  validateHeadingLevel() {
+    if (!checkNonEmpty(this, 'headingLevel')) {
```
alizedebray, line 29: There is a convention for functions that return a boolean: https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig Can you try and apply it here? 

Otherwise would you be able to know if the property is required or not by inspecting the `this`? If so the check empty/non empty could be done automatically in the checkers and not manually in the components.

packages/components/src/components/post-accordion/post-accordion.tsx
```diff
@@ -25,33 +25,33 @@ export class PostAccordion {
...
+      checkOneOf(
+        this,
+        'headingLevel',
+        HEADING_LEVELS,
+        'The `heading-level` property of the `post-accordion` must be a number between 1 and 6.',
```
alizedebray, line 34: Is this custom message necessary? The goal of this PR would be to remove them.

packages/components/cypress/fixtures/post-back-to-top.test.html
alizedebray, line 1: This file should probably not be deleted.

packages/styles/src/variables/components/_tag.scss
alizedebray, line 1: Is that change really part of your PR?

alizedebray (changes requested):
Is it possible to split the PR into smaller chunks? 
Reviewing it is a little complex at the moment and I am not sure all changes are wanted.

PR #4893, opened 2025-03-06
fix(documentation): stepper progress on firefox

leagrdv:


packages/documentation/src/stories/components/stepper/stepper.stories.ts
```diff
@@ -104,10 +113,17 @@ function getStepperItem(args: Args, step: string, index: number, updateStep: (ne
...
 
-
 function renderStepper(args: Args) {
   const [_, updateArgs] = useArgs();
 
+  // Hack to force a reflow when the current step changes
+  const stepperEl = document.querySelector('.stepper') as HTMLElement;
+  if (stepperEl) {
+    stepperEl.style.display = 'none';
+    console.log(stepperEl.offsetHeight);
```
gfellerph, line 123: There are other ways to make the browser flush the queue of styles to apply. Check out `queueMicrotask` or `requestAnimationFrame` (https://stackoverflow.com/questions/23253976/how-to-properly-wait-for-browser-reflow-repaint-to-finish/23254061).

packages/documentation/src/stories/components/stepper/stepper.stories.ts
```diff
@@ -104,10 +113,17 @@ function getStepperItem(args: Args, step: string, index: number, updateStep: (ne
   `;
 }
 
-
 function renderStepper(args: Args) {
   const [_, updateArgs] = useArgs();
 
+  // Hack to force a reflow when the current step changes
```
gfellerph, line 119: I'd be ok with implementing this hack, but it nags me a bit that this is fighting symptoms instead of solving the root cause.

gfellerph (changes requested):

packages/documentation/src/stories/components/stepper/stepper.stories.ts
leagrdv: The issue was indeed not only on storybook, what's weird is that after the reflow it did work. I've updated the css selector and it seem to work fine on firefox now

gfellerph (approved):

PR #4898, opened 2025-03-07
chore(components): add max-width css var on `post-popover`

leagrdv:
## 📄 Description

Added a max-width css variable on the `post-popover` component to allow users to add a max-width to the popover.


.changeset/metal-files-smell.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': patch
+'@swisspost/design-system-components': patch
+---
+
+Added a max-width customizable css variable on the `post-popover` component.
```
gfellerph, line 6:
```suggestion
Added the CSS variable `--post-popover-max-width` that enables configuring the max-width of the `post-popover` component.
```

packages/components/src/components/post-popovercontainer/post-popovercontainer.scss
```diff
@@ -35,6 +35,7 @@
     background-color: var(--post-current-palette-bg);
     border: 2px solid transparent; // Keeping the default border for HCM
     border-radius: commons.$border-radius;
+    max-width: var(--post-popover-max-width);
```
gfellerph, line 38: The post-popover.scss also has a max-width defined. Setting the variable on the popovercontainer has some side effects, namely that it's available for other components using the popovercontainer like tooltip, menu, etc. I think this specific variable should be set on the popover component specifically.

gfellerph (changes requested):

packages/components/src/components/post-popover/post-popover.scss
```diff
@@ -19,10 +22,10 @@
   padding: 0.5em;
 
   min-width: 160px;
-  max-width: 280px;
+  max-width: var(--post-popover-max-width);
```
gfellerph, line 25: How about using this rule?

```suggestion
  max-width: min(var(--post-popover-max-width, 280px), 100%);
```

**Breakdown**
- min(`x`, 100%): Ensure that max width does not exceed 100%, no matter what's set in `x`
- var(`x`, 280px): Use `--post-popover-max-width` if it's set, otherwise fall back to 280px as the default max-width

This would enable you to remove the declaration on the host and the max-width on the popovercontainer. It likely also makes the mobile declaration obsolete.

packages/documentation/src/stories/components/popover/popover.stories.ts
```diff
@@ -46,6 +47,14 @@ const meta: MetaComponent = {
         category: 'General',
       },
     },
+    maxWidth: {
+      name: 'Max width of the popover',
+      description:
+        'Value can either be in `vw`, `px` or `%`. If no max-width is defined, the popover will extend to the width of its content.',
+      table: {
+        category: 'General',
```
gfellerph, line 55: Could be helpful to know for some people.

```suggestion
        category: 'General',
        defaultValue: { summary: '280px' }
```

packages/documentation/src/stories/components/popover/popover.stories.ts
```diff
@@ -30,6 +30,7 @@ const meta: MetaComponent = {
     placement: 'top',
     arrow: true,
     title: true,
+    maxWidth: '80vw',
```
gfellerph, line 33: Should we leave this empty by default so the popover uses the default 280px max-width if nobody tweaks it?

gfellerph (changes requested):

packages/components/src/components/post-popover/post-popover.scss
leagrdv: It didn't work with the 100% as it was adding a large margin on the right if used in the `min()` but I managed to fix it using 100vw instead

gfellerph (approved):

PR #4906, opened 2025-03-10
fix(components): add scroll to megadropdown on desktop

leagrdv:
## 📄 Description

Added a max-height to the `post-megadropdown` on desktop to allow for scrolling in case the content is too long.

Also added an extra 48px of space underneath the open megadropdown, otherwise it looked kind of weird with the double scrollbar (one for the page, one for the megadropdown), and hiding the page overflow seemed like a bad idea as the megadropdown will not always take the whole space.


gfellerph (approved):
Neat solution!

PR #4907, opened 2025-03-11
chore(eslint): add migration rule for `rounded` utility classes

leagrdv:
## 📄 Description

Added a migration rule for the renaming of the `rounded` class to `rounded-4` (as well as the `rounded-*` ones).


packages/eslint/src/rules/html/migrations/no-rounded-borders.ts
```diff
@@ -0,0 +1,60 @@
...
+  { old: 'rounded-bottom', new: 'rounded-bottom-4' },
+  { old: 'rounded-start', new: 'rounded-start-4' },
+  { old: 'rounded-end', new: 'rounded-end-4' },
+];
+
+function getRemovedRoundedClassMsgs(): Record<string, string> {
+  return classesMap.reduce(
+    (o, key) =>
+      Object.assign(o, {
+        [key.old]: `The "${key.old}" class have been deleted. Please remove it or replace it with "${key.new}".`,
```
gfellerph, line 18:
```suggestion
        [key.old]: `The "${key.old}" class has been deleted. Please remove it or replace it with "${key.new}".`,
```

packages/eslint/src/rules/html/migrations/no-rounded-borders.ts
```diff
@@ -0,0 +1,60 @@
+import { createRule } from '../../../utils/create-rule';
+import { HtmlNode } from '../../../parsers/html/html-node';
+
+export const name = 'no-rounded-borders';
```
gfellerph, line 4: Not sure if that name is describing what's going on here. To me, this would rather be something like "no-unnumbered-border-radius". What do you think?

gfellerph (approved):

PR #4910, opened 2025-03-11
chore(components): fixed popover not opening when clicking on elemets nested in trigger (v9)

schaertim:
## 📄 Description

Changed the globalToggleHandler method of the `post-popover` component. Before it only checked if the event target had the 'data-popover-target' attribute. When for example an icon is nested in the trigger, this would result to false and the popover would not open. In the new implementation instead of just checking the e.target we traverse the DOM tree upwards searching for an element with the 'data-popover-target' attribute. Now clicking on an icon nested in the popover trigger, the popover will be opened/closed as expected.

## 🚀 Demo

https://github.com/user-attachments/assets/e5448e03-75c9-42cb-af49-fe726b4f1e44

---

## 📝 Checklist

### DIFF
--- a/.changeset/lazy-tables-agree.md
+++ b/.changeset/lazy-tables-agree.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Adapted the trigger of the `post-popover` component to be able to handle nested elements.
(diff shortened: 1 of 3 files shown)


gfellerph (changes requested):
Can you add an e2e test that covers this use case?

gfellerph (approved):

PR #4925, opened 2025-03-11
chore(components): expose header height to root

myrta2302:
## 📄 Description

This PR moves the `--post-header-height` (visible height) and `--post-header-expanded-height` custom properties to the `:root` via the header/index.scss. All the custom properties involved and thus also exposed to the `:root` have also been updated with the `post-` prefix.
Additionaly, the `scroll-padding-top` property is set as the value of the (visible) header height.

## 📝 Checklist

- ✔️ Existing unit tests pass locally with my changes


packages/styles/src/components/header/index.scss
```diff
@@ -9,51 +9,56 @@
...
-    --local-header-min-height: 56px;
+    --post-local-header-min-height: 56px;
   }
 
   @include media.max(md) {
-    --local-header-min-height: 48px;
+    --post-local-header-min-height: 48px;
   }
 
   &:not(:has([slot='title'])) {
```
alizedebray, line 50:
```suggestion
  &:not(:has(post-header [slot='title'])) {
```

packages/styles/src/components/header/index.scss
```diff
@@ -9,51 +9,56 @@
...
+    var(--post-global-header-reduced-height) + var(--post-main-navigation-height)
+  );
+  --post-header-height: calc(
+    var(--post-header-expanded-height) - min(
+        var(--post-header-scroll-top, 0px),
+        calc(var(--post-header-expanded-height) - var(--post-header-reduced-height))
       )
   );
-  height: var(--header-height);
+  scroll-padding-top: var(--post-header-height);
```
alizedebray, line 25: We usually use the the root for custom properties only, I would expect the `scroll-padding-top` to be set on the body instead.

alizedebray (changes requested):

oliverschuerch (changes requested):

oliverschuerch (approved):

alizedebray (approved):

PR #4934, opened 2025-03-12
fix(styles): remove placeholder mixin

leagrdv:
## 📄 Description

There was an invalid `::input-placeholder` on the placeholder mixin which I removed. The remaining `::placeholder` seemed sufficent as it is widely available. Therefore the `@mixin placeholder()` did not seem relevant anymore as it only contained the `::placeholder` CSS selector so I removed it as well.

### DIFF
--- a/.changeset/chilled-crabs-march.md
+++ b/.changeset/chilled-crabs-march.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-styles': minor
+---
+
+Removed the `@mixin placeholder` as using the `::placeholder` CSS selector is now widely available.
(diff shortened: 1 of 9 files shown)


.changeset/chilled-crabs-march.md
gfellerph, line 1: Can you add a patch changeset that says:

"Removed the usage of the deprecated pseudo-element `::input-placeholder`. This fixes an issue with CSS validation tools that don't allow deprecated selectors."

To inform the PPNL team that this is now fixed.

gfellerph (approved):
Nice!

PR #4936, opened 2025-03-12
fix(internet-header): add self admin url prop to override menu links

leagrdv:
## 📄 Description

In order to let users override the calculated menu links origin URL in the kpl login widget, a `selfAdminOrigin` prop has been added to the internet-header.


leagrdv:
Closing this one, as the internet-header package has been removed from v8. I created a new PR with the same changes targetting v9 #5941

PR #4948, opened 2025-03-13
feat(components): use native scroll in the main navigation

alizedebray:
## 📄 Description

This PR is to refactor the main navigation component using native scroll. This was made possible by the design decision of insetting the focus ring and will make the component robust and easy to maintain.

## 🚀 Demo

No visual changes are expected.

---

## 📝 Checklist


oliverschuerch (changes requested):
1. The scroll buttons do not appear immediately after a DOM mutation, but only after resizing the window.
2. I wanted to scroll to one side and kept pressing the left mouse button. Then I accidentaly pressed the right mouse button, which destroyed the functionality (probably a timeout/interval which is not cleared properly). I could only get rid of that behavior, by reloading the whole page.
This for sure is an edge case but we should test what happens if the user is pressing the scroll button, while the mouse leaves the button or as sayed, what happens if a different mouse button or a key on the keyboard is pressed at the same time.

https://github.com/user-attachments/assets/43817cfe-92f1-42b2-b84f-67ffa6760de0

oliverschuerch (changes requested):
Since you're using the native scroll, you could check if https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-type helps reducing the needed JS logic.

alizedebray:
> Since you're using native scrolling, you could check if https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-type helps reducing the needed JS logic.

I started using this CSS scroll snap , by I removed it in the end. 
What happens is that the `scrollBy()` function requires number of pixels to scroll the nav.
When the scroll snap is activated, this number of pixels should be at least half the width of the element you want to scroll to and should therefor be calculated. 
Given I had to calculate a number of pixels anyways, I though it was easier to directly calculate the correct position of the scroll instead of relying half on the Javascript calculation, half on the CSS scroll snap behavior.

oliverschuerch (changes requested):

oliverschuerch (approved):

oliverschuerch:
@alizedebray Great work!

PR #4957, opened 2025-03-17
fix(components): remove transform property from scroll-locked parent

oliverschuerch:
## 📄 Description

This needs to be done to ensure, the offset parent of all elements in the doc will still be the html/body element.

---

## 📝 Checklist


packages/styles/src/components/header/index.scss
```diff
@@ -10,7 +10,16 @@ tokens.$default-map: elements.$post-link;
 
 [data-post-scroll-locked] {
   overflow: hidden !important;
-  transform: translate3d(0, 0, 0) !important;
+
+  &:not(body) {
+    transform: translate3d(0, 0, 0) !important;
```
gfellerph, line 15: Still not sure about this. I think we should refactor the header to not rely on this to be there. What are our options here?

gfellerph (changes requested):

alizedebray (approved):

PR #4971, opened 2025-03-18
fix(styles): implement consistent local header padding on mobile for wrapped and not wrapped titles

alionazherdetska:
## 📄 Description

This PR addresses an issue where the `.local header` padding behaves differently depending on whether the title wraps or not:
      1. Updated padding-block for `.local-header-mobile-extended`.
      2. Modified padding values for interactive elements to align with design specs.

## 📝 Checklist


packages/styles/src/components/header/index.scss
```diff
@@ -140,6 +140,14 @@ post-header {
     }
   }
 
+  a:not(post-mainnavigation *, post-language-switch *) {
+    display: flex;
+
+    @include media.max(sm) {
+      padding: 4px 10px;
```
alizedebray, line 147: I guess this padding should be applied to buttons as well, you can probably include it to this mixin: https://github.com/swisspost/design-system/blob/main/packages/styles/src/components/header/_mixins.scss#L3

alizedebray, line 147: I guess this padding should be applied to buttons as well, you can probably include it to this mixin: https://github.com/swisspost/design-system/blob/main/packages/styles/src/components/header/_mixins.scss#L3

alizedebray (approved):

PR #4978, opened 2025-03-19
chore(components-angular): update package.json to delete .angular (cache) on `clean`.

myrta2302:
## 📄 Description

Added .angular to "clean": "rimraf dist .angular" in package.json.


.changeset/twelve-ghosts-vanish.md
oliverschuerch, line 1: Since this is only important for us, a changeset is not needed/wanted.

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #5098, opened 2025-03-25
chore(tests): added header tests simulating react navigation behaviour

schaertim:
## 📄 Description
Added two tests simulating react navigation behavior by removing and reattaching the header from the DOM. The first test checks that the mobile menu is correctly closed after the header is reattached and the second one checks that the scroll lock is added and released at the correct times.

There was no way to test the correct removal of event listeners. I tried all the following approaches:

- Use the getEventListener available in chrome dev environment
- Spy on function calls to Add & RemoveEventListener
- Overwrite Add & RemoveEventListener with trackable version
- Check heap space usage to detect memory leaks

## 🚀 Demo

Not really helpful here.
---

## 📝 Checklist


packages/components/cypress/e2e/header.cy.ts
```diff
@@ -0,0 +1,42 @@
...
+        const headerElement = $header[0];
+        headerElement.remove();
+        cy.document().then(doc => {
+          doc.body.prepend(headerElement);
+        });
+      });
+      cy.get('div.local-header-mobile-extended').should('not.exist');
+    });
+
+    it.skip('should release scroll lock after reattaching header', () => {
```
leagrdv, line 24: You forgot to remove the `.skip` here

packages/components/cypress/e2e/header.cy.ts
```diff
@@ -0,0 +1,42 @@
...
+      cy.get('[data-post-scroll-locked]').should('not.exist');
+      cy.get('post-togglebutton').click();
+      cy.get('[data-post-scroll-locked]').should('exist');
+
+      // Remove and re-attach the header
+      cy.get('@header').then($header => {
+        const headerElement = $header[0];
+        headerElement.remove();
+        cy.get('[data-post-scroll-locked]').should('not.exist');
```
leagrdv, line 33: I think you can skip that step since you're already checking it after, and then you could make a common function for this header part to share between this test and the previous one

leagrdv (dismissed):

leagrdv:
Approved :)

oliverschuerch (approved):

PR #5099, opened 2025-03-25
chore: update pnpm to the latest version

alionazherdetska:
## 📄 Description

This PR updates the project’s package manager from pnpm@9.15.3 to pnpm@10.6.5.

To address issues caused by the upgrade:

✅ Added eslint-plugin-react to fix ESLint config resolution in documentation

✅ Added cypress as a dev dependency to ensure the binary installs correctly. The problem was introduced after upgrading to pnpm@10, which no longer runs dependency lifecycle scripts (like cypress install) by default. https://github.com/cypress-io/cypress/issues/31111

---

## 📝 Checklist


leagrdv:
Approved

packages/documentation/package.json
```diff
@@ -68,6 +68,7 @@
     "eslint": "9.18.0",
     "eslint-plugin-markdown": "5.1.0",
     "eslint-plugin-mdx": "3.1.5",
+    "eslint-plugin-react": "7.37.4",
```
oliverschuerch, line 71: Does this really belong to this pr?

package.json
```diff
@@ -125,6 +125,9 @@
         "zone.js": ">=0.14"
       }
     },
-    "overrides": {}
+    "overrides": {},
+    "onlyBuiltDependencies": [
```
oliverschuerch, line 129: Why is this necessary?

oliverschuerch (changes requested):

package.json
oliverschuerch: Can you please test what happens if you delete all node_modules along with all output folders in all packages (dist, www, loader, loaders, storybook-static, components-react/src/components/stencil-generated, icons/public/post-icons, etc.):

- [ ] Does the command `pnpm bootstrap` still work?
- [ ] Does the command `pnpm start` still work?
- [ ] Does the command `pnpm e2e` still work?
- [ ] Does the command `pnpm unit` still work?
- [ ] Does the command `pnpm lint` still work?

oliverschuerch (changes requested):

PR #5100, opened 2025-03-25
fix(styles): datepicker arrows size and disabled state

leagrdv:


oliverschuerch (changes requested):
The Arrows look good in general, but when I hove one of them, it shows, that the buttons are not square. I'd like it, if the buttons would be always exact squares.
What do you think?
image

oliverschuerch (approved):

PR #5102, opened 2025-03-25
chore(setup): added cypress eslint plugin to enable linting of cypress folders and files

schaertim:
## 📄 Description

Installed the cypress eslint plugin (https://www.npmjs.com/package/eslint-plugin-cypress) globally and added the config to the eslint config files of the packages using cypress. I also fixed all the linter errors which appeared after activating the linting.

## 🚀 Demo

Not applicable.
---

## 📝 Checklist


packages/internet-header/cypress/downloads/downloads.htm
leagrdv, line 1: What is this file for and why did it move?

packages/internet-header/cypress/e2e/breadcrumbs.cy.ts
```diff
@@ -3,7 +3,7 @@ import rawTestConfiguration from '../fixtures/internet-header/test-configuration
 import { prepare } from '../support/prepare-story';
 import { BREADCRUMBS } from './shared/variables';
 
-const testConfiguration: IPortalConfig = rawTestConfiguration as any;
+const testConfiguration: IPortalConfig = rawTestConfiguration as unknown as IPortalConfig;
```
leagrdv, line 6: This is not really clean as it forces a type that might not be correct.
For instance, `loginWidgetId` on the header's model of the configuration is defined as `string`, but it's set to `null` on the rawTestConfiguration.
We use typing to be able to detect these issues and `as unknown as MyType` is pretty much the same as any at the end. Could you try and fix the typing issues instead?

packages/internet-header/cypress/e2e/language-switch-2.cy.ts
```diff
@@ -130,9 +130,9 @@ describe('language-switch-2', () => {
     });
 
     it('should not be rendered with only one language', () => {
-      let config: IPortalConfig = <any>{ en: testConfiguration.en };
+      const config: IPortalConfig = { en: testConfiguration.en } as unknown as IPortalConfig;
```
leagrdv, line 133: Same here

packages/internet-header/cypress/e2e/meta-navigation.cy.ts
```diff
@@ -93,7 +93,7 @@ describe('meta-navigation', () => {
     describe('navMeta config not set', () => {
       it(`removes meta navigation`, () => {
         // Cast the imported JSON object to the IPortalConfig interface
-        let config: IPortalConfig = <any>testConfiguration;
+        const config: IPortalConfig = testConfiguration as unknown as IPortalConfig;
```
leagrdv, line 96: On this file as well

leagrdv (changes requested):

packages/internet-header/cypress/support/prepare-story.ts
```diff
@@ -4,7 +4,7 @@ import mockAuth from '../fixtures/internet-header/auth.json';
 import { IPortalConfig } from '../../src/models/general.model';
 
 export const installInterceptors = (
-  config: Object = testConfiguration,
+  config: object = testConfiguration,
```
leagrdv, line 7: Shouldn't the type here be IPortalConfig rather than object?

packages/internet-header/cypress/support/prepare-story.ts
```diff
@@ -14,7 +14,7 @@ export const installInterceptors = (
 
 interface PrepareOptions {
   loggedIn?: boolean;
-  config?: Object;
+  config?: object;
```
leagrdv, line 17: Same here

leagrdv (changes requested):

packages/internet-header/cypress/e2e/breadcrumbs.cy.ts
schaertim: I discuessed this with philipp and he said that we cannot change the interfaces without risking stuff breaking and for simplicity reasons it does also not make sense to adapt the json.

package.json
```diff
@@ -107,7 +107,8 @@
   },
   "devDependencies": {
     "@changesets/cli": "2.27.11",
-    "start-server-and-test": "2.0.9"
+    "start-server-and-test": "2.0.9",
+    "eslint-plugin-cypress": "4.2.0"
```
leagrdv, line 111: Since you've added the plugin to the other package.json files, it can be removed from here.

leagrdv (changes requested):

leagrdv (commented):
Approved

packages/documentation/package.json
```diff
@@ -83,7 +83,8 @@
     "storybook": "8.4.7",
     "typescript": "5.8.2",
     "typescript-eslint": "8.20.0",
-    "vite": "6.2.6"
+    "vite": "6.1.3",
```
oliverschuerch, line 86: Is there a special reason, why you downgrade vite here?
If not, update it to "6.2.6" as it was on "main"

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #5105, opened 2025-03-26
fix(styles): updated the post-icon-scss-mixin to take the package version into account

alionazherdetska:
## 📄 Description

This update ensures that the post-icon mixin uses the correct version of the design-system-icons package in the generated URLs, improving consistency and reducing the risk of mismatched asset versions.

What changed:
      1. Added a new Gulp task generate-icon-version-scss to generate a SCSS utility file (_post-icon-version.scss) containing the current package.json version.
      2. Updated the post-icon mixin in _icons.scss to dynamically reference the icon version using the new $post-icon-version variable.
      3. Included the new Gulp task in the default build pipeline.

## 📝 Checklist


packages/styles/src/utilities/_post-icon-version.scss
```diff
@@ -0,0 +1 @@
+$post-icon-version: '9.0.0-next.32';
```
leagrdv, line 1: Should this generated file be added? I feel like it should be ignored on git

leagrdv (commented):
Approved

gfellerph (approved):

PR #5109, opened 2025-03-27
chore(components): improved URL handling for `post-icon` component

schaertim:
## 📄 Description

Improved URL handling in `post-icon` component:

I changed the order of which "sources" are used to determine the origin and the slug. Now the order is the following: base property > base tag > data-post-icon-base meta attribute. This means that if we have both a base tag and a meta attribute present the url built using the information in the base tag first and only if either the origin/slug part are missing it takes them from the meta attribute. 
I also changed how a root path ("/") is interpreted. Before it would not count as a slug and would therefore be ignored.

## 🚀 Demo

---

## 📝 Checklist


packages/components/cypress/e2e/icon.cy.ts
```diff
@@ -177,7 +177,7 @@ describe('Icon', () => {
       );
     });
 
-    it('should use "meta[name="design-system-settings"]" tag for "slug", before "base[href]" tag', () => {
+    it('should use "base[href]" tag for "slug", before "meta[name="design-system-settings"]" tag', () => {
```
leagrdv, line 180: Your test result uses the base[href] **instead** of the meta one, not **before**.

On the ticket:
_In general, icon URLs should be constructed as:
current domain + base[href] + meta[name="design-system-settings"][data-post-icon-base]) (overridden by a potential base property set on the component)_
Which means this test should result in a `*/base/tag/meta/tag/*` instead of having one override the other.

packages/components/cypress/e2e/icon.cy.ts
```diff
@@ -177,7 +177,7 @@ describe('Icon', () => {
...
       cy.get('@meta')
```
leagrdv, line 181: Also, it might be nice to add a test or two with a mix of relative and absolute path between `base[href]` and `meta` to make sure it works correctly.

leagrdv (changes requested):

pnpm-lock.yaml
leagrdv, line 1: Why did this file changed? Otherwise, PR is approved :)

leagrdv:
Approved

packages/components/src/components/post-icon/post-icon.tsx
```diff
@@ -104,71 +98,60 @@ export class PostIcon {
...
+      : '';
+    const isBaseHrefAbsolute = /^https?:\/\//.test(baseHref);
+
+    const iconBase =
+      this.base ||
+      (IS_BROWSER
+        ? document
+          .querySelector('meta[name="design-system-settings"]')
+          ?.getAttribute('data-post-icon-base') || ''
+        : '');
```
oliverschuerch, line 125: Extract that to a standalone variable for readability:

```ts
const metaIconBase = IS_BROWSER ? document.querySelector('meta[name="design-system-settings"]')?.getAttribute('data-post-icon-base') || '' : '';
```

packages/components/src/components/post-icon/post-icon.tsx
```diff
@@ -104,71 +98,60 @@ export class PostIcon {
...
+    if (baseHref) {
+      normalizedBaseHref = baseHref.endsWith('/') ? baseHref : `${baseHref}/`;
     }
 
-    return url;
-  }
+    let normalizedIconBase = '';
+    if (iconBase) {
+      normalizedIconBase = iconBase.endsWith('/') ? iconBase : `${iconBase}/`;
+    }
```
oliverschuerch, line 136: This could be simplified with a sub-function:

```ts
const normalizedBaseHref = normalizeUrl(baseHref);
const normalizedIconBase = normalizeUrl(iconBase);

function normalizeUrl(url: string) {
  if (!url) return '';
  return url.endsWith('/') ? url : `${url}/`; 
} 
```

oliverschuerch (changes requested):

oliverschuerch (approved):

leagrdv (approved):

PR #5165, opened 2025-04-01
refactor(styles): generate not-defined selectors by using the components/src/components/ subfolder names

oliverschuerch:
## 📄 Description

This PR handles the follwing things:
1. It removes the `not-defined.scss` definition from the styles package.
2. It adds a global style output in the components package which includes the same logic. The `:not(:defined)` selector has been  updated, to ensure, prehydrated ssr components can show up from the very beginning. All csr components will only become visible, as soon as they have been hydrated. It's possible there is still room for optimization. I've used [this](https://codepen.io/oliverschuerch/pen/mydZoEE) codepen to test how the selector works.
3. The global style output is manually added to the `components-angular` and `components-react` packages, where I tried to make sure it can be used from the release package. However, there is for sure room for optimization here.
4. I tryed to set up needed scripts in our `package.json` files, so we can still seemlessly develop. Since this was not the case before, there meight be some cases, when one has to prebuild a local dependency, before another package can be started.
However, I tested the following things:
- `pnpm components (play|start|build)`
- `pnpm components-react build`
- `pnpm components-angular (start|build)`
- `pnpm docs:start`
- `pnpm nextjs:start`

---

## 📝 Checklist


oliverschuerch:
TODO: create issue to document how to use the global style output!

oliverschuerch:
I do not expect, this pr will be accepted as is, that's why I added both of you as reviewers.

packages/components-angular/package.json
```diff
@@ -27,12 +29,14 @@
     "zone.js": "0.15.0"
   },
   "devDependencies": {
+    "@swisspost/design-system-components-angular": "workspace:9.0.0-next.33",
```
oliverschuerch, line 32: This is only used for the consumer-app and does not end up in the lib package.json.

.changeset/old-teachers-deliver.md
```diff
@@ -0,0 +1,14 @@
...
+'@swisspost/design-system-components': major
+---
+
+Removed the `not-defined.scss` file, which controlled when web components become visible:
+
+- for server-side rendered components: immediately
+- for client-side rendered components: after hydration
+
+On the other side, we introduced the global styles file `post-components.css` in all component packages, which repalces the styling logic which was defind in the `not-defined.scss` in our styles package before.
+This file must be included in any project using our component packages to ensure proper functionality of the web components.
```
gfellerph, line 14:
```suggestion
Moved the file responsible for preventing flashes of unstyled content for components from the styles to the components package. The flashes occur when non-hydrated components are rendered before JavaScript has loaded on a client. These styles hide the non-hydrated component, preventing the flash.

**Changes:**
Removed: `@swispost/design-system-styles/components/not-defined.(scss|css)`
Added: `@swisspost/design-system-components/post-components/post-components.css`
Added: `@swisspost/design-system-components-angular/post-components.css`
Added: `@swisspost/design-system-components-react/post-components.css`
```

packages/components/prebuild.ts
gfellerph, line 1: Please add tests for this file. [`mock-fs`](https://www.npmjs.com/package/mock-fs) can be used to mock a file system and check the expected output.

packages/components/src/styles/utilities/not-defined.scss
```diff
@@ -0,0 +1,7 @@
+/* Components rendered on the client-side only, need to be invisible, as long as they match the css selector :not(:defined), so they are not showing up, before they got hydrated. */
+@use 'sass:list';
+@use '../generated/component-names' as *;
+
+:where(#{$component-names}):not([data-hydrated]):not(:defined) {
```
gfellerph, line 5:
```suggestion
:where(#{$component-names}):not([data-hydrated], :defined) {
```

packages/components/stencil.config.ts
```diff
@@ -9,6 +9,7 @@ export const config: Config = {
   namespace: 'post-components',
   buildDist: true,
   sourceMap: false,
+  globalStyle: 'src/styles/index.scss',
```
gfellerph, line 12: Is that even necessary? If I remember correctly, this will be applied with JS and thus take no effect.

packages/components/tsconfig.ts-node.json
gfellerph, line 1: explain

gfellerph (changes requested):

packages/components/stencil.config.ts
oliverschuerch: This is the definition of the globalStyle entry file. If we do not declare it, stencil does not know there is a global styles file which should be processed and therefore will not output a global styles file.

packages/components/tsconfig.ts-node.json
oliverschuerch: The ts-node script command in the package.json file only worked, after setting the module to commonjs. I tried to work around that without luck, so I added this tsconfig file specific for ts-node.
I will try again to get rid of it...

oliverschuerch: I was now able to get rid of this config file. But since I changed the tsconfig according to the stencil docs, there are still some configs that need to be different for ts-node to work. Have a look into the tsconfig.json file, I hope this will clearify why this is necessary. If not, ask me again 😉

gfellerph:
I think this is good to go after fixing the conflicts. I'll remove my assignment however.

gfellerph (approved):

PR #5174, opened 2025-04-02
chore: group only patch updates

gfellerph:
## 📄 Description

As the system dependencies grow, it becomes less and less likely, that a grouped minor and patch update PR can be merged without issues. This change would only group patch updates into one PR and submit minor updates individually, keeping the existing exclusion rules in place.

It also updates the label assigned to renovate PRs to not mess with our package labels.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": ["config:recommended"],
-  "labels": ["📦 {{{parentDir}}}"],
+  "labels": ["⛓️ dependencies"],
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -14,18 +14,19 @@
...
+        "matchPackageNames": [
+          "*",
+          "!typescript",
+          "!bootstrap",
+          "!@storybook/*",
+          "!@angular/*",
+          "!@stencil/*"
+        ]
+      },
     },
```
oliverschuerch, line 30: Remove the outer object.

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #5180, opened 2025-04-02
refactor(styles): align file structure for global header and footer styles

oliverschuerch:
## 📄 Description
This pr prepares the file structure in the styles package for the future, when we'll add more global styles for the mainnavigation and the megadropdown components.

---

## 📝 Checklist


oliverschuerch:
@gfellerph I think we need to inform the cargo team about this change, because they meight import the post-footer file which is now in a different folder.

.changeset/real-cars-mate.md
```diff
@@ -0,0 +1,10 @@
+---
+'@swisspost/design-system-styles': major
+---
+
+Updated the file structure for the `post-footer` component global styles:
+
+- dist/components/globals/post-footer.scss -> dist/components/footer/index.scss
+- dist/components/globals/post-footer.css -> dist/components/footer/index.css
```
gfellerph, line 8:
```suggestion
- dist/components/globals/post-footer.scss -> dist/components/footer/index.[s]css
```

gfellerph (approved):

.changeset/real-cars-mate.md
oliverschuerch: This feels like a custom notation imo.
If we change this, wouldn't it be better to write it like so:
```md
- dist/components/globals/post-footer.(css|scss) -> dist/components/footer/index.(css|scss)
```
???

PR #5262, opened 2025-04-10
4791:The migration guide mentions that some variables not used for the post button were removed or refactored.

hugomslv:
## 📄 Description

Append removal of old button variables to the migration guide

## 🚀 Demo

---

## 📝 Checklist

### DIFF
--- a/packages/documentation/src/stories/misc/migration-guide/migrationv8-9.component.ts
+++ b/packages/documentation/src/stories/misc/migration-guide/migrationv8-9.component.ts
@@ -128,6 +128,49 @@ export class MigrationV89Component extends LitElement {
                   </p>
                 </li>
               </ul>
+              <h5>post-button</h5>
+              <ul>
+                <li class="mb-16">
+                  <p>
+                    Some <code>post-button</code> scss variables have been removed
+                    <span class="tag tag-sm tag-danger">breaking</span>
+                  </p>
+                  <ul>
+                    <li><code>$btn-height-sm</code></li>
+                    <li><code>$btn-height-md</code></li>
+                    <li><code>$btn-disabled-opacity</code></li>
+                    <li><code>$btn-focus-width</code></li>
+                    <li><code>$btn-focus-box-shadow</code></li>
+                    <li><code>$btn-icon-size-sm</code></li>
+                    <li><code>$btn-icon-size-md</code></li>
+                    <li><code>$btn-icon-size-rg</code></li>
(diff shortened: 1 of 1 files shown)


gfellerph:
@hugomslv thanks for the PR! We're using [conventional commit formatting](https://www.conventionalcommits.org/en/v1.0.0/) for our PR titles. The current title can be changed to something like this to pass the failing check: `chore: add missing migration entry for removed button variables` (chore because it's not a bugifx, not a new feature and not a breaking change).

oliverschuerch:
@alizedebray and @leagrdv Since I did not work on the migration guide at all, I wanted to ask you if you think this is something we just missed to document or if we planned to do so in a different ticket?

leagrdv:
> @alizedebray and @leagrdv Since I did not work on the migration guide at all, I wanted to ask you if you think this is something we just missed to document or if we planned to do so in a different ticket?

The documenting is what is asked on this ticket. The PR #4702 had removed those variables and the removal needs to be documented in the migration.

I think the removal ticket might have been done too early and the variables that were refactored will probably end up being deleted at the end, once we update the timepicker, so I don't think I would document those. What do you think @oliverschuerch and @alizedebray ?

Also, some (if not all) of the button variables that were deleted are overrides of the bootstrap variables. Since we'll remove bootstrap at some point, all of those variables overrides (there are more than a hundred) will also be removed and I doubt it'd be a good idea to document each of them as the migration guide will just be endless and it will just discourage developers to migrate.

oliverschuerch:
@leagrdv 
> ... some (if not all) of the button variables that were deleted are overrides of the bootstrap variables. Since we'll remove bootstrap at some point, all of those variables overrides (there are more than a hundred) will also be removed and I doubt it'd be a good idea to document each of them as the migration guide will just be endless and it will just discourage developers to migrate.

That's totally true. At the moment it seems legit to add the detailed information about the removed button variables, but as soon as we get rid of bootstrap (completely) we would need to mention houndreds of removed variables, etc.

I would not like to do that in such a detail. But it could still be good to collect all scss variables coming from bootstrap in a list (e.g. bootstrap scss variables, no longer available) so DEVs can easily check if a variable has been removed.

What do you think?

gfellerph:
@hugomslv, we've discussed this in the Dev Roundtable and decided to document the removal of Bootstrap in general in the migration docs and don't list all of the variables individually. There's documentation from bootstrap that lists all the variables and we can just refer to that one.

PR #5268, opened 2025-04-10
chore(components): fix logo disappearing on page locale change

myrta2302:
## 📄 Issue Description

The logo disappearing issue is caused by the missing custom property `--post-header-scroll-top` within the `--logo-height` calculation. The property has been so far set in the connectedCallback. 

## Testing
When tested within the Cargo implementation:

a) on first load it works as expected
b) on language /country change the page is re-rendered but the post-header connectedCallback seems to be called before the React renders (tested with useEffect() inside the PostHeader component) and the custom property is unset.  
c) When moved to componentWillRender the custom property is set successfully.

## Proposed Solution

This PR moves the handleScrollEvent function from the connectedCallback to the componentWillRender method. Additionally, a fallback value is added for the `--post-header-scroll-top` within the `--logo-height` calculation.

## 📝 Checklist

### DIFF
--- a/.changeset/new-spoons-exercise.md
+++ b/.changeset/new-spoons-exercise.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Updated `post-header` by moving handleScrollEvent, handleScrollParentResize and lockBody methods from connectedCallback to componentDidRender. Added fallback value to the `--post-header-scroll-top` property within the `--logo-height` calculation.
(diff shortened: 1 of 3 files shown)


oliverschuerch:
> ...
> ## Proposed Solution
> This PR moves the handleScrollEvent, handleScrollParentResize, lockBody functions from the connectedCallback to the componentDidRender method.
> ...

This will probably cause a rerendering and therefore impact the components performance.
Do we need to move all of the three function from `connectedCallback` to `componentDidRender`?

And also, wouldn't the `componentDidLoad` be the better hook?
Because `componentDidRender` is called also after every rendering update, while `componentDidLoad` is called only once.

If moving these functions is the only solution, I'm fine with it, I just think we need to optimize everything related as far as possible.

oliverschuerch (approved):

PR #5281, opened 2025-04-14
fix(component): sanitized and encoded config request parameters in `post-internet-header` to prevent injection

alionazherdetska:
## 📄 Description

Addressed a potential injection vulnerability in the Internet Header's config.json request mechanism by validating and encoding the input parameters (projectId, environment, and version).

## 📝 Checklist

### DIFF
--- a/packages/internet-header/cypress/e2e/header.cy.ts
+++ b/packages/internet-header/cypress/e2e/header.cy.ts
@@ -42,4 +42,194 @@ describe('header', () => {
         });
     });
   });
+
+  describe('header security', () => {
+    context('malicious input handling', () => {
+      it('should prevent malicious project IDs from modifying the config request URL', () => {
+        const maliciousProjectIds = [
+          'project-id-1?malicious=1&extra=2',
+          'project-id-1/../../../malicious', 
+          'project-id-1%3Fextra=malicious',
+          'project-id-1"><script>alert(1)</script>',
+          'project-id-1\';DROP TABLE users;--'
+        ];
+
+        maliciousProjectIds.forEach(maliciousProjectId => {
+          cy.intercept('GET', '**/api/**', (req) => {
+            throw new Error('API call should not be made with invalid project ID');
+          }).as('configRequest');
+          
+          cy.window().then(win => {
+            const headerElement = win.document.createElement('swisspost-internet-header');
+            headerElement.setAttribute('project', maliciousProjectId);
+            headerElement.setAttribute('environment', 'int01');
+            
+            cy.on('uncaught:exception', (err) => {
+              expect(err.message).to.include('Internet Header project key is invalid');
+              return false;
+            });
+            
+            win.document.body.appendChild(headerElement);
+            
+            cy.wait(500).then(() => {
+              cy.get('@configRequest.all').should('have.length', 0);
+              headerElement.remove();
+            });
+          });
+        });
+      });
+
+      it('should prevent malicious environment values from affecting the config request', () => {
+        const maliciousEnvironments = [
+          'PROD; rm -rf /',
+          'INT01?injected=true',
+          '<script>alert(1)</script>',
+          '../../../etc/passwd'
+        ];
+
+        maliciousEnvironments.forEach(maliciousEnv => {
+          cy.window().then(win => {
+            const headerElement = win.document.createElement('swisspost-internet-header');
+            headerElement.setAttribute('project', 'valid-project');
+            headerElement.setAttribute('environment', maliciousEnv);
+            
+            cy.on('uncaught:exception', (err) => {
+              expect(err.message).to.include('Internet Header environment');
+              expect(err.message).to.include('is not valid');
+              return false;
+            });
+            
+            win.document.body.appendChild(headerElement);
+            
+            cy.wait(500).then(() => {
+              headerElement.remove();
+            });
+          });
+        });
+      });
+
+      it('should prevent malicious language parameters from being injected', () => {
+        const maliciousLanguages = [
+          'de?malicious=1',
+          'fr"><script>alert(1)</script>',
+          'en;DROP TABLE users;--',
+          'it%00',
+          '../../../etc/passwd'
+        ];
+
+        maliciousLanguages.forEach(maliciousLang => {
+          cy.changeArg('language', maliciousLang);
+          
+          cy.get('swisspost-internet-header').then(async el => {
+            const currentLang = await el[0].getCurrentLanguage();
+            if (currentLang !== maliciousLang) {
+              expect(['de', 'fr', 'it', 'en']).to.include(currentLang);
+            }
+          });
+        });
+      });
+
+      it('should properly encode custom parameters to prevent injection', () => {
+        const maliciousInputs = {
+          languageCookieKey: 'cookie"><script>alert(1)</script>',
+          languageLocalStorageKey: 'storage?malicious=true'
+        };
+      
+        cy.window().then(win => {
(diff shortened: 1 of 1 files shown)


packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
```diff
@@ -152,10 +152,28 @@ export class PostInternetHeader {
...
+
   constructor() {
     if (this.project === undefined || this.project === '' || !isValidProjectId(this.project)) {
       throw new Error(
-        `Internet Header project key is "${this.project}". Please provide a valid project key.`,
+        `Internet Header project key is "${this.project}". Please provide a valid project key.`
+      );
+    }
+    
+    const validEnvironments = ['prod', 'int01', 'dev'];
```
gfellerph, line 168: These are not all valid environments. Have a look at https://github.com/swisspost/design-system/blob/main/packages/internet-header/src/models/general.model.ts#L39.

gfellerph (changes requested):
Please add a test case where a "malicious" or invalid payload is trying to modify the URL of the config request where we can see that these changes are preventing the malicious request from being sent. I think it will be valuable in the future to understand the methods used to manipulate the config requests so we can react to them not only in the header, but in other places where we send requests based on properties on a web component.

packages/internet-header/cypress/e2e/header.cy.ts
gfellerph, line 1: Solid suite of tests!

packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
```diff
@@ -460,13 +570,14 @@ export class PostInternetHeader {
     const renderLanguageSwitch = config.header.navLang.length > 1;
 
     const initialLogoScale = renderMetaNavigation ? getLogoScale(this.host) : '1';
+    const safeVersion = this.isValidVersion(packageJson.version) ? packageJson.version : 'unknown';
```
gfellerph, line 577: I don't think this check is effective here. The `packageJson.version` imported from the actual package.json file cannot be manipulated. I think the issue here is, that after the initial render with the correct version, the version is tampered with and when analytics grabs the value, it grabs the tampered value. Not really an issue for us after all and certainly not something we could fix. Should we revert the changes here to keep code simpler?

packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
```diff
@@ -151,15 +151,34 @@ export class PostInternetHeader {
...
+    if (!validEnvironments.includes(sanitizedEnvironment)) {
       throw new Error(
-        `Internet Header project key is "${this.project}". Please provide a valid project key.`,
+        `Internet Header environment "${newValue}" is not valid. Please use one of: ${validEnvironments.join(', ')}`
       );
     }
   }
 
+  constructor() {}
+
```
gfellerph: Empty constructors are not necessary. Please remove this.

```suggestion
```

packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
```diff
@@ -151,15 +151,34 @@ export class PostInternetHeader {
...
-  constructor() {
-    if (this.project === undefined || this.project === '' || !isValidProjectId(this.project)) {
+  @Watch('project')
+  handleProjectChange(newValue: string) {
+    if (!isValidProjectId(newValue)) {
+      throw new Error(
+        `Internet Header project key is invalid. Please provide a valid project key.`
+      );
+    }
+  }
```
gfellerph, line 179: A project change requires a page reload (and is not a real scenario, since one project only ever has one project id). Throwing an error here might suggest that changing project id on the fly is supported. I'd remove this check.

packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
```diff
@@ -151,15 +151,34 @@ export class PostInternetHeader {
   private lastWindowWidth: number = window.innerWidth;
   private updateLogoAnimation: () => void;
   private scrollParent: Element | Document;
+  
+  private isValidVersion(str: string): boolean {
+    if (typeof str !== 'string') return false;
+    return /^[a-z0-9.-]+$/.test(str);
+  }
```
gfellerph, line 161: Possibly not needed since we can't really validate the version at the right time.

packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
alizedebray:
```suggestion
```

packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
```diff
@@ -174,48 +193,132 @@ export class PostInternetHeader {
...
+      throw new Error(
+        `Internet Header environment "${this.environment}" is not valid. Please use one of: ${validEnvironments.join(', ')}`
+      );
+    }
+
+    if (!this.isValidVersion(packageJson.version)) {
+      throw new Error(
+        `Invalid data-version format: "${packageJson.version}". Version should only contain lowercase letters, numbers, and dashes.`
+      );
+    }
```
alizedebray: Can you call the `handleProjectChange` and `handleEnvironmentChange` methods here to avoid code duplication?

packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
```diff
@@ -174,48 +193,132 @@ export class PostInternetHeader {
...
+    this.validateAndSetEnvironment();
+    this.validateAndSetLanguage();
+    this.parseLanguageSwitchOverrides();
+    this.parseOsFlyoutOverrides();
+    this.parseCustomConfig();
+  }
+
+  private validateAndSetEnvironment() {
+    const validEnvironments = ['dev01', 'dev02', 'devs1', 'test', 'int01', 'int02', 'prod'];
+    const sanitizedEnvironment = this.environment.toLowerCase();
```
alizedebray: This is also a duplicate. Can you check the whole file?

alizedebray (changes requested):

packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
```diff
@@ -174,48 +201,127 @@ export class PostInternetHeader {
...
+      throw new Error(
+        `Internet Header project key is invalid. Please provide a valid project key.`
+      );
+    }
+
+    if (!this.isValidEnvironment(this.environment)) {
+      throw new Error(
+        `Internet Header environment "${this.environment}" is not valid. Please use one of: ${PostInternetHeader.VALID_ENVIRONMENTS.join(', ')}`
+      );
+    }
```
alizedebray, line 215: If you give a default value to the parameter of the `handleEnvironmentChange` function like so: 
```ts
  handleEnvironmentChange(newValue = this.environment) {
```

You can completely reuse it in the `componentWillLoad`, intead of duplicating the error message: 
```suggestion
    this.handleEnvironmentChange();
```

It's usually how we do it in our components, here the example of the `post-banner`: 
https://github.com/swisspost/design-system/blob/main/packages/components/src/components/post-banner/post-banner.tsx#L87

alizedebray (changes requested):

alionazherdetska:
Closing this PR as outdated and reopening the issue. When I was working on this issue, I have noticed that I luck the knowledge about `post-internet-header` to be able to resolve the issue.

PR #5288, opened 2025-04-15
fix(docs):  fixed the accessibility issue in the Form labels section under Guidelines by properly associating labels with radio inputs

alionazherdetska:
## 📄 Description

This update improves the accessibility of the radio button examples in the Form Labels guidelines by replacing plain text labels with `<label>` elements that reference input IDs using for.

## 📝 Checklist


gfellerph (commented):
Looking at this page, there are some more things that should be fixed and I think it would be ok to add them to this PR:
1. This example contains an unnecessary div. Even if it's a wrong example, that should not be there

  image
2. I think it would be good to use the actual Design System Radios here, the examples look kind of broken
3. The text above the radio examples has some HTML entities that should be decoded 

  image

gfellerph (approved):

PR #5311, opened 2025-04-16
chore: update fetch-icons workflow

alizedebray:
## 📄 Description

This PR is to update the fetch-icons workflow so that the icon PR body and change sets are correctly written.

## 🚀 Demo

The icon PR's will now look like this: https://github.com/swisspost/design-system/pull/5310
(the example PR is just to see the description and change sets, other changes are not relevant)

---

## 📝 Checklist


.github/workflows/scripts/icons.mjs
gfellerph, line 1: Would be nice to add tests to each of these functions so we can be sure that the output is correct.

gfellerph (approved):

.github/workflows/scripts/icons.mjs
alizedebray: I created a ticket for this: https://github.com/swisspost/design-system/issues/5314
We would then probably need a new package

PR #5321, opened 2025-04-17
chore(tokens): transform px to rem for font-sizes

leagrdv:
## 📄 Description

## 🚀 Demo

---

## 📝 Checklist


packages/tokens/_build/style-dictionary.js
```diff
@@ -50,6 +50,34 @@ StyleDictionary.registerTransform({
...
+    const usesDtcg = token.$type && token.$value;
+    return token[usesDtcg ? '$type' : 'type'] === PX_TO_REM_TOKEN_TYPE;
+  },
+  transform: token => {
+    const usesDtcg = token.$type && token.$value;
+    const baseFontSize = 16;
+    let value = token[usesDtcg ? '$value' : 'value'];
+    if (typeof value === 'string' && value.includes('px')) {
+      // Remove 'px' and convert the result to a number
+      value = parseFloat(value.replace('px', ''), 10);
```
oliverschuerch, line 75: [parseFloat](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/parseFloat) only knows one parameter which need to be of type string.
And since it is parsing a string searching for a number, the `value.replace('px', '')` command is obsolete, because if `value` iquals `10px` it will return `10` as a unitless number anyway.

```suggestion
      value = parseFloat(value);
```

packages/tokens/_build/style-dictionary.js
```diff
@@ -50,6 +50,34 @@ StyleDictionary.registerTransform({
...
+StyleDictionary.registerTransform({
+  name: 'swisspost/px-to-rem',
+  type: 'value',
+  filter: token => {
+    const usesDtcg = token.$type && token.$value;
+    return token[usesDtcg ? '$type' : 'type'] === PX_TO_REM_TOKEN_TYPE;
+  },
+  transform: token => {
+    const usesDtcg = token.$type && token.$value;
+    const baseFontSize = 16;
```
oliverschuerch, line 71: Should we define this as a constant like the `PX_TO_REM_TOKEN_TYPE` in the constants.js file as well?
What do you think?

oliverschuerch (changes requested):
At first I thought, we need to update also the tailwind and utilities config files and add the new transformer function there too. But since they are only outputting component-level tokens that point to the core tokens, which are transformed in the configs/all.js file already, this seems not to be necessary.

packages/tokens/_build/style-dictionary.js
```diff
@@ -50,6 +50,34 @@ StyleDictionary.registerTransform({
...
+StyleDictionary.registerTransform({
+  name: 'swisspost/px-to-rem',
+  type: 'value',
+  filter: token => {
+    const usesDtcg = token.$type && token.$value;
+    return token[usesDtcg ? '$type' : 'type'] === PX_TO_REM_TOKEN_TYPE;
+  },
+  transform: token => {
+    const usesDtcg = token.$type && token.$value;
+    const baseFontSize = BASE_FONT_SIZE;
```
oliverschuerch, line 71: Unnecessary, but ok.

PR #5338, opened 2025-04-23
test(documentation): added basic e2e tests for the new palette implementation

oliverschuerch:
## 📄 Description

This PR implements some basic e2e test for the new palettes integration.
However, they are not complete, but are meant to stabilize the current implementation so that we can figure out if anything goes horribly wrong during further development.

---

## 📝 Checklist

### DIFF
--- a/packages/documentation/src/shared/snapshots/schemes.ts
+++ b/packages/documentation/src/shared/snapshots/schemes.ts
@@ -19,7 +19,7 @@ export function schemes(renderFn: (scheme: string) => TemplateResult, options: I
     ${schemes.map(
       scheme => html`
         <div data-color-scheme="${scheme}">
-          <div class="p-16 palette palette-default">
+          <div class="palette palette-default">
             <p>Color Scheme: ${scheme}</p>
             ${renderFn(scheme)}
           </div>


packages/documentation/src/shared/snapshots/schemes.ts
```diff
@@ -17,8 +17,6 @@ export function schemes(renderFn: (scheme: string) => TemplateResult, options: I
   return html`${[...additionalSchemes, ...Object.values(COLOR_SCHEMES)]
     .filter(filter)
     .map(
-      scheme => html` <div data-color-scheme="${scheme}" class="p-16 palette-default">
-        ${renderFn(scheme)}
-      </div>`,
+      scheme => html` <div data-color-scheme="${scheme}" class="p-16">${renderFn(scheme)}</div>`,
```
alizedebray, line 20: It would be good to test how the palettes behave without a defined `data-color-scheme`, I have seen this article about dark mode in Cypress: https://www.cypress.io/blog/test-your-web-app-in-dark-mode

packages/documentation/.storybook/preview-body.html
```diff
@@ -1 +1,3 @@
-<script src="/assets/scripts/storybook-preview-events.js"></script>
+<body data-color-scheme="light">
```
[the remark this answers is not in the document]
oliverschuerch: It's not necessary, but since our docs styles are not ready yet to be used in dark mode, and people (like me) prefer dark over light mode, we should keep this attribute on the html element until we actively decide to allow a dark mode in our docs.

Btw: our stories can still be watched in dark mode by switching the menu in the docs top bar, that you've once created.
image

packages/documentation/src/shared/snapshots/schemes.ts
oliverschuerch: You're right, we should test preferred colors too and using chrome with the command line flag `--force-dark-mode=true` could be an easy solution to do so.

packages/tokens/tokensstudio-generated/tokens.json
alizedebray, line 1: Is this change expected? Shouldn't it be changes in tokens studio?

oliverschuerch: I think they need to be resetted anyway, because Sandra is about to define the colors from scratch.

packages/documentation/src/shared/snapshots/schemes.ts
```diff
@@ -14,13 +14,18 @@ export const COLOR_SCHEMES = {
...
-        <p>Color Scheme: ${scheme}</p>
-        ${renderFn(scheme)}
-      </div>
-    </div>`,
-  )}`;
+  return html`
+    ${schemes.map(
+      scheme => html`
+        <div data-color-scheme="${scheme}">
+          <div class="p-16 palette palette-default">
```
alizedebray:
```suggestion
          <div class="palette palette-default">
```

The snapshots sometimes need to be full-screen, for example the full-width tabs.
Adding an option to remove the padding would also work.

alizedebray (approved):

PR #5339, opened 2025-04-23
fix(styles): restore validation icons on browser-autocompleted form fields

alionazherdetska:
## 📄 Description

- Added dedicated `.validation-icon` span element
- Background images were substituted with mask images on these elements

## 🚀 Demo

---

## 📝 Checklist


packages/documentation/src/stories/components/forms/validation/validation.stories.ts
```diff
@@ -123,6 +122,7 @@ export const Input: Story = {
         placeholder="Placeholder"
       />
       <label class="form-label" for="Input_1${props.validationState}${props.scheme}">Label</label>
+      <span class="validation-icon"></span>
```
gfellerph, line 125: Changing the structure of the HTML for HTML/CSS components is a major breaking change because every project has to update every input field in their whole project for the changes to take effect. It would be great to have a solution that does not alter the markup required.

Since input fields cannot have pseudo elements, I was thinking that we maybe can put a pseudo element on the wrapper element around each input field and use `:has` to determine the validity state:

```css
.form-input:has(input[type="text"].is-invalid)::before { content ''; background-image: /*invalid icon*/}
```

gfellerph (changes requested):

packages/styles/gulpfile.js
```diff
@@ -235,15 +235,17 @@ gulp.task(
...
-      'generate-not-defined-components-scss',
-      'map-icons',
-      'copy',
-      'generate-icon-version-scss',
-      'autoprefixer',
-      'transform-package-json',
-    ),
-    gulp.series('temporarily-copy-token-files', 'sass'),
+  gulp.series(
+    'generate-icon-version-scss',
```
oliverschuerch, line 239: I know this has been done in a former PR, but I'm asking myself, if we should rename the task `generate-icon-version-scss` and also the output file name and path.

The package version is used in the post-icon mixin only (at the moment).
Actually, it's just a global information that could be used from any file. But for sure it is not similar to our other scss utilities modules and has nothing in common with them.

IMO it would match much better, if we'd rename the gulp task to something like `prebuild-env-vars` and the output file to something like `environment.scss` or `env-variables.scss`.

@alionazherdetska what do you think?

oliverschuerch (changes requested):

packages/styles/gulpfile.js
alionazherdetska: Renaming to `prebuild-env-vars` and `environment.scss` is a good idea since leaves room for adding other environment variables in the future.

oliverschuerch (approved):

gfellerph (approved):

PR #5341, opened 2025-04-24
chore(setup): upgraded `styles:build` task to use turborepo

schaertim:
## 📄 Description

This PR is the result of the POC for Turborepo. It adds turborepo as a dependency and upgrades the `styles:build` command to make us of local computation caching. To enable the caching the build task was defined in the turbo.json file with all the outputs which need to be cached for the specific package.

## 🚀 Demo
To test local caching, run pnpm styles:build twice in a row :)

---

## 📝 Checklist


turbo.json
```diff
@@ -0,0 +1,13 @@
+{
+  "tasks": {
+    "build": {
+      "dependsOn": ["^build"]
+    },
+    "icons#build": {
+      "outputs": ["public/post-icons/**", "public/report.json", "public/report-min.json"]
+    },
```
gfellerph, line 8: If we're focusing on the styles package, this should not be here, no? Or is it necessary for the styles build? If so, the tokens might be missing.

gfellerph (changes requested):

turbo.json
schaertim: I thought it would be needed since the style build also builds the icons. I removed it and the caching seems to work fine.

schaertim:
Here is the preview link: https://preview-5377--swisspost-design-system-next.netlify.app/?path=/docs/edfb619b-fda1-4570-bf25-20830303d483--docs

turbo.json
oliverschuerch: If at all, the styles package build script is creating a scss-icon-map, but this has nothing to do with the icons package.

gfellerph (approved):

PR #5350, opened 2025-04-26
chore(components): update language switch accessible names and roles

myrta2302:
## 📄 Description
This PR initial goal was to fix non conformity with the https://www.w3.org/WAI/WCAG22/Understanding/label-in-name accessibility guideline. Instead of modifying the `aria-label` content this PR replaces the `aria-labels` set on the contained button elements with descriptive visually hidden span texts. List variant 'role' attrs are modified to correctly recognize list items. An info box is added in the language switch page of the documentation to inform about proper usage of the `name` language option prop.

## 📝 Checklist


.changeset/stale-dogs-hope.md
alionazherdetska, line 1: Maybe you could split it into two changesets:

1. **'@swisspost/design-system-components': patch**
Enhanced accessibility of `post-language-switch` component by replacing `aria-labels` with `visually-hidden` spans and properly assigning semantic `role` attributes.

2. **'@swisspost/design-system-documentation': patch**
Added comprehensive accessibility guidance to `post-language-switch` documentation with practical examples showing how to correctly match visible text with the `name` prop to ensure speech recognition compatibility.

packages/documentation/src/stories/components/language-switch/language-switch.docs.mdx
```diff
@@ -1,21 +1,32 @@
...
 <Canvas sourceState="shown" of={languageSwitchStories.Default} />
 
 <Controls of={languageSwitchStories.Default} />
 
+<PostBanner type="info">
+  <p>
+    Please note that for each language switch option, a readable accessible `name` must be determined so that assistive technologies can clearly announce the specific language.
+
+  </p>
+</PostBanner>
```
alionazherdetska, line 28: To make it even more helpful for developers, we could possibly add examples:

Suggestion:
image
@alizedebray Do you find useful to add it?

packages/components/src/components/post-language-switch/post-language-switch.tsx
```diff
@@ -108,13 +108,18 @@ export class PostLanguageSwitch {
...
 
   private renderList() {
     return (
-      <Host data-version={version} role="list" aria-label={this.caption}>
-        <div class="post-language-switch-list" role="group" aria-label={this.description}>
+      <Host data-version={version}>
+        <div
+          class="post-language-switch-list"
+          role="list"
+          aria-label={this.caption + this.description}
```
alionazherdetska, line 121: Directly concatenating these strings without any separator might cause screen readers to read them as one run-on phrase, potentially creating confusion and harming accessibility.

**Suggestion:** 
```tsx
aria-label={`${this.caption}: ${this.description}`}

File: packages/components/src/components/post-language-switch/post-language-switch.tsx [hunk 3]
```diff
@@ -125,8 +130,9 @@ export class PostLanguageSwitch {
     return (
       <Host data-version={version}>
         <post-menu-trigger for={this.menuId}>
-          <button class="post-language-switch-trigger" aria-label={this.description}>
+          <button class="post-language-switch-trigger">
             {this.activeLang}
+            <span class="visually-hidden">{this.caption + this.description}</span>
```
alionazherdetska, line 135: Directly concatenating these strings without any separator might cause screen readers to read them as one run-on phrase, potentially creating confusion and harming accessibility.

**Suggestion:** 
```tsx
aria-label={`${this.caption}: ${this.description}`}

File: packages/documentation/src/stories/components/language-switch/language-switch.docs.mdx (hunk 1 above)
Comment (myrta2302) [reply packages/documentation/src/stories/components/language-switch/language-switch.docs.mdx:28]: I think that this suggestion (e.g. De-Deutch) would definitely be applicable if the `name` was set via the `aria-label` in the button/a elements. Now that it is set in the `visually-hidden` span, i think it's enough to communicate that this must a description of the specific language and that it has to be descriptive of the visible text.

File: packages/components/src/components/post-language-switch/post-language-switch.tsx (hunk 2 above)
Comment (myrta2302) [reply packages/components/src/components/post-language-switch/post-language-switch.tsx]: According to @alizedebray, concatenated texts are not a good practice, since the text could be anything and the result very unpredictable. Thus, the code was updated to no longer make use of concatenated texts.

File: .changeset/stale-dogs-hope.md
Comment (myrta2302) [reply .changeset/stale-dogs-hope.md:1]: Nice idea!  :)

File: packages/components/src/components/post-language-switch/post-language-switch.scss [hunk 4]
```diff
@@ -61,3 +64,9 @@ tokens.$default-map: components.$post-button;
     }
   }
 }
+
+.post-language-switch-list {
+  .visually-hidden {
```
leagrdv: Does this `.visually-hidden` class need to be set twice in this file?

leagrdv (changes requested):

PR #5354, opened 2025-04-28
chore(components, styles, internet-header): prefix CSS variables with `post-`

leagrdv:
## 📄 Description

## 🚀 Demo

---

## 📝 Checklist


.changeset/fast-mayflies-argue.md
gfellerph, line 1: The internet header is in a kind of "don't release major versions" state right now. As soon as we're ready to go forward with a new config based header, we'll completely refactor how that package is working. Basically, we'll use the components from the components package and a config from PPNL to construct the config based header. The package and the `<swisspost-internet-header>` tags will stay the same to reduce breaking changes as much as possible. Until then, I'd like to avoid any breaking change. Also, the internet header package is currently not aligned in versions with the other packages. That will change as well, so the internet-header will jump from 1.x to 10.x or so.

I'd therefore revert any changes to the internet header package in this commit, since we'll not use the current internal components any more.

gfellerph (changes requested):
Please revert changes to the internet header package, other things would be good to go.

PR #5376, opened 2025-04-30
chore: update renovate config

gfellerph:
## 📄 Description

Updates renovate config to use more pre-defined grouping rules from https://docs.renovatebot.com/presets-group/. This should make it easier to handle the many dependencies that are necessary for our integration testing packages.

I've also re-ordered some of the props so each ruleset uses the same order.


renovate.json
```diff
@@ -24,47 +43,28 @@
...
     {
-      "matchUpdateTypes": ["minor", "patch"],
-      "groupName": "Angular",
-      "matchPackageNames": ["@angular/*"]
-    },
-    {
-      "matchFileNames": ["packages/demo/package.json"],
-      "allowedVersions": "~18.17",
-      "matchPackageNames": ["@types/node"]
+      "groupName": "Stylelint major",
```
alizedebray, line 60:
```suggestion
      "groupName": "Stylelint",
```

renovate.json
```diff
@@ -24,47 +43,28 @@
         "!@storybook/*",
         "!@angular/*",
         "!@stencil/*"
```
alizedebray, line 45: You probably need to exclude `tailwindcss`, `eslint`, and `nextjs` as well.

gfellerph:
```suggestion
        "!@storybook/*",
        "!@angular/*",
        "!@stencil/*",
        "!tailwindcss",
        "!@eslint/*",
        "!next"
```

alizedebray (approved):

PR #5377, opened 2025-04-30
chore(docs): include `raw-components` and `health` stories in the local build

schaertim:
## 📄 Description

## 🚀 Demo

---

## 📝 Checklist

### DIFF
--- a/.changeset/rotten-groups-crash.md
+++ b/.changeset/rotten-groups-crash.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': minor
+---
+
+Only include stories located in the `raw-components` or `health` directory in local builds. This will also prevent them from showing up in the search of the deployed version.
--- a/packages/documentation/.storybook/main.ts
+++ b/packages/documentation/.storybook/main.ts
@@ -5,6 +5,26 @@ import { mergeConfig } from 'vite';
 import rehypeSlug from 'rehype-slug';
 import rehypeAutolinkHeadings from 'rehype-autolink-headings';
 
+const isDev = process.env.NODE_ENV === 'development';
+
+const allowedDirs = [
+  'components',
+  'foundations',
+  'guidelines',
+  'misc',
+  'modules',
+  'packages',
+  'templates',
+  'utilities',
+];
+
+const stories = isDev
+  ? ['../src/stories/**/*.mdx', '../src/stories/**/*.stories.@(ts|tsx)'] // include all stories (local)
+  : allowedDirs.flatMap(dir => [
+      `../src/stories/${dir}/**/*.stories.@(ts|tsx)`,
+      `../src/stories/${dir}/**/*.mdx`,
+    ]); // don't include helath and raw-componets stories (int/prod)
+
 const config: StorybookConfig = {
   logLevel: 'info',
   core: {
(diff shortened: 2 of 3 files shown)


oliverschuerch (changes requested):

.changeset/rotten-groups-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': minor
+---
+
+Only show `healt` and `raw-components` stories in development environment and filter them from the search in production mode.
```
oliverschuerch:
```suggestion
Only show `health` and `raw-components` stories in development environment and filter them from the search in production mode.
```

packages/documentation/.storybook/manager.ts
```diff
@@ -1,11 +1,44 @@
...
 
-if (process.env.NODE_ENV) document.documentElement.setAttribute('data-env', process.env.NODE_ENV);
+const storedDevMode = localStorage.getItem('devModeEnabled');
 
+let initialEnv;
+if (storedDevMode !== null) {
+  initialEnv = JSON.parse(storedDevMode) ? 'development' : 'production';
+} else {
+  initialEnv = process.env.NODE_ENV || 'production';
+}
```
oliverschuerch: While your solution is totally fine, that's how I would write it (personal preference), because it uses less code and you can read it stright forward.
It's totally ok to keep your implementation if you prefer it!

```suggestion
let initialEnv  = process.env.NODE_ENV || 'production';

if (storedDevMode !== null) {
  initialEnv = JSON.parse(storedDevMode) ? 'development' : 'production';
}
```

packages/documentation/.storybook/manager.ts
oliverschuerch, line 1: Just did a little research and found out, that it should be possible to add whatever we want in the [meta.parameter](https://storybook.js.org/docs/writing-stories/parameters#story-parameters) object.
```ts
// @storybook/web-components/dist/index.d.ts

interface Parameters {
    [name: string]: any;
}

type Meta = {
  ...
  /**
   * Custom metadata for a story.
   * @see [Parameters](https://storybook.js.org/docs/writing-stories/parameters)
   */
  parameters?: Parameters;
  ...
}
```
What do you mean, would this be the better place to specify `devOnly`?
I'd expect parameters would be available in the filter function beside tags like so `item.parameters`.

```ts
// any.stores.ts

const meta: MetaComponent = {
  ...
  parameters: {
    customSidebar: {
      itemFilter: 'devOnly'
    }
  },
  ...
};
```

```ts
// manager.ts

const excludeDevOnlyFilter = (
  item: API_PreparedIndexEntry & { status: Record<string, API_StatusObject | null> },
): boolean => !item.parameters?.customSidebar?.ItemFilter === 'devOnly';
```

oliverschuerch: If you feel it's not worth to try it, or if you're facing any difficulties, just leave it as it is, I'm totally fine with your current implementation!

oliverschuerch (changes requested):

packages/documentation/.storybook/manager.ts
schaertim: I think I would stick to the tags for two reasons:

Performance - Tags are available at the index level so filtering is instant, while parameters require loading full story data for every story just to check a boolean flag.
Simplicity - Tags are purpose-built for filtering and keep the implementation clean without needing API dependencies or addon registration complexity in the manager.ts file.

oliverschuerch (approved):

PR #5424, opened 2025-05-05
feat(styles, components): target group html/css component

leagrdv:
## 📄 Description

This PR adds the target group html/css component with its documentation.

## 🚀 Demo

---

## 📝 Checklist

### DIFF
--- a/packages/documentation/src/stories/components/target-group/target-group.stories.ts
+++ b/packages/documentation/src/stories/components/target-group/target-group.stories.ts
@@ -1,0 +1,84 @@
+import type { Args, StoryObj } from '@storybook/web-components';
+import { html } from 'lit/static-html.js';
+import { MetaComponent } from '@root/types';
+
+const meta: MetaComponent = {
+  id: 'f321d60b-cf7e-4a64-90af-7b90e6869ee6',
+  title: 'Components/Target group',
+  tags: ['package:HTML'],
+  parameters: {
+    badges: [],
+    design: {
+      type: 'figma',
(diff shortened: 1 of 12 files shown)


oliverschuerch (commented):
Since this component is adding styles specially meant to work when used in the post-header (at least on mobile devices), we should document it under "raw-components" in my opinion.

packages/documentation/src/stories/components/target-group/target-group.stories.ts
```diff
@@ -0,0 +1,84 @@
...
+type Story = StoryObj;
+
+export const Default: Story = {
+  render: (args: Args) => {
+    return html`
+      <ul class="target-group">
+        ${[args.label_1, args.label_2, args.label_3].map(
+          (arg, i) => html`
+            <li>
+              <a href="#" class="${i === args.active ? 'active' : ''}">${arg}</a>
```
oliverschuerch, line 77: When clicking on the links on the docs page, the page is loading an iframe url.
You could avoid that, by adding a story decorator, which adds a click event listener and then call event.preventDefault() within.

packages/documentation/src/stories/modules/header/header.stories.ts
```diff
@@ -42,6 +43,16 @@ const meta: MetaComponent = {
...
+    targetGroup: {
+      name: 'Target group',
+      description: 'Whether or not the target group buttons are visible.',
+      control: {
+        type: 'boolean',
+      },
+      table: {
+        category: 'Content',
+      },
+    },
```
oliverschuerch, line 55: This arg is not used in the render function, so it does not hide/show the target group at all.

oliverschuerch (changes requested):

PR #5428, opened 2025-05-05
chore(styles): remove unused placeholder + cleanup

leagrdv:
## 📄 Description

I noticed the `%btn-transparent-background` was never used so I removed it and cleaned up the other placeholder files to remove unnecessary imports.

## 🚀 Demo

---

## 📝 Checklist

### DIFF
--- a/.changeset/loose-jeans-hug.md
+++ b/.changeset/loose-jeans-hug.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-styles': minor
+---
+
+Removed the unused placeholder `%btn-transparent-background`.
--- a/packages/documentation/src/stories/misc/migration-guide/migrationv8-9.component.ts
+++ b/packages/documentation/src/stories/misc/migration-guide/migrationv8-9.component.ts
@@ -575,6 +575,12 @@ export class MigrationV89Component extends LitElement {
                 </li>
                 <li class="mb-16">
                   <p>
+                    The <code>%btn-transparent-background</code> CSS placeholder has been removed.
+                    <span class="tag tag-sm tag-danger">breaking</span>
+                  </p>
+                </li>
+                <li class="mb-16">
+                  <p>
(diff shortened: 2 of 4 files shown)


oliverschuerch (changes requested):
In general, I'm fine with the update, but...

We never documented this placeholder, which means it's interal code that should not be in use by any project. Therefore, removing it, because it's not used anymore, would not be a changeset worthy update, imo.

If I'm right, we would not need to add a changeset and we would also not need to document it in the migration guide as a breaking change.

If I'm wrong, it should be `major` instead of `minor`.

Please double check with @gfellerph or @alizedebray

alizedebray:
> In general, I'm fine with the update, but...
> 
> We never documented this placeholder, which means it's interal code that should not be in use by any project. Therefore, removing it, because it's not used anymore, would not be a changeset worthy update, imo.
> 
> If I'm right, we would not need to add a changeset and we would also not need to document it in the migration guide as a breaking change.
> 
> If I'm wrong, it should be `major` instead of `minor`.
> 
> Please double check with @gfellerph or @alizedebray

I don’t have a strong opinion because it's probably not used at all, but I would consider it a major change since something was removed. It will be grouped with other major changes in v10 anyway. Also, since it was never documented, it doesn't need to be mentioned in the migration guide.

.changeset/loose-jeans-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': major
+---
+
+Removed the unused placeholder `%btn-transparent-background`.
```
oliverschuerch: If this has never been documented in the past, we do not need to add a changeset. Removing internal code is not causing any verion bumps at all.

leagrdv: I deleted it :)

oliverschuerch (approved):

PR #5466, opened 2025-05-09
fix(components): remove back button rotation on `post-header`

leagrdv:
## 📄 Description

When the header was first developed, we didn't have as many icons available and the `arrowleft` icon was missing. Therefore we had used the `arrowright` icon with a 180 deg rotation on whatever the user was putting in the back button slot.

Now that we have the correct icon, this PR removes the 180deg rotation so that users can use the correct icon which is `arrowleft`.

### DIFF
--- a/.changeset/brave-garlics-lead.md
+++ b/.changeset/brave-garlics-lead.md
@@ -1,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': minor
+'@swisspost/design-system-components': minor
+---
+
+Removed the rotation of the back button in the `post-header`, which means users need to change the back button's icon `arrowright` to `arrowleft` to make sure it's pointing in the right direction.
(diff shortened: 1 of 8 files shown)


.changeset/brave-garlics-lead.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': minor
+'@swisspost/design-system-components': minor
```
alizedebray, line 3: You simply updated the docs but did not add a feature or page, therefore it is a simple patch. 
On the contrary, this change to the header will require projects to update there code once its published so it's a breaking change for the components package. 
```suggestion
'@swisspost/design-system-documentation': patch
'@swisspost/design-system-components': major
```

alizedebray (approved):

PR #5475, opened 2025-05-12
chore(documentation): minor updates for `back to top`, `card-control` and `popover`.

myrta2302:
## 📄 Description

- Back-top-top: Removed unnecessary`fullscreen` property.
- Popover: Added missing `close-button-caption` attribute 

## 📝 Checklist

### DIFF
--- a/.changeset/clever-bears-pay.md
+++ b/.changeset/clever-bears-pay.md
@@ -2,5 +2,5 @@
 '@swisspost/design-system-documentation': patch
 ---
 
-Added `close-button-caption` attribute in `post-popover`.
+Added `close-button-caption` attribute in `post-popover` component.
 Removed an obsolete parameter from the `post-back-to-top` component metadata object.
--- a/.changeset/clever-bears-pay.md
+++ b/.changeset/clever-bears-pay.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': patch
+---
+
+Implemented minor changes in `back-to-top`, `card-control` and `popover` documentation pages.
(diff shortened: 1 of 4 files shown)


.changeset/clever-bears-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': patch
+---
+
+Implemented minor changes in `back-to-top`, `card-control` and `popover` documentation pages.
```
alionazherdetska, line 5: **Suggestions:**
- The removal of `layout: 'fullscreen'` for `back-to-top` docs page has no user-facing impact (no visible changes in docs), it should not be included in the changeset.
- Maybe we could add a more precise changeset description that clearly communicates to users what's actually changing, e.g. Updated documentation: added `is-valid` validation state support to `post-card-control` component and included `close-button-caption` attribute in `post-popover` component.

alionazherdetska (commented):
Well done!!!

packages/documentation/src/stories/components/card-control/standard-html/card-control.stories.ts
```diff
@@ -146,6 +146,7 @@ export const Default = {
     const cardClasses = mapClasses({
       'checked': args.checked,
       'disabled': args.disabled,
+      'is-valid': args.validation === 'is-valid',
```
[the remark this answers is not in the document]
myrta2302: Thank you for the comment! I actually made this change because i was getting an error from a test, but now that i re-checked i think it is not needed and i will revert it. The card-control component does not yet have a v2 version, so i think it should be better updated overall when there is a new design.

.changeset/clever-bears-pay.md
myrta2302: Changeset has been updated! I am not completly sure if the `back-to-top` change should be omitted, so i have left it for the moment.

PR #5490, opened 2025-05-14
feat(component): implemented breakpoints utlity in the `post-header` component

alionazherdetska:
## 📄 Description

The PR makes these key changes:
  - Removes the `DEVICE_SIZE` custom type
  - Removes the `postUpdateDevice` event emitter
  - Removes the custom event handling in the component interface
  - Changes the device state management to use `breakpoint events` directly instead of manual detection
  - Simplifies the handleResize method to only update header height

## 📝 Checklist


packages/components/src/components/post-header/post-header.tsx
```diff
@@ -272,29 +289,8 @@ export class PostHeader {
       }
     }
   }
-
   private handleResize() {
```
myrta2302, line 292: Maybe the handleResize() is not needed anymore?

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -110,6 +109,8 @@ export class PostHeader {
     });
     document.addEventListener('postToggleMegadropdown', this.megedropdownStateHandler);
```
myrta2302, line 110: Not directly relevant to this PR. but maybe this typo could be also corrected:  `meg**e**dropdownStateHandler`  -> `meg**a**dropdownStateHandler`

.changeset/angry-rules-win.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': major
+---
+
+Updated `post-header` component to use the `breakpoint` utility for responsive breakpoints.
```
alizedebray: Can you please update this message? 

"the `breakpoint` utility" probably does not mean anything for developers using the Design System as they are internal and not documented. 

What actually changed for them is: 
1.the header responsive behavior has been improved
2. an event has been removed, therefore they can no longer use it

PR #5492, opened 2025-05-14
chore(docs): add click blocker on stories

leagrdv:
## 📄 Description

## 🚀 Demo

---

## 📝 Checklist

### DIFF
--- a/packages/documentation/src/stories/components/footer/footer.stories.ts
+++ b/packages/documentation/src/stories/components/footer/footer.stories.ts
@@ -1,6 +1,7 @@
 import { Args, StoryObj } from '@storybook/web-components';
 import { html } from 'lit';
 import { MetaComponent } from '@root/types';
+import { clickBlocker } from '@/shared/click-blocker';
 
 const GRID_CELLS = [1, 2, 3, 4];
 const LINKS_PER_CELL = [6, 8, 8, 5];
@@ -10,6 +11,7 @@ const meta: MetaComponent = {
   title: 'Components/Footer',
   component: 'post-footer',
   tags: ['package:WebComponents'],
+  decorators: [clickBlocker],
   parameters: {
     layout: 'fullscreen',
     badges: [],
--- a/packages/documentation/src/stories/components/header/header.stories.ts
+++ b/packages/documentation/src/stories/components/header/header.stories.ts
@@ -2,6 +2,7 @@ import type { StoryObj } from '@storybook/web-components';
 import { MetaComponent } from '@root/types';
 import { html } from 'lit';
 import { fakeContent } from '@/utils';
+import { clickBlocker } from '@/shared/click-blocker';
 
 const meta: MetaComponent = {
   id: '27a2e64d-55ba-492d-ab79-5f7c5e818498',
@@ -57,7 +58,7 @@ const meta: MetaComponent = {
     story =>
       html` <div class="header-story-wrapper">
         <div class="virtual-body">${story()} ${fakeContent()}</div>
-      </div>`,
+      </div>`, clickBlocker
   ],
 };
 
--- a/packages/documentation/src/stories/raw-components/breadcrumbs/breadcrumbs.stories.ts
+++ b/packages/documentation/src/stories/raw-components/breadcrumbs/breadcrumbs.stories.ts
@@ -4,6 +4,7 @@ import { spread } from '@open-wc/lit-helpers';
 import { getAttributes } from '@/utils';
 import customItems from './custom-items';
 import { MetaComponent } from '@root/types';
+import { clickBlocker } from '@/shared/click-blocker';
 
 const meta: MetaComponent<HTMLSwisspostInternetBreadcrumbsElement> = {
   id: '4347e5bf-8bf2-4f44-9075-9faaa53591ed',
@@ -11,7 +12,7 @@ const meta: MetaComponent<HTMLSwisspostInternetBreadcrumbsElement> = {
   component: 'swisspost-internet-breadcrumbs',
   tags: ['package:InternetHeader'],
   render: renderInternetBreadcrumbs,
-  decorators: [hiddenHeader],
+  decorators: [hiddenHeader, clickBlocker],
   parameters: {
     badges: [],
     design: {
--- a/packages/documentation/src/stories/raw-components/internet-header/header.stories.ts
+++ b/packages/documentation/src/stories/raw-components/internet-header/header.stories.ts
@@ -5,6 +5,7 @@ import customConfig from './config/custom-config';
 import osFlyoutOverrides from './config/os-flyout-overrides';
 import languageSwitchOverrides from './config/language-switch-overrides';
 import { MetaComponent } from '@root/types';
+import { clickBlocker } from '@/shared/click-blocker';
 
 const meta: MetaComponent = {
   id: 'ebb11274-091b-4cb7-9a3f-3e0451c9a865',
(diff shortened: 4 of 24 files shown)


alionazherdetska (commented):
There are still clickable links that aren't covered by the current implementation:

1. card.stories.ts -  `<p><a href="http://google.com">Interactive card</a></p>`
2. teaser.stories.ts -  `<a href="#" class="btn btn-tertiary px-0">Button label <post-icon name="arrowright"></post-icon></a>`
3. popover.stories.ts -  `<a href="#">Links</a>`

alionazherdetska:
It could possibly be good to make the link click-blocking behavior consistent across the entire codebase. 
Looking at the anchor tags in `typography.blocks.tsx` (https://github.com/swisspost/design-system/blob/main/packages/documentation/src/stories/foundations/typography/typography.blocks.tsx), I see it's currently handling click prevention individually for each link with:
```jsx
onClick={e => e.preventDefault()}

**Comment by leagrdv** [comment]
> I like the idea to have one click prevention logic for all our stories. Anyway, your solution is stopping propagation and preventing default behaviour on all elements within the decorator element. I fear this may lead to unwanted side effects.
> 
> Personally I would prefer to add a global eventListener (e.g. in the `preview-head.html` file) and check the target and the `href` value.
> 
> ```js
> document.addEventListener('click', e => {
>   if (e.target.tagName === 'A' && e.target.getAttribute('href') === '#') {
>     e.stopPropagation();
>     e.preventDefault();
>   }
> }, {
>   capture: true
> });
> ```
> 
> Not sure if this would cover everything, or if we should, but it would mean we do not need to configure the blocker in every story in need.
> 
> Anyway, I would like to hear what @alizedebray thinks about it... and if she's ok with it, I'll not block your PR ;)

I understand the intention behing a global solution but I feel like handling this at the story level might be a better and safer approach:
- Applying a global handler will interfere with components that intentionally use `<a href="#">` for internal logic (e.g.: subnavigation)
- Adding the decorator manually to the story makes it easier for developers to understand what's happening. A global override could confuse devs on why some links might not be working.
- Not all stories need link-blocking: For instance, I've intentionally left the link in the `Foundations > Typography > Link` and there might be others in the future that would not benefit from a global override.

@gfellerph What do you think?
I'm assigning this PR to you as Oli is on holidays

**Comment by leagrdv** [comment]
@oliverschuerch I'm assigning the PR back to you since you're back 😉

**Comment by gfellerph** [comment]
@leagrdv in case you still want to hear my opinion: I'm in favour of having to declare if a story should have a click blocker. Implicitly blocking on all stories (or explicitly opting out) is harder to find.

File: packages/documentation/src/stories/components/footer/footer.stories.ts [hunk 1]
```diff
@@ -10,6 +11,7 @@ const meta: MetaComponent = {
   title: 'Components/Footer',
   component: 'post-footer',
   tags: ['package:WebComponents'],
+  decorators: [clickBlocker],
```
oliverschuerch, line 14: This breaks the accordion which is part of the footer (only on mobile) and is exactly what I mentioned [here](https://github.com/swisspost/design-system/pull/5492#pullrequestreview-3010392587):

"...your solution is stopping propagation and preventing default behaviour on all elements within the decorator element. I fear this may lead to unwanted side effects."

Since you use the option `capture: true`, it get's called before the JS eventListener on the accordion and with `e.stopPropagation()` the JS eventListener there is never being called anymore.

packages/documentation/src/stories/components/header/header.stories.ts
```diff
@@ -58,6 +59,7 @@ const meta: MetaComponent = {
       html` <div class="header-story-wrapper">
         <div class="virtual-body">${story()} ${fakeContent()}</div>
       </div>`,
+    clickBlocker,
```
oliverschuerch, line 62: Same here. For example the mobile menu, the megadropdown toggles and the language-switch components are not working anymore.

packages/documentation/src/stories/components/list-interactive/list-interactive.stories.ts
```diff
@@ -1,13 +1,14 @@
...
 import { MetaComponent } from '@root/types';
+import { clickBlocker } from '@/shared/click-blocker';
 
 const meta: MetaComponent = {
   id: '5a47ba70-7831-4e59-b83e-81b6e6c32372',
   title: 'Components/List Interactive',
   tags: ['package:HTML'],
   render: renderListGroup,
-  decorators: [gridDecorator],
+  decorators: [gridDecorator, clickBlocker],
```
oliverschuerch, line 11: This breaks the `.list-interactive-switch`.

packages/documentation/src/stories/components/popover/popover.stories.ts
```diff
@@ -91,6 +92,7 @@ const meta: MetaComponent = {
       name: 'Placement',
     },
   },
+  decorators: [clickBlocker],
```
oliverschuerch, line 95: Popover close button does not work anymore.

packages/documentation/src/stories/raw-components/breadcrumbs/breadcrumbs.stories.ts
```diff
@@ -4,14 +4,15 @@ import { spread } from '@open-wc/lit-helpers';
...
+import { clickBlocker } from '@/shared/click-blocker';
 
 const meta: MetaComponent<HTMLSwisspostInternetBreadcrumbsElement> = {
   id: '4347e5bf-8bf2-4f44-9075-9faaa53591ed',
   title: 'Raw Components/Internet Header/Breadcrumbs',
   component: 'swisspost-internet-breadcrumbs',
   tags: ['package:InternetHeader', 'devOnly'],
   render: renderInternetBreadcrumbs,
-  decorators: [hiddenHeader],
+  decorators: [hiddenHeader, clickBlocker],
```
oliverschuerch, line 15: This is breaking the modal which should open on the help and the contact button.

packages/documentation/src/stories/raw-components/internet-header/header.stories.ts
```diff
@@ -158,15 +159,15 @@ const meta: MetaComponent = {
...
-      `,
+    story => html`
+      <div
+        class="header-story-wrapper"
+        style="--header-z-index: 1;overflow: auto;max-height: 100svh;"
+      >
+        ${story()} ${fakeContent()}
+      </div>
+    `,
+    clickBlocker,
```
oliverschuerch, line 170: This is breaking the mobile menu of the header.

packages/documentation/src/stories/raw-components/megadropdown/megadropdown.stories.ts
```diff
@@ -121,7 +122,7 @@ function render() {
 type Story = StoryObj<HTMLPostLanguageOptionElement>;
 
 export const Default: Story = {
-  decorators: [megadropdownDecorator],
+  decorators: [megadropdownDecorator, clickBlocker],
```
oliverschuerch, line 125: Same here.

packages/documentation/src/stories/raw-components/menu-button/menu-button.stories.ts
```diff
@@ -1,12 +1,14 @@
...
 import { html, nothing } from 'lit';
 import { MetaComponent } from '@root/types';
+import { clickBlocker } from '@/shared/click-blocker';
 
 const meta: MetaComponent = {
   id: '8ca2bd70-56e6-4da9-b1fd-4e55388dca88',
   title: 'Raw Components/Menu Button',
   tags: ['package:WebComponents', 'devOnly'],
   component: 'post-menu',
+  decorators: [clickBlocker],
```
oliverschuerch, line 11: and here

PR #5502, opened 2025-05-15
fix(component): fix animation error in post-collapsible when inside inactive tabs

alionazherdetska:
## 📄 Description

This PR adds visibility check before committing animation styles in PostCollapsible.toggle() to prevent InvalidStateError when the target element is hidden. This resolves errors occurring when animating elements inside hidden tab panels.

---

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-collapsible/post-collapsible.tsx
+++ b/packages/components/src/components/post-collapsible/post-collapsible.tsx
@@ -77,7 +77,10 @@ export class PostCollapsible {
 
     await animation.finished;
 
-    animation.commitStyles();
+    const isVisible = this.host.offsetParent || this.host.offsetWidth || this.host.offsetHeight;
+    if (isVisible) {
+      animation.commitStyles();
+    }
 
     return open;
   }
--- a/.changeset/tricky-buckets-sniff.md
+++ b/.changeset/tricky-buckets-sniff.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Fixed an issue where `post-accordion` placed inside inactive tabs would throw animation errors.


alizedebray (changes requested):
The `expand` and `collapse` animations are set with the `fill: 'forwards'` option, which ensures that the final keyframe of the animation remains applied to the element once the animation completes. However, using fill modes to persist an animation's effect is generally not recommended  for several reasons (see the [Web Animation API documentation](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API/Using_the_Web_Animations_API#persisting_animation_styles)). 

Therefore, whenever the `expand` or`collapse` animation are triggered, the styles must be explicitly committed using `animation.commitStyles();`. 

Currently, if the element is not visible, the collapse animation will still run, but the animation effects are not properly committed.

packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -40,7 +40,15 @@ export class PostCollapsible {
...
     );
-
+    
+    if (!this.isLoaded) {
+      Object.assign(
+        this.host.style, 
+        this.collapsed ? collapsedKeyframe : expandedKeyframe
+      );
+      this.isOpen = !this.collapsed;
+    }
```
alizedebray, line 50: A `return` or `else` statement is missing here as you do not want the `toggle` method to be called when the collapsible is not loaded.

packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -40,7 +40,15 @@ export class PostCollapsible {
       'boolean',
       'The `collapsed` property of the `post-collapsible` must be a boolean.',
     );
-
+    
+    if (!this.isLoaded) {
+      Object.assign(
+        this.host.style, 
+        this.collapsed ? collapsedKeyframe : expandedKeyframe
```
alizedebray, line 47: Applying the `expandedKeyframe` here is not necessary since `height: 'auto'` and `overflow: 'visible'` are the initial value for any HTML element.

alizedebray (changes requested):

packages/components/src/components/post-collapsible/post-collapsible.tsx
alionazherdetska: The issue is that during the `!this.isLoaded` phase, we need to ensure the visual state matches the collapsed property value:

If `collapsed = true` → element should visually appear collapsed
If `collapsed = false` → element should visually appear expanded

Without explicitly applying expandedKeyframe when `collapsed = false`, the element could remain in whatever state it was in initially, which might not match the expected expanded appearance.

packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -40,7 +40,15 @@ export class PostCollapsible {
...
+      );
+      this.isOpen = !this.collapsed;
```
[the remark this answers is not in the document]
alionazherdetska: After investigating, the `isOpen` property actually can't be removed. While `collapsed` represents the target state (what the component should become), `isOpen` tracks the current visual state during transitions. 

Without `isOpen`, we lose the ability to prevent animation conflicts since `collapsed` gets updated immediately while the element is still transitioning. This causes the expanding/collapsing logic to break.

Both properties serve distinct purposes for proper animation handling, so removing `isOpen` would unfortunately break functionality.

packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -40,8 +40,18 @@ export class PostCollapsible {
       'boolean',
       'The `collapsed` property of the `post-collapsible` must be a boolean.',
     );
-
-    void this.toggle(!this.collapsed);
+    
+    const expandedKeyframe: Keyframe = { height: 'auto', overflow: 'visible' };
+
+    if (!this.isLoaded) {
```
alizedebray:
```suggestion
    if (!this.isLoaded) {
    
      const expandedKeyframe: Keyframe = { height: 'auto', overflow: 'visible' }; 
```

alizedebray (approved):

PR #5540, opened 2025-05-23
feat(component): replace EventGuard utility with @EventGuard decorator

alionazherdetska:
## 📄 Description

This PR refactors the event handling system by replacing the `eventGuard` utility function with `@EventFrom` decorator. 
### Changes Made:

- **Replaced utility function with decorator**: Converted `eventGuard()` calls to `@EventFrom()` decorator usage
- **Simplified event handlers**: Removed nested callback structures in favor of direct method decoration
- **Updated components**: Refactored 8 components to use the new decorator pattern:
  - `post-accordion-item`
  - `post-accordion` 
  - `post-collapsible-trigger`
  - `post-header`
  - `post-language-switch`
  - `post-megadropdown-trigger`
  - `post-menu`

## 📝 Checklist


packages/components/src/utils/event-guard.ts
```diff
@@ -1,48 +1,96 @@
...
+      descriptor.value = function (event: CustomEvent) {
+        if (!event || !event.target) return;
+
+        const eventTarget = event.target as HTMLElement;
+        
+        if (eventTarget.localName !== options.targetLocalName) return;
+        
+        if (options.delegatorSelector) {
+          const closest = shadowClosest(eventTarget, options.delegatorSelector);
+          if (closest !== this.host) return;
```
alizedebray: These two lines means that `option.delegatorSelector` is actually equal to `this.host.localName`, therefore `option.delegatorSelector` can be removed.

Then you can also remove the `shadowClosest` function and simplify with:
```suggestion
          if (!this.host.contains(eventTarget)) return;
```

packages/components/src/utils/event-guard.ts
```diff
@@ -1,48 +1,96 @@
...
-  callback: () => void
-): void {
-  const target = event.target as HTMLElement | null;
+/**
+ * Decorator that guards event handlers to only execute when the event meets certain criteria.
+ * @param options - Configuration options for the event guard
+ * @param options.targetLocalName - The local name of the element that should be the target of the event
+ * @param options.delegatorSelector - Optional CSS selector to check if the event was delegated from a specific ancestor
+ */
+export function EventGuard(options: { targetLocalName: string; delegatorSelector?: string }) {
```
alizedebray: I would like to rename the guard to make it easier to understand and use. 

For example: 
```suggestion
export function EventFrom(tag: string, option?: { ignoreNestedComponents: boolean }) {
```

Then the event can be used like this:
```typescript
// by default ignoreNestedComponents is true since it's the most common way the guard is used
@EventFrom('post-collapsible')
handleEvent(e: Event) { ... }

@EventFrom('post-accordion-item', { ignoreNestedComponents: false })
handleOtherEvent(e: Event) { ... }
```

What do you think?

packages/components/src/utils/event-guard.ts
```diff
@@ -1,48 +1,96 @@
...
+      });
+
+      Object.defineProperty(target, propertyKey, {
+        get() {
+          return this[privateKey];
+        },
+        set(originalFunction: (event: CustomEvent) => void) {
+          if (typeof originalFunction === 'function') {
+            this[privateKey] = (event: CustomEvent) => {
+              if (!event || !event.target) return;
```
alizedebray: Since the code is not trivial, it would be nice to have comments to explain it.

packages/components/src/utils/event-guard.ts
```diff
@@ -1,48 +1,96 @@
...
 
-  if (!target) return;
+              const eventTarget = event.target as HTMLElement;
+              
+              if (eventTarget.localName !== options.targetLocalName) return;
+              
+              if (options.delegatorSelector) {
+                const closest = shadowClosest(eventTarget, options.delegatorSelector);
+                if (closest !== this.host) return;
+              }
```
alizedebray: This code is duplicated, can you extract it into a function?

packages/components/src/utils/event-guard.ts
```diff
@@ -1,48 +1,96 @@
...
-  const target = event.target as HTMLElement | null;
+/**
+ * Decorator that guards event handlers to only execute when the event meets certain criteria.
+ * @param options - Configuration options for the event guard
+ * @param options.targetLocalName - The local name of the element that should be the target of the event
+ * @param options.delegatorSelector - Optional CSS selector to check if the event was delegated from a specific ancestor
+ */
+export function EventGuard(options: { targetLocalName: string; delegatorSelector?: string }) {
+  return function (
+    target: unknown,
```
alizedebray:
```suggestion
    target: object,
```

packages/components/src/utils/event-guard.ts
```diff
@@ -1,48 +1,96 @@
...
+    propertyKey: string,
+    descriptor?: PropertyDescriptor
+  ) {
+    if (descriptor) {
+      const originalMethod = descriptor.value;
+      
+      descriptor.value = function (event: CustomEvent) {
+        if (!event || !event.target) return;
+
+        const eventTarget = event.target as HTMLElement;
```
alizedebray: You don’t know if the decorator will be correctly applied to a method that takes a custom event as a parameter, nor can you be sure if the event’s target will be an HTMLElement. To make your method foolproof, you need to test the types:
```typescript
      descriptor.value = function (event: unknown) {
        if (!(event instanceof CustomEvent && event.target instanceof HTMLElement)) return;
```

Also: will the decorator be used only for custom events, or can it also be applied to standard events?

alizedebray (changes requested):

packages/components/src/utils/event-guard.ts
alizedebray: The second part of my comment isn't correct, `this.host.contains` does not mean the host is the _closest_ parent so you can only simplify with:
```suggestion
          const closest = shadowClosest(eventTarget, this.host.localName);
          if (closest !== this.host) return;
```

alizedebray: And to improve readability, you can refactor the `shadowClosest(eventTarget, this.host.localName)` to `isClosestParent(this.host, eventTarget)` since it's what you actually want to check.

packages/components/src/components/post-accordion-item/post-accordion-item.tsx
```diff
@@ -52,15 +52,10 @@ export class PostAccordionItem {
 
   // Capture to make sure the "collapsed" property is updated before the event is consumed
   @Listen('postToggle', { capture: true })
+  @EventFrom('post-accordion-item')
+  @Listen('postToggle', { capture: true })
```
alizedebray: The `@Listen` is duplicated here:
```suggestion
  @Listen('postToggle', { capture: true })
  @EventFrom('post-accordion-item')
```

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -187,19 +187,10 @@ export class PostHeader {
...
-        targetLocalName: 'post-megadropdown',
-        delegatorSelector: 'post-header',
-      },
-      () => {
-        this.megadropdownOpen = event.detail.isVisible;
-      },
-    );
-  };
+  @EventFrom('post-megadropdown')
+  private megadropdownStateHandler(event: CustomEvent) {
```
alizedebray: You turned the property into a method. Can you check if it is still working as expected?

Otherwhise: 
```suggestion
  private megadropdownStateHandler = (event: CustomEvent) => {
```

PR #5551, opened 2025-05-28
chore(components): update tooltip with accessible name

myrta2302:
## 📄 Description

## 🚀 Demo

---

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-tooltip/post-tooltip.tsx
+++ b/packages/components/src/components/post-tooltip/post-tooltip.tsx
@@ -12,6 +12,8 @@ import { checkType } from '@/utils';
 export class PostTooltip {
   private popoverRef: HTMLPostPopovercontainerElement;
 
+  private slotEl!: HTMLSlotElement;
+
   @Element() host: HTMLPostTooltipElement;
 
   /**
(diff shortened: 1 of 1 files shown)


packages/components/src/components/post-tooltip-trigger/post-tooltip-trigger.tsx
oliverschuerch, line 1: Nothing really changed in this file, if I'm not wrong. Therefore we should reset the format changes.

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -43,6 +43,19 @@ export class PostTooltip {
...
+    if (node.nodeType === Node.TEXT_NODE) {
+      return node.textContent?.trim() ?? '';
+    }
+    if (node.nodeType === Node.ELEMENT_NODE) {
+      const element = node as Element;
+      return Array.from(element.childNodes)
+        .map(child => this.extractText(child))
+        .join(' ');
+    }
+    return '';
```
oliverschuerch: Imo a switch statement would fit your needs 100% here and also increase readability and performence.
https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Statements/switch

But also, where is this method used, I couldn't find any execution of it?

myrtasakellariou: You are right. I had a different implementation before (that added the text in an aria-label) and forgot to remove this code.

myrtasakellariou:
> We did the exact opposite in the the commit [b16ba09](https://github.com/swisspost/design-system/commit/b16ba09710d930a54615d51cee1059d88a1e4b68#diff-0accc57a09ea3cddf962f5d1e8ead7b5abd4a036150469488778a088406969a7). That's why I would like to hear @gfellerph opinion about this change.

image  

I am not sure why it was moved before. The accessible name is visible in the Accessibility tab only when the `role="tooltip"` is set to the `<popovercontainer>`.

gfellerph:
When it's on the tooltip host, the tooltip is empty on startup because the popover is hidden. This gives an accessibility error of empty tooltip. With the role on the container, the thing works. My change was a regression.

oliverschuerch (approved):

PR #5557, opened 2025-05-30
feat(tokens): use typescript in the tokens package

myrta2302:
## 📄 Description

This PR implements Typescript in the Tokens Package.

## 📝 Checklist


packages/tokens/package.json
```diff
@@ -6,6 +6,7 @@
   "license": "Apache-2.0",
   "private": false,
   "type": "module",
+  "types": "dist/index.d.ts",
```
leagrdv, line 9: I can't seem to find this file, even after building the package.

packages/tokens/package.json
```diff
@@ -20,7 +21,7 @@
     "linkDirectory": true
   },
   "scripts": {
-    "build": "pnpm clean && node ./build.js && pnpm copy-files",
+    "build": "pnpm clean && tsc -p . && node ./dist/build.js && pnpm copy-files",
```
leagrdv, line 24: The `build.js` does not exist anymore, I think you changed it to a typescript file and it's in another folder than the `dist` one.

packages/tokens/src/_build/configs/all.ts
```diff
@@ -55,8 +59,9 @@ registerConfigMethod((tokenSets, { sourcePath, buildPath }) => {
  */
 StyleDictionary.registerFilter({
   name: 'swisspost/scss-filter',
-  filter: (token, { meta }) => {
-    return token.filePath.includes(`/output/${meta.filePath}`);
+  filter: (token, options) => {
+    const configOptions = options as Config & LocalOptions;
```
leagrdv: I don't think the type here is LocalOptions, this passes because `LocalOptions` has the following key: `[key: string]: any;`. I think it'd be cleaner to just extend the `Config` interface and add the `meta` object within.

packages/tokens/src/_build/configs/all.ts
```diff
@@ -81,7 +86,7 @@ StyleDictionary.registerFormat({
     return (
       header +
       meta.setNames
-        .map(setName => {
+        .map((setName: string) => {
```
leagrdv: As mentioned before, adding the correct type above would help here as you wouldn't have to define it's a string here.

packages/tokens/src/_build/configs/tailwind.ts
```diff
@@ -65,7 +65,7 @@ StyleDictionary.registerFormat({
       const tokenObj = token.path
         .slice(token.path.indexOf(TOKENSET_PREFIX) + 1)
         .reverse()
-        .reduce((res, p) => ({ [p]: res }), getTokenValue(options, token));
+        .reduce((res, p) => ({ [p]: res }), getTokenValue(options, token)) as object;
```
leagrdv: Could we be more specific here?

packages/tokens/src/_build/methods.ts
```diff
@@ -125,6 +143,7 @@ function createTokenSets(tokensFile) {
           (isComponent && TOKENSET_LAYERS.component) ||
           TOKENSET_LAYERS.semantic,
         filePath: `${groupName ?? setName}.json`,
+        sets: {}, // MYRTA: is this needed?
```
leagrdv: Is it? 😉

packages/tokens/src/_build/methods.ts
```diff
@@ -309,20 +330,28 @@ export function getSetName(_options, setName) {
...
   const { meta } = options;
-  let tokenSet = [];
+
+  let tokenSet: TransformedToken[] = [];
 
   if (meta.layer === 'semantic') {
     const baseSetName = meta.setNames[0];
-    const overrideSetNameIndex = meta.setNames.findIndex(setName => setName === currentSetName);
+    const overrideSetNameIndex = meta.setNames.findIndex(
+      (setName: string) => setName === currentSetName,
```
leagrdv: Same here, the `: string` would not be necessary if options are of the correct type.

packages/tokens/src/_build/style-dictionary.ts
```diff
@@ -92,24 +99,25 @@ StyleDictionary.registerPreprocessor({
...
-    function traverse(context) {
+    function traverse(context: TokenGroup) {
       Object.entries(context).forEach(([key, value]) => {
-        const usesDtcg = context[key].$type && context[key].$value;
-        const isToken = context[key][usesDtcg ? '$type' : 'type'] !== undefined;
-
-        if (isToken) {
-          const tokenType = context[key][usesDtcg ? '$type' : 'type'];
-          const tokenValue = context[key][usesDtcg ? '$value' : 'value'];
+        const usesDtcg = (context[key] as any).$type && (context[key] as any).$value;
```
leagrdv: It's not good practice to set `as any`. If the context we receive on the `traverse` function is of type `TokenGroup` then `context[key]` should be of type `TokenGroup | TokenProperty` which both don't seem to contain the `$type` key, meaning that the context received on the `traverse` function is probably of another type.

packages/tokens/src/_build/utils/object-deepmerge.ts
```diff
@@ -0,0 +1,43 @@
...
+          ? DeepMerged<T[K], U[K]>
+          : U[K]
+        : U[K]
+      : U[K]
+    : K extends keyof T
+    ? T[K]
+    : never;
+};
+
+export default function deepmerge<T extends object, U extends object>(
```
leagrdv, line 16: Could you comment this method to explain a bit its purpose?

leagrdv (changes requested):

packages/tokens/package.json
myrtasakellariou: It seems to be removed at the end of the `node dist/build.js` process.

packages/tokens/package.json
myrtasakellariou: It seems to be removed at the end of the node dist/build.js process.

packages/tokens/src/_build/configs/tailwind.ts
```diff
@@ -60,17 +60,20 @@ StyleDictionary.registerFormat({
...
-      return allTokens;
-    }, {});
+
+    const themeVariables = dictionary.allTokens.reduce<Record<string, TokenProperty>>(
+      (allTokens, token) => {
+        const tokenPath = token.path.slice(token.path.indexOf(TOKENSET_PREFIX) + 1);
+
+        const cssVarName = `--${tokenPath.join('-')}`;
+
+        const tokenValue: TokenProperty = getTokenValue(options, token);
```
leagrdv: This is redundant and already implied since the type of the return of getTokenValue is `TokenProperty`

packages/tokens/src/_build/configs/utilities.ts
```diff
@@ -48,7 +49,7 @@ StyleDictionary.registerFormat({
     const utilityTokens = new Map();
 
     dictionary.allTokens.forEach(token => {
-      const { subitem, state } = token.attributes;
+      const { subitem, state } = token.attributes || {};
```
leagrdv: This hides the typing error but it doesn't really fix it.
`token.attributes` is generally defined as `Record<string, unknown>` but we should specify it more to reflect what we get, which is something like this:
```
  attributes: {
    category: 'typo',
    type: 'post',
    item: 'utility',
    subitem: 'font-size',
    state: '11'
  },
  ```

File: packages/tokens/src/_build/methods.ts [hunk 12]
```diff
@@ -11,13 +16,21 @@ import {
...
-let CLI_OPTIONS;
-let tokenSets;
-let registeredConfigMethods = [];
+let tokenSets: TokenSets;
 
-export async function setup() {
+let registeredConfigMethods: Array<
+  (tokenSets: TokenSets, options: { sourcePath: string; buildPath: string }) => Config[]
+>;
+registeredConfigMethods = [];
```
leagrdv:
```suggestion
let registeredConfigMethods: Array<
  (tokenSets: TokenSets, options: { sourcePath: string; buildPath: string }) => Config[]
> = [];
```

packages/tokens/src/_build/methods.ts
```diff
@@ -211,11 +233,11 @@ export async function createOutputFiles() {
           preprocessors: [
             'swisspost/box-shadow-keep-refs-workaround',
             'tokens-studio',
-            ...(config.proprocessors ?? []),
+            ...(config.preprocessors ?? []),
           ],
         });
 
-        config.platforms = Object.entries(config.platforms).reduce(
+        config.platforms = Object.entries(config.platforms || {}).reduce(
```
leagrdv: This would hide a bug if config.platforms were to not be defined. It might be better to throw an error if config.platforms isn't defined instead. What do you think?

Something like this beforehand:
`if (!config.platforms) return;`

packages/tokens/src/_build/methods.ts
```diff
@@ -309,20 +331,28 @@ export function getSetName(_options, setName) {
  * | a       |        | b      | desktop/tablet = a, mobile = b
  * | a       | b      | c      | desttop = a, tablet = b, mobile = c
  */
-export function getSet(options, dictionary, currentSetName) {
+
+export function getSet(
+  options: Config & LocalOptions,
```
leagrdv: options here should be of type `ConfigMeta` instead of `Config`, otherwise `meta` is set as `any`

packages/tokens/src/_build/methods.ts
leagrdv: the `:string` is still there

packages/tokens/src/_build/methods.ts
```diff
@@ -309,20 +331,28 @@ export function getSetName(_options, setName) {
...
+      (setName: string) => setName === currentSetName,
+    );
     const overrideSetNames = meta.setNames.slice(1, overrideSetNameIndex + 1);
 
     tokenSet = dictionary.allTokens
       .filter(token => token.path[0] === baseSetName)
       .map(normalizeToken);
 
-    overrideSetNames.forEach(overrideSetName => {
+    overrideSetNames.forEach((overrideSetName: string) => {
```
leagrdv: same here, it could be removed

packages/tokens/src/_build/constants.ts
```diff
@@ -36,11 +36,13 @@ export const EXPLICIT_COMPONENT_LAYER_GROUPNAMES = [
...
 export const EXPLICIT_FIGMAONLY_GROUPNAMES = ['figmaonly'];
 export const EXPLICIT_FIGMAONLY_SETNAMES = ['figmaonly'];
+
 export const TOKENSET_LAYERS = {
   core: 'core',
-  semantic: 'semantic',
   component: 'component',
-};
+  semantic: 'semantic',
+} as const;
```
leagrdv: I think it would make more sense to set an enum here rather than force an object into a const

```
export enum TOKENSET_LAYERS {
  core = 'core',
  component = 'component',
  semantic = 'semantic',
}
```

packages/tokens/src/_build/style-dictionary.ts
```diff
@@ -34,9 +39,11 @@ StyleDictionary.registerTransform({
   type: 'value',
   filter: token => {
     const usesDtcg = token.$type && token.$value;
-    const transformType = NO_UNITLESS_ZERO_VALUE_TOKEN_TYPES.includes(
-      usesDtcg ? token.$type : token.type,
-    );
+    let transformType = false;
+    if (typeof token.$type === 'string' && typeof token.type === 'string') {
```
leagrdv: Here, in case `$type` is a string and `usesDtcg` is true, but `type` is not a string, we would no go inside the if statement, even though we could.
I think I'd go with forcing the type instead

```
const transformType = NO_UNITLESS_ZERO_VALUE_TOKEN_TYPES.includes(
      usesDtcg ? token.$type as string : token.type as string
    );
```

leagrdv (changes requested):

packages/tokens/src/_build/style-dictionary.ts
```diff
@@ -34,9 +39,11 @@ StyleDictionary.registerTransform({
   type: 'value',
   filter: token => {
     const usesDtcg = token.$type && token.$value;
-    const transformType = NO_UNITLESS_ZERO_VALUE_TOKEN_TYPES.includes(
-      usesDtcg ? token.$type : token.type,
-    );
+    let transformType = false;
+    if (token.$type) {
```
leagrdv: If there is no `token.$type` but there is a `token.type`, you would not get inside this if statement and `token.type` would not be used.

leagrdv (changes requested):

leagrdv (approved):
Could you please add a changeset?

PR #5578, opened 2025-06-02
chore(components): fix screen reader focus issue in `<post-language-switch>` with proper ARIA roles

myrta2302:
## 📄 Description

This PR fixes arrow navigation focus issue on the <post-language-switch> component by:

- Setting `role="menu"` on the nested `<post-menu>` element
- Adding `role="menuitem"` to all the `<post-language-option>`’s internal `<button>` elements if variant="menu".

These ARIA roles prompt screen readers like NVDA and JAWS to enter `focus mode` automatically when users navigate between language options using the Arrow Up and Arrow Down keys, ensuring proper keyboard interaction.

## 📝 Checklist


packages/components/src/components/post-language-switch/post-language-switch.tsx
```diff
@@ -111,6 +113,14 @@ export class PostLanguageSwitch {
     });
   }
 
+  // Update post-language-options to have role="menuitem"
+  private updateLanguageOptionsRole() {
+    const languageOptionBtns = this.host.querySelectorAll('post-language-option > button');
+    languageOptionBtns.forEach(langOptionBtn => {
+      langOptionBtn.setAttribute('role', 'menuitem');
+    });
+  }
```
gfellerph: This could be rendered inside the `post-language-option` component, that would be way easier and more resilient when things get re-rendered than to update it from here. Also check the accessibility if the language options are links.

myrtasakellariou: 1) Initially, I set the `menuitem` role directly on the options, but there's a test for `post-language-option` that fails accessibility due to the missing parent `role="menu"`. I discussed this with Alizée, and she suggested setting the roles from `<post-language-switch>`, since they’re only needed when the parent is present. But if you prefer i can change the test to be for the whole component.  2) Added the links case.

.changeset/cute-feet-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Improved accessibility by adding ARIA roles to trigger Screen Readers' focus mode navigation.
```
alizedebray:
```suggestion
Improved accessibility of the `post-language-switch` component by adding ARIA roles to trigger Screen Readers' focus mode navigation.
```

packages/components/src/components/post-language-switch/post-language-switch.tsx
alizedebray: Currently, `post-language-switch` sets both `role` and `variant` on its `post-language-option` children. But since `role` depends on `variant`, setting both independently risks them getting out of sync.

I agree with Philipp that `post-language-option` should determine the correct `role` itself, based solely on `variant`.

This also fixes test issues. Currently, `post-language-option` is tested without a `variant` property, so no `variant="menu"` means no `role` is set and no accessibility issues arise.

packages/components/src/components/post-language-option/post-language-option.tsx
```diff
@@ -130,6 +130,7 @@ export class PostLanguageOption {
       <Host data-version={version}>
         {this.url ? (
           <a
+            role={this.variant === 'menu' ? 'menuitem' : undefined}
```
alizedebray: Can you extract the role in a constant declared within the render function?

alizedebray (approved):

PR #5586, opened 2025-06-04
chore(components): use animation API in the popover container 

hugomslv:
## 📄 Description

## 🚀 Demo

---

## 📝 Checklist


.changeset/rotten-ears-judge.md
alizedebray, line 1: There is no need for a changeset here: developers using the Design System do not need do know how the components are built internally, the changesets are there to explain changes that could potentially have an impact on their code or on their application.

packages/components/src/animations/pop-in.ts
```diff
@@ -0,0 +1,19 @@
+const DEFAULT_DURATION = 250;
+const DEFAULT_EASING = 'cubic-bezier(0.34, 1.56, 0.64, 1)';
+const FILL_MODE = 'forwards';
+
+export function popInAnimation(el: Element, duration = DEFAULT_DURATION, easing = DEFAULT_EASING) {
```
alizedebray: In other animations we usually don't make the duration and easing configurable, this prevents having discrepancies. Can you align this animation and remove the `duration` and `easing` parameters?

alizedebray:
```suggestion
export function popIn(el: Element, duration = DEFAULT_DURATION, easing = DEFAULT_EASING) {
```

packages/components/src/components/post-popovercontainer/post-popovercontainer.scss
```diff
@@ -37,10 +37,6 @@
     background-color: var(--post-current-palette-bg);
     border: 2px solid transparent; // Keeping the default border for HCM
     border-radius: commons.$border-radius;
-    
-    &.animate-pop-in {
-      @include animation.pop-in-animation(0.25s);
```
alizedebray, line 41: Can you also delete the `pop-in-animation` from the styles package?

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -348,11 +354,9 @@ export class PostPopovercontainer {
   }
 
   render() {
-    const animationClass = this.animation ? `animate-${this.animation}` : '';
```
alizedebray, line 378: Before the animation only ran when `this.animation` was not null. Can you add this check in you changes as well?

packages/components/src/animations/pop-in.ts
```diff
@@ -0,0 +1,19 @@
+const DEFAULT_DURATION = 250;
+const DEFAULT_EASING = 'cubic-bezier(0.34, 1.56, 0.64, 1)';
```
alizedebray: Can you use the same timing function as previously declared in the `pop-in-animation`? 
Here you are using a `cubic-bezier` intead of a `linear` easing.

To know more about timing functions (also called easing), here is the documentation: https://developer.mozilla.org/en-US/docs/Web/CSS/transition-timing-function

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
alizedebray: This issue seems not to be resolved. The `popIn` animation on line 211 is started even when `this.animation` is `undefined`.

packages/components/src/animations/pop-in.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  return el.animate(
+    [
+      { transform: 'scale(0.9)', opacity: 0 },
+      { transform: 'scale(1)', opacity: 1 },
+    ],
+    {
+      duration,
+      easing,
+      fill: FILL_MODE,
```
alizedebray:
```suggestion
      fill: 'forwards',
```

packages/components/src/animations/pop-in.ts
```diff
@@ -0,0 +1,19 @@
+const duration = 250;
+const easing = 'ease-out';
+const FILL_MODE = 'forwards';
```
alizedebray:
```suggestion
```

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -204,7 +205,12 @@ export class PostPopovercontainer {
 
     const isOpen = e.newState === 'open';
     if (isOpen) {
+      const content = this.host.querySelector('.popover-content');
       this.startAutoupdates();
+      if (content && this.animation === `animate-${this.animation}`) {
```
alizedebray: `this.animation` can never be equal to `animate-${this.animation}`. 
Please check on [line 99](https://github.com/swisspost/design-system/pull/5586/files#diff-df6410ed94624cc721ffe1939a316d2e4d2de16b0dc709ddacfc2c2fbc56c46dR99) the values `this.animation` can take.

alizedebray (changes requested):

alizedebray (approved):

PR #5622, opened 2025-06-12
chore(documentation): required and optional fields on form elements

leagrdv:
## 📄 Description

Added the documentation for the required and optional fields. As decided with the team (see ticket's comments), the arguments were added to each form element but the general documentation was put in "Form creation" guidelines.

The required/optional were added separately from the valid/invalid arguments as a field can be required/optional as well as valid/invalid so it did not make much sense to me to have them together.

Also fixed a small issue in the documentation were the slider valid/invalid feedback didn't have an `id` and was therefore not connected to the `aria-describedby` of the input.


packages/documentation/src/stories/components/input/input.snapshot.stories.ts
gfellerph, line 1: It would be nice if we could rely on the standard `required` on form controls as additional indicator because that will allow tools to collect infos about required form fields via standard mechanisms. Adding the string "(required)" to the label will not tell assistive tech that this field is required.

If we're doing `<input type="text" required><label>Name (required)` however, screenreaders will read "required" twice because they already announce the standard required attribute on the input. What we'll have to do instead is something like

```html
<input type="text" required>
<label>
  Name
  <span aria-hidden="true">(required)</span>
</label>
```
For optional fields it's the same since an input field without the `required` attribute is already optional. It's a bit counterintuitive but complies better with standards. What do you think?

gfellerph (dismissed):

packages/documentation/src/stories/components/input/input.snapshot.stories.ts
leagrdv: I agree :) I've updated the docs accordingly.

PR #5632, opened 2025-06-13
feat(components): add default styles to the post-menu

alizedebray:
## 📄 Description

This PR is to set default styles to the `post-menu` component matching the designed menu.
**At the moment the `post-menu` remains a raw component and is not yet public, this is why no changeset was added.** 

This PR will be followed by other PRs in order to update the component's logic and then make it publicly available by publishing a documentation page.

## 🚀 Demo

---

## 📝 Checklist

### DIFF
--- a/.changeset/cute-feet-serve.md
+++ b/.changeset/cute-feet-serve.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Improved accessibility of the `post-language-switch` component by adding ARIA roles to trigger Screen Readers' focus mode navigation.
--- a/.changeset/forty-rings-refuse.md
+++ b/.changeset/forty-rings-refuse.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': patch
+---
+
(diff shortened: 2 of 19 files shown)


packages/components/src/components/post-popovercontainer/post-popovercontainer.scss
```diff
@@ -35,12 +35,16 @@
     max-width: 100%;
     color: var(--post-current-palette-fg);
     background-color: var(--post-current-palette-bg);
-    border: 2px solid transparent; // Keeping the default border for HCM
     border-radius: commons.$border-radius;
-    
+    overflow: hidden;
```
oliverschuerch: This creates a bug in the post-popup and post-tooltip components, if you want to show them with an pointy arrow.

oliverschuerch (changes requested):

leagrdv (changes requested):
The example in the docs shows an element within the menu that is a `div` which therefore looks broken since it's neither a `a` or a `button`. Should we remove it, since only the buttons and links are being styled here?

Also, merging this PR will close the ticket associated to it. Should we already create the other tickets (sub-issues maybe) for the tests, better documentation and making the component public?

PR #5642, opened 2025-06-13
chore(docs): add dynamic tooltips for angular

leagrdv:


packages/documentation/src/stories/components/tooltip/dynamic-tooltip-code.sample.html
gfellerph, line 1: While this will work, we were discussing about adding a single `<post-tooltip>` element outside of the loop that changes it's content based on which item is hovered. @alizedebray which version should we go with? I imagine the setup for the one-tooltip-solution to be more complicated.

alizedebray: The simplest way I see to have a one-tooltip solution with the current `post-tooltip` is using bindings:
```html
  @for (item of items; track item.id) {
    <post-tooltip-trigger for="itemDetails">
      <button
        (pointerenter)="currentItem = item"
        (focusin)="currentItem = item"
        class="btn btn-secondary"
      >
        {{ item.title }}
      </button>
    </post-tooltip-trigger>
  }
    
  <post-tooltip id="itemDetails">
    <p><strong>Status:</strong> {{ currentItem.shippingInfo }}</p>
    <p><strong>Quantity:</strong> {{ currentItem.quantity }}</p>
  </post-tooltip>
```

However the `post-tooltip-trigger` also listens to `long-press` custom events so you will end up with incorrect tooltip contents on mobile.

A possible solution would be to add an custom event to the `post-tooltip-trigger`, which would also make the code more readable:
```html
  @for (item of items; track item.id) {
    <post-tooltip-trigger for="itemDetails" (postTrigger)="currentItem = item">
      <button class="btn btn-secondary">
        {{ item.title }}
      </button>
    </post-tooltip-trigger>
  }
    
  <post-tooltip id="itemDetails">
    <p><strong>Status:</strong> {{ currentItem.shippingInfo }}</p>
    <p><strong>Quantity:</strong> {{ currentItem.quantity }}</p>
  </post-tooltip>
```

The best way will be to manually trigger the tooltip via an Angular service that creates the tooltip on demand with the expected context. That is what ng-bootstrap is offering but probably to Angular-specific for our Design System.

PR #5648, opened 2025-06-16
fix(styles): fixed the `back-button` disalignment in the mobile menu

alionazherdetska:
## 📄 Description

This PR fixes the disalignment of the back button in the mobile menu by restructuring how the target group and navigation elements are positioned. The main changes include:

- Removed absolute positioning of the `back-button` that was causing overlap issues with the target group
- Added `target-group slot`  to post-mainnavigation to be able to style it from there

## 🚀 Demo

---

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-header/post-header.scss
+++ b/packages/components/src/components/post-header/post-header.scss
@@ -265,6 +265,14 @@
     gap: 32px;
   }
 
+  @include media.max(sm) {
+    :has([slot="target-group"]) {
+      ::slotted(post-mainnavigation) {
+        gap: 24px;
+      }
+    }
+  }
+
   .navigation-footer {
     background-color: #f0efed;
     gap: 24px;
@@ -276,9 +284,5 @@
   ::slotted(post-mainnavigation),
   .navigation-footer {
     padding-inline: 16px;
-  }
-
-  ::slotted(post-mainnavigation) {
-    gap: 24px;
   }
 }
--- a/packages/components/src/components/post-header/post-header.scss
+++ b/packages/components/src/components/post-header/post-header.scss
@@ -230,8 +230,7 @@
   }
 
   ::slotted(post-mainnavigation),
-  .navigation-footer,
-  .navigation-target-group {
+  .navigation-footer {
     display: none;
     flex-direction: column;
 
@@ -277,5 +276,9 @@
   ::slotted(post-mainnavigation),
   .navigation-footer {
     padding-inline: 16px;
+  }
+
+  ::slotted(post-mainnavigation) {
+    gap: 24px;
   }
 }
--- a/packages/components/src/components/post-header/post-header.tsx
+++ b/packages/components/src/components/post-header/post-header.tsx
@@ -354,6 +354,9 @@ export class PostHeader {
             </div>
           </div>
           <div class="global-sub">
+            {this.device === 'desktop' && <slot name="target-group"></slot>}
+          </div>
+          <div class="global-sub">
             {this.device === 'desktop' && <slot name="meta-navigation"></slot>}
             <slot name="global-controls"></slot>
             {this.device === 'desktop' && <slot name="post-language-switch"></slot>}
--- a/packages/components/src/components/post-header/post-header.tsx
+++ b/packages/components/src/components/post-header/post-header.tsx
@@ -23,6 +23,7 @@ import { eventGuard } from '@/utils/event-guard';
  * @slot title - Holds the application title.
  * @slot default - Custom controls or content, right aligned in the local header.
  * @slot post-mainnavigation - Has a default slot because it's only meant to be used in the `<post-header>`.
+ * @slot target-group - Holds the list of buttons to choose the target group.
  */
 
 @Component({
--- a/packages/components/src/components/post-header/post-header.tsx
+++ b/packages/components/src/components/post-header/post-header.tsx
@@ -23,7 +23,6 @@ import { eventGuard } from '@/utils/event-guard';
  * @slot title - Holds the application title.
  * @slot default - Custom controls or content, right aligned in the local header.
  * @slot post-mainnavigation - Has a default slot because it's only meant to be used in the `<post-header>`.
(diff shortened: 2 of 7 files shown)


packages/components/src/components/post-header/post-header.scss
```diff
@@ -230,8 +230,7 @@
   }
 
   ::slotted(post-mainnavigation),
-  .navigation-footer,
-  .navigation-target-group {
+  .navigation-footer {
```
gfellerph: That selector already exists on line 226/227, please merge them. This will also fix the sonar code smell.

packages/components/src/components/post-mainnavigation/post-mainnavigation.tsx
```diff
@@ -31,6 +37,14 @@ export class PostMainnavigation {
...
   }
 
+  connectedCallback() {
+    window.addEventListener('postBreakpoint:name', this.breakpointChange);
+  }
+
+  private readonly breakpointChange = (e: CustomEvent) => {
+    this.device = e.detail;
+  };
+
```
gfellerph: Why was this added to this component? The state is not being used anywhere.

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -333,11 +333,6 @@ export class PostHeader {
         style={{ '--post-header-navigation-current-inset': `${mobileMenuScrollTop}px` }}
       >
         <div class="mobile-menu" ref={el => (this.mobileMenu = el)}>
-          <div class="navigation-target-group">
```
gfellerph, line 332: Moving around markup here is a breaking change. Is there another way to make this work?

gfellerph (changes requested):

packages/components/src/components/post-header/post-header.tsx
oliverschuerch: Can you explain why changing shadow dom would produce a breaking change?

packages/components/src/components/post-mainnavigation/post-mainnavigation.tsx
alionazherdetska: Removed as redundant

packages/components/src/components/post-header/post-header.scss
```diff
@@ -113,11 +113,16 @@
...
+    transition: height 0.2s ease-in-out;
     height: var(--post-logo-height);
   }
 }
 
+::slotted(post-mainnavigation) {
+    @include media.min(lg) {
+    display: none;
+  }
+}
```
oliverschuerch: As far as I can see, this is immediately overridden by a style defined either in the `post-mainnavigation` component itself or somewhere in a global style sheet. So why is this even needed?
image


Please at least fix indentation.

packages/components/src/components/post-mainnavigation/post-mainnavigation.scss
```diff
@@ -11,7 +11,22 @@
...
 post-mainnavigation {
-  // reset links and buttons
+
+  .mobile-header-section {
+    @include media.min(lg) {
+      display: none;
+    }
+    
+    @include media.max(lg) {
+      display: none;
```
oliverschuerch: Is it necessary to nest this property definition in two separate mediaqueries?

packages/components/src/components/post-mainnavigation/post-mainnavigation.tsx
```diff
@@ -100,8 +100,10 @@ export class PostMainnavigation {
   }
 
   private handleBackButtonClick() {
-    const header = this.navbar.closest<HTMLPostHeaderElement>('post-header');
-    if (header) header.toggleMobileMenu();
+    const header = document.querySelector<HTMLPostHeaderElement>('post-header');
```
oliverschuerch: The `post-mainnavigation` and the `post-header` are coupled by design (software desgin) and therefore should always ensure they really communicate to/with the right parent/child component.
Use for example `this.host.closest('post-header')` instead of implying that there can only be one `post-header` element per web page (may be right 99.9%, but if you do it correctly you will have no issue with the other 0.01% 😉).

packages/styles/src/components/target-group.scss
oliverschuerch, line 1: Why is that file even existing, when we reverted it and there is a PR ready to add it back?
- #5687

packages/components/src/components/post-mainnavigation/post-mainnavigation.tsx
alionazherdetska: @oliverschuerch  I initially tried implementing your suggestion to use `this.host.closest('post-header')`, but I discovered a technical limitation that prevents this approach from working.

The issue is that `post-mainnavigation` is rendered inside the Shadow DOM of `post-header` (since `post-header` uses `shadow: true`). The `closest()` method cannot traverse up through shadow boundaries, so `this.host.closest('post-header')` always returns `null`.

That's why I've implemented the reference approach instead!

packages/styles/src/components/target-group.scss
alionazherdetska: I reverted the changes to this file

.changeset/hot-birds-swim.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-components': major
```
alizedebray, line 2: This is also a breaking change for angular and react component packages.

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -329,11 +335,16 @@ export class PostHeader {
...
-            {(this.device === 'mobile' || this.device === 'tablet') && (
-              <slot name="target-group"></slot>
-            )}
-          </div>
+          {((this.device === 'mobile' || this.device === 'tablet') && this.mobileMenuExtended) && (
```
alizedebray, line 338:
```suggestion
          {((this.device !== 'desktop') && this.mobileMenuExtended) && (
```

Issue #4834, opened 2025-03-03
chore(deps): update stencil (`@stencil/angular-output-target`, `@stencil/core`, `@stencil/react-output-target`, `@stencil/sass`, `@stencil/store`)


gfellerph:

oliverschuerch:
@stencil/store is not installed yet. I'll skip that.

oliverschuerch:
# Components SSR State

Legend:
✅ Runs without hydration errors
⚠️ Runs without hydration errors depending on the implementation
❗ Does NOT run without hydration error. After fixing the current hydration error, different errors can occur!

| Name | State | Description |
| :- | :- | :- |
| post-accordion | ✅ | |
| post-accordion-item | ✅| |
| post-avatar | ✅| |
| post-back-to-top | ✅| |
| post-banner | ✅| |
| post-breadcrumbs | ✅| |
| post-breadcrumb-item | ✅| |
| post-card-control | ✅| |
| post-closebutton | ✅| |
| post-collapsible | ✅| |
| post-collapsible-trigger | :warning: | The aria attributes need to be defined on the button already, otherwise nextjs will report a hydration error, see [example](#user-content-post-collapsible-trigger) below |
| post-footer | ✅| |
| post-header | ❗ | Hydration error, because it adds style attribute to the html-tag, no clue how to solve that by now |
| post-icon | ✅| |
| post-language-switch| ✅| |
| post-language-option| ✅| |
| post-linkarea | ✅| |
| post-list | ❗ | Hydration error, because it is a none-shadow component and adds "slotted" elements to the DOM on runtime. Make it a shodow component, should fix the hydration error |
| post-list-item | ✅ | |
| post-logo | ✅| |
| post-mainnavigation | ❗ | Hydration error, because it is a none-shadow component and adds "slotted" elements to the DOM on runtime. Make it a shodow component, should fix the hydration error |
| post-megadropdown-trigger | ✅| |
| post-megadropdown | ❗ | Hydration error, because it is a none-shadow component and adds "slotted" elements to the DOM on runtime. Make it a shodow component, should fix the hydration error |
| post-menu-trigger | ❗ | Hydration error, because it is a none-shadow component and adds "slotted" elements to the DOM on runtime. Make it a shodow component, should fix the hydration error |
| post-menu | ❗ | Hydration error, because it is a none-shadow component and adds "slotted" elements to the DOM on runtime. Make it a shodow component, should fix the hydration error |
[...]

PR #5650, opened 2025-06-16
feat(components): update stencil and implement the hydration app for SSR compatibility

oliverschuerch:
## 📄 Description

This PR updates stenciljs in our components package to the latest version and outputs the hydration app.
It also adds the hydration app to our react-output-target, so that our web-components can be handled differently when rendered on the server than on the client.

However, for the moment the solution only works when we stick to the @stencil/react-output-target package version 0.8.2. Any newer Version breaks the whole functionality. To fix this, we need to investigate what changes there and if our setup is correctly done.

---

## 📝 Checklist

- ⚠️ ~~My changes generate no new warnings or errors~~
   -> Hydration errors occur in a nextjs setup (see: https://github.com/swisspost/design-system/issues/4834#issuecomment-2975293100)


packages/components/cypress/e2e/menu-button.cy.ts
```diff
@@ -56,6 +56,7 @@ describe('menus', { baseUrl: null, includeShadowDom: true }, () => {
       cy.get('post-menu-trigger[for="menu-three"]').eq(1).as('triggerD');
       cy.get('post-menu#menu-three').as('menuC');
       cy.get('post-menu#menu-four').as('menuD');
+      cy.wait(500);
```
gfellerph: That does not seem like the proper way to handle this. An element selector would be more appropriate since cypress is waiting for it, for example `cy.get('post-menu#menu-four[data-hydrated]')` or something along those lines.

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -68,6 +68,7 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
     });
 
     it('should open and close with the API', () => {
+      cy.wait(1000);
```
gfellerph: Same here, wait for elements being hydrated instead of an arbitrary timeout.

packages/styles/gulpfile.js
```diff
@@ -11,6 +11,13 @@ const autoprefixer = require('autoprefixer');
 const { globSync } = require('glob');
 const options = require('./package.json').sass;
 
+const COMPONENT_SOURCE_PATHS = [path.join(__dirname, '../components/src/components')];
+const COMPONENT_NOT_DEFINED_TEMPLATE_PATH = path.join(
+  __dirname,
+  'src/templates/_not-defined.template.scss',
+);
+const COMPONENT_NOT_DEFIEND_OUTPUT_PATH = path.join(__dirname, 'src/utilities/_not-defined.scss');
+
```
gfellerph: It does not seem like these constants are being used. Should we remove them?

gfellerph (changes requested):

packages/components/cypress/e2e/menu-button.cy.ts
oliverschuerch: I just wanted to have a quick way to see if the rest of the PR works and then probably forgot about this...
I've changed it now, let's hope the built still becomes green🤞🏼

packages/documentation/.storybook/helpers/register-web-components.ts
```diff
@@ -1,13 +1,12 @@
+import '@swisspost/design-system-components';
 import { defineCustomElements as defineHeader } from '@swisspost/internet-header/loader/index.es2017.js';
-import { defineCustomElements as defineComponents } from '@swisspost/design-system-components/loader';
 import { setStencilDocJson } from '@kurbar/storybook-addon-docs-stencil';
 import { StencilJsonDocs } from '@kurbar/storybook-addon-docs-stencil/dist/types';
 import postComponentsDocJson from '@swisspost/design-system-components/dist/docs.json';
 import internetHeaderDocJson from '@swisspost/internet-header/dist/docs.json';
 import '../../src/shared/link-design/link-design.component';
 
 defineHeader();
```
alizedebray, line 9: Why do we need to define the header and the components?

alizedebray (approved):

PR #5655, opened 2025-06-16
chore: update storybook to 9

gfellerph:
## 📄 Description

Updates storybook to the new major version of storybook, v9. All suggested auto updates have been applied, thus the many changes in the files.

Known bugs:
- There is a react error connected to the `sourceState="shown"` attribute. If it's not there, code samples are not open by default and the console error is gone. If the attribute is set, the error appears in the console but the page generally seems to work. Further investigation is necessary, but the issue does not seem blocking.

---

## 📝 Checklist


packages/documentation/.storybook/main.ts
```diff
@@ -90,13 +85,27 @@ const config: StorybookConfig = {
...
     });
   },
+
+  features: {
+    actions: false,
+    backgrounds: false,
+    highlight: false,
+    outline: false,
+    measure: false,
+    viewport: false,
```
alizedebray, line 99: We should try this addon, it was broken in the previous version but would be useful in my opinion!

alizedebray (approved):

packages/nextjs-integration/tsconfig.tsbuildinfo
alizedebray, line 1: Is this change intended?

alizedebray (changes requested):

packages/nextjs-integration/tsconfig.tsbuildinfo
gfellerph: No, reverted it.

packages/documentation/.storybook/main.ts
gfellerph: Can we do that in a separate issue?

alizedebray (approved):

packages/nextjs-integration/tsconfig.tsbuildinfo
alizedebray: Changes are still here.

packages/documentation/.storybook/main.ts
alizedebray: Done.

PR #5666, opened 2025-06-17
fix(setup): versions.json updating within release workflow

schaertim:
## 📄 Description
There was an issue in how the release workflow was updating the versions.json. There are essentially two cases:

**Minor release** (e.g. 7.2 -> 7.3):
This case worked fine. The entry for version 7.x in the versions.json was updated as well as the dependencies.

**Major releases** (e.g. 7.x -> 8.0):
Here is where the problem was. The code not only added a new entry for the 8.0 version but also wrongfully updated the dependencies of the 7.x entry to be the same as the 8.0 dependencies. That way every time a new major release was done the dependencies of the previous major version are messed up.

This pr fixes this. For major releases the dependencies of the previous versions are left alone and the new entry is just added at the start. I tried to test this by mocking the workflow to see how the versions.json would look like and it seems correct. The fix will only adress future releases so the current versions.json is still messed up. Thats why i corrected it manually using the information form the packages.json files.

## 🚀 Demo

---

## 📝 Checklist


.github/workflows/release.yaml
```diff
@@ -275,14 +275,15 @@ jobs:
             const versions = JSON.parse(fs.readFileSync(versionsPath, 'utf8'))
 
             const currentVersion = versions.find(version => version.version.startsWith(release.old.major))
+
+            if (!release.isMajor) {
             currentVersion.version = release.new.version
             currentVersion.dependencies = release.new.dependencies
+            }
```
alizedebray:
```suggestion
            if (!release.isMajor) {
              currentVersion.version = release.new.version
              currentVersion.dependencies = release.new.dependencies
            }
```

packages/documentation/public/assets/versions.json
```diff
@@ -59,12 +59,12 @@
...
     "title": "Version 7",
-    "version": "7.4.0",
+    "version": "7.4.13",
     "description": "Pattern documentation, code snippets and implementation guidelines for the Design System Styles.",
     "url": "https://design-system-version-7.netlify.app",
     "dependencies": {
-      "@angular/core": "^18.0.0",
-      "@ng-bootstrap/ng-bootstrap": "^17.0.0",
+      "@angular/core": "^16.0.0 || ^17.0.0",
+      "@ng-bootstrap/ng-bootstrap": "^15.0.0 || ^16.0.0",
```
alizedebray, line 67: This is a good idea, but is the workflow able to write version range? If not, it would be good to open a ticket for this. 

Also the version selector in the docs is not showing both version, which can be misleading. Can you open a ticket for this as well?

schaertim: The workflow should be able to write the version range without an issue. I will create a ticket for the version selector.

PR #5669, opened 2025-06-17
5508 generate dependency tables from versionsjson

schaertim:
## 📄 Description
This only implements part of the ticket. The information available in the `versions.json` does not allow us to dynamically generate the current release line table so only the dependencies table is filled dynamically. For this a small react component is introduced which reads uses the contents of the `versions.json` file, formats the versions and generates the table.

## 🚀 Demo
Before:
image

After:
image

---

## 📝 Checklist


alizedebray:
> The information available in the `versions.json` does not allow us to dynamically generate the current release line table

Can you create a ticket listing the information missing so that we add them and generate the table in the future?

packages/documentation/src/stories/packages/styles/dependenciesTable.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const match = version.match(/^(\d+)\./);
+    return match ? `${match[1]}.x` : version;
+  };
+
+  return (
+    <div className="table-responsive table-fit-content text-start">
+      <table aria-describedby="dependencies" className="table">
+        <thead>
+          <tr>
+            {['Version', 'Bootstrap', 'Angular', 'Ng-Bootstrap'].map(header => (
```
alizedebray:
```suggestion
            {['Version', 'Bootstrap', 'Angular', 'ng-Bootstrap'].map(header => (
```

pnpm-lock.yaml
alizedebray, line 1: Why do you have changes in the lock file?

PR #5670, opened 2025-06-17
fix(workflow): fix release action on main

alionazherdetska:
## Changes
- Adds temporary Git commit step before `pnpm publish --dry-run` to ensure:
  - Package manager sees a clean working tree (matching real publish conditions)
  - All build artifacts are properly included in validation
  - No false negatives from uncommitted changes

---

## 📝 Checklist


.github/workflows/dry-run-publish.yaml
```diff
@@ -22,5 +22,12 @@ jobs:
       - name: Install & Build Dependencies
         run: pnpm bootstrap
 
+      - name: Stage and commit build artifacts
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
```
alizedebray: Can you set GitHub user using this (new) action: https://github.com/swisspost/design-system/blob/main/.github/actions/configure-github-user/action.yaml

You can update the action if you need, adding inputs for example.

alizedebray (changes requested):

PR #5698, opened 2025-06-24
chore(docs): added component package icons to the sidebar navigation

schaertim:
## 📄 Description

This pr adds icons to the storybook sidebar navigation which indicate which type of component (CSS, WC or Angular) it is. Because some components could have multiple versions (CSS & WC for example) the icons were added after the component name. This way multiple icons can be displayed after each other. 

## 🚀 Demo
image

---

## 📝 Checklist

### DIFF
--- a/.changeset/pink-schools-double.md
+++ b/.changeset/pink-schools-double.md
@@ -1,0 +1,13 @@
+---
+'@swisspost/design-system-styles': major
+'@swisspost/design-system-documentation': minor
+---
+
+Removed deprecated classes and SCSS variables for the spinner:
+
+- `.loader-xs`
+- `.loader-sm`
+- `$spinner-size-xs`
+- `$spinner-size-sm`
+- `$spinner-border-width-xs`
+- `$spinner-border-width-sm`
--- a/packages/documentation/src/stories/components/spinner/spinner.docs.mdx
+++ b/packages/documentation/src/stories/components/spinner/spinner.docs.mdx
@@ -12,11 +12,6 @@ import StylesPackageImport from '@/shared/styles-package-import.mdx';
 
 <p className="lead">Indicate a loading state and prevent interactivity behind it</p>
 
-<div className="banner banner-warning">
-  `.loader-xs` and `loader-sm` classes for sizes are deprecated and will be replace with
-  respectively `.loader-16` and `.loader-40`.
-</div>
-
 <Canvas sourceState="shown" of={SpinnerStories.Default} />
 
 <div className="hide-col-default">
--- a/packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
+++ b/packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
@@ -576,7 +576,8 @@ export class MigrationV99Component extends LitElement {
                 </li>
                 <li class="mb-16">
                   <p>
-                    Removed the dependency to ng-bootstrap as well as all of the components built on it:
+                    Removed the dependency to ng-bootstrap as well as all of the components built on
+                    it:
                     <span class="tag tag-sm tag-danger">breaking</span>
                   </p>
                   <ul>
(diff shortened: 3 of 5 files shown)


alizedebray:
**Proposed Resolution**
- add a different color to each icon to better differentiate them 
- align the icon to the left in front of the text

Ask the designer's opinion in the pattern discussion

oliverschuerch:
Resolution:
- [x] keep icons on the right side of the text but only render them in development mode
- [x] Add tags with icons on the docs page below the title (html, web-component, etc.) for our customers
- [x] Align package names with corresponding svg icons (package:styles -> CSS-Icon, package:NgBootstrap -> NgBootstrap Logo, etc.)

leagrdv:
@oliverschuerch As discussed, we will do the package type information in the documentation page in another ticket (already created here #5768)

I've also removed the changeset, as the changes are only for us working in development mode.

packages/documentation/.storybook/manager.ts
oliverschuerch, line 1: Since there are some changes in the manager.scss anyway, we could extract the inline styles on the image and the span elements wrapping those images and define them in the styles.

Please decide for yourself if it's worth or not.

packages/documentation/src/stories/components/banner/standard-html/banner.stories.ts
```diff
@@ -7,7 +7,7 @@ import { MetaComponent } from '@root/types';
 const meta: MetaComponent = {
   id: '105e67d8-31e9-4d0b-87ff-685aba31fd4c',
   title: 'Components/Banner',
-  tags: ['package:HTML'],
+  tags: ['package:Styles'],
```
oliverschuerch: We have a CSS and a WebComponent for the banners. Will that change in the near future? If not, I would add `package:WebComponents` here as well.

packages/documentation/src/stories/components/card-control/standard-html/card-control.stories.ts
oliverschuerch, line 1: There are two different story files for this component and one of them is used to show the icon in the navigation.
As long as we do not remove one of these components (CSS or WC), we should reflect that through the icons.

Therefore, I suggest to add both tags (`package:Styles` and `package:WebCompnents` in one of the files, so both icons can be shown in the navigation.

oliverschuerch (changes requested):
Some of the docs pages are not classified (have no tags):
- Foundations/Highlighted Text
- Foundations/Lead Text
- Foundations/Legend
- Foundations/Lists
- Foundations/Paragraph
- Foundations/Layout/*
- Raw Components/Internet Header/*
- Utilities/*

I don't know if this is intentionally or not. Decide for yourself or ask the team.

packages/documentation/.storybook/manager.ts
leagrdv: I moved the styles into the scss, and added the web components icon to the elements within the Internet Header package.

Regarding Foundations and Utilities, I don't feel like it makes sense to add it as they're all within the styles, so it's a bit redundant to add the same icon for each.

packages/documentation/src/stories/components/banner/standard-html/banner.stories.ts
oliverschuerch: I still don't see both icons on the banner navigation point.
image

packages/documentation/src/stories/components/card-control/standard-html/card-control.stories.ts
oliverschuerch: Same here...
image

packages/documentation/.storybook/manager.ts
oliverschuerch: It seems, that none of the files have changed in the pr since I last reviewed it. last commit is from 8th of august (merge main into current branch). Did you push your changes?

Regarding Foundation pages, we should either classify all of the or none. Since we have a mixup between web-components and css components, I would recommend to classify them.
On the other side, I'm fine if we do not classify Utility pages.
image

oliverschuerch (approved):

Issue #5697, opened 2025-06-24
[bug]: Investigate zusammen's footer issue


leagrdv:
### 🐞 Describe the bug

On the https://zusammen.post.ch/en website (mobile), there are two empty accordions that appear even though there is no content within. Is the problem on our side or theirs?

**Edit: The cause of the issue is on our side. We always expect to have 4 columns in the footer and if we don't receive the 4 of them, we still show the 4 accordions in mobile.**

It would be nice to add tests for that as well.

### 📷 Screenshots

![Image](https://github.com/user-attachments/assets/6cca4bca-21fd-4ad4-aa6c-e859d00d6648)

alizedebray:
First check with the PPNL team if the cause of the issue was already found.

PR #5709, opened 2025-06-25
chore(components): added e2e test for the `post-footer` component

schaertim:
## 📄 Description

Added more comprehensive e2e test for the `post-footer`. I also found a typo where it misspelled `post-accor**dd**ion` in the render function of the footer. Surprisingly it still partially worked because the browser just created a custom tag instead of using the `post-accordion` component.
## 🚀 Demo

---

## 📝 Checklist

### DIFF
--- a/.changeset/olive-pans-talk.md
+++ b/.changeset/olive-pans-talk.md
@@ -1,5 +1,0 @@
----
-'@swisspost/design-system-components': minor
----
-
-Added e2e test for the `post-footer` and fixed the render function to use the correct `post-accordion` component.


leagrdv (dismissed):
I took the fix for the `post-accordion` and moved it to its own PR #5723 in order for us to merge it faster. Looking at your tests made us realise that the bug on the zusammen's website was coming from our side. The footer should allow less than 4 columns and accordions should appear only if there is content. A fix for that is in progress by Alizé (#5697) so I think this PR should wait for that and be updated accordingly after this has been done.

PR #5736, opened 2025-07-01
feat(components): implement full-width tabs layout 

myrta2302:
## 📄 Description

This PR adds a `fullwidth` prop to the `post-tabs` component. When  `fullwidth` is set to `true`, the tabs header wrapper becomes full-width. It is extended to fit the 100% width of the screen, while the tabs content Remains aligned to the container.

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-tabs/post-tabs.scss
+++ b/packages/components/src/components/post-tabs/post-tabs.scss
@@ -3,3 +3,22 @@
 :host {
   display: block;
 }
+
+$container-max-width: var(--post-core-dimension-1280);
+
+:host([fullwidth]) .tabs-wrapper {
+  // Extra space outside the container
+  $extra-space: calc((100vw - #{$container-max-width}) / 2);
+  $space-to-compensate: max(0px, $extra-space);
+  $total-offset-per-side: calc(
+    max(var(--post-core-dimension-40), 0.5 * var(--post-grid-gutter-x)) + $space-to-compensate
+  );
+  // Negative margin to make the tabs-wrapper fullwidth
+  margin-inline: calc(-1 * $total-offset-per-side);
+  padding-inline: 0;
+
+  //Internal padding to re-align the tabs header to the container
+  .tabs {
+    padding-inline: $total-offset-per-side;
+  }
+}
--- a/packages/components/src/components/post-tabs/post-tabs.tsx
+++ b/packages/components/src/components/post-tabs/post-tabs.tsx
@@ -38,9 +38,16 @@ export class PostTabs {
   @Prop() readonly activePanel?: HTMLPostTabPanelElement['name'];
 
   /**
+   * When this property is set to true, the tabs container should extend across the
+   * * full width of the screen, spanning from edge to edge.
+   */
+  @Prop({ reflect: true }) fullWidth?: boolean;
+
+  /**
    * An event emitted after the active tab changes, when the fade in transition of its associated panel is finished.
    * The payload is the name of the newly shown panel.
    */
+
   @Event() postChange: EventEmitter<string>;
 
   componentDidLoad() {
@@ -194,7 +201,7 @@ export class PostTabs {
 
   render() {
     return (
-      <Host data-version={version}>
+      <Host data-version={version} fullwidth={this.fullWidth}>
         <div class="tabs-wrapper" part="tabs">
           <div class="tabs" role="tablist">
             <slot name="tabs" onSlotchange={() => this.enableTabs()} />
--- a/packages/components/src/components.d.ts
+++ b/packages/components/src/components.d.ts
@@ -448,6 +448,10 @@ export namespace Components {
          */
         "activePanel"?: HTMLPostTabPanelElement['name'];
         /**
+          * When this property is set to true, the tabs container should extend across the * full width of the screen, spanning from edge to edge.
+         */
+        "fullWidth"?: boolean;
+        /**
           * Shows the panel with the given name and selects its associated tab. Any other panel that was previously shown becomes hidden and its associated tab is unselected.
          */
         "show": (panelName: string) => Promise<void>;
@@ -1291,6 +1295,10 @@ declare namespace LocalJSX {
           * The name of the panel that is initially shown. If not specified, it defaults to the panel associated with the first tab.  **Changing this value after initialization has no effect.**
          */
         "activePanel"?: HTMLPostTabPanelElement['name'];
+        /**
+          * When this property is set to true, the tabs container should extend across the * full width of the screen, spanning from edge to edge.
+         */
+        "fullWidth"?: boolean;
         /**
           * An event emitted after the active tab changes, when the fade in transition of its associated panel is finished. The payload is the name of the newly shown panel.
          */
--- a/packages/components/src/components/post-tabs/readme.md
+++ b/packages/components/src/components/post-tabs/readme.md
@@ -7,9 +7,10 @@
 
 ## Properties
 
-| Property      | Attribute      | Description                                                                                                                                                                           | Type     | Default     |
-| ------------- | -------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- | ----------- |
-| `activePanel` | `active-panel` | The name of the panel that is initially shown. If not specified, it defaults to the panel associated with the first tab.  **Changing this value after initialization has no effect.** | `string` | `undefined` |
+| Property      | Attribute      | Description                                                                                                                                                                           | Type      | Default     |
+| ------------- | -------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------- | ----------- |
+| `activePanel` | `active-panel` | The name of the panel that is initially shown. If not specified, it defaults to the panel associated with the first tab.  **Changing this value after initialization has no effect.** | `string`  | `undefined` |
+| `fullWidth`   | `full-width`   | When this property is set to true, the tabs container should extend across the * full width of the screen, spanning from edge to edge.                                                | `boolean` | `undefined` |
 
 
 ## Events


.changeset/common-parts-stick.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': minor
+'@swisspost/design-system-components': minor
+---
+
+Added `full-width` prop to `post-tabs`component to allow tabs container to span full screen width.
```
alizedebray: As the change set end up in the change log, which is part of the documentation, it is better to use full sentences:
```suggestion
Added a `full-width` property to the `post-tabs` component to allow the tabs container to span the full screen width.
```

packages/components/src/components/post-tabs/post-tabs.tsx
```diff
@@ -194,7 +201,7 @@ export class PostTabs {
 
   render() {
     return (
-      <Host data-version={version}>
+      <Host data-version={version} fullwidth={this.fullWidth}>
```
alizedebray: This will be set automatically because of the `{ reflect: true }` option.
```suggestion
      <Host data-version={version}>
```

packages/components/src/components/post-tabs/post-tabs.tsx
```diff
@@ -37,10 +37,17 @@ export class PostTabs {
...
+   * When set to true, this property allows the tabs container to span the
+   * full width of the screen, from edge to edge.
+   */
+  @Prop({ reflect: true }) fullWidth?: boolean;
+
   /**
    * An event emitted after the active tab changes, when the fade in transition of its associated panel is finished.
    * The payload is the name of the newly shown panel.
    */
+
```
alizedebray:
```suggestion
```

packages/documentation/cypress/snapshots/components/segmented-button.snapshot.ts
alizedebray, line 1: Can you revert the changes to this file?

packages/documentation/src/stories/components/tabs/tabs.stories.ts
```diff
@@ -22,14 +22,26 @@ const meta: MetaComponent<HTMLPostTabsElement> = {
       control: 'select',
       options: ['first', 'second', 'third'],
     },
+    fullWidth: {
```
alizedebray: Is this necessary? The controls are usually automatically defined from the component.

packages/components/src/components/post-tabs/post-tabs.tsx
```diff
@@ -37,10 +37,17 @@ export class PostTabs {
    */
   @Prop() readonly activePanel?: HTMLPostTabPanelElement['name'];
 
+  /**
+   * When set to true, this property allows the tabs container to span the
+   * full width of the screen, from edge to edge.
+   */
+  @Prop({ reflect: true }) fullWidth?: boolean;
```
alizedebray: Setting boolean attribute to `false` by default helps users not having to define the property every time they use the component. 
```suggestion
  @Prop({ reflect: true }) fullWidth: boolean = false;
```

packages/documentation/src/stories/components/tabs/tabs.stories.ts
```diff
@@ -22,14 +22,26 @@ const meta: MetaComponent<HTMLPostTabsElement> = {
...
 };
 
 export default meta;
 
 function renderTabs(args: Partial<HTMLPostTabsElement>) {
   return html`
-    <post-tabs active-panel="${ifDefined(args.activePanel)}">
+    <post-tabs
+      active-panel="${ifDefined(args.activePanel)}"
+      fullwidth="${args.fullWidth ?? undefined}"
```
alizedebray:
```suggestion
      full-width="${args.fullWidth ?? undefined}"
```

packages/components/src/components/post-tabs/post-tabs.tsx
alizedebray: Please also update the documentation text and stories accordingly.

packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -3,3 +3,22 @@
 :host {
   display: block;
 }
+
+:host([fullwidth='true']) .tabs-wrapper {
```
alizedebray:
```suggestion
:host([full-width='true']) .tabs-wrapper {
```

packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -3,3 +3,22 @@
...
+  // Extra space outside the container
+
+  $extra-space: calc((100vw - var(--post-core-dimension-1280)) / 2);
+  $space-to-compensate: max(0px, $extra-space);
+  $total-offset-per-side: calc(
+    max(var(--post-core-dimension-40), 0.5 * var(--post-grid-gutter-x)) + $space-to-compensate
+  );
```
alizedebray: Core token should never be used to style a component directly. Please check [how the containers are build](packages/styles/src/layout/_containers.scss) and use the same tokens to create the style for the full-width tabs.

Also, the tabs aren't currently full-width on tablet, can you correct that?

alizedebray (changes requested):

packages/documentation/src/stories/components/tabs/tabs.styles.scss
alizedebray, line 1: Can you check if using the `layout: 'fullscreen'` parameter works? This way you don't have to duplicate this styles.

It is for example used in the header stories: https://github.com/swisspost/design-system/blob/main/packages/documentation/src/stories/components/header/header.stories.ts#L12
But you should be able to use it at story level as well.

packages/documentation/src/stories/components/tabs/tabs.stories.ts
```diff
@@ -2,6 +2,7 @@ import { StoryObj } from '@storybook/web-components';
 import { html } from 'lit';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { MetaComponent } from '@root/types';
+import './tabs.styles.scss';
 
 const meta: MetaComponent<HTMLPostTabsElement> = {
```
alizedebray: Can you add a `args: { fullWidth: false }` to the metadata so that the control is correctly initialized?

packages/documentation/src/stories/components/tabs/tabs.stories.ts
```diff
@@ -29,7 +30,10 @@ export default meta;
 
 function renderTabs(args: Partial<HTMLPostTabsElement>) {
   return html`
-    <post-tabs active-panel="${ifDefined(args.activePanel)}">
+    <post-tabs
+      active-panel="${ifDefined(args.activePanel)}"
+      full-width="${args.fullWidth ? true : undefined}"
```
alizedebray: This renders an empty `full-width` attribute in the code snippet: 
image

packages/documentation/src/stories/components/tabs/tabs.snapshot.stories.ts
alizedebray, line 1: To see the real result, the snapshot stories should not be contained in padded containers

packages/documentation/src/stories/components/tabs/tabs.snapshot.stories.ts
```diff
@@ -17,11 +17,16 @@ export const Tabs: Story = {
...
-          ${bombArgs({
-            activePanel: [undefined, 'tria'],
-          }).map((args: Args) => meta.render?.({ ...context.args, ...args }, context))}
-        </div>
+        ${['container', 'container-fluid', ''].map(
+          containerClass => html`
+            <div class="${containerClass}">
+              ${bombArgs({
+                activePanel: [undefined, 'third'],
+                fullWidth: [false, true],
```
alizedebray: If there is no container, the full-width and not full-width tabs look the same. Can you filter out the use case where `containerClass === '' && fullWidth === true`?

packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -1,5 +1,38 @@
...
+      $gutter: tokens.get('grid-gutter-#{$breakpoint}', components.$post-container);
+
+      $container-inline-padding: max(#{$padding}, calc(0.5 * #{$gutter}));
+
+      // Extra space outside the container
+      $extra-space: calc((100vw - $content-width) / 2);
+      $space-to-compensate: max(0px, $extra-space);
+      $total-offset-per-side: calc($container-inline-padding + $space-to-compensate);
+
+      .tabs-wrapper {
```
alizedebray: The concept behind the styles is correct, butyou never verify if the tabs are inside a container.
Right now, the same negative margin is applied regardless of whether the tabs are within a `.container`, a `.container-fluid`, or not inside any container at all.

packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -1,5 +1,38 @@
...
   display: block;
 }
+
+:host([full-width='true']) {
+  @each $breakpoint in map.keys(breakpoints.$grid-breakpoints) {
+    @include media.min($breakpoint) {
+      $content-width: tokens.get('grid-max-width', components.$post-container);
+      $padding: tokens.get('grid-padding-#{$breakpoint}', components.$post-container);
+      $gutter: tokens.get('grid-gutter-#{$breakpoint}', components.$post-container);
+
```
alizedebray: It would be better to define these values as CSS custom properties directly on the containers. This way, `$total-offset-per-side` can be calculated dynamically based on the tabs’ parent container, rather than using fixed values.

packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -1,5 +1,38 @@
...
+      $container-inline-padding: max(#{$padding}, calc(0.5 * #{$gutter}));
+
+      // Extra space outside the container
+      $extra-space: calc((100vw - $content-width) / 2);
```
alizedebray: This actually is the container's margin: 
```suggestion
      $container-margin: calc((100vw - $content-width) / 2);
```

packages/documentation/src/stories/components/tabs/tabs.styles.scss
myrtasakellariou: There is still a p-16 added to the scheme div. `<div class="p-16 palette-default" data-color-scheme="light">`

alizedebray, line 1: The padding from the palette will be removed as soon as this PR is merged: https://github.com/swisspost/design-system/pull/5250/files#diff-429f5ed6ca203bdf9ebdb0d30f7e84123f661d938703bc4e2d2c4feb262f9d87R10

You can therefore remove this file.

packages/styles/src/layout/_containers.scss
```diff
@@ -7,26 +7,39 @@
 
 tokens.$default-map: components.$post-container;
 
-.container,
-.container-fluid {
-  width: 100%;
-  margin-inline: auto;
+@each $class in (container, container-fluid) {
```
alizedebray: By adding this loop, you create two declaration blocks instead of one, increasing the size of the CSS output. 
Instead, you can leave the common styles in a block and create other blocks for styles that differs, like: 
```scss
.container,
container-fluid {
  // common styles
}

.container {
  // container offset and max-width  for the container
}

.container-fluid {
  // container offset for the fluid container
}
```

packages/styles/src/layout/_containers.scss
```diff
@@ -7,26 +7,39 @@
...
-  @each $breakpoint in map.keys(breakpoints.$grid-breakpoints) {
-    @include media.min($breakpoint) {
-      $gutter: tokens.get('grid-gutter-#{$breakpoint}');
-      $padding: tokens.get('grid-padding-#{$breakpoint}');
+    @each $breakpoint in map.keys(breakpoints.$grid-breakpoints) {
+      @include media.min($breakpoint) {
+        $gutter: tokens.get('grid-gutter-#{$breakpoint}');
+        $padding: tokens.get('grid-padding-#{$breakpoint}');
 
-      --post-grid-gutter-x: #{$gutter};
```
alizedebray: Why did you remove this variable?

packages/styles/src/layout/_containers.scss
```diff
@@ -7,26 +7,39 @@
...
+        // The `.row` has a negative margin equal to half the gutter size.
+        // To prevent content from overflowing, the container needs to have
+        // a padding of at least half the gutter size to offset this negative margin.
+        padding-inline: max(#{$padding}, calc(0.5 * #{$gutter}));
```
alizedebray: This calculation is repeated 3 times, you can create a custom property out of it and reuse it wherever necessary.

alizedebray (changes requested):

packages/styles/src/layout/_containers.scss
myrtasakellariou: This was by mistake.

packages/styles/src/layout/_containers.scss
```diff
@@ -22,11 +27,33 @@ tokens.$default-map: components.$post-container;
...
+      $gutter: tokens.get('grid-gutter-#{$breakpoint}');
+      $padding: tokens.get('grid-padding-#{$breakpoint}');
+
+      --post-grid-gutter-x: #{$gutter};
+
+      // The `.row` has a negative margin equal to half the gutter size.
+      // To prevent content from overflowing, the container needs to have
+      // a padding of at least half the gutter size to offset this negative margin.
+      --post-container-padding-inline: max(#{$padding}, calc(0.5 * #{$gutter}));
+      padding-inline: var(--post-container-padding-inline);
```
alizedebray: This code is duplicated, can you refactor to have it only once?

alizedebray (changes requested):

packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -1,5 +1,23 @@
...
 
 :host {
   display: block;
 }
+
+:host([full-width='true']) {
+  .tabs-wrapper {
+    // Negative margin to make the tabs-wrapper fullwidth
+    margin-inline: calc(-1 * var(--post-container-offset));
+    padding-inline: 0;
```
alizedebray: Is this padding necessary?
```suggestion
```

myrtasakellariou: No, i removed it.

alizedebray (approved):

PR #5755, opened 2025-07-04
feat(styles): update chips component to v2

leagrdv:
## 📄 Description

Updated the chip component and its documentation according to v2 of the design.

## 🚀 Demo

---

## 📝 Checklist


packages/documentation/src/stories/components/chip/chip.stories.ts
```diff
@@ -244,14 +258,13 @@ export const FilterRadioChip: Story = {
...
     radio: true,
-    size: 'Small',
     badge: true,
   },
 };
 
 export const Dismissible: Story = {
   render: ({ dismissed, ...args }, context) => html`
-    <div class="d-flex flex-wrap gap-8">
+    <div class="chip-dismissible-group">
```
alizedebray, line 257: You can add the class to the fieldset directly.

packages/styles/src/components/chip.scss
```diff
@@ -1,135 +1,191 @@
...
-  max-width: 100%;
-
-  @include utilities.focus-style() {
-    border-radius: chip.$chip-border-radius;
-    width: fit-content;
-  }
+  border-radius: tokens.get('chip-border-radius');
+  @include utilities.focus-style();
 
   &-label {
```
alizedebray: I wrote this code originally, but I’ve realized it’s better not to use the `&-` notation since it makes class names harder to search for. Can you please replace all instances of `&-` in the file?
```suggestion
  .chip-selectable-label {
```

alizedebray: Also I wonder if these classes are necessary we could simply target the `label` and `input` elements, this simplify the code. What do you think?

packages/styles/src/components/chip.scss
```diff
@@ -1,135 +1,191 @@
...
-.chip-dismissible {
+tokens.$default-map: components.$post-chip;
+
+button.chip-dismissible {
   position: relative;
-  padding-inline-start: chip.$chip-dismissable-padding-start;
-  gap: chip.$chip-gap;
+  padding-inline: tokens.get('chip-dismissible-padding-inline');
+  gap: tokens.get('chip-dismissible-gap-inline');
+  flex-direction: row-reverse;
```
alizedebray: I guess you added the `row-reverse` to avoid a breaking change here but it makes the chip difficult to predict. If a user adds an icon in front of the chip text and it is shown they will have a hard time understanding why.

packages/styles/src/components/chip.scss
```diff
@@ -1,135 +1,191 @@
...
+    width: tokens.get('close-size', components.$post-close);
+    margin-right: calc(
+      -1 * (
+          tokens.get('close-size', components.$post-close) - tokens.get(
+              'close-icon-size',
+              components.$post-close
+            )
+        ) / 2
+    );
   }
```
alizedebray:
```suggestion
  &::before {
    $button-size: tokens.get('close-size', components.$post-close);
    $icon-size: tokens.get('close-icon-size', components.$post-close);
    
    content: '';
    border-radius: tokens.get('close-border-radius', components.$post-close);
    height: $button-size;
    width: $button-size;
    margin-right: calc(-1 * ($button-size - $icon-size) / 2);
  }
```

packages/styles/src/components/chip.scss
leagrdv: I've updated the rest so that I can remove the `row-reverse`

packages/documentation/src/stories/components/chip/chip.stories.ts
leagrdv: The dismissible chips are `button` elements and are not within a fieldset. But if you meant the selectable ones, I've made the change!

packages/styles/src/components/chip.scss
```diff
@@ -138,27 +186,16 @@
...
 
-    > .badge {
-      @extend %badge-sm;
-      line-height: chip.$chip-badge-line-height;
-      border-width: chip.$chip-badge-border-width;
-      padding-inline: calc(var(--post-badge-padding-x) * 1.2);
-    }
-  }
+.chip-selectable-group {
+  gap: tokens.get('filter-chips-group-selectable-gap', components.$post-filter);
```
alizedebray, line 200: Why isn't it defined as a chip token like `chip-selectable-group-gap`?

packages/styles/src/components/chip.scss
```diff
@@ -1,135 +1,183 @@
...
-          background-color: chip.$chip-bg;
-        }
+      + label {
+        color: tokens.get('chip-selectable-selected-enabled-fg');
+        background-color: tokens.get('chip-selectable-selected-enabled-bg');
+        border-color: tokens.get('chip-selectable-selected-enabled-stroke');
+        padding-left: calc(
+          tokens.get('chip-selectable-padding-inline') + tokens.get('chip-selectable-icon') +
+            tokens.get('chip-selectable-gap-inline-selected-icon')
+        );
```
alizedebray, line 97: Personal preference — I find it more readable with separate variables rather than a multi-line `calc()`:
```suggestion
        $base-padding: tokens.get('chip-selectable-padding-inline');
        $icon-size: tokens.get('chip-selectable-icon');
        $icon-margin:  tokens.get('chip-selectable-gap-inline-selected-icon');
        padding-left: calc(#{$base-padding} + #{$icon-size} + #{$icon-margin});
```

alizedebray (approved):

packages/styles/src/components/chip.scss
leagrdv: In figma, Sandra separated the selectable chips group into another section called filter pattern, so to the design team, it seems like two different components.

PR #5772, opened 2025-07-08
fix(component): fix `<post-header>` component gap regression

alionazherdetska:
## 📄 Description

# Fix post-header gap regression caused by #5490

## 🐛 Problem Description

After PR #5490 "feat(component): implemented breakpoints utility in the post-header component", users are experiencing a visual gap between the global and local headers on desktop during initial page load. The gap appears briefly and then disappears once the page is fully loaded.

## 🔍 Root Cause Analysis

The regression was introduced in PR #5490 when the breakpoint detection logic was refactored. The issue stems from calling `this.updateLocalHeaderHeight()` in `connectedCallback()` before the shadow DOM is fully rendered:

**Problematic code from #5490:**
```typescript
connectedCallback() {
  // ... event listeners setup ...
  
  this.updateLocalHeaderHeight(); // ❌ Called too early!
  this.handleScrollParentResize();
  this.lockBody(false, this.mobileMenuExtended, 'mobileMenuExtended');
}
```

**What happens:**
1. `connectedCallback()` runs before shadow DOM elements exist
2. `updateLocalHeaderHeight()` tries to query `.local-header` but gets `null`
3. Sets `--post-local-header-height` to `0px` 
4. Creates visible gap between global and local headers
5. Later, ResizeObserver correctly updates the height and gap disappears

## ✅ Solution

Remove the premature `this.updateLocalHeaderHeight()` call from `connectedCallback()`. The ResizeObserver set up in `handleLocalHeaderResize()` properly handles height updates when the shadow DOM is ready.

**Fixed code:**
```typescript
connectedCallback() {
  // ... event listeners setup ...
  
  // ✅ Removed: this.updateLocalHeaderHeight();
  this.handleScrollParentResize();
  this.lockBody(false, this.mobileMenuExtended, 'mobileMenuExtended');
}
```

**Why this works:**
- ResizeObserver in `handleLocalHeaderResize()` triggers when shadow DOM is ready
- Height updates happen at the correct timing with valid dimensions
- No premature setting of `--post-local-header-height` to `0px`

## 🧪 Testing

**Before fix:**
```
🔌 connectedCallback → 📏 updateLocalHeaderHeight (element: false) → 🎨 height: '0px' → GAP!
```

**After fix:**
```
🔌 connectedCallback → 🎉 componentDidRender → 📏 ResizeObserver → 🎨 height: '112px' → ✅ No gap!
```

## 📋 Changes Made

- **File:** `packages/components/src/components/post-header/post-header.tsx`
- **Change:** Removed `this.updateLocalHeaderHeight();` call from `connectedCallback()`
- **Impact:** Eliminates FOUC (Flash of Unstyled Content) gap on initial page load

## 🎯 Verification

1. Load page and observe no gap between global and local headers
2. Resize window to ensure ResizeObserver still works correctly  
3. Test mobile/tablet breakpoints to ensure responsive behavior intact
4. Verify `--post-local-header-height` CSS custom property has correct values

## 📖 Related Issues

- Resolves header gap regression introduced in #5490
- Maintains all breakpoint utility improvements from #5490
- Preserves responsive behavior and functionality

## 🔄 Breaking Changes

None. This is a pure bug fix that restores the expected visual behavior.

---

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-header/post-header.tsx
+++ b/packages/components/src/components/post-header/post-header.tsx
@@ -38,6 +38,7 @@ export class PostHeader {
   private readonly throttledResize = throttle(50, () => this.updateLocalHeaderHeight());
   private scrollParentResizeObserver: ResizeObserver;
   private localHeaderResizeObserver: ResizeObserver;
+
   get scrollParent(): HTMLElement {
     const frozenScrollParent: HTMLElement | null = document.querySelector(
       '[data-post-scroll-locked]',
@@ -112,7 +113,6 @@ export class PostHeader {
     window.addEventListener('postBreakpoint:device', this.breakpointChange);
     this.switchLanguageSwitchMode();
 
-    this.updateLocalHeaderHeight();
     this.handleScrollParentResize();
     this.lockBody(false, this.mobileMenuExtended, 'mobileMenuExtended');
   }
--- a/.changeset/curvy-clowns-flow.md
+++ b/.changeset/curvy-clowns-flow.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Fixed `<post-header>` component layout shift on load. 


packages/components/src/components/post-header/post-header.tsx
oliverschuerch, line 1: At the first sight, for me this looks like we‘re only fighting the symptoms, instead of fixing the cause.

We should ask our self, why this function is calculating a wrong height for a millisecond, before it gets reset correctly?
Also, why is it reset correctly, because we‘re not calling this method twice, right?

oliverschuerch (changes requested):

packages/components/src/components/post-header/post-header.tsx
alionazherdetska: You are right. The function `updateLocalHeaderHeight` calculates wrong height because it's called in connectedCallback() before shadow DOM exists: `querySelector('.local-header')` returns null, so we get 0px.

- Early in connectedCallback() → 0px (fails)
- Later via ResizeObserver → 112px (works)

My updated fix moves the call of `updateLocalHeaderHeight` to componentDidLoad() when shadow DOM is guaranteed to exist, plus adds a guard inside the function to check if the localHeader exists.

Does this approach make sense for you?

oliverschuerch: Love it!

PR #5798, opened 2025-07-14
fix(docs): fixed `hide/show code` button by removing Prettier transform from `preview.ts`

alionazherdetska:
## 📄 Description

**Problem:**
The hide/show code button in Storybook docs was not functioning due to a breaking change in Prettier 3.6.2. The format() function now returns a Promise instead of a string, causing the source code transformation to fail.

**Root Cause:**
Prettier's format() function became asynchronous in version 3.6.2, but Storybook's source.transform expects a synchronous function that returns a string. This mismatch caused the code toggle functionality to break.

**Solution:**
Temporarily removed the Prettier formatting from `preview.ts`. 

---

## 📝 Checklist

### DIFF
--- a/packages/documentation/.storybook/preview.ts
+++ b/packages/documentation/.storybook/preview.ts
@@ -1,6 +1,5 @@
 import type { Preview } from '@storybook/web-components';
 import { extractArgTypes, extractComponentDescription } from '@kurbar/storybook-addon-docs-stencil';
-import { format } from 'prettier';
 import DocsLayout from './blocks/layout/layout';
 import {
   fullScreenUrlDecorator,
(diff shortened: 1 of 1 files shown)


oliverschuerch (commented):
The current Storybook version specifies that `transform` is indeed async:
https://storybook.js.org/docs/api/doc-blocks/doc-block-source#transform

Sadly we're still on storybook v8
https://storybook.js.org/docs/8/api/doc-blocks/doc-block-source#transform

oliverschuerch (approved):
I would like to keep this unmerged, unless we have no choice than releasing the current version! As the code panels will show unformatted code with this solution.
I'm only approving it, to unblock it for now.

alizedebray:
> I would like to keep this unmerged, unless we have no choice than releasing the current version! As the code panels will show unformatted code with this solution. I'm only approving it, to unblock it for now.

Actually merging this PR would help with the reviews, we can still wait for the release.
@oliverschuerch is it okay we you to merge it?

alionazherdetska:
Closing the issue, because we have decided to revert prettier update as a temporary solution before Storybook upate:
https://github.com/swisspost/design-system/pull/5806

PR #5809, opened 2025-07-15
chore: update issue templates

alizedebray:
Update issue templates using GitHub UI.


.github/ISSUE_TEMPLATE/component---base.md
```diff
@@ -1,10 +1,10 @@
 ---
 name: Component - Base
 about: Parent task to organize related sub-issues for a component.
-title: '[base] componentnamehere'
-labels: '🧩 base component'
+title: "[base] componentnamehere"
+labels: "\U0001F9E9 base component"
```
oliverschuerch: As far as I understand this has been updated with a tool. Anyway, it would be more human readable if we keep the icon character, instead of using the unicode value.

.github/ISSUE_TEMPLATE/component---base.md
```diff
@@ -1,10 +1,10 @@
...
 assignees: ''
-type: 'Feature'
```
oliverschuerch: Since we're not properly using/setting the `type` anyway, we should consider to hide it also on our tickets, because it adds no value.
What do you think?

oliverschuerch:
FYI: I added my feedback as general comments, so not to block the PR if I may not have the time to review again until end of week 😉

.github/ISSUE_TEMPLATE/component---base.md
alizedebray: I agree. Do you know how to hide this?

oliverschuerch: I thought it could be managed in the board's settings. But that might not be possible. If so, we'll just leave it as it is.

.github/ISSUE_TEMPLATE/6-component-tokens.md
```diff
@@ -2,9 +2,9 @@
 name: Component - Tokens
 about: Sub-issue to create design tokens for a component.
 title: '[tokens] componentnamehere'
-labels: '🔮 tokens', 'needs: design'
+labels: 'needs: design, 🔮 tokens'
```
oliverschuerch, line 5: Is the format of having two different labels in the same string, comma separated within the string really how this need to be defiend?

.github/ISSUE_TEMPLATE/7-component-html-css.md
```diff
@@ -2,9 +2,9 @@
 name: Component - HTML/CSS
 about: Sub-issue to implement an HTML/CSS component in the styles package.
 title: '[html/css] componentnamehere'
-labels: '📦 styles', 'needs: tokens'
+labels: 'needs: tokens, 📦 styles'
```
oliverschuerch, line 5: Check label format here as well.

.github/ISSUE_TEMPLATE/9-web-component-styles.md
```diff
@@ -2,9 +2,9 @@
 name: Component - Web Component Styles
 about: Sub-issue to implement styles for a web component.
 title: '[web component styles] componentnamehere'
-labels: '📦 components', 'needs: tokens', 'needs: logic'
+labels: '📦 components, needs: logic, needs: tokens'
```
oliverschuerch, line 5: And here

.github/ISSUE_TEMPLATE/6-component-tokens.md
alizedebray: Yes, that's what I had to correct

oliverschuerch (approved):

.github/ISSUE_TEMPLATE/component_request.yml
```diff
@@ -1,28 +0,0 @@
-name: New component request
```
leagrdv, line 1: How come you're removing that one?

leagrdv (approved):

PR #5813, opened 2025-07-17
feat(components): add post-user-menu

alizedebray:
## 📄 Description

This PR adds a new raw component: the `post-user-menu`.
Since this is a raw component, no changeset was added.
Tests will be added in another PR.

## 🚀 Demo

---

## 📝 Checklist


packages/components/src/components/post-user-menu/post-user-menu.tsx
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * An Object containing the personal data of the user currently logged-in.
+   */
+  @Prop() user?: User;
+
+  @Watch('user')
+  validateUser() {
+    checkEmptyOrType(this, 'user', 'object');
+  }
+
```
myrtasakellariou, line 57: `componentWillLoad() {
   this.validateUser();
}`

packages/components/src/components/post-user-menu/post-user-menu.tsx
```diff
@@ -0,0 +1,62 @@
...
+      ></post-avatar>
+    );
+  }
+
+  render() {
+    return (
+      <Host data-version={version}>
+        {this.user ? [
+          <post-menu-trigger for={this.menuId}>
+            <button>{this.renderAvatar()}</button>
```
myrtasakellariou: The `<button>` would maybe need an `aria-label`  a) as the accessible name and b) to trigger the focus mode of the screen readers and be able to open on `Enter` (tested with NVDA). 

There is an issue related to the `<post-menu>`. For the key navigation to work when Screen Readers are running, the focusable elements  inside <post-menu-item> must have `role="menuitem"` and the `<div part="menu">` must have `role="menu"`. 

Also i am not sure if this requirement is covered: _"Menu trigger and login should have a tooltip when no accessible name is being displayed"_ . In what case would the accessible name be missing? Should maybe both be used, to cover sighted and screen reader users?

alizedebray: > There is an issue related to the `<post-menu>`. For the key navigation to work when Screen Readers are running, the focusable elements inside must have `role="menuitem"` and the `<div part="menu">` must have `role="menu"`.

Since this issue is related to the `post-menu`, could you please create a separate ticket?

> Also i am not sure if this requirement is covered: _"Menu trigger and login should have a tooltip when no accessible name is being displayed"_ . In what case would the accessible name be missing? Should maybe both be used, to cover sighted and screen reader users?

In our case, the `post-user-menu` isn't documented as a standalone component and won’t be in the future. It will only be used in the header, where it will always have an accessible name.

myrtasakellariou: Ticket for the `<post-menu>` : https://github.com/swisspost/design-system/issues/5837

packages/components/src/components/post-avatar/post-avatar.scss
```diff
@@ -1,17 +1,21 @@
 @use '@swisspost/design-system-styles/tokens/components';
 @use '@swisspost/design-system-styles/functions/tokens';
+@use '@swisspost/design-system-styles/variables/animation';
 @use '@swisspost/design-system-styles/core' as post;
 
 tokens.$default-map: components.$post-avatar;
 
 :host {
+  --post-avatar-size: #{tokens.get('avatar-size')};
```
oliverschuerch, line 9: Should we document the possibility to override the size also in the docs?
Because bg & fg custom-properties are documented.

packages/components/src/components/post-icon/post-icon.scss
oliverschuerch, line 1: 1. Do we want users to override the default icon size by custom-property?
  If so, we should maybe document it or at least create a changeset.
2. Switching from `em` to `rem` prevents the icons from beeing as big as the current font-size, which changes their whole functionality and would (imo) be a breaking change.

Left: PR preview
Right: Next docs
image

packages/components/src/components/post-menu-item/post-menu-item.scss
```diff
@@ -65,4 +58,8 @@ post-menu-item {
       }
     }
   }
+
+  post-icon {
+    --post-icon-size: 1.5rem;
```
oliverschuerch: As described in the icons docs, you could do the exact same thing by just using `font-size: 1.5rem`.

packages/components/src/components/post-menu-item/post-menu-item.scss
oliverschuerch, line 1: Just so we are aware about that, this change also changes the styles for `post-breadcrumbs` and maybe other components using the same component under the hood.

Left: PR preview
Right: Next docs
image

packages/components/src/components/post-user-menu/post-user-menu.tsx
```diff
@@ -0,0 +1,88 @@
...
+  email?: string;
+}
+
+@Component({
+  tag: 'post-user-menu',
+  styleUrl: 'post-user-menu.scss',
+  shadow: true,
+})
+export class PostUserMenu {
+  private readonly menuId = `p${nanoid(11)}`;
```
oliverschuerch, line 18: I'm pretty sure this will cause a hydration error.
It would be great if you could check that. If it's the case, we need to find a general solution for this problem, because I totally see the need for that sometimes.

packages/components/src/components/post-user-menu/post-user-menu.tsx
```diff
@@ -0,0 +1,88 @@
...
+    this.validateDescription();
+    this.validateUser();
+  }
+
+  private renderAvatar() {
+    return (
+      <post-avatar
+        firstname={this.user.name}
+        lastname={this.user.surname}
+        email={this.user.email}
```
oliverschuerch, line 63: The `post-avatar` also has the possibility to add a prop:
- `userid`
  This is currently not working, because the intranet team updated to M365.
  @gfellerph and I had once a talk with them, but without a final solution. I feel like now would be a good moment to discuss this with @gfellerph and the intranet team again, maybe as a follow up task of this one here.
- `imageSrc`
  Since the `email` property is tied to gravatar, this property could quickly become the preffered one for most of the projects that will use the user-menu. Therefore it would be good to implement it.

packages/documentation/src/shared/decorators/click-blocker.ts
oliverschuerch, line 1: I'd expect this will be removed in favor of https://github.com/swisspost/design-system/pull/5492, befor the PR is merged.

packages/styles/src/components/avatar.scss
oliverschuerch, line 1: These are post-avatar global styles.
I'm not sure if we already decided about that, but it would be worth to check with the team, if the file could move over to the components package and become part of the `post-components.scss` file there.

What do you think?

alizedebray: I agree.

packages/documentation/src/shared/decorators/click-blocker.ts
alizedebray: I’ll leave this as is for now and check the status of the other PR before merging:
- If it’s already merged, I’ll make the necessary adjustments.
- If it’s not merged yet, I’ll leave a comment for @leagrdv to update her PR accordingly.

packages/components/src/components/post-user-menu/post-user-menu.tsx
alizedebray: I created the post-user-menu to work with the current user object used by the header on post.ch, which doesn’t include `userId` or `imageSrc`.

Since I’m unsure if these properties will be added to the object or what their names might be, I suggest keeping the user menu as is for now and adding `userId` and `imageSrc` later in a follow-up ticket if needed.

packages/components/src/components/post-user-menu/post-user-menu.tsx
alizedebray: I tried the PostUserMenu in our NextJs package and there was no hydration error.

alizedebray:
> I got your PR to review again, but there are still some unresolved issues. I'm going to set it as reviewed again until we have solutions or follow up tickets linked on these issues.

I kept them open so that you can see my answers, now everything is closed.

packages/components/src/components/post-menu-item/post-menu-item.scss
oliverschuerch: This is still the case, but I expect it's ok for the designers, since your implementation visually is much closer or exactly how I saw it on [Figma](https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-Next-Level?node-id=597-28993&t=PzXoGsgMR4NMYX0N-0).

packages/components/src/components/post-user-menu/post-user-menu.tsx
oliverschuerch: Do you have any insight for that? Is it just working as expected?

packages/components/src/components/post-user-menu/post-user-menu.tsx
oliverschuerch: There is still no possibility to define the post-avatar `imageSrc` value in the post-user-menu `user` object.

packages/styles/src/components/avatar.scss
oliverschuerch: Did you create a separate ticket to do that?

oliverschuerch (commented):
> > I got your PR to review again, but there are still some unresolved issues. I'm going to set it as reviewed again until we have solutions or follow up tickets linked on these issues.
> 
> I kept them open so that you can see my answers, now everything is closed.

Sorry, I didn't understand that.

PR #5830, opened 2025-07-21
chore(styles, docs): remove ng-bootstrap dependency

leagrdv:
## 📄 Description

Removed the dependency to `ng-bootstrap` on the styles package, as well as all of the components built on it any mentions of `ng-bootstrap` in the docs.


packages/documentation/src/shared/nb-bootstrap/ngb-component-banner.mdx
alizedebray, line 1: You can simply remove the docs, we will not be able to reuse them in the future anyways.

packages/documentation/src/stories/getting-started/developer/developer.docs.mdx
```diff
@@ -88,11 +88,7 @@ The design-system is formed with several packages and different compatibilities
           Angular components:
           <developer-list-component packageType={PackageType.Angular}></developer-list-component>
         </td>
-        <td>N/A</td>
-        <td>
-          <a href="https://ng-bootstrap.github.io">Ng-Bootstrap</a> module specific to the component
-        </td>
-        <td >N/A</td>
+        <td colSpan="3">Currently unavailable as the ng-bootstrap dependency was removed. We are working on implementing those components as web components which will be available in a future release of this version.</td>
```
alizedebray: Same here, you can remove the ng-bootstrap components altogether.

leagrdv:
> Did you check if the version switcher still works correctly with a version that has no ng-bootstrap dependency?

Just checked with the local versions.json file, and yes it does! But I see that it's actually using the versions.json file from https://design-system.post.ch/, so I will create a branch to change it there to remove both ng-bootstrap and bootstrap once I'm done with removing that as well :)

PR #5844, opened 2025-07-22
chore(docs): remove bootstrap and ng-bootstrap dependencies from versions.json

leagrdv:
## 📄 Description

Removed the bootstrap and ng-bootstrap dependencies from the versions.json file to update the VersionSwitcher.


.changeset/early-hats-do.md
alizedebray, line 1: If this has no impact on the publish documentation I would remove the changeset.

alizedebray (approved):

PR #5850, opened 2025-07-22
chore(styles, docs): remove bootstrap dependency

leagrdv:
## 📄 Description

Removed the bootstrap package from the dependencies of the `styles`, `documentation` and `internet-header` packages.
Removed all mentions of bootstrap within the docs as well as all imports of the bootstrap css in the `styles` package.

Some broken components were fixed in this PR (`table`, `product-card` and `button-group`) by internalizing the necessary code from bootstrap.

I will create tickets for the two remaining ones: `Slider` and `Card footer/header`.

The PR #5857 handles the removal of all of the variable overrides in the `styles` package that are no longer needed nor used, as well as the deletion of the slider (form range) component.


packages/documentation/.storybook/addons/addons.scss
```diff
@@ -1,11 +1,12 @@
 @use '@swisspost/design-system-styles/core' as post;
 @use '@swisspost/design-system-styles/functions/tokens' as tokens;
 @use '@swisspost/design-system-styles/tokens/utilities' as utilities;
+@use '@swisspost/design-system-styles/tokens/elements' as elements;
 
 tokens.$default-map: utilities.$post-spacing;
 
 .addon-label {
-  color: post.$body-color !important;
+  color: tokens.get('body-color', elements.$post-body) !important;
```
alizedebray: It does not exist yet, but once the palettes are merged this variable should be used instead of the body color: 
```suggestion
  color: var(--post-current-fg) !important;
```

packages/documentation/src/stories/foundations/layout/grid/grid.module.scss
```diff
@@ -1,6 +1,7 @@
 @use 'sass:list';
 @use 'sass:map';
 @use '@swisspost/design-system-styles/core' as post;
+@use '@swisspost/design-system-styles/layout' as layout;
```
alizedebray, line 4: Aren't layout variables exported from the `core` file?

packages/documentation/src/stories/introduction.data.ts
```diff
@@ -50,14 +48,6 @@ export const technologies: ITechnology[] = [
       alt: 'Gulp Logo',
     },
   },
-  {
-    name: 'Bootstrap',
-    href: `https://getbootstrap.com/docs/${getVersion(DEPENDENCIES['bootstrap'], 'Mm')}`,
-    img: {
-      src: '/assets/images/technologies/logo-bootstrap.svg',
```
alizedebray, line 57: Can you delete the file as well?

packages/styles/package.json
```diff
@@ -30,24 +30,15 @@
     "format": "prettier src/**/*.scss --write"
   },
   "peerDependencies": {
-    "@angular/core": "^19.0.0",
-    "@ng-bootstrap/ng-bootstrap": "^18.0.0",
-    "bootstrap": "~5.3.0"
+    "@angular/core": "^19.0.0"
```
alizedebray: The peer dependency to Angular can be removed as well, it was only there because of ng-bootstrap

alizedebray (changes requested):

packages/documentation/src/stories/foundations/layout/grid/grid.module.scss
leagrdv: No, only the contents of the function, variables and mixins folder are exported from core. Should I add the layout to the core file, or keep it as it?

packages/documentation/.storybook/addons/addons.scss
leagrdv: I've done the update, and I'll wait for the merge of the palettes PR to merge that one.

packages/documentation/src/stories/foundations/layout/grid/grid.module.scss
alizedebray: Ah! True, this is new that we have variables and mixin in the component folder. We will need to decide on what is exported from the styles package as some point, but it is probably to early for this since the styles are still subject to a lot of changes. For the moment you can keep it as it is

alizedebray (approved):

PR #5857, opened 2025-07-23
chore(styles, docs): remove bootstrap variables and classes cleanup

leagrdv:
## 📄 Description

Cleaned up the styles package and removed all of the leftover SCSS variables overrides as well as CSS classes that are no longer used within the project.
Deleted the slider (form range) component from the docs and styles.


packages/documentation/src/stories/components/select/select.stories.ts
```diff
@@ -222,6 +222,7 @@ const Template: Story = {
     const ariaDescribedBy =
       args.hint || args.validation !== 'null' ? ariaDescribedByParts.join(' ') : nothing;
 
+    console.log('idis', context.id);
```
alizedebray:
```suggestion
```

PR #5860, opened 2025-07-23
chore(components): update post avatar to handle errors if no gravatar is found

myrta2302:
## 📄 Description

This PR updates the `<post-avatar>` component to better handle cases where no Gravatar image is found, with the following:

1. The name initials are displayed by default before any image fetch.
2. The `getImageByProp()` function now first checks for a cached failed image url to avoid unnecessary fetch.
3. Introduces the storageKey as a combination of `userid` and `email` props to track changes and refresh the avatar accordingly.

## 📝 Checklist


packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -92,31 +107,66 @@ export class PostAvatar {
...
-        const r = await fetchImage();
+    const cachedImageRes = await this.getStorageItem(prop);
+    if (cachedImageRes?.failed) return false;
 
-        imageResponse.ok = r.ok;
-        imageResponse.url = r.url;
+    if (cachedImageRes?.ok && cachedImageRes.url) {
+      this.imageUrl = cachedImageRes.url;
+      this.imageAlt = `${this.firstname} ${this.lastname} avatar`;
+      this.avatarType = AvatarType.Image;
```
alizedebray: Can you refactor to avoid code duplication?

packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -92,31 +107,66 @@ export class PostAvatar {
...
+      return true;
+    }
 
-        this.imageUrl = imageResponse.url;
-        this.imageAlt = `${this.firstname} ${this.lastname} avatar`;
-        this.avatarType = AvatarType.Image;
+    const maxRetries = 2;
+    let delayMs = 100;
```
alizedebray: Configuration constants like this are usually declared outside the class.

packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -92,31 +107,66 @@ export class PostAvatar {
...
+        }
       } catch (error) {
-        this.removeStorageItem(prop);
-        console.info(`Loading avatar by type "${AvatarType.Image}" failed.`);
+        console.info(`Attempt ${attempt}: Network error or fetch failed.`, error);
+      }
+
+      if (attempt < maxRetries) {
+        await this.delay(delayMs);
+        delayMs *= 2; // exponential backoff (3 attempts)
```
alizedebray: If I'm not mistaken there is only 2 attempt currently. Can you either increase `maxRetries` or update the comment?

packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -163,6 +213,12 @@ export class PostAvatar {
     window?.sessionStorage?.removeItem(key);
   }
 
+  private async refreshAvatar() {
+    await this.removeStorageItem(this.userid);
+    await this.removeStorageItem(this.email);
+    this.getAvatar();
+  }
```
alizedebray: I'm not sure about this function: `getAvatar` includes logic to retrieve a cached image, but the cached image is being removed right before it's called. This seems to defeat the purpose of the cache.

packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -92,31 +107,66 @@ export class PostAvatar {
       if (!imageLoaded) this.getAvatarByInitials();
     }
   }
+  private async delay(ms: number) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+  }
```
alizedebray: This method is actually a utility function.

alizedebray (dismissed):

packages/components/src/components/post-avatar/post-avatar.tsx
myrtasakellariou: Yes, correct i forgot to update the comment.

packages/components/cypress/e2e/avatar.cy.ts
```diff
@@ -1,5 +1,22 @@
...
+  return `https://www.gravatar.com/avatar/${hash}?s=${GRAVATAR_SIZE}&d=${GRAVATAR_DEFAULT}&r=${GRAVATAR_RATING}`;
+}
+
+async function cryptify(key: string) {
+  return await crypto.subtle.digest('SHA-256', new TextEncoder().encode(key)).then(buffer => {
+    return Array.from(new Uint8Array(buffer))
+      .map(bytes => bytes.toString(16).padStart(2, '0'))
+      .join('');
+  });
+}
```
oliverschuerch: It would be nice, if we could share these constants and logic somehow between the component and the e2e test.
For example store and export them in a dedicated file aside the post-avatar.tsx, so you can use the same source on both sides.

packages/components/cypress/e2e/avatar.cy.ts
```diff
@@ -40,23 +57,26 @@ describe('Avatar', () => {
...
-      cy.get('@avatar').find('slot img').should('not.exist');
-      cy.get('@avatar').find('.initials').should('exist');
+        if (response.status === 200) {
+          cy.get('@avatar').find('slot img').should('exist');
+          cy.get('@avatar').find('.initials').should('not.exist');
+        } else {
+          cy.get('@avatar').find('slot img').should('not.exist');
+          cy.get('@avatar').find('.initials').should('exist');
+        }
+      });
```
oliverschuerch, line 64: As far as I understand, this will run only once, therefore the status code will never be 200.
My suggestion: keep the tests as they were implemented before but extend them by checking also the response status code.
Even if this was somehow already tested. Because if the component does not get any gravatar, it will not create the img in the slot.

packages/components/src/components/post-avatar/post-avatar.tsx
oliverschuerch, line 1: Imo this is way more change that @alizedebray requested in the ticket.

✨ Expected behavior
As a developer using the post-avatar component, I expect it to:

- Attempt to fetch the Gravatar image.
  This already happens with `imageLoaded = await this.getImageByProp(this.email, this.fetchImageByEmail.bind(this));`.
  The real problem is, that we need to stop the component from ending up in a life cycle loop, when the image response is not 200.
  This status is btw exactly reflected by the return value of the `getImageByProp` function.
  
  To make sure the same email property is not used twice, something simple like `if (!imageLoaded && this.email && this.email !== this.oldEmail)` could probably do the trick already.
- If no Gravatar is found, gracefully fall back to displaying the user’s initials.
  This is already the case.
- Avoid throwing or logging errors in the console when no image is available.
  Since we use the [fetch()](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) method, I don't see a way how to do this, but feel free to try it.
  
**All I had to change to avoid the life cycle loop**

image

packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -77,46 +89,118 @@ export class PostAvatar {
...
+        const reason = `Fetch failed with status ${response.status}`;
+        await this.updateStorage({
+          [`${type}ImageFailed`]: true,
+          [`${type}ImageFailedReason`]: reason,
+        });
+        return false;
       }
+    } catch (error) {
+      const reason = (error as Error).message;
+      console.info(`Network error for ${type}`, error);
```
oliverschuerch: Since it is an error, we should output it like one.

```suggestion
      console.error(`Network error for ${type}`, error);
```

oliverschuerch (changes requested):
Since it's working now, I'm fine with it generally.

However, I have two concerns:
1. It's not only fixing a bug, but adding features, therefore, please update the changeset file or split the pr into a bugfix and a feature (if possible).
2. Let @alizedebray review it as well, because I expect, she could be not fully ok with the change, since it is not exactly what she demanded for.

myrtasakellariou:
**Functionality Changes**

- Add AvatarStorage interface
- Introduce **@State() storageKey**
- Add a double **watcher on two props (userId, email)**; when either changes:
  a) Reset the image
  b) Initialize storage to an empty object
  c) Set avatar type to default (initials)
  d) Call this.getAvatar() to fetch the image
- On **onSlotChange**, update avatar type and perform related actions
- **Load avatar initials** on componentWillLoad
- Fetch avatar image on componentWillRender
- **Store fetch results** (URL or error) in sessionStorage under the storage key
- Session storage entries are never removed or reused to cache the already fetched keys (pairs of userId/email)

packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -70,53 +76,61 @@ export class PostAvatar {
...
+      return false;
+    }
+    if (!imageResponse?.ok) {
+      return false;
+    }
+    if (imageResponse.ok) {
+      this.imageUrl = imageResponse.url;
+      this.imageAlt = `${this.firstname} ${this.lastname} avatar`;
+      this.avatarType = AvatarType.Image;
+      return true;
```
oliverschuerch: This should be an if/else statement.
if imageResponse.ok use it, else return false;

Or keep the `return imageResponse.ok;` as it was.

packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -171,17 +173,40 @@ export class PostAvatar {
...
+      });
+    });
+    this.slottedImageObserver.observe(img, { attributes: true, attributeFilter: ['src'] });
   }
 
-  componentWillRender() {
-    this.slottedImage = this.host.querySelector('img');
-    this.getAvatar();
+  connectedCallback() {
+    this.getAvatarInitials();
```
oliverschuerch, line 214: Please add a comment in the code to explain, why this is necessary.

packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -171,17 +173,40 @@ export class PostAvatar {
...
+    this.getAvatarImage();
   }
 
-  componentDidLoad() {
+  componentWillLoad() {
```
oliverschuerch: I think there was a reason, why we had to validate only on `componentDidLoad`.
Have a look into the other components to find out, if this change should be reverted.

packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -171,17 +173,40 @@ export class PostAvatar {
...
+    assignedNodes.forEach(node => {
+      if (node.nodeType === Node.ELEMENT_NODE) {
+        const el = node as Element;
+        if (el.tagName === 'IMG') {
+          this.observeImageSrcChanges(el as HTMLImageElement);
+        }
+      }
+    });
+    this.getAvatarImage();
+  }
```
oliverschuerch, line 195:
```suggestion
  }

```

packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -70,53 +76,61 @@ export class PostAvatar {
...
-        this.imageAlt = `${this.firstname} ${this.lastname} avatar`;
-        this.avatarType = AvatarType.Image;
-
-        this.setStorageItem(this.userid, JSON.stringify(imageResponse));
-      } catch (error) {
-        this.removeStorageItem(prop);
-        console.info(`Loading avatar by type "${AvatarType.Image}" failed.`);
-      }
+    let imageResponse: Response;
+    try {
```
oliverschuerch:
```suggestion
    let imageResponse: Response;

    try {
```

packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -70,53 +76,61 @@ export class PostAvatar {
...
+    let imageLoaded = false;
+    this.slottedImage = this.host.querySelector('img');
+    const imageUrl = this.slottedImage?.getAttribute('src');
+    if (!imageUrl) {
+      if (this.userid) {
         imageLoaded = await this.getImageByProp(this.userid, this.fetchImageByUserId.bind(this));
-
-      if (!imageLoaded && this.email)
+      }
+      if (!imageLoaded && emailPattern.exec(this.email ?? '') !== null) {
```
oliverschuerch: It would be much more readable (imo), and also more consistent with the rest of the code, if you'd write it like so:

`!imageLoaded && this.email && this.email.match(emailPattern)`
Meaning: imageLoaded is not loaded, this.email is defiend and matches pattern

While `emailPattern.exec(this.email ?? '') !== null` is way harder to understand.
Meaning: execute pattern on this.email or if not defined use '', must be not null.

packages/components/src/components/post-avatar/post-avatar.tsx
oliverschuerch, line 1: Please add some empty lines between var definition and if statements which do not belong together, in `getAvatarImage` function, to improve readability.

packages/components/src/components/post-avatar/post-avatar.tsx
```diff
@@ -70,53 +76,61 @@ export class PostAvatar {
...
   validateUserid() {
     checkEmptyOrType(this, 'userid', 'string');
+    this.getAvatarImage();
   }
 
   @Watch('email')
   validateEmail() {
     if (this.email) checkEmptyOrPattern(this, 'email', emailPattern);
+    this.getAvatarImage();
   }
```
oliverschuerch: Leaving this as is, and calling the `validateUserId` and `validateEmail` function on `componentWillLoad` causes the `getAvatarImage` function to be called three times on startup.

1. In the `connectedCallback` (which is the only one that should be called initially)
2. In `validateUserId` triggered by `componentWillLoad`
3. In `validateEmail` triggered by `componentWillLoad`

Please make it possible to call the validation functions without triggering the `getAvatarImage`.

For example:
```ts
// like so, `updateUserId()` is only called when 'userid' gets an update
@Watch('userid')
updateUserid() {
  this.validateUserid();
  this.getAvatarImage();
}

// while the `validateUserid()` function can be called initially and when the `userid` chages
private validateUserid() {
  checkEmtpyOrType(this, 'userid', 'string');
}

connectedCallback() {
  // this should be the only inital call to `getAvatarImage()`
  this.getAvatarImage();
}

componentDidLoad() {
  // initially only call the validation
  this.validateUserid();
}
```

packages/components/cypress/e2e/avatar.cy.ts
```diff
@@ -66,11 +86,25 @@ describe('Avatar', () => {
...
+      cy.get('@avatar').invoke(
+        'append',
+        '<img src="/assets/images/logo-swisspost.svg" alt="Swiss Post Logo" />',
+      );
+      cy.get('@avatar').find('img').invoke('remove');
       cy.get('@avatar').find('img').should('not.exist');
       cy.get('@avatar').find('.initials').should('exist');
     });
+
+    it('should not show image bur fallback to initials, when slotted image is invalid', () => {
```
oliverschuerch:
```suggestion
    it('should not show image but fallback to initials, when slotted image is invalid', () => {
```

packages/components/cypress/e2e/avatar.cy.ts
```diff
@@ -66,11 +86,25 @@ describe('Avatar', () => {
...
+      cy.get('@avatar').invoke(
+        'append',
+        '<img src="/assets/images/invalid-image.svg" alt="Invalid image" />',
+      );
+      cy.get('@avatar').find('.initials').should('exist');
```
oliverschuerch, line 93: You should also check if the slotted image is not visible

packages/components/cypress/e2e/avatar.cy.ts
oliverschuerch: What about that topic? Is something not clear?
I can't see any change, nor any message, why we should not do this.

oliverschuerch (changes requested):

packages/components/src/components/post-avatar/post-avatar.tsx
myrtasakellariou: correct

packages/components/cypress/e2e/avatar.cy.ts
myrtasakellariou: Sorry, i missed that point.

packages/components/cypress/e2e/avatar.cy.ts
oliverschuerch: What I meant is:
- If the slotted image is invalid
- and you show the fallback (which is nice)
- check as well, if the image is really not visible
  e.g. `cy.get('@avatar').find('img').should('be.not.visible');`

The test below is ok and can stay.

packages/components/src/components/post-avatar/post-avatar.tsx
oliverschuerch: I think you probably missed this one, right?
Otherwise close it and keep it as it is.

packages/components/src/components/post-avatar/avatar-utils.ts
oliverschuerch, line 1: I like ❤️

packages/components/src/components/post-avatar/avatar-utils.ts
```diff
@@ -0,0 +1,20 @@
...
+const GRAVATAR_DEFAULT = '404';
+const GRAVATAR_RATING = 'g';
+const GRAVATAR_SIZE = 80;
+
+export function getGravatarUrl(email: string): string {
+  const hash = cryptify(email.trim().toLowerCase());
+  return `https://www.gravatar.com/avatar/${hash}?s=${GRAVATAR_SIZE}&d=${GRAVATAR_DEFAULT}&r=${GRAVATAR_RATING}`;
+}
+
+export async function cryptify(key: string) {
```
oliverschuerch, line 12: If this function is not used outside this file, the `export` statement can be removed.

oliverschuerch (changes requested):

packages/components/src/components/post-avatar/avatar-utils.ts
myrtasakellariou: it is used also in the post-avatar.tsx

packages/components/src/components/post-avatar/post-avatar.tsx
myrtasakellariou: When i add it like this -> `if (!imageLoaded && this.email && this.email?.match(emailPattern))`  i get msg from sonarqube (_Prefer using an optional chain expression instead, as it's more concise and easier to read._) Then if i do this-> ` if (!imageLoaded && this.email?.match(emailPattern)) ` i get other sonarqube msg (_Use the "RegExp.exec()" method instead._)

packages/components/src/components/post-avatar/avatar-utils.ts
oliverschuerch: Sorry, didn't see that.

oliverschuerch (approved):

PR #5907, opened 2025-07-28
feat(internet-header): allow to use p tags instead of heading tags

gfellerph:
Added the possibility to use paragraphs instead of headings for the post-header component. This enables project to fully control the heading structure of the document.


packages/components/src/index.html
```diff
@@ -97,9 +97,9 @@ <h2>Main Navigation</h2>
                 Back
               </button>
               <post-closebutton slot="close-button">Schliessen</post-closebutton>
-              <h2 slot="megadropdown-title">Briefe title</h2>
+              <p slot="megadropdown-title" class="h2">Briefe title</p>
               <post-list>
-                <h3>Briefe senden</h3>
+                <p class="h3">Briefe senden</p>
```
leagrdv, line 112: I feel it's a bit dangerous to let people put whichever class they want here. In our CSS for instance, we set `.megadropdown h2, .megadropdown > h2` to `display: none` and letting users choose which title size they want might result in a hidden title. Also, it opens the possibility for users to derive from the intended design.

What do you think? Should we skip those classes and target the "fake" headings another way in CSS?

packages/documentation/src/stories/components/header/header.stories.ts
```diff
@@ -42,6 +44,17 @@ const meta: MetaComponent = {
...
+    useHeadings: {
+      name: 'Use headings',
+      description: 'Whether or not to use headings for the title and megadropdown titles.',
+      control: {
+        type: 'boolean',
+      },
+      table: {
+        category: 'Content',
+      },
+      defaultValue: { summary: true },
```
leagrdv: This value is not shown in the arguments table, also is there really a default value here?

gfellerph:
Closing as probably not the right approach. Ticket re-triaged.

PR #5933, opened 2025-08-04
feat(components): add one-pager header variant

alizedebray:
## 📄 Description

This PR updates the header to support the one-pager variant.

## 🚀 Demo

https://preview-5933--swisspost-design-system-next.netlify.app/?path=/docs/27a2e64d-55ba-492d-ab79-5f7c5e818498--docs&story=Default&args=mainNavigation:false#27a2e64d-55ba-492d-ab79-5f7c5e818498--default

---

## 📝 Checklist


packages/components/src/components/post-header/post-header.scss
```diff
@@ -16,21 +16,25 @@
...
+    var(--post-global-header-reduced-height) - var(--post-global-header-expanded-height)
   );
   --post-local-header-top: calc(
-    -1 * (var(--post-local-header-height) - var(--post-main-navigation-height)) + var(--post-global-header-reduced-height)
+    var(--post-global-header-reduced-height) + var(--post-local-header-reduced-height) - var(--post-local-header-expanded-height)
+  );
+  --post-global-controls-top: clamp(
+    1000 * var(--post-global-header-top), /* 0 when header is static, large negative when header shrinks to enable offset */
+    calc(-1 * min(var(--post-header-scroll-top, 0px) / 4, var(--post-global-header-reduced-height))),
+    0px
```
alizedebray: @leagrdv can you please check if this is still working as intended?

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -345,6 +356,10 @@ export class PostHeader {
   }
 
   render() {
+    const localHeaderClasses = ['local-header'];
+    if (this.mobileMenuExtended) localHeaderClasses.push('local-header-mobile-extended');
+    if (!this.hasNavigation) localHeaderClasses.push('no-navigation');
```
leagrdv, line 367: If I manually hide the main navigation in the args table, it doesn't add the `no-navigation` class to the local header.

packages/components/src/index.html
```diff
@@ -1,187 +1,52 @@
...
-    <script nomodule src="/build/post-components.js"></script>
-  </head>
-
-  <body style="min-height: 200vh">
-    <post-header>
-      <!-- Logo -->
-      <post-logo slot="post-logo">Homepage</post-logo>
-
-      <!-- Meta navigation -->
-      <ul class="list-inline" slot="meta-navigation">
```
leagrdv: Should we maybe keep the meta/menu button/main navigation/custom controls in here? I feel like it's the most common use case so it would make sense to have the "full" header here.

packages/components/src/components/post-header/post-header.scss
leagrdv: For desktop, it's all good.

I think there is an inconsiscenty in mobile though. I see you've added the animation to hide the local header on scroll, which corresponds to the design in figma but:
- The animation is both for the full header and the one-pager header, which makes sense for the former but in the one-pager, since the local header is never hidden in desktop, I feel like it does not make sense to hide it in mobile. What do you think?
- For the full header, on mobile if you've scrolled and the local header is hidden, then you open the menu, you still don't see the local header here but there is a gap where you can see the page's content. I think when the mobile menu is expanded, it'd make sense to see the local header, no matter if we've scrolled or not.

Capture d'écran 2025-08-06 162728

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -30,6 +30,11 @@ export class PostHeader {
   private readonly throttledResize = throttle(50, () => this.updateLocalHeaderHeight());
   private scrollParentResizeObserver: ResizeObserver;
   private localHeaderResizeObserver: ResizeObserver;
+
+  private get isLanguageSwitchInMainHeader(): boolean {
```
leagrdv: The check should also be done for the meta navigation, otherwise if they add a meta navigation without the main navigation, the buttons disappear on mobile.
Either that or we keep the Menu button if the meta navigation or the main navigation is present.

packages/components/src/index.html
alizedebray: Right, I forgot to revert my changes to the file

packages/components/src/components/post-header/post-header.scss
alizedebray: Good catch! This was an involuntary change but I adapted it to match the design.

packages/components/src/components/post-header/post-header.tsx
alizedebray: For now I left the meta navigation in the global header on mobile but I add this point to the pattern discussion.

alizedebray:
> It would be nice to have some tests for some of the use cases (meta/no meta, main nav/no main nav) to check that the menu button is displayed when intended and that the links are displayed on mobile and desktop

I created a separate ticket to not block the release of the one-pager header since it has be requested for a  while now: #5954

packages/components/src/components/post-header/post-header.scss
leagrdv: When scrolled, the local header is back again when the menu is opened but the space for it has been removed, resulting in hidden navigation elements.

Capture d'écran 2025-08-07 132624

packages/components/src/components/post-header/post-header.tsx
leagrdv: The `no-navigation` is still not added when the navigation is removed manually in the storybook args table, resulting in a weird behavior on mobile.

packages/components/src/components/post-header/post-header.scss
alizedebray: I couldn't reproduce this issue and if I am not mistaken you shouldn't be able to scroll with an open menu: https://github.com/swisspost/design-system/blob/main/packages/components/src/components/post-header/post-header.tsx#L69

Can you tell me on which browser and platform you saw this behavior?

alizedebray: This was due to the `top` property override on focus (the menu button gets the focus when you click on it). I restricted this behavior to only happen on deskop: https://github.com/swisspost/design-system/pull/5933/files#diff-4357be46c16d336c7da3edaf20f66477d06d37481955a4641da3998d75fcfd2cR32

alizedebray:
Inputs from @sandra-post:

- Global header
  - Div which contains "Jobs" etc. - Should it really contain "Jobs" and "About us"? Bc thats something Post-Portal specific and I at least do not have it in the design for the onepager. 
  - Padding right is 12, should be 8 (needs to align with content in Local header)
- Local header
  - Change placeholder title "Application title" to "OnePager title" (just for the sake of minimizing confusion)
  - Padding right is 8, should be 16 (needs to align with content in Global header)
  - Spacing between "Search" and "Login" is 16, should be 4

sandra-post:
Reviewed adjusted version, is looks good now.

leagrdv (approved):
I'm approving the PR cause I don't want to block it for things that are not directly related to the one-page header itself, but I still feel like they should be addressed at some point, either on that PR if you have some time, or to have those in another ticket for a later time.

- There are still some issues in the storybook documentation when switching between the options (main nav/no main nav, meta/no meta, etc.) when we resize the header:
Capture d'écran 2025-08-13 120333
Capture d'écran 2025-08-13 120412

- On mobile -> scroll down to hide the local header -> open the menu -> we don't have the local header. I feel like it should probably be displayed when the mobile menu is opened, as user might not understand why they need to scroll up the page to see the search and login buttons. It's not related to the one-pager itself but I don't know if it might be considered a regression? Since the hide on scroll animation was added on this PR.

PR #5951, opened 2025-08-06
chore(styles, docs): removed `.loader-xs` and `.loader-sm` classes from spinner stylesheet

bucknatt:
…spinner component

## 📄 Description

Removed loader-xs and loader-sm from styles for spinner component. Consequently removed information about deprecated loader-xs and loader-sm in the documentation.

### DIFF
--- a/packages/components/src/components/post-popovercontainer/post-popovercontainer.scss
+++ b/packages/components/src/components/post-popovercontainer/post-popovercontainer.scss
@@ -33,8 +33,8 @@
     box-shadow: elevation.$elevation-300;
     width: max-content;
     max-width: 100%;
-    color: var(--post-current-palette-fg);
-    background-color: var(--post-current-palette-bg);
+    color: var(--post-current-fg);
+    background-color: var(--post-current-bg);
     border-radius: commons.$border-radius;
 
     &.animate-pop-in {
--- a/packages/documentation/.storybook/styles/preview.scss
+++ b/packages/documentation/.storybook/styles/preview.scss
@@ -368,7 +368,7 @@ $monospace: SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier
 
   &::after {
     content: '';
-    background: var(--post-current-palette-bg);
+    background: var(--post-current-bg);
     width: 33%;
     height: 1.7rem;
     display: block;
--- a/packages/styles/src/components/form-textarea.scss
+++ b/packages/styles/src/components/form-textarea.scss
@@ -156,7 +156,7 @@ textarea.form-control {
 
     &:disabled ~ label {
       color: tokens.get('textarea-disabled-fg');
-      background-color: var(--post-current-palette-bg);
+      background-color: var(--post-current-bg);
     }
 
     &:not(:disabled):hover ~ label {


packages/styles/src/components/spinner.scss
```diff
@@ -38,15 +38,6 @@
     }
   }
 
-  // Deprecated sizes
-  &.loader-sm {
-    font-size: spinners.$spinner-size-sm;
```
leagrdv, line 43: These CSS classes were using the SCSS variables `$spinner-size-sm` and `$spinner-size-xs`. Check if these SCSS variables are used elsewhere. If not, you can also delete them.
Also, where those SCSS variables are located, there are two correspondig SCSS variables for the XS/SM border width. Check if they're still used, otherwise you can also delete them.

Don't forget to update the changeset to document the removal of the SCSS variables as well.

packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -51,6 +51,23 @@ export class MigrationV99Component extends LitElement {
             <section>
               <h4>Styles</h4>
 
+              <h5>Spinner</h5>
```
leagrdv: The migration list is kind of set in order of importance (sort of), and the removal of those classes doesn't make much sense at the top of it. If you go further in the guide, you'll see a "Other styles" section, it would make more sense to document it there. 

Also don't forget to close your HTML tags. Here, you've opened a `<ul>` which is never closed.

.changeset/pink-schools-double.md
```diff
@@ -0,0 +1,7 @@
+---
+'@swisspost/design-system-styles': major
+'@swisspost/design-system-documentation': minor
+---
+
+Removed `.loader-xs` and `.loader-sm` classes from spinner stylesheet and documentation
+Removed deprecated `$spinner-size-xs` and `$spinner-size-sm` variables and related border-with from spinner stylesheet.
```
leagrdv: It's easier for the user to read if they're in a list format, and to have them all written down.
```suggestion
Removed deprecated classes and SCSS variables for the spinner:
- `.loader-xs`
- `.loader-sm`
- `$spinner-size-xs`
- `$spinner-size-sm`
- `$spinner-border-width-xs`
- `$spinner-border-width-sm`
```

packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -758,6 +759,23 @@ export class MigrationV99Component extends LitElement {
                   </p>
                 </li>
               </ul>
+              <ul>
```
leagrdv: No need to open a new `<ul>` tag here, you can just append your `<li>` content to the "Other styles" one.

packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -758,6 +759,23 @@ export class MigrationV99Component extends LitElement {
...
+                <li class="mb-16">
+                  <p>
+                    Depredated loader classes and related scss variables have been removed
```
leagrdv: *Deprecated

packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -758,6 +759,23 @@ export class MigrationV99Component extends LitElement {
...
+                    <span class="tag tag-sm tag-danger">breaking</span>
+                  </p>
+                  <ul>
+                    <li><code>.loader-xs</code></li>
+                    <li><code>.loader-sm</code></li>
+                    <li><code>$spinner-size-xs</code></li>
+                    <li><code>$spinner-size-sm</code></li>
```
leagrdv: You can remove those two lines. It's a bit complicated but generally what we do is we document all the deleted things in the changeset, but on the migration guide to avoid having something too long (which is already the case), we only document the removal/changes of things that had been documented in the documentation before. So in the case here, on the `Spinner documentation`, we documented the `.loader-xs` and `.loader-sm` classes, so we need to inform users that we deleted them, but the SCSS variables were never documented so we can leave them out of the migration guide.

leagrdv (changes requested):
Some comments have not been addressed.

leagrdv (approved):

PR #5966, opened 2025-08-08
chore(styles, docs): remove elements from card component

leagrdv:
## 📄 Description

Following the removal of bootstrap, this PR removes the following classes as most of the CSS was coming from bootstrap and we just add a few overrides on them:
- `.card-header`
- `.card-footer`
- `.card-img`
- `.card-img-top`
- `.card-img-bottom`

The image classes got removed as our CSS overrides were not relevant anymore, and an image is correctly displayed without the need for such classes.


packages/components/cypress/fixtures/post-linkarea.test.html
```diff
@@ -9,10 +9,10 @@
   <body>
     <post-linkarea id="without-data-link">
       <div class="card">
-        <div class="card-body">
-          <h5 class="card-title">Titulum</h5>
+        <div class="p-16">
```
alizedebray: It would be good to add a padding directly to the card styles and remove the `p-16` div's.
In the current design the padding is 24px.

alizedebray (changes requested):
In the ticket it is mentionned to add auto-migrations

packages/components/cypress/fixtures/post-linkarea.test.html
leagrdv: The problem with adding a padding in the card itself is that if users add an image within, it will not be flushed with the card, and they'll have to manually add negative margin to it to revert that (or we target it on our side `.card > img` and do it ourselves?).
But if we say that cards with images should be using the `teaser card` component, then we could add the 24px padding to the `.card` class and remove it on the `.teaser-card` one. What do you think? If we go with that solution, I'd remove the images arguments on the card stories though.

alizedebray: From what I see in Figma, I'd go with you second proposal: card + image = teaser card

packages/documentation/src/stories/components/linkarea/linkarea.stories.ts
```diff
@@ -64,10 +64,8 @@ function renderLinkarea(args: Args) {
     <post-linkarea>
       <div class="card">
```
alizedebray, line 65: Can we set the `card` class on the `post-linkarea` to save on markup? (just an idea, it can be done is another ticket)
```suggestion
<post-linkarea class="card">
```

leagrdv: Good idea, I've created the ticket #6110 to address it.

PR #5984, opened 2025-08-13
feat(docs): add package type information to the components pages

bucknatt:
## 📄 Description

The information about package type following the installation guide of the component is added to the documentation.
The package type information is added as a tag component to the component pages.
The tag variant gives the information, if the component is of package
- HTML *(Default variant)*
- WebComponents *(Yellow)*.


.changeset/hungry-pumas-repair.md
alizedebray, line 1: There are two change set files, you can delete one.

.changeset/unlucky-peaches-join.md
```diff
@@ -0,0 +1,5 @@
+---
+"@swisspost/design-system-documentation": patch
+---
+
+ Add package type information to the components pages #5984
```
alizedebray: We usually start change set messages with a past tense verb and end them with a period.
By the way, what does “#5984” mean?

packages/documentation/src/shared/package-tag.mdx
```diff
@@ -0,0 +1,3 @@
+<div className={'tag tag-sm mb-24 tag-' + (props.meta.tags[0] == 'package:HTML' ? "" : "yellow")}>
```
alizedebray: Here, if the tag is "package:HTML" you will end up with an unwanted `tag-` class. You can see it here by inspecting the package tag here: https://preview-5984--swisspost-design-system-next.netlify.app/?path=/docs/bec68e8b-445e-4760-8bd7-1b9970206d8d--docs

alizedebray: Also, because the "package:HTML" is not always on position 0 you can use: 
```suggestion
<div className={'tag tag-sm mb-24 tag-' + (props.meta.tags.includes('package:HTML') ? "" : "yellow")}>
```

Can you change this everywhere?

alizedebray (changes requested):

.changeset/hungry-pumas-repair.md
```diff
@@ -0,0 +1,4 @@
+---
+'@swisspost/design-system-documentation': minor
+---
+feat(docs): added package-info to compo on docs.
```
alizedebray: Only the **commit messages** start with `feat`, `fix`, or `chore`: these are short messages meant for Design System developers to quickly understand what a commit is about. 

Here you are writing a **change set** meant for external users to understand our changes. These users do not know our code base and need complete sentences with no abbreviations or technical jargon.

packages/documentation/src/shared/package-tag.mdx
alizedebray, line 1: If you decide to go with that solution you must use a `.tsx` extension and give a type to the `props` argument.

alizedebray (changes requested):

packages/documentation/src/shared/package-tag.tsx
```diff
@@ -0,0 +1,22 @@
+type PackageTagProps = {
+  meta: {
+    tags: string[];
+  };
+};
+
+export default function PackageTag({ meta }: PackageTagProps) {
+  let packageName = '';
+  const tagColor = '';
```
alizedebray: The tag color is never updated.

packages/documentation/src/shared/package-tag.tsx
```diff
@@ -0,0 +1,22 @@
+type PackageTagProps = {
+  meta: {
+    tags: string[];
+  };
+};
+
+export default function PackageTag({ meta }: PackageTagProps) {
```
alizedebray, line 7: The Meta type already exists, please reuse the type used in the `*.stories.ts` files.

.changeset/hungry-pumas-repair.md
```diff
@@ -0,0 +1,10 @@
+---
+'@swisspost/design-system-documentation': minor
+---
+
+Added the information about package-type following the installation guide of the component,
+for all components on the documentation.
+The package type information is added as a tag component to the component pages.
+The tag variant gives the information, if the component is of package:
+- HTML (Default variant)
+- WebComponents (Yellow)
```
alizedebray:
```suggestion
Enhanced component documentation to include a badge at the top of each page showing the package where the component can be found, clarifying which dependency needs to be installed to make it available in a project
```

alizedebray (changes requested):
Can you also add the badge to the utility documentation pages?

packages/documentation/src/shared/package-tag.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+  if (meta.tags.includes('package:Styles')) {
+    packageName = '@swisspost/design-system-styles';
+  } else if (meta.tags.includes('package:WebComponents')) {
+    packageName = '@swisspost/design-system-components';
+    tagColor = 'yellow';
+  }
+
+  return (
+    <div className={`tag tag-sm mb-24 tag-${tagColor}`}>
```
alizedebray: With the styles package `tagColor` is an empty string, therefore here you are adding an unnecessary `tag-` class.

alizedebray (changes requested):

PR #6002, opened 2025-08-15
feat(docs, styles): improve position utilities.

bucknatt:
## 📄 Description

The _position_ utility classes allow defining the inset of an element. The classes available are
- top-[position]
- bottom-[position]
- start-[position]
- end-[position]

where [position] can be 0, 50, or 100.

The goal was to improve these _position_ utilities and their documentation by:

- `inline-*` and `block-*` properties: Replaced `top`, `bottom`, `right` and `left` with
`inset-block-start`, `inset-block-end`, `inset-inline-start` and `inset-inline-end`.
- Added a control for the `bottom-*` and `end-* classes` in the _position_ utility documentation.

Consequently classes for `bottom` and `end` were added to the SCSS variables on styles.


packages/documentation/src/stories/utilities/position/position.stories.ts
```diff
@@ -31,59 +33,43 @@ const meta: MetaExtended = {
       control: {
         type: 'select',
       },
-      options: ['', '0', '50', '100'],
+      options: ['undefined', '0', '50', '100'],
```
alionazherdetska: I suggest renaming `undefined` to `unset` to avoid confusion with JavaScript's `undefined type` and better indicate that no positioning class will be applied.

Can you please replace all instances of 'undefined' with 'unset'.

packages/documentation/src/stories/utilities/position/position.stories.ts
```diff
@@ -9,8 +9,10 @@ const meta: MetaExtended = {
   title: 'Utilities/Position',
   args: {
     position: 'absolute',
-    start: '50',
     top: '0',
+    bottom: '',
+    start: '50',
+    end: '',
```
alionazherdetska:
```suggestion
    bottom: 'unset',
    start: '50',
    end: 'unset',
```

alionazherdetska: There's an issue with the default args that's causing invalid CSS class generation. The current default args use empty strings for bottom and end:
image
But the render logic checks against the string 'undefined' (or now 'unset'). Since empty string '' is not equal to 'undefined', the condition evaluates to true and generates invalid CSS classes like bottom- and end- in the HTML output.
image
The default args should be updated to use 'unset'.

.changeset/busy-lines-pick.md
```diff
@@ -0,0 +1,9 @@
+---
+'@swisspost/design-system-documentation': minor
+'@swisspost/design-system-styles': minor
+---
+
+Replaced `top`, `bottom`, `right` and `left` with
+`inset-block-start`, `inset-block-end`, `inset-inline-start` and `inset-inline-end`.
+Added controls for `bottom` and for `end` property for position compo in utilities.
+Added classes for `bottom` and for `end` property for position compo in styles.
```
alionazherdetska:
```suggestion
Added missing `bottom` and `end` controls to position utilities for complete configuration. Replaced empty label option with `unset` for better clarity across all position controls (`top`, `bottom`, `start`, `end`).
```

alionazherdetska (changes requested):

.changeset/busy-lines-pick.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': minor
+'@swisspost/design-system-styles': minor
+---
+
+Added missing `bottom` and `end` controls to position utilities for complete configuration. Replaced empty label option with `unset` for better clarity across all position controls (`top`, `bottom`, `start`, `end`).
```
alionazherdetska:
```suggestion
Added missing `bottom` and `end` controls to position utilities for complete configuration. Replaced empty label option with `unset` for better clarity across all position controls (`top`, `bottom`, `start`, `end`). Position utilities now use logical CSS inset properties (inset-block-start, inset-block-end, inset-inline-start, inset-inline-end). This means `start` and `end` follow the writing direction — left in LTR and right in RTL.
```
```

**Review by alionazherdetska — APPROVED** [review]

File: packages/documentation/src/stories/utilities/position/position.stories.ts [hunk 5]
```diff
@@ -31,59 +33,43 @@ const meta: MetaExtended = {
...
       description: 'Sets the distance from bottom. ',
       control: {
         type: 'select',
       },
-      options: ['', '0', '50', '100'],
+      options: ['unset', '0', '50', '100'],
       table: {
         category: 'General',
       },
-      if: {
```
leagrdv, line 53: I'm not sure why this has been changed. I think the display of the top/bottom/start/end arguments should still depend on their counterpart (e.g.: hide the end option if the start is not set to "unset"), otherwise you can have opposite classes together `start-100 end-100`, which should not be the case. Either that, or programmatically only keep the last selected one and put the opposite one to unset.

bucknatt: `top`, `bottom`, `start`, `end` work for position values that are not static (i.e. relative, absolute, fixed, sticky)

e.g. `<div class="my-element position-relative start-50 end-100 top-100 bottom-0 translate-middle"></div>`

The result is:
image

So you can have them both. That is the reason why the if-statement has been removed.
⚠️The values must logically match each other, otherwise the layout may disappear or overlap.
I can see your point. 👉It often makes more sense to only set `top` + `start` (with fixed sizes), or `top` + `bottom` (for height), or `start` + `end` (for width).
As I understood, it is possible to set `top`, `bottom`, `start`, `end` at the same time.

leagrdv (approved):

PR #6022, opened 2025-08-18
fix(tokens): allow references in all multi-value tokens

oliverschuerch:
## 📄 Description

- The tokens build script did not handle multi-value tokens (except from type `typography` tokens) well, which led to CI errors sometimes. I've copied over the files we have on `main` and just updated the `constants.js` file, to match the current setup.
- Since we store the `versions.json` file only on the latest release branch (release/v9) and on main, we must make sure not to use it locally, when we release a next major version. Therefore, the dependencies table component on the styles package page had to be updated.

---

## 📝 Checklist

### DIFF
--- a/packages/documentation/src/stories/getting-started/packages/styles/dependenciesTable.tsx
+++ b/packages/documentation/src/stories/getting-started/packages/styles/dependenciesTable.tsx
@@ -1,5 +1,9 @@
-import React from 'react';
-import versionsData from '../../../../../public/assets/versions.json';
+import React, { useEffect, useState } from 'react';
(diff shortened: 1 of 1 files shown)


pnpm-lock.yaml
gfellerph, line 1: Changes here are likely accidental. Should we enforce using the exact same version of pnpm?

oliverschuerch: I'm using corepack, so I use the exact version specified in the root package.json, specified in the packageManager field.
How do you like to proceed?

Issue #5992, opened 2025-08-14
[bug]: Element with role menu is empty


gfellerph:
### 🐞 Describe the bug

When the menu is closed, there's an empty container with `role=menu` in the DOM. We had the same issue with tooltips. The solution there was to move the role definition to the `<post-popovercontainer>` which is hidden initially unlike the wrapping `<post-menu>` component.

### 🔗 Additional context

- https://jira.post.ch/browse/POW-177

> h2. 📝 Description: The element has the ARIA role menu but does not contain any menuitem elements. According to [WAI-ARIA 1.1: 5.2.5 Required Owned Elements|https://www.w3.org/TR/wai-aria/#aria-roles], certain container roles\- such as menu - require specific child roles, like menuitem, to meet accessibility standards. h2. 🔍 Obervation Current markup:   This element is *empty* and lacks any menuitem elements. h2. 🧠 *Impact* Assistive technologies (e.g., screen readers) expect a menu to contain menuitem elements. Without them, the component is not announced or navigated correctly, creating a significant accessibility issue for users relying on keyboard navigation or screen readers. h2. 📘 *References* * Siteimprove issue: [Container element is empty|https://my2.siteimprove.com/Accessibility/1223337/NextGen/Issue/1?ruleId=68&pageSegments=&issueKind=1&tagFilterType=2&conformance=] * WAI-ARIA 1.1: [5.2.5 Required Owned Elements|https://www.w3.org/TR/wai-aria/#aria-roles] * Siteimprove Rule: [SIA-R68 Technical Documentation|https://dequeuniversity.com/rules/axe/4.8/aria-required-children]\\ h2. ✅ *Acceptance Criteria* * The post-menu contains a structured list with role="menuitem", e.g.:
> 
> English
> Deutsch
> * List items are focusable and keyboard-navigable * Accessibility tools (Siteimprove, WAVE, axe) no longer report this issue * Manual screen reader testing confirms correct focus, structure, and announcement\\

myrtasakellariou:
Closing because the issue is addressed by #5837

PR #6042, opened 2025-08-19
chore(components): fix `post-menu` related accessibility issues

myrta2302:
## 📄 Description

This PR updates:

1. `<post-menu>` and `<post-menu-item>`:

- `menu` and `menuitem` roles are now assigned only after the popovercontainer is displayed.
- The menu role is applied to the <post-menu> host, and the menuitem role is applied to the focusable elements inside each `<post-menu-item>`.
- Applying the `menuitem` role in the focusable children corrects focus behavior during keyboard navigation with NVDA.
- Deferring role assignment until the menu content is visible resolves the “empty container” error (see #5992).
- A `label` prop is introduced for the accessible name of the menu. It is optional in the scope of this PR. To become required in this PR #6343 

2.  `<post-language-switch>` and `<post-language-option>`:

- Removes the obsolete `role="menuitem"` previously set on `<post-language-switch>`, since focus is now working correctly.
- `<post-language-option>` now assigns `role="listitem"` to the host of the `<post-language-option>` items with `variant="list"`, addressing the siteimprove accessibility check's “empty container” error.

3.  `<post-breadcrubs>`:

- Removes the "More Breadcrumbs" `aria-label`, which was causing the following error:

image

- Added `role="none"` to the trigger wrapper `<div>` to reflect its presentational purpose.

## 📝 Checklist


packages/components/src/components/post-menu/post-menu.tsx
```diff
@@ -144,7 +149,15 @@ export class PostMenu {
         if (this.isVisible) {
           this.lastFocusedElement = this.root?.activeElement as HTMLElement;
           const menuItems = this.getSlottedItems();
+          this.focusableChildren = menuItems;
```
alizedebray: Why is this in the `handlePostToggle`? This roles is static, we probably don't want it to be reset anytime the menu is opened or closed.

packages/components/src/components/post-menu/post-menu.tsx
```diff
@@ -144,7 +149,15 @@ export class PostMenu {
...
+
           if (menuItems.length > 0) {
+            // Add role="menu" to the popovercontainer
+            this.host.setAttribute('role', 'menu');
+            // Add role="menuitem" to the focusable elements
+            menuItems.forEach(item => {
+              item.setAttribute('role', 'menuitem');
```
alizedebray: Can this attribute be set in a lifecycle hook of the `post-menu-item` for a good separation of concerns?

packages/components/src/components/post-language-switch/post-language-switch.tsx
```diff
@@ -136,12 +135,7 @@ export class PostLanguageSwitch {
             <post-icon aria-hidden="true" name="chevrondown"></post-icon>
           </button>
         </post-menu-trigger>
-        <post-menu
-          id={this.menuId}
-          class="post-language-switch-dropdown-container"
-          aria-label={this.caption}
```
alizedebray, line 142: An `aria-label` is required on elements with a `role="menu"`: https://www.w3.org/WAI/ARIA/apg/patterns/menubar/

If you remove it from you must add it as a property for the `post-menu`.

myrtasakellariou: In the end i did not manage to reproduce an announcement of the aria-label text, so i removed it.

alizedebray: Is it an issue to have it? Otherwise it is still preferable to follow the guidelines.

myrtasakellariou: It is possible, but the label must be added together with the role, otherwise, it causes an error. Therefore, the label should be included inside the `@EventFrom('post-popovercontainer')` in `post-menu.tsx`. So this would require adding a new `label` prop to `post-menu.tsx` and pass it form the language-switch.

packages/components/src/components/post-menu/post-menu.tsx
```diff
@@ -57,6 +57,16 @@ export class PostMenu {
...
 
+  /**
+   * Holds the current focusable children
+   */
+  @State() focusableChildren: Element[];
+
+  /**
+   * A flag to know when popovercontainer has been rendered for the first time.
+   */
+  @State() popovercontainerRendered: boolean = false;
```
alizedebray: This is not a state: changes to a state will lead to a re-render of the component and that is not what you want here.

packages/components/src/components/post-menu/post-menu.tsx
```diff
@@ -57,6 +57,16 @@ export class PostMenu {
    */
   @State() isVisible: boolean = false;
 
+  /**
+   * Holds the current focusable children
+   */
+  @State() focusableChildren: Element[];
```
alizedebray: This state seems not to be used anywhere. Can you remove it?

packages/components/src/components/post-menu/post-menu.tsx
```diff
@@ -54,6 +54,11 @@ export class PostMenu {
     checkEmptyOrOneOf(this, 'placement', PLACEMENT_TYPES);
   }
 
+  /**
+   * Required label providing an accessible name for the menu.
```
alizedebray:
```suggestion
   * An accessible name for the menu.
```

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -74,11 +74,12 @@ export class PostPopovercontainer {
   private eventTarget: Element;
   private clearAutoUpdate: () => void;
   private toggleTimeoutId: number;
+  private firstOpen: boolean = true;
 
   /**
-   * Fires whenever the popovercontainer gets shown or hidden, passing the new state in event.details as a boolean
+   * Fires whenever the popovercontainer gets shown or hidden, passing in event.detail an object containing two booleans: `isOpen`, which is true if the popovercontainer was opened and false if it was * closed, and `first`, which is true if it was opened for the first time.
```
alizedebray:
```suggestion
   * Fires whenever the popovercontainer gets shown or hidden, passing in event.detail an object containing two booleans: `isOpen`, which is true if the popovercontainer was opened and false if it was closed, and `first`, which is true if it was opened for the first time.
```

PR #6050, opened 2025-08-19
chore(docs): show docs in dev mode in the deploy preview

bucknatt:
## 📄 Description

This PR adds functionality to read a specific URL parameter `devModeEnabled` and use it to control the development mode in the Storybook manager (`manager.ts`).

- The existing `initialEnv` variable is still initialized from `process.env.NODE_ENV`. _(already implemented)_
    
- New: If the URL contains the parameter `?devModeEnabled=true`, it overrides `initialEnv` and sets the mode to `development`.
    
- If the URL parameter is not present, the mode falls back to the existing logic using the value from `localStorage` (`storedDevMode`). _(already implemented)_
    
- The CI workflow (`deploy-documentation.yaml`) has been updated to append `?devModeEnabled=true` to the generated preview URL, ensuring the preview deploy runs in development mode.
    
- Make sure to test locally to verify the mode toggling from `production` to `development` works correctly.


packages/documentation/.storybook/manager.ts
oliverschuerch, line 1: Setting the `devModeEnabled` to `true` in the URL works as expected.
However, setting it to `false` does not work at all.
Since we store the latest state in a sessionStorage item, it is absolutely impossible to switch from `development` back to `production` afterwards.

Also some of your comments seem to be placed wrongly.

I would suggest to turn around the logic a little bit:

```ts
// Get existing URL params
const params = new URLSearchParams(window.location.search);
// Get value, that states if URL devMode is enabled (String 'true' or 'false', or null if devModeEnabled is not specified in URL)
const devModeFromUrl = params.get('devModeEnabled');

// Get value, that states if localStorage devMode is enabled (String 'true' or 'false', or null if the item is not specified in localStorage)
const storedDevMode = localStorage.getItem('devModeEnabled');

// Get initial mode (e.g. `development` or `production` from environment variable
let initialEnv = process.env.NODE_ENV || 'production';

// Override initialEnv by priority (url param is more important than the stored state, because this represents the users explicit wish)
if (devModeFromUrl !== null) {
  initialEnv = JSON.parse(devModeFromURL) ? 'development' : 'production';
} else if (storedDevMode !== null) {
  initialEnv = JSON.parse(storedDevMode) ? 'development' : 'production';
}
```
Nice to know: Updating the localStorage item is not necessary, because this is done everytime the `data-env` attribute on the `html` element is updated.
See MutationObserver at the end of the file.

Btw. I don't know if the code would work like that, try it out and change whatever you want ;)

.changeset/brown-chicken-ring.md
oliverschuerch, line 1: I would not add a changeset here, since this is internal logic only for us.

Otherwise, something more general like the following would be enough:
`Added the possibility to override the used environment mode (`development` or `production`) through the URL, by setting the `devModeEnabled` parameter to `true` or `false`.

packages/documentation/.storybook/manager.ts
```diff
@@ -14,12 +14,40 @@ const TECH_ICONS: Record<string, string> = {
...
+  // Default to 'production' if no URL param and no stored value
+  initialEnv = 'production';
+}
+
+// In case of no URL parameter, reset localStorage value (fallback)
+if (devModeFromUrl === null) {
+  // We clear only localStorage if it's set to 'development' (to reset it when switching back to production)
+  if (storedDevMode === 'true') {
+    localStorage.removeItem('devModeEnabled');
+  }
```
oliverschuerch: Logically this makes no sense...

Test your solution with the following steps:
1. Open the docs (locally) and remove the localStorage item `devModeEnabled`
  image

2. Check if `<html data-env="production">` is set by NODE_ENV
3. Add the URL parameter `&devModeEnabled=true` and hit enter
4. Check if `<html data-env="development">` is set by URL parameter
5. Remove the URL parameter and hit enter
6. Check if  `<html data-env="development">` is set by the localStorage item (which should store the last state)
7. Add the URL parameter `&devModeEnabled=false` and hit enter
8. Check if `<html data-env="production">` is set by URL parameter
9. Remove the URL parameter and hit enter
10. Check if  `<html data-env="production">` is set by the localStorage item (which should store the last state)

If you have questions, please call me ;)

.changeset/brown-chicken-ring.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': minor
+---
+
+Added the possibility to override the used environment mode (`development`or `production`) through the URL, 
+by setting the `devModeEnabled` parameter to `true`or `false`.
```
oliverschuerch:
```suggestion
Added the possibility to override the used environment mode (`development`or `production`) through the URL, by setting the `devModeEnabled` parameter to `true`or `false`. 
```

packages/documentation/.storybook/manager.ts
```diff
@@ -14,14 +14,44 @@ const TECH_ICONS: Record<string, string> = {
...
+  initialEnv = JSON.parse(devModeFromUrl) ? 'development' : 'production';
+} else {
+  // If no URL param, check if there's a stored devMode
+  if (storedDevMode !== null) {
+    // If the stored devModeEnabled is 'true', reset to 'production'
+    localStorage.removeItem('devModeEnabled');
+  }
+  // Default to 'production' if no URL param and no stored value
+  initialEnv = 'production';
 }
```
oliverschuerch, line 40:
```suggestion
if (devModeFromUrl !== null) {
  initialEnv = JSON.parse(devModeFromUrl) ? 'development' : 'production';
} else if (storedDevMode !== null) {
  initialEnv = JSON.parse(storedDevMode) ? 'development' : 'production';
}

localStorage.setItem('devModeEnabled', initialEnv);
```

packages/documentation/.storybook/manager.ts
```diff
@@ -14,14 +14,44 @@ const TECH_ICONS: Record<string, string> = {
...
+const devModeFromUrl = params.get('devModeEnabled');
 
-let initialEnv = process.env.NODE_ENV || 'production';
+// Get value, that states if localStorage devMode is enabled (String 'true' or 'false', or null if the item is not specified in localStorage)
+const storedDevMode = localStorage.getItem('devModeEnabled');
 
-if (storedDevMode !== null) {
-  initialEnv = JSON.parse(storedDevMode) ? 'development' : 'production';
+// Default fallback: always 'production'
+let initialEnv = 'production';
```
oliverschuerch:
```suggestion
let initialEnv = process.env.NODE_ENV || 'production'
```

.changeset/brown-chicken-ring.md
alizedebray, line 1: I agree with @oliverschuerch and would remove this changeset file altogether.

alizedebray (changes requested):

PR #6076, opened 2025-08-21
fix(docs): fix `<post-banner>` Web Component examples behavior in documentation

alionazherdetska:
## 📄 Description

This update corrects the behavior of the banner examples in the documentation.

- The **Reset Banner** button is now hidden initially.  
- It only becomes visible after the banner has been dismissed via the close button or programmatically

---

## 📝 Checklist


packages/documentation/src/stories/components/banner/web-component/post-banner.stories.ts
myrtasakellariou, line 1: Everything works fine on `localhost`, but in the preview link, the button appears right away when `dismissible` is set to `true`. It could be that `queueMicrotask(()` behaves a bit differently there. Might be worth simplifying it and just handling the display styles.

myrta2302 (changes requested):

packages/documentation/src/stories/components/banner/banner.stories.ts
```diff
@@ -44,63 +51,53 @@ const meta: MetaComponent<PostBannerControls> = {
...
+  const [, updateArgs] = useArgs();
 
-    if (banner.parentNode) {
-      await banner.dismiss();
-    } else {
-      bannerContainer.appendChild(banner);
-      button.hidden = true;
-    }
-  };
+  if (dismissed && prevDismissible !== undefined && prevDismissible !== dismissible) {
```
myrtasakellariou: Maybe this could be simplified to `if (dismissed && !dismissible) ` and `prevDismissible` could be removed.

leagrdv (approved):
This is more a fix than a chore, so if you could update the PR title before merging to something like **fix(docs): `<post-banner>` web component ex...**, that'd be great :)

PR #6078, opened 2025-08-21
chore(components, styles): remove standard html banners

alizedebray:
## 📄 Description

This PR remove the HTML banner examples and integrated the banner styles to the `post-banner` web component.

---

## 📝 Checklist


packages/components/src/components/post-banner/post-banner.tsx
```diff
@@ -98,36 +95,29 @@ export class PostBanner {
...
-        </div>
+      <Host data-version={version} role="alert">
+        {this.dismissible && (
+          <button class="btn-close" onClick={() => this.dismiss()}>
+            <span>{this.dismissLabel}</span>
+          </button>
+        )}
+
+
+        <slot name="heading" />
```
leagrdv, line 112: Since we're no longer adding a class around the heading slot, do you think it makes sense to still have a specific slot for it? We could simply remove it and only have the main `slot` and the actions one. What do you think?

leagrdv (changes requested):
Could you add a changeset and an entry in the migration guide, since it's a breaking change?

packages/components/src/components/post-banner/post-banner.tsx
alizedebray: I agree but I wanted to check with design first: if I remove it it's a breaking change so I want to make sure we will not have to add it back again later. I'll add the PR to the pattern discussion.

alizedebray: Can we merge the PR as-is and wait for design input about the banner heading? I created a separate ticket #6102

PR #6098, opened 2025-08-26
chore: fix syntax in deploy demo yaml

gfellerph:
Let's see if this fixes the failing workflow runs.


alizedebray (approved):

alizedebray (changes requested):
The demo no longer exists on `main`. This workflow should be on the branch from where we want to release the demo, or at least check out the correct branch.

gfellerph:
Close as wrong target branch

PR #6123, opened 2025-08-29
feat(documentation): add accessibility reference relationship stories and demo components

myrta2302:
## 📄 Description

This PR adds documentation for the following cross-component referencing scenarios under the (private) Accessibility section:

1. for / label
2. aria-labelledby
3. aria-describedby
4. role="list"

To support these examples, a new demo-components folder has been added under `documentation/src`, containing the following native web components:

- demo-button
- demo-label
- demo-list-item-group
- demo-list-item
- demo-list
- demo-span
- demo-target

## 📝 Checklist

### DIFF
--- a/packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
+++ b/packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
@@ -25,11 +25,8 @@ import WorkingExampleIV from './examples/working-example-IV.sample.html?raw';
   This is useful when a simple text label isn't enough or when the description needs to be combined
   from different pieces of text.
 </p>
-<p>
-  ℹ️ The `aria-describedby` attribute is not intended to reference content from external sources.
-  Its value must consist of one or more element IDs (space-separated if multiple), and each ID must
-  correspond to an element within the same DOM document.
-</p>
+<p>ⓘ The `aria-describedby` attribute is not intended to reference content from external sources. Its value must consist of one or more element IDs (space-separated if multiple), and each ID must
+  correspond to an element within the same DOM document.</p>
 
 ### I. Referencing Within the Same DOM Tree
 
@@ -78,13 +75,17 @@ import WorkingExampleIV from './examples/working-example-IV.sample.html?raw';
 
 #### Workaround: `Element:ariaDescribedByElements`
 
-<p>`ariaDescribedByElements` is a property of the <a href="https://developer.mozilla.org/fr/docs/Web/API/Element" target="blank">Element API</a> that returns an array of elements
+<p>
+  `ariaDescribedByElements` is a property of the <a href="https://developer.mozilla.org/fr/docs/Web/API/Element" target="blank">Element API</a> that returns an array of elements
   describing the element. It provides direct references to these elements rather than just their ID
-  strings. <a
+  strings.
+  <a
     target="blank"
     href="https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaDescribedByElements"
   >
-    🔗ariaDescribedByElements</a></p>
+    🔗ariaDescribedByElements
+  </a>
+</p>
 
 <div className="lead">Basic usage of `ariaDescribedByElements` </div>
 <Source language="html" dark code={AriaDescribedbyElementsPropertyHTML}></Source>
--- a/packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-labelledby/aria-labelledby.docs.mdx
+++ b/packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-labelledby/aria-labelledby.docs.mdx
@@ -21,16 +21,11 @@ import WorkingExampleIV from './examples/working-example-IV.sample.html?raw';
   In the Light DOM, the `aria-labelledby` attribute assigns an accessible name to an element by
   referencing the ID of one or more elements that provide the descriptive text.
 </div>
-<p>
-  This is particularly useful when a simple text label isn’t directly linked to the element, or when
-  the accessible name needs to be composed from multiple pieces of text.
-</p>
+<p>This is particularly useful when a simple text label isn’t directly linked to the element, or when
+  the accessible name needs to be composed from multiple pieces of text.</p>
 
-<p>
-  ℹ️ The `aria-labelledby` attribute is not intended to reference content from external sources.
-  Its value must consist of one or more element IDs (space-separated if multiple), and each ID must
-  correspond to an element within the same DOM document.
-</p>
+<p>ⓘ The `aria-labelledby` attribute is not intended to reference content from external sources.
+  Its value must consist of one or more element IDs (space-separated if multiple), and each ID must correspond to an element within the same DOM document.</p>
 
 ### I. Referencing Within the Same DOM Tree
 
@@ -41,12 +36,7 @@ import WorkingExampleIV from './examples/working-example-IV.sample.html?raw';
 • <b>Both elements in same Shadow DOM</b> ✔️
 
 <Source language="html" dark code={DefaultExample2}></Source>
-<p>
-  <b>
-    The above relationships are valid because in both cases the label and its associated target
-    element are part of the same tree.
-  </b>
-</p>
+<p><b>The above relationships are valid because in both cases the label and its associated target element are part of the same tree.</b></p>
 
 <p>
   Below is a working example of an `aria-labelledby`/`id` relationship within the shadowDom of a
@@ -79,15 +69,8 @@ import WorkingExampleIV from './examples/working-example-IV.sample.html?raw';
 
 #### Workaround: `Element:ariaLabelledByElements`
 
-<p>`ariaLabelledByElements` is a property of the <a href="https://developer.mozilla.org/fr/docs/Web/API/Element" target="blank">
-    Element API</a> that returns an array of elements labeling the element. It provides direct references to these
-  elements rather than just their ID strings.<a
-    target="blank"
-    href="https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaLabelledByElements"
-  >
-    🔗ariaLabelledByElements
-  </a></p>
-
+<p>`ariaLabelledByElements` is a property of the <a href="https://developer.mozilla.org/fr/docs/Web/API/Element" target="blank">Element API</a> that returns an array of elements labeling the element. It provides direct references to these
+  elements rather than just their ID strings. <a target="blank" href="https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaLabelledByElements">🔗ariaLabelledByElements</a></p>
 <div className="lead">Basic usage of `ariaLabelledByElements` </div>
 <p>
   Instead of adding the `aria-labelledby` attribute on the element, we can set it using the


packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-labelledby/aria-labelledby.docs.mdx
```diff
@@ -0,0 +1,160 @@
...
+</div>
+<p>
+  This is particularly useful when a simple text label isn’t directly linked to the element, or when
+  the accessible name needs to be composed from multiple pieces of text.
+</p>
+### I. Referencing Within the Same DOM Tree
+
+<Canvas of={CCR.ExampleHTML} sourceState="hide" />
+
+• <b>Same Light DOM</b> ✔️
```
oliverschuerch: Because there is only one light dom
```suggestion
• <b>Both elements in Light DOM</b> ✔️
```

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-labelledby/aria-labelledby.docs.mdx
```diff
@@ -0,0 +1,160 @@
...
+
+<Source language="html" dark code={DefaultExample}></Source>
+
+• <b>Same Shadow DOM</b> ✔️
```
oliverschuerch:
```suggestion
• <b>Both elements in same Shadow DOM</b> ✔️
```

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-labelledby/examples/defaultExample2.sample.html
```diff
@@ -0,0 +1,11 @@
...
+  <!-- #shadow-root -->
+   
+    <!-- This is a label element in the "shadow DOM" -->
+    <span id="labelId">My Text</span>
+
+    <!-- This is a div element in the same "shadow DOM" -->
+    <div part="button" role="button" tabindex="0" aria-labelledby="labelId">
+      <icon></icon>
+    </div>
+</custom-component>
```
oliverschuerch, line 7: Can you try out how this looks in the code preview?

```suggestion
<custom-component>
  ⏷ #shadow-root (open)
    <span id="labelId">My Text</span>
    <div part="button" role="button" tabindex="0" aria-labelledby="labelId">
      <custom-icon />
    </div>
</custom-component>
```

myrtasakellariou:
> The preview URL is not working.

I merged the conflicts to main and it is working now.

packages/documentation/src/demo-components/demo-button.ts
```diff
@@ -0,0 +1,83 @@
...
+        elementToLink = document.querySelector(`#${id}`);
+      }
+    } else if (this.buttonVersion === '2' || this.buttonVersion === '4') {
+      if (this.slotEl) {
+        const assignedElements = this.slotEl.assignedElements({ flatten: true });
+        elementToLink = assignedElements.find(el => el.tagName === 'SPAN') || null;
+      }
+    }
+
+    const ariaPropertyName = isLabelled ? 'ariaLabelledByElements' : 'ariaDescribedByElements';
```
oliverschuerch: It's a so called "good practice" to define variables at the beginning of a function. Please move this to line 32.

packages/documentation/src/demo-components/demo-button.ts
```diff
@@ -0,0 +1,83 @@
...
+    this.render();
+  }
+
+  attributeChangedCallback(name: string, _oldValue: string, newValue: string) {
+    if (name === 'workaround') this.workaround = newValue;
+    if (name === 'arialabelledby-id') this.arialabelledbyId = newValue;
+    if (name === 'ariadescribedby-id') this.ariadescribedbyId = newValue;
+    if (name === 'button-version') this.buttonVersion = newValue as '1' | '2' | '3';
+    this.render();
+  }
```
oliverschuerch: Wouldn't it be easier to use some kind of generic logic here?

```ts
attributeChangedCallback(name: string, _oldValue: string, newValue: string) {
  const memberName = toCamelCase(name);
  if (this.hasOwnProperty(memberName)) this[memberName] = this[] = newValue;
  this.render();
}

private toCamelCase(name: string) {
  return name.replace(/-./g, m => m[1].toUpperCase());
}
```

oliverschuerch, line 34: Since only the observed attributes will execute this callback, you could easily make it generic.

```ts
attributeChangedCallback(name: string, _oldValue: string, newValue: string) {
  const member = name.replace(/-./g, m => m[1].toUpperCase());
  
  this[member] = newValue;
  this.render();
}
```

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/examples/default-example.sample.html
oliverschuerch, line 1: To be inline with the given file naming, I suggest to rename all the sample files:

1. Write names in kebab-case
2. Add context for the html samples and only use numbers/letters if there is no better word to describe the context
(e.g. `default-example-lightdom.sample.html` and `default-example-shadowdom.sample.html` or `not-working-example-1.sample.html` and `not-working-example-2.sample.html`).

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
```diff
@@ -0,0 +1,171 @@
...
+    are part of the same tree.
+  </b>
+</p>
+
+<p>
+  Below is a working example of an `aria-describedby`/`id` relationship within the shadowDom of a
+  custom button component. You can <b>test it with NVDA</b> to confirm the correct description
+  announcement.
+</p>
+<Canvas of={CCR.ExampleHTML} sourceState="hide" />
```
oliverschuerch: The text says it's a working `shadowdom` example, but the story contains a `lightdom` example.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
```diff
@@ -0,0 +1,171 @@
...
+
+
+### II. Referencing From Outside a Shadow DOM Into That Shadow DOM
```
oliverschuerch: The `aria-describedby` attribute, which is in `shadowdom` references the `id` attribute, which is in the `lightdom`. Therefore either the title or the example is wrong.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
```diff
@@ -0,0 +1,171 @@
...
+import NonWorkingExampleIIb from './examples/nonWorkingExample-IIb.sample.html?raw';
+import WorkingExampleIIa from './examples/workingExample-IIa.sample.html?raw';
+import WorkingExampleIIb from './examples/workingExample-IIb.sample.html?raw';
+import NonWorkingExampleIII from './examples/nonWorkingExample-III.sample.html?raw';
+import WorkingExampleIII from './examples/workingExample-III.sample.html?raw';
+import NonWorkingExampleIV from './examples/nonWorkingExample-IV.sample.html?raw';
+import WorkingExampleIV from './examples/workingExample-IV.sample.html?raw';
+
+<Meta of={CCR} />
+
```
oliverschuerch, line 17: I would add something like the following as a general statement:

> Note: The aria-describedby attribute is not designed to reference descriptions from external resources. As its value is one or more ids (space-separated if multiple), it must reference elements in the same DOM document.

Source: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-describedby#example

If you find a better description about this fact, that be even better.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/examples/non-working-example-IIb.sample.html
```diff
@@ -0,0 +1,8 @@
+<custom-component>
+  ⏷ #shadow-root (open)
+    <div role="button" tabindex="0" aria-describedby="descriptionId">
+      <custom-icon />
+    </div>
+    <slot name="description-slot"> </slot>
+  <span slot="description-slot" id="descriptionId">My Description</span>
```
oliverschuerch, line 8: It is a bit hard to distinguish between `shadowdom` and slotted `lightdom` elements in the code example. Maybe adding a blank line inbetween helps.

```suggestion
    <slot name="description-slot"> </slot>
  
  <span slot="description-slot" id="descriptionId">My Description</span>
```

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/examples/nonWorkingExample-IV.sample.html
```diff
@@ -0,0 +1,12 @@
+<custom-button>
+  ⏷ #shadow-root (open)
+    <div role="button" tabindex="0" aria-describedby="descriptionId">
+      <custom-icon />
+    </div>
+</custom-button>
+
+<custom-description>
+  ⏷ #shadow-root (open)
+  <span id="descriptionId">My Description</span>
```
oliverschuerch: Check if the indentation is correct. If the `<span>` is a `shadowdom` element, increase indent. If it is a `lightdom` element, remove the `#shadow-root (open)` node.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/examples/workaround-IIa.sample.html
oliverschuerch, line 1: IIa and IIb sample files only differ by the newline at the end of the file.
Can we get rid of one of these files?

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/examples/working-example-IIa.sample.html
```diff
@@ -0,0 +1,16 @@
+<span id="descriptionId">My Description</span>
+
+<custom-component descriptionId="descriptionId">
+  ⏷ #shadow-root (open)
+    <div ref="targetRef" role="button" tabindex="0" aria-describedby>
+      <!--The aria-describedby attribute is set automatically by the browser -->
+        <custom-icon />
```
oliverschuerch: Wrong indentation.
The `<custom-icon />` tag should be at the same level as the comment above.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/examples/workingExample-IIb.sample.html
```diff
@@ -0,0 +1,19 @@
+<custom-component>
+  ⏷ #shadow-root (open)
+  <slot name="description-slot"></slot>
+  <div ref="targetRef" role="button" tabindex="0" aria-describedby>
+    <!--The aria-describedby attribute is set automatically by the browser -->
+    <custom-icon />
+  </div>
```
oliverschuerch: Increase indentation (same as like in other files).

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/examples/workingExample-IIb.sample.html
```diff
@@ -0,0 +1,19 @@
...
+  <!-- slotted description -->
```
oliverschuerch:
```suggestion

```

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
```diff
@@ -0,0 +1,171 @@
...
+### II. Referencing From Outside a Shadow DOM Into That Shadow DOM
+
+• <b>Light DOM → Shadow DOM</b> ❌
+
+<Source language="html" dark code={NonWorkingExampleIIa}></Source>
+
+• <b>Slotted Content → Shadow DOM</b> ❌
+
+<Source language="html" dark code={NonWorkingExampleIIb}></Source>
```
oliverschuerch: It's the `aria` attribute, which is in `shadowdom`, that references one or more `id` attributes.

Eiter fix the wording or update the examples.
e.g.
**II. Referencing from a specific shadowdom, to an element outside this exact shadowdom**
Shadow DOM -> Light DOM ❌
Shadow DOM -> Slotted Content ❌
Shadow DOM A -> Shadow DOM B ❌

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
```diff
@@ -0,0 +1,171 @@
...
+    IDs are only valid within their own scope.
+  </b>
+</p>
+
+#### Workaround: `Element:ariaDescribedByElements`
+
+<p>
+  `ariaDescribedByElements` is a property of the Element interface that returns an array of elements
+  describing the element. It provides direct references to these elements rather than just their ID
+  strings.
```
oliverschuerch: It could be helpful for some DEVs, if you add the information, that this is a [WebAPI](https://developer.mozilla.org/en-US/docs/Web/API) and therefore JavaScript related.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
```diff
@@ -0,0 +1,171 @@
...
+    🔗ariaDescribedByElements
+  </a>
+</p>
+
+<div className="lead">Basic usage of `ariaDescribedByElements` </div>
+<Source language="html" dark code={AriaDescribedbyElementsPropertyHTML}></Source>
+<p>
+  Instead of adding the `aria-describedby` attribute on the element, we can set it using the
+  `ariaDescribedByElements` property.
+</p>
```
oliverschuerch, line 95: I'd like to have the small explainer before I dive into the code example.

```suggestion
<p>
  Instead of adding the `aria-describedby` attribute on the element, we can set it using the
  `ariaDescribedByElements` property.
</p>
<Source language="html" dark code={AriaDescribedbyElementsPropertyHTML}></Source>
```

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/examples/ariadescribedbyElementsProperty.sample.html
```diff
@@ -0,0 +1,3 @@
+<element id="elementId"></element>
+
+<span id="descriptionId"></span>
```
oliverschuerch: I would add a little content (text) to both of these elements.

Is the `<element>` tag representing a custom element?
If yes, it needs to have a dash in the name `<my-element>`.
If no, why not using a native element like `<div>` or so?

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/examples/workingExample-IIa.sample.html
```diff
@@ -0,0 +1,16 @@
+<span id="descriptionId">My Description</span>
```
oliverschuerch: In the first two examples on the page you implement the description after the element that references it.
Should we keep the order to make it more comparable or is it mandatory to have the referenced elmeent before the one which references it?

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
```diff
@@ -0,0 +1,171 @@
...
+  In this case a description element in the light DOM is directly assigned to a target element inside a
+  custom component’s shadow DOM using `ariaDescribedByElements`.
+</p>
+
+<Source language="html" dark code={WorkingExampleIIa}></Source>
+<p>Use the control below to toggle the `ariaDescribedByElements` workaround implementation.</p>
+<Canvas of={CCR.Example2} sourceState="hide" />
+<div className="hide-col-default">
+  <Controls of={CCR.Example2} />
+</div>
```
oliverschuerch, line 115: Something like "checkout our real life example" would be enough for me. Why should i try the example, which does not work?

Imo, we can remove all these control tables, if they do not provide any benefit.
What do you think?

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
```diff
@@ -0,0 +1,171 @@
...
+<Source language="html" dark code={WorkingExampleIIb}></Source>
+<p>Use the control below to toggle the `ariaDescribedByElements` workaround implementation.</p>
+<Canvas of={CCR.Example3} sourceState="hide<p>" />
+<div className="hide-col-default">
+  <Controls of={CCR.Example3} />
+</div>
+
+⚠️ In the above scenarios, the relationship is correctly announced by NVDA, passes accessibility checks on Axe, WAVE and NVDA but partially fails on Siteimprove and Accessibility Insights tools.
+
+### III. Referencing From Inside a Shadow DOM Out to the Light DOM
```
oliverschuerch: Again, the `aria` attribute is the one that references to the `id` attributes.
There for the example shows how you reference from the `lightdom` to the `shadowdom`.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.styles.scss
```diff
@@ -0,0 +1,22 @@
+::part(button) {
```
oliverschuerch, line 2: If this belongs only to the `demo-button` custom-element, please prefix it with the component tag or at least leave a comment, so we know also in a year, to what component these styles blong.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-labelledby/examples/arialabelledbyElementsProperty.sample.html
```diff
@@ -0,0 +1,3 @@
+<span id="labelId"></span>
+
+<element id="elementId"></element>
```
oliverschuerch: Add some text content and use either a native element or a valid custom-element name for the `<element>` tag.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/examples/default-example.sample.html
oliverschuerch, line 1: Please keep file naming consistant.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/examples/default-example-3.sample.html
```diff
@@ -0,0 +1,7 @@
+<custom-component role="list">
+  ⏷ #shadow-root (open)
+    <slot></slot>
+  <div role="listitem">item 1</div>
```
oliverschuerch, line 5:
```suggestion

  <div role="listitem">item 1</div>
```

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/aria-role-list.docs.mdx
```diff
@@ -0,0 +1,137 @@
...
+<p>
+  Same as above, but contained within a custom component’s Shadow DOM. The `role="list"` still has
+  direct `role="listitem"` children, so the structure remains valid.
+</p>
+
+• <b>`list` role on the host and slotted `listitem` children</b> ✔️
+
+<Source language="html" dark code={DefaultExample3}></Source>
+
+<Canvas of={CCR.Example1b} sourceState="hide" />
```
oliverschuerch, line 50: Do we need the stories? These are both common cases and they work just find. Imo, the code examples are enough.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/aria-role-list.docs.mdx
```diff
@@ -0,0 +1,137 @@
...
+• <b>Light DOM → Shadow DOM</b> ✔️
+
+<Source language="html" dark code={WorkingExampleIIa}></Source>
+
+<Canvas of={CCR.Example2a} sourceState="hide" />
+
+<p>
+  In the case a component with `role="list"` in the host with `listitem` elements within its Shadow
+  DOM, the relationship is valid.
+</p>
```
oliverschuerch, line 68: But why? Should be explained, as it seems to be an exception.
Or what's the difference to the aria attributes, which can't cross different DOM borders.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/examples/working-example-II.sample.html
oliverschuerch, line 1: This examples seems to be unnecessary. Which DEV would think this could work? 😆

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/aria-role-list.docs.mdx
```diff
@@ -0,0 +1,137 @@
...
+  relationship is preserved correctly.
+</p>
+
+### II. Parent in the Light DOM-children Into The Shadow DOM
+
+• <b>Light DOM → Shadow DOM</b> ✔️
+
+<Source language="html" dark code={WorkingExampleIIa}></Source>
+
+<Canvas of={CCR.Example2a} sourceState="hide" />
```
oliverschuerch: I don't think we need that.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/examples/working-example-IIIa.sample.html
oliverschuerch, line 1: Same here.
Unlike the `aria` attributes, which are not bound to a nesting structure, this is absolutely the case for the `role` attributes.

I don't think we need to explain that fact to developers.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/aria-role-list.docs.mdx
oliverschuerch, line 1: Is there a workaround here as well?
Maybe with the upcoming [ariaOwnsElements](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaOwnsElements) property?

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/for/examples/nonWorkingExample-IIb.sample.html
```diff
@@ -0,0 +1,6 @@
+<custom-component>
+  ⏷ #shadow-root (open)
+    <slot name="label-slot"></slot>
+    <input id="inputId">
+  <label for="inputId" slot="label-slot">My Text</label>
```
oliverschuerch, line 6:
```suggestion

  <label for="inputId" slot="label-slot">My Text</label>
```

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/for/examples/workaround-IIb.sample.html
```diff
@@ -0,0 +1,6 @@
+<custom-input>
+  ⏷ #shadow-root (open)
+  <input id="inputId" aria-labelledby="labelId" />
+
+  <label id="labelId" for="labelId">My Text</label>
+</custom-input>
```
oliverschuerch: Fix indentation

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/for/examples/working-example-IIa.sample.html
```diff
@@ -0,0 +1,16 @@
+<label for="inputId">My Text</label>
+
+<custom-input labelId="labelId">
+  ⏷ #shadow-root (open)
+    <input ref="targetRef" id="inputId" aria-labelledby>
+      <!--The aria-labelledby attribute is set automatically by the browser -->
+        <custom-icon />
+    </div>
+</custom-input>
```
oliverschuerch, line 7: Fix missing opening `div`.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/for/examples/workingExample-IIb.sample.html
```diff
@@ -0,0 +1,18 @@
+<custom-component>
+  ⏷ #shadow-root (open)
+    <slot name="label-slot"></slot>
+    <!--The aria-labelledby attribute is set automatically by the browser -->
+    <div ref="targetRef" role="button" tabindex="0" aria-labelledby>
+
+  <!-- slotted label element -->
```
oliverschuerch:
```suggestion
```

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/for/examples/non-working-example-IIa.sample.html
oliverschuerch, line 1: Keep file naming consistant.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/for/examples/workingExample-IV.sample.html
```diff
@@ -0,0 +1,16 @@
+<custom-label id="labelId">
+  ⏷ #shadow-root (open)
+    <label for="inputId">My Text</label>
+</custom-label>
+
+<custom-button>
+  ⏷ #shadow-root (open)
+  <input ref="targetRef" id="inputId" aria-labelledby />
```
oliverschuerch:
```suggestion
    <input ref="targetRef" id="inputId" aria-labelledby />
```

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/for/for.docs.mdx
```diff
@@ -0,0 +1,170 @@
...
+  attributes directly.
+</p>
+<p>
+  <b>
+    Attribute element references can only point to elements in the same DOM or shadow DOM, because
+    IDs are only valid within their own scope.
+  </b>
+</p>
+
+#### Workaround: `Element:ariaLabelledByElements`
```
oliverschuerch, line 72: This complete section reminds me almost 100% of the aria-labelledby page.
Do we need to keep both or would it be enough to say here, that you can use the `ariaLabelledByElements` workaround on the other page?

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/reference-table/reference-table.docs.mdx
oliverschuerch, line 1: Check the links. some of them did not work.

packages/documentation/src/demo-components/demo-button.ts
myrtasakellariou: ok

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/examples/default-example.sample.html
myrtasakellariou: I applied the kebab-case but kept the numbers in e.g. Ia, IIa to be consistent with the doc sections.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/examples/workaround-IIa.sample.html
myrtasakellariou: All workaround-* named files where not used at the end. I removed them.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
myrtasakellariou: ok, I ve added something very similar to this.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
myrtasakellariou: Correct. I updated the example.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
myrtasakellariou: Updated except for aria-labelledby and aria-describedby.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.docs.mdx
myrtasakellariou: I would still like to have the ability to toggle between a working and not working state. Because e.g. i would maybe want to see what the screenreader announces in any case.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/aria-describedby.styles.scss
myrtasakellariou: Added a comment.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/examples/working-example-II.sample.html
myrtasakellariou: (ha ha .. me maybe? ) If you want i can remove it. But i dont find the harm in keeping it.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/examples/working-example-IIIa.sample.html
myrtasakellariou: Ok. But i still feel it does not harm to keep even some more 'stupid' examples.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/aria-role-list.docs.mdx
myrtasakellariou: Why would we need a workaround since it already works?

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/aria-role-list.docs.mdx
myrtasakellariou: I would like to have the possibility to inspect and test with NVDA.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/for/for.docs.mdx
myrtasakellariou: I would like to keep both, to have the specific examples.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/reference-table/reference-table.docs.mdx
myrtasakellariou: Correct, when the heading changes, the anchor changes.

myrtasakellariou:
> Currently, the guidelines are always visible, no matther which mode the docs are rendered in (development | production). Shouldn't we hide the guidelines in production mode, or is the content also relevant for our users?
> 
> The docs pages `aria-describedby` and `aria-labelledby` are almost the same (compared by page structur and content). Please update the CRs requested for one also on the other.
> 
> During the review, I asked my self if the examples are just to specific? At least it's a lot of repeating information on all of the example pages.
> 
> I think it could be beneficial to add a page with general information about, how different HTML attributes or JS properties can cross the DOM borders or not. And then additionally only document exceptions.
> 
> e.g.
> 
> * light    -> light ✔️
> * light    -> shadow ❌
> * shadow   -> light ❌
> * shadow   -> slotted ❌
> * shadow A -> shadow A ✔️
> * shadow B -> shadow B ❌
> * ...
> 
> if we'd do that and add examples, most of the content of the other pages could become obsolete.
> 
> Since you've already put a lot of effort into it, I'll leave it up to you if you want to do that, ask the team about it or completely ignore this idea.

> Currently, the guidelines are always visible

It is because of the current Storybook bug https://github.com/swisspost/design-system/issues/6631. Raw components are also visible right now. 

>  The docs pages `aria-describedby` and `aria-labelledby` are almost the same

I would prefer to keep both pages.

> I asked my self if the examples are just to specific

Specific examples was exactly my intention for a quick reference for every case. I personally  benefit from some "stupid" examples too.

> _a page with general information_

That is i would say what the table reference for. Do you think it needs more text?

packages/components/src/components/post-popover-trigger/readme.md
oliverschuerch, line 1: Why is there a change in this file?
Please update your working branch.

packages/documentation/src/demo-components/demo-button.ts
```diff
@@ -0,0 +1,96 @@
...
+    super();
+    this.attachShadow({ mode: 'open' });
+  }
+
+  connectedCallback() {
+    this.render();
+  }
+
+  attributeChangedCallback(name: string, _oldValue: string, newValue: string) {
+    const member = name.replace(/-([a-z])/g, (_, c) => c.toUpperCase());
```
oliverschuerch: I have no idea what `c` means.

But I know that the function you implemented is a [replacer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_function_as_the_replacement) function, which receives a `match` object as its first parameter and then one `pattern` parameter per capturing group in the RegExp.

Because of that, you should rename `c` to `p` or use the `match` object parameter as I showed you [here](https://github.com/swisspost/design-system/pull/6123/files#r2498303277).

It is technically doing the same thing.

Image

packages/documentation/src/demo-components/demo-button.ts
```diff
@@ -0,0 +1,96 @@
+export class DemoButton extends HTMLElement {
+  static get observedAttributes() {
+    return ['button-version', 'workaround', 'arialabelledby-id', 'ariadescribedby-id'];
+  }
+  private buttonVersion?: '1' | '2' | '3' | '4';
```
oliverschuerch: Please create a reusable type for `buttonVersion`.

e.g.
```ts
type buttonVersion = '1' | '2' | '3' | '4';

export class DemoButton extends HTMLElement {
  private buttonVersion?: buttonVersion;

  constructor() {
    super();
    this.attachShadow({ mode: 'open' });
  }

  attributeChangedCallback(name: string, _oldValue: string, newValue: buttonVersion | string) {
    this.buttonVersion = newValue as buttonVersion;
  }
}
```

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-describedby/examples/working-example-IIa.sample.html
oliverschuerch: I think you missed that one, didn't you?

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-labelledby/examples/non-working-example-IIb.sample.html
```diff
@@ -0,0 +1,8 @@
+<custom-component>
+  ⏷ #shadow-root (open)
+  <slot name="aria-slot"> </slot>
+  <div role="button" tabindex="0" aria-labelledby="labelId">
+    <custom-icon />
+  </div>
+  <span slot="aria-slot" id="labelId">My Text</span>
```
oliverschuerch, line 8: Fix indentation and add an empty line between shadowdom and lightdom elements.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-labelledby/examples/non-working-example-IV.sample.html
```diff
@@ -0,0 +1,11 @@
+<custom-label>
+  ⏷ #shadow-root (open)
+  <span id="labelId">My Text</span>
```
oliverschuerch, line 3: Indentation

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-labelledby/examples/working-example-IIb.sample.html
```diff
@@ -0,0 +1,19 @@
+<custom-component>
+  ⏷ #shadow-root (open)
+  <slot name="aria-slot"></slot>
+  <div ref="targetRef" role="button" tabindex="0" aria-labelledby>
+    <!--The aria-labelledby attribute is set automatically by the browser -->
+    <custom-icon />
+  </div>
+  <!-- slotted label element -->
+  <span slot="aria-slot">My Text</span>
```
oliverschuerch, line 9: Indentation

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-labelledby/aria-labelledby.docs.mdx
```diff
@@ -0,0 +1,172 @@
...
+# aria-labelledby
+
+<div className="lead">
+  In the Light DOM, the `aria-labelledby` attribute assigns an accessible name to an element by
+  referencing the ID of one or more elements that provide the descriptive text.
+</div>
+<p>
+  This is particularly useful when a simple text label isn’t directly linked to the element, or when
+  the accessible name needs to be composed from multiple pieces of text.
+</p>
```
oliverschuerch: Should we add the following message here as well?

"⚠️ The `aria-labelledby` attribute is not intended to reference content from external sources. Its value must consist of one or more element IDs (space-separated if multiple), and each ID must correspond to an element within the same DOM document."

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/examples/working-example-II.sample.html
oliverschuerch: I'm sorry if I offended you, that was not my intention.

This is what the example is producing in the browser:
```html
<custom-list>
  <ul></ul>
  <li>item 1</li>
  <li>item 2</li>
  <li>item 3</li>
<custom-list>
```

It has nothing to do with light- or shadowdom barriers, but this is just invalid HTML.
Image
Source: https://validator.w3.org
`ul` specs: https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element
`li` specs: https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element

In other words, you probably won't show the following example to anyone with JS knowledge and tell them, this won't work:

```js
var my-var = 'abc' + function() {};
```

Fun fact: assigning `'abc' + function() {};` to a variable indeed works in JS without any issue 😆

With all these new facts, I'm even more sure we should remove these examples.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/examples/working-example-IIIa.sample.html
oliverschuerch: That's what it would output.
```html
<custom-list>
  <ul>test</ul>
</custom-list>

<li>item 1</li>
<li>item 2</li>
<li>item 2</li>
```

This is not even valid HTML!
Image
Source: https://validator.w3.org

I want to have it removed, because it is invalid HTML.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/examples/working-example-IVa.sample.html
```diff
@@ -0,0 +1,11 @@
+<custom-list>
+  ⏷ #shadow-root (open)
+    <div role="list">
+      <custom-list-item-group>
+        ⏷ #shadow-root (open)
+        <div role="listitem">item 1</div>
+        <div role="listitem">item 2</div>
+        <div role="listitem">item 3</div>
```
oliverschuerch: Indentation

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/aria-role-list.docs.mdx
oliverschuerch: Sorry, you're right. I think I just missed the fact, that it is actually a page describing `role="list"`, not another `aria` attribute.
This brings me to the fact, that we should rename the page in the navigation and also the file itself.

The `role` attribute indeed belongs to the [WAI-ARIA spec](https://www.w3.org/TR/wai-aria-1.2/#roles), but to keep the navigation consistant, we should not prefix it with `aria-`.

What do you think?

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/for/examples/non-working-example-IIb.sample.html
```diff
@@ -0,0 +1,7 @@
+<custom-component>
+  ⏷ #shadow-root (open)
+    <slot name="label-slot"></slot>
+    <input id="inputId">
+    
+    <label for="inputId" slot="label-slot">My Text</label>
```
oliverschuerch: Indentation

packages/documentation/src/demo-components/demo-button.ts
myrtasakellariou: c holds the character matched by the capture group. I changed it to p.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/aria-role-list.docs.mdx
myrtasakellariou: Yes. I renamed it.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/examples/working-example-II.sample.html
myrtasakellariou: i 've removed the invalid markup examples.

packages/documentation/src/stories/accessibility-practices/foundational-structure-and-semantics/reference-crossing-the-shadowdom/aria-role-list/examples/working-example-II.sample.html
```diff
@@ -0,0 +1,10 @@
+<custom-list>
+  ⏷ #shadow-root (open)
+    <slot name="parent-slot"></slot>
+    <div role="listitem">item 1</div>
+    <div role="listitem">item 2</div>
+    <div role="listitem">item 3</div>
+
+ <!-- Slotted parent with role="list"-->   
+ <div role="list" slot="parent-slot"></div>
+</custom-list>
```
oliverschuerch, line 10:
```suggestion
<custom-list>
  ⏷ #shadow-root (open)
    <slot name="parent-slot"></slot>
    <div role="listitem">item 1</div>
    <div role="listitem">item 2</div>
    <div role="listitem">item 3</div>

  <!-- Slotted parent with role="list"-->   
  <div role="list" slot="parent-slot"></div>
</custom-list>
```

oliverschuerch (approved):

myrtasakellariou:
To be merged when the storybook navigation issue (that causes the hidden sections to be public) is resolved.

PR #6136, opened 2025-09-01
chore(docs): POC full width stories

leagrdv:
## 📄 Description

POC to show what full width stories would look like on docs that are within a `container`.

### DIFF
--- a/packages/documentation/.storybook/blocks/layout/layout.tsx
+++ b/packages/documentation/.storybook/blocks/layout/layout.tsx
@@ -18,7 +18,7 @@ export default (props: PropsWithChildren<DocsContainerProps>) => {
   const { children, context } = props;
   const container =
     context.channel.data.docsPrepared[0].parameters.layout === 'fullscreen'
-      ? 'container-fluid'
+      ? 'container custom-container'
       : 'container';
   const pathToStoryFile = context?.storyIdToCSFFile?.values()?.next()?.value?.meta
     ?.parameters?.fileName;
--- a/packages/documentation/.storybook/styles/preview.scss
+++ b/packages/documentation/.storybook/styles/preview.scss
@@ -309,6 +309,14 @@ $monospace: SFMono-Regular, Menlo, Monaco, Consolas, 'Liberation Mono', 'Courier
   overflow-x: hidden;
 }
 
+@include media.min(xl) {
+  .custom-container {
+    .sbdocs-preview {
+      margin-inline: calc(-1 * ((100vw - var(--post-container-max-width)) / 2));
+    }
+  }
+}
+
 .sbdocs-preview {
   &:not([data-color-scheme]) {
     display: none;


packages/documentation/src/stories/templates/error-page/error-page.stories.ts
```diff
@@ -7,6 +7,11 @@ const meta: Meta = {
   tags: ['status:InProgress'],
   parameters: {
     layout: 'fullscreen',
+    docs: {
+      story: {
+        inline: false,
```
myrtasakellariou: Is this necessary here? If so, perhaps we should adjust the iframe height so that the main content is visible.

packages/documentation/.storybook/styles/preview.scss
```diff
@@ -322,6 +322,14 @@ $toolbar-icons: (
   overflow-x: hidden;
 }
 
+@include media.min(xl) {
+  .custom-container {
```
myrtasakellariou: It would be maybe nice to have a more descriptive class name, such as “fullscreen-container” or “extended-container.

packages/documentation/src/stories/templates/error-page/error-page.stories.ts
leagrdv: Not sure anymore why I had added that in the past but indeed, it's not useful here. I removed it.

.changeset/fair-groups-make.md
myrtasakellariou, line 1: Not sure if this needs to be communicated via a changeset, since it is only an internal styling update.

oliverschuerch: You're right. I think we can remove the changeset for this pr.

@myrta2302 if you add a cr to a pr, please do not approve it in the meantime. This was a decision @alizedebray, @gfellerph and I once made.

packages/documentation/.storybook/blocks/layout/layout.tsx
```diff
@@ -18,7 +18,7 @@ export default (props: PropsWithChildren<DocsContainerProps>) => {
   const { children, context } = props;
   const container =
     context.channel.data.docsPrepared[0].parameters.layout === 'fullscreen'
-      ? 'container-fluid'
+      ? 'container extended-container'
```
oliverschuerch: I would turn around the class name (to follow our common naming convention`) and prefix it, so it becomes obvious it's a custom docs class:

`extended-container` --> `sb-container-extended`

@leagrdv I don't think we need to change anything in the POC. Instead, I would discuss in the DEV Roundtable, if we should follow our naming convention here or if we should add a storybook specific prefix for classes only defined in the docs CSS.

packages/documentation/.storybook/styles/preview.scss
oliverschuerch, line 1: On the first sight it looks very nice (for example on the header docs page).
After scrolling through the page a bit, I started thinking, what if ...:

- ... only one or two specific stories should be extended and the rest sticks to the common content width?
- ... some other content need to have an extended with?
  e.g. a code snippet, the controls table, etc.

@leagrdv I don't think we need to change anything in the POC. Instead, I would discuss in the DEV Roundtable, if these edge cases should be covered as well.

oliverschuerch (changes requested):

PR #6146, opened 2025-09-01
chore(eslint): propose common rule creator

alizedebray:
This PR implements a rule common rule creator for all rules that aim to replace a class with another. 
The `createClassUpdateRule` could be used with simple properties: 
```ts
{
  name: string; // name of the rule 
  description: string; // description of the rule 
  messages: Record<string, string>;  // error messages as { messageId: 'message text' }
  mutations: Record<string, [string] | [string, string]>; // actual class updates as { messageId: ['oldClass', 'newClass'] }
  type?: 'suggestion' | 'layout' | 'problem'; // rule type (defaults to 'problem')
}
```

**Advantages:**
- Simpler development for the rule
- Easy to update all rule at once

**Disadvantages:**
- Less flexible rules 

**Vote:**
🎉 to use the `createClassUpdateRule` creator
🚀 to right individual rules


packages/eslint/src/rules/html/migrations/no-deprecated-btn-rg.ts
```diff
@@ -1,43 +1,16 @@
...
-                const fixedNode = $node.removeClass('btn-rg').addClass('btn-sm');
-                return fixer.replaceTextRange(node.range, fixedNode.toString());
-              },
-            });
-          }
-        }
-      },
-    };
+  mutations: {
+    deprecatedBtnRg: ['btn-rg']
```
leagrdv:
```suggestion
    deprecatedBtnRg: ['btn-rg', 'btn-sm']
```

The `btn-rg` is smaller than the default button so it makes sense to replace it with `btn-sm`

packages/eslint/test/rules/html/migrations/no-deprecated-btn-rg.spec.ts
oliverschuerch, line 1: It would be nice to have a utility to create the test as well.
What do you think?

alizedebray: Good idea, I'll create a follow-up ticket for this.

PR #6165, opened 2025-09-04
chore(docs): move post-logo into raw components

leagrdv:
## 📄 Description

Moved the `post-logo` component into the `Raw components` folder in the documentation.


.changeset/stupid-bananas-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': minor
+---
+
+Moved the `post-logo` component into `Raw components`.
```
alizedebray: Raw components are hidden in prod, so it’s best not to mention them in the change log since users may not know what they are:

```suggestion
Removed documentation for the `post-logo` component, as it is not intended for standalone use outside the header.
```

PR #6173, opened 2025-09-04
chore(styles): disable automatic dark scheme

leagrdv:
## 📄 Description

Temporarily disabled the automatic dark scheme based on the user's preferred color scheme.


packages/styles/src/tokens/_schemes-static.scss
```diff
@@ -10,7 +10,7 @@
 // prefered root color-scheme
 :root {
   @include supports.light-dark(true) {
-    color-scheme: light dark;
+    color-scheme: light;
```
oliverschuerch, line 13: This is fine, but it does not set the fallback (below) to `light`.
I think to do this we would need to set the `--post-fallback-prefers-light` custom-property in the entire file like so:
`--post-fallback-prefers-light: ;`

Please add comments on the affected lines in the file directly, which explain how to reset that.

leagrdv:
> Do you really want to keep the changes in the styles playground? Imo, they where only there for testing reasons, but we should revert them now.
> 
> Instead we should remove the `data-color-scheme="light"` attribute on the `html` tag in the docs (see preview-body.html). Because this was the "old" way to ensure, the docs always were renderd in light mode and would need to be removed as soon as the dark mode is ready anyway.

I intentionally put the changes in the styles playground so that we can test components when they don't have a `data-color-scheme` on a parent, as we had found a bug recently with this issue

Also, removing the data-color-scheme on the preview-body breaks all of the tests of the palette as it's needed to test the palettes on the different `data-color-scheme`. Since storybook is not in dark mode anyway, I think we can leave it there for now and tackle it if we ever do a dark mode documentation.

leagrdv:
@oliverschuerch With the help of Alizé, I seem to have found a solution by simply removing the `@media (prefers-color-scheme: light/dark)` in the `:root` and defining the light theme by default for both cases. I tested it on a recent Edge as well as Edge 120 and it seems to all work as intended now!

PR #6194, opened 2025-09-09
chore(styles): rename loader to spinner

bucknatt:
## 📄 Description

Classes for the spinner component are renamed: 
`loader-*` is renamed to `spinner-*` in styles, docs and migration guide.

### DIFF
--- a/packages/components/cypress/e2e/popovercontainer.cy.ts
+++ b/packages/components/cypress/e2e/popovercontainer.cy.ts
@@ -36,7 +36,6 @@ describe('popovercontainer', { baseUrl: null, includeShadowDom: true }, () => {
           cy.get('@container').should('be.visible');
           cy.get(selector).should('exist');
           await container.hide();
-          cy.wait(1000);
           cy.get('@container').should('not.be.visible');
           cy.get(selector).should('not.exist');
           await container.toggle(trigger);


.changeset/many-wings-call.md
```diff
@@ -0,0 +1,7 @@
+---
+'@swisspost/design-system-documentation': major
```
leagrdv:
```suggestion
'@swisspost/design-system-documentation': minor
```

It's a major for the styles but not for the documentation

.changeset/many-wings-call.md
```diff
@@ -0,0 +1,7 @@
...
+'@swisspost/design-system-styles': major
+---
+
+Classes for the spinner component are renamed:
+loader-* is renamed to spinner-* in styles, docs and migration guide.
```
leagrdv:
```suggestion
Renamed the loader classes from `loader-*` to `spinner-*`.
```

There is no need to specify the packages (styles, docs, etc.) since we're stating at the beginning of the file which package it affects

packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -797,6 +797,16 @@ export class MigrationV99Component extends LitElement {
                     Replaced by the <code>post-banner</code> component.
                   </p>
                 </li>
+
+                <li class="mb-16">
+                  <p>
+                    Renamed loader classes and related scss variables (<code>.loading-modal</code>, <code>.loader</code>).
```
leagrdv: No need to write about the scss variables as they were not documented, but the sizes `loader-*` classes should be mentioned (no need to list them all)

leagrdv: No need to write about the scss variables as they were not documented, but the sizes `loader-*` classes should be mentioned (no need to list them all)

packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -797,6 +797,16 @@ export class MigrationV99Component extends LitElement {
                     Replaced by the <code>post-banner</code> component.
                   </p>
                 </li>
+
+                <li class="mb-16">
+                  <p>
+                    Renamed loader classes (<code>.loading-modal</code>, <code>.loader</code>).
```
leagrdv: Please mention the `.loader-*` classes as well (no need to list them all, but add the `.loader-*`). Since there are 3 elements, I think it'd be better to put them in a list and do as the other migration lines (e.g.: `...` is now `...`) as it's clearer to understand.
Like here for instance:
Capture d'écran 2025-09-11 130614

packages/components/cypress/e2e/popovercontainer.cy.ts
```diff
@@ -36,6 +36,7 @@ describe('popovercontainer', { baseUrl: null, includeShadowDom: true }, () => {
           cy.get('@container').should('be.visible');
           cy.get(selector).should('exist');
           await container.hide();
+          cy.wait(1000);
```
leagrdv: Please remove this change

PR #6198, opened 2025-09-10
fix(styles): fixed picker input display issues

bucknatt:
## 📄 Description

- Fixes the placement of the selected file value by updating `padding-inline-start` and `padding-inline-end` in SCSS of `form-input`.
- Adjusts the styling of the validation icon to not overlap the `::file-selector-button` by by adding `padding-inline-end`.


.changeset/twenty-toes-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Fixed picker input display issues of `type='file'` in `form-input` SCSS-file for **Form Creation Guidelines** on documentation.
```
leagrdv: There is no need to state in which file you make your changes. The changeset is for users of the design system and they don't need this information.
Also the changes you've done are for all input of type file, just because they're only documented in the **Form creation guidelines** does not mean you've made the changes for this only. Again, this is for users of the design system and they need to know what has been changed.

```suggestion
Fixed file input alignment and prevented validation icon from overlapping the selector button on valid/invalid states.
```

packages/styles/src/components/form-input.scss
```diff
@@ -149,11 +150,22 @@ input.form-control {
...
+        padding-inline-end: calc(
+          #{tokens.get('input-padding-inline-end')} + #{tokens.get(
+              'input-validation-icon-position-inline-end'
+            )} + #{tokens.get('input-sizing-icon')}
+        );
+      }
     }
   }
 
+  &:not(:disabled):not([readonly]) {
```
leagrdv: Looking at the indentation of this, it's still wrongly placed. This should be on the `type='file'` selector, right now it's on the `input.form-control` one.

packages/styles/src/components/form-input.scss
```diff
@@ -299,7 +311,8 @@ input.form-control {
     &[type='file'] {
       &:focus,
       &:not(:placeholder-shown) {
-        padding-block-start: calc(#{tokens.get('input-padding-block-start')} * 1.33);
+        padding-inline-start: tokens.get('input-label-empty-padding-inline-start');
```
leagrdv, line 314: How come the padding-block-start has changed? Was this needed?

bucknatt: Yes, we wanted them to have the same padding as in the input field above for `input.form-control ~ label` on line 253.
Before the padding by getting `'input-padding-block-start' * 1.33` was of value *33.915*. With the adapted padding from line 253, now both are of value *25.500*. (see screenshots)
image
image

PR #6199, opened 2025-09-10
feat(nextjs): add playright setup and basic ssr and csr tests

oliverschuerch:
## 📄 Description

This adds tests to ensure we add new components to the nextjs integration package and also checks if the components are throwing any hydration errors.

Currently the Github Workflow will not fail, even if errors occur during the test run.
Instead, annotations are added to the workflow summary, which show if there where errors or not.
This can be changed, as soon as all of our components are fully ssr compatible.

---

## 📝 Checklist


.changeset/proud-adults-bet.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': minor
+---
+
+Added a new component-names.json file, which contains a list of all components the package is outputting.
```
alizedebray, line 5: Can this file be exported from all component packages?
That way you could access it from `components-angular` and `components-react` without having to add `components` as a dev dependency.

oliverschuerch: Since we will probably never release the nextjs-integration package and components-names.json is really only used internally (I don't see why anyone else would need it) I felt it is not work the work to copy & paste it from components to components-angular and components-react.

Do you see any use case?

alizedebray: Not really, I just asked because I came across the same issue when trying to use the "not defined" styles. Since the file is not copied over, I had to add `components` as a dependency on top of the `components-angular`. 

It would be convenient to have a generic solution to copy additional files in all component packages but I agree here it's not mandatory. We can do it as an improvement in another ticket.

oliverschuerch: These specific styles should be implemented in the post-components.css file which is indeed copied over to components-angular and components-react and also ends up in the released packages. Did you see that?

alizedebray: Ah ok, I didn't see that!

packages/components/prebuild.ts
```diff
@@ -1,35 +1,70 @@
 import fs from 'fs';
 import path from 'path';
+import { format } from 'prettier';
 
-export function createComponentNamesScssMap() {
+interface IComponentNameOutputOptions {
```
alizedebray: Can you just rename the interface?
```suggestion
interface ComponentNameOutputOptions {
```

PR #6206, opened 2025-09-11
fix(docs): fix dialog markup on dialog text content and button intendation

bucknatt:
## 📄 Description

- Added `<p>`-tag around dialog content in stories, required for accessibility.
- Separated `<button>`-tags to be on two different lines.


packages/components/cypress/e2e/popovercontainer.cy.ts
alionazherdetska, line 1: Could you please revert the changes to this file? It's unrelated to this PR.

You were probably confused by the e2e test failing for `popovercontainer`:
Image

This failure isn't related to your PR. It's most likely an issue with the Cypress e2e test setup. This test has been failing intermittently on many PRs recently. You can just trigger a rerun and the test should pass. We're planning to migrate to Playwright tests since we've been having ongoing issues with Cypress.

.changeset/four-monkeys-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': patch
+---
+
+Added `<p>`-tag around dialog content in stories, required for accessibility.
```
alionazherdetska:
```suggestion
Improved `dialog` component accessibility by adding proper paragraph markup around text content.
```

alionazherdetska (changes requested):

alizedebray (changes requested):
The code snippet is still showing the two buttons on a single line: 
image

packages/documentation/src/stories/components/dialog/dialog.stories.ts
```diff
@@ -120,7 +120,7 @@ const getCloseButton = () => {
 };
 
 const getControls = () => {
-  return html`<button class="btn btn-primary">OK</button>
+  return html`<button class="btn btn-primary">OK</button><br />
     <button class="btn btn-secondary">Cancel</button>`;
```
alizedebray: The code should remain the same, we do not want to change the dialog component. 
The ticket is about code formatting, for a good indentation the button tags must be on two different lines:
```suggestion
  return html`
    <button class="btn btn-primary">OK</button>
    <button class="btn btn-secondary">Cancel</button>
  `;
```

PR #6208, opened 2025-09-11
chore(docs): bootstrap and ng-bootstrap removal in migration guide

leagrdv:
## 📄 Description

Updated the migration guide:

- Added information regarding the removal of bootstrap and ng-bootstrap in one place
- Cleaned up scattered informations regarding bootstrap removal in the rest of the page
- Removed all informations of removed elements that were never previously documented

## 📝 Checklist

### DIFF
--- a/packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
+++ b/packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
@@ -60,9 +60,9 @@ export class MigrationV99Component extends LitElement {
                 If you encounter any broken styles or issues after upgrading, you have two options:
               </p>
               <ul>
-                <li>Install Bootstrap as a dependency in your own project</li>
                 <li>Internalize the specific mixins or utilities you still rely on
                 </li>
+                <li>If issues persist, open a <a href="https://github.com/swisspost/design-system/issues">GitHub issue</a></li>
               </ul>
               <p>Please review your components and styles to ensure compatibility.
               </p><br/>


packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -48,6 +48,45 @@ export class MigrationV99Component extends LitElement {
...
+                As part of the latest migration, Bootstrap and Ng-Bootstrap have been fully removed from the design system. This means that any variables, classes, mixins, or utilities originating from Bootstrap and all components from Ng-Bootstrap are no longer available.
+              </p>
+              <p>
+                However, <span class="fw-bold">many commonly used features from Bootstrap — such as the grid system (columns) and most utility classes — have been internalized into the design system</span>. You can continue using them through the design system without needing Bootstrap.<br/>
+              </p>
+              <p>
+                If you encounter any broken styles or issues after upgrading, you have two options:
+              </p>
+              <ul>
+                <li>Install Bootstrap as a dependency in your own project</li>
```
alizedebray: Are we sure this is something we want to recommend? Since we’ve kept many Bootstrap class names, adding Bootstrap on top of the Design System styles could easily cause conflicts.

I’d rather recommend that users open a GitHub issue if they run into this situation.

leagrdv: I think we had discussed adding bootstrap internally with Philipp but I do agree with you that we'd probably run into more issues if projects were to add bootstrap as a library alongside ours. I've updated the text

alizedebray (approved):

PR #6209, opened 2025-09-11
feat(components): create post popover trigger component

myrta2302:
## 📄 Description

This PR introduces a `<post-popover-trigger>` web component to replace the former `data-popover-target=" "` implementation.
Aria attribute and focus management logic is now handeled within the trigger component.
Documentation and tests are updated accordingly.

## 📝 Checklist

- 🧪 I have udpated the existing tests
- Manual Accessibility Test

### DIFF
--- a/packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
+++ b/packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
@@ -75,6 +75,7 @@ export class MigrationV99Component extends LitElement {
       accordion_heading: false,
       hydrated_flag: false,
       accordion_item_part: false,
+      popover_trigger: false,
     },
   };
 
@@ -1278,6 +1279,22 @@ export class MigrationV99Component extends LitElement {
                           Update your styles to use <code>::part(button)</code> for the header
                           trigger and <code>::part(body)</code> for the content area instead.
                         </span>
+                      </label>
+                    </div>
+                  </li>
+                  <li class="mb-16">
+                    <div class="form-check">
+                      <input
+                        id="components.popover_trigger"
+                        class="form-check-input"
+                        type="checkbox"
+                        ?checked="${this.state.components.popover_trigger}"
+                      />
+                      <label class="form-check-label" for="components.popover_trigger">
+                        The <code>post-popover</code> now uses its own
+                        <code>post-popover-trigger</code> component instead of an element with a
+                        <code>data-popover-target</code> attribute.
+                        <span class="tag tag-sm tag-danger">breaking</span>
                       </label>
                     </div>
                   </li>
--- a/packages/documentation/src/stories/misc/migration-guide/types.ts
+++ b/packages/documentation/src/stories/misc/migration-guide/types.ts
@@ -148,6 +148,7 @@ export interface V910Checks {
     accordion_heading: boolean;
     hydrated_flag: boolean;
     accordion_item_part: boolean;
+    popover_trigger: boolean;
   };
 }
 


.changeset/forty-jokes-sin.md
```diff
@@ -0,0 +1,7 @@
+---
+'@swisspost/design-system-components-angular-workspace': patch
```
leagrdv: This package is private so there is no need to add it to the changeset

.changeset/forty-jokes-sin.md
```diff
@@ -0,0 +1,7 @@
...
+'@swisspost/design-system-documentation': patch
+'@swisspost/design-system-components': patch
```
leagrdv:
```suggestion
'@swisspost/design-system-components': major
```

This is a breaking change

packages/components-angular/projects/consumer-app/cypress/e2e/popover.cy.ts
leagrdv, line 1: There is already a test file for the popover in the `components` package. How come you've added one here?
I feel like it'd be clearer to have all popover tests within the same place.

packages/components-angular/projects/consumer-app/src/app/routes/popover/popover.component.html
```diff
@@ -0,0 +1,18 @@
+<h2>Popover</h2>
```
leagrdv: All web components are documented within the `home.component.html` page, the popover is already there and just needs to have trigger replaced with the new `post-popover-trigger`.

.changeset/forty-jokes-sin.md
leagrdv, line 1: Not related to this file but I see two remaining usage of `data-popover-target`, one in the angular consumer app, and one in the nextjs-integration package.

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
```diff
@@ -0,0 +1,126 @@
...
+        this.trigger.setAttribute('role', 'button');
+      }
+
+      // set aria attributes
+      this.trigger.setAttribute('ariahaspopup', 'true');
+
+      this.trigger.setAttribute('ariacontrols', this.for);
+
+      // add event listeners for click and keydown
+      this.trigger.addEventListener('click', () => {
```
leagrdv: You're adding the event listener on `slot change` which would add it every time the slot changes but you only remove it on `disconnectedCallback`

leagrdv: Also, because you're using an arrow function in the eventListener, `removeEventListener` will not actually remove it as it's intended to only remove the same reference (the way you did it for the `handleKeyDown` is the correct way.

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
```diff
@@ -0,0 +1,126 @@
...
+      this.trigger.setAttribute('aria-expanded', this.ariaExpanded.toString());
+
+      // check if its not focusable and add aria role and tabindex
+      if (!isFocusable(this.trigger)) {
+        this.trigger.setAttribute('tabindex', '0');
+        this.trigger.setAttribute('role', 'button');
+      }
+
+      // set aria attributes
+      this.trigger.setAttribute('ariahaspopup', 'true');
```
leagrdv:
```suggestion
      this.trigger.setAttribute('aria-haspopup', 'true');
```

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+      this.trigger.setAttribute('ariacontrols', this.for);
```
leagrdv:
```suggestion
      this.trigger.setAttribute('aria-controls', this.for);
```

packages/components/src/components/post-tooltip-trigger/post-tooltip-trigger.tsx
```diff
@@ -66,7 +66,6 @@ export class PostTooltipTrigger {
   }
 
   componentDidLoad() {
-    this.setupTrigger();
```
leagrdv, line 69: Why did you remove it here?
If the slot content doesn't change, it won't be triggered. Same thing in the `post-popover-trigger` component

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -156,13 +156,12 @@ export class PostPopovercontainer {
 
   /**
    * Programmatically display the popovercontainer
-   * @param target An element with [data-popover-target="id"] where the popovercontainer should be shown
+   * @param target A <post-popover-trigger> component that controls the popover
```
leagrdv: The target is not actually the `post-popover-trigger` (as it's also used for the tooltip trigger` but rather the focusable element within

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -7,10 +7,19 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
...
       // Aria-expanded is set by the web component, therefore it's a good measure to indicate the component is ready
-      cy.get('[data-popover-target="popover-one"][aria-expanded]').as('trigger');
+      cy.get('post-popover-trigger[data-hydrated][for="popover-one"]')
+        .children()
+        .first()
+        .as('trigger');
       cy.get('#testtext').as('popover');
     });
 
+    it('should contain an HTML element inside the trigger, not just plain text', () => {
```
leagrdv: Maybe you could more specifically check that there is focusable content within the trigger

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -7,10 +7,19 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
       cy.get('post-popover[data-hydrated]');
 
       // Aria-expanded is set by the web component, therefore it's a good measure to indicate the component is ready
-      cy.get('[data-popover-target="popover-one"][aria-expanded]').as('trigger');
+      cy.get('post-popover-trigger[data-hydrated][for="popover-one"]')
```
leagrdv, line 8: It would be good to also test:
- the `aria-expanded` does switch to `true` when clicked
- the first element within the popover gets focused when opened

.changeset/forty-jokes-sin.md
myrtasakellariou: Regarding the Angular app, i removed the separate popover page and updated the markup in home.html, as you suggest. For the nextjs-integration package, i ve included the markup change in the related ticket: https://github.com/swisspost/design-system/pull/6404/files

packages/components/src/components/post-tooltip-trigger/post-tooltip-trigger.tsx
myrtasakellariou: Just because i noticed that it is running twice on load.

packages/components-angular/projects/consumer-app/cypress/e2e/popover.cy.ts
myrtasakellariou: There is a task to add Angular functional tests, thats why i thought of adding them also in the Angular package, similar to the card-control. Not sure if they should test something else though.

packages/components/cypress/e2e/popover.cy.ts
myrtasakellariou: Added the check for the first element being focused on Enter

packages/components-angular/projects/consumer-app/cypress/e2e/popover.cy.ts
leagrdv: We have discussed that last week during the dev roundtable and here is the output of the discussion: https://github.com/swisspost/design-system/discussions/6462
So there is no need to add all of those specific tests on Angular which are already tested on stencil

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -7,8 +7,29 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
...
+
+      // Alternative popover trigger with no interactive element.
+      cy.get('post-popover-trigger[data-hydrated][for="popover-two"]')
+        .children()
+        .first()
+        .as('trigger2');
+    });
+
+    it('should contain an HTML element inside the trigger', () => {
+      cy.get('@trigger2').should('have.length.at.least', 1);
```
leagrdv: I'm not sure this test is necessary, as it only tests the implementation of a post-popover, and not the actual functionality of the popover itself.

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
```diff
@@ -0,0 +1,136 @@
...
+    // remove event listeners
+    this.trigger.removeEventListener('click', this.boundHandleToggle);
+    this.trigger.removeEventListener('keydown', this.boundHandleKeyDown);
+    this.popover.removeEventListener('postToggle', this.boundHandlePopoverPostToggle);
+  }
+
+  render() {
+    return (
+      <Host data-version={version}>
+        <slot onSlotchange={() => this.handleSlotChange()}></slot>
```
leagrdv:
```suggestion
        <slot onSlotchange={() => this.setUpTrigger()}></slot>
```

That way you can remove the extra function

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
```diff
@@ -0,0 +1,136 @@
...
+      }
+
+      // set aria attributes
+      this.trigger.setAttribute('aria-haspopup', 'true');
+      this.trigger.setAttribute('aria-controls', this.for);
+
+      // add event listeners
+      this.trigger.addEventListener('click', this.boundHandleToggle);
+      this.trigger.addEventListener('keydown', this.boundHandleKeyDown);
+      if (this.popover) {
```
leagrdv: Do you need to listen to the `postToggle` event here? `post-popover-trigger` is the one that defines whether or not the popover is opened, meaning you have that information in the component without having to listen to the event. This creates an unnecessary loop

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -173,25 +172,34 @@ export class PostPopovercontainer {
   @Method()
   async hide() {
     if (!this.toggleTimeoutId) {
+      if (this.eventTarget && this.eventTarget instanceof HTMLElement) {
```
leagrdv, line 243: Do you need to focus the first element here? This is the `hide()` method

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -203,30 +211,41 @@ export class PostPopovercontainer {
...
 
       // Emit event with `first` flag only true on the first open
       if (this.firstOpen) {
-        this.postToggle.emit({ isOpen, first: this.firstOpen });
+        this.postToggle.emit({ isOpen: true, first: true });
         this.firstOpen = false;
         return;
       }
     } else {
+      // Return focus to the trigger on close
```
leagrdv: When I close the popover, the focus does not go back to the trigger.

packages/components/cypress/e2e/popover.cy.ts
myrtasakellariou: Correct, i will remove it.

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
myrtasakellariou: OK. I added a popoverOpen state to the trigger instead.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
myrtasakellariou: It was supposed to focus back to the trigger (before i passed the whole post-popover-trigger). I updated it now to focus directly the target element.

packages/components-angular/projects/consumer-app/cypress/e2e/popover.cy.ts
leagrdv, line 1: Based on this PR from Alona: https://github.com/swisspost/design-system/pull/6471/files, you should be able to completely remove that file as there will be a single test file which will test the existence and visibility of all components there.

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -63,12 +79,16 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
...
-    it('should open and close on enter', () => {
+    it('should open on enter and first focusable element should be focused', () => {
       cy.get('@popover').should('not.be.visible');
       cy.get('@trigger').focus().type('{enter}');
       cy.get('@popover').should('be.visible');
-      cy.get('@trigger').type('{enter}');
-      cy.get('@popover').should('not.be.visible');
+
+      // find the first focusable element (e.g., button, input, link, etc.)
+      cy.get('@popover')
```
leagrdv, line 88: If I remove the link within the popover, the focus stays on the `popover-trigger` instead of going to the first focusable element which is then the close button

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
```diff
@@ -0,0 +1,138 @@
...
+      }
+
+      // Set aria attributes
+      this.trigger.setAttribute('aria-haspopup', 'true');
+      this.trigger.setAttribute('aria-controls', this.for);
+
+      this.trigger.addEventListener('click', this.boundHandleToggle);
+      this.trigger.addEventListener('keydown', this.boundHandleKeyDown);
+    } else {
+      console.warn(
```
leagrdv: This should be a `console.error` rather than a warning because no content means the popover will not work at all.

packages/components/cypress/e2e/popover.cy.ts
myrtasakellariou: added a getDeepFocusableChildren function to find focusable also accross Shadow

myrtasakellariou:
> Since it's a breaking change, could you add a line in the migration guide regarding the change from the data attribute to the new trigger component? And if you could add a line for the tooltip trigger as well, that'd be great!

Since post-tooltip-trigger is already in v9, i added the line for post-popover-trigger.

PR #6214, opened 2025-09-13
chore(icons): refactor post-icon mixin to dynamicallly load icons as css files

alionazherdetska:
## 📄 Description

## 🚀 Demo

---

## 📝 Checklist

### DIFF
--- a/packages/documentation/src/stories/foundations/icons/icon.styles.scss
+++ b/packages/documentation/src/stories/foundations/icons/icon.styles.scss
@@ -28,13 +28,13 @@ $resizer-max-width: 64px;
 }
 
 .my-1022-icon {
-  @include post.post-icon($name: '1022', $base: '/post-icons');
+  @include post.post-icon-load($name: '1022');
 }
 
 .my-accessibility-icon {
-  @include post.post-icon($name: 'accessibility', $base: '/post-icons');
+  @include post.post-icon-load($name: 'accessibility');
 }
 
 .my-handvictory-icon {
-  @include post.post-icon($name: 'handvictory', $base: '/post-icons', $color: #666666);
+  @include post.post-icon-load($name: 'handvictory', $color: #666666);
 }
--- a/packages/styles/src/mixins/_icons.scss
+++ b/packages/styles/src/mixins/_icons.scss
@@ -1,5 +1,6 @@
 @use 'sass:map';
 @use 'sass:meta';
+@use 'sass:list';
 
 @use './../variables/type';
 @use './../variables/color';
@@ -14,6 +15,12 @@
   background-color: currentColor;
   color: currentColor; // Required in this case with usage of forced-color-adjust: preserve-parent-color
   forced-color-adjust: preserve-parent-color;
+}
+
+@mixin load-icon-css($name, $path: '@swisspost/design-system-icons/dist/custom-properties') {
+  @at-root {
+    @include meta.load-css("#{$path}/#{$name}.css");
+  }
 }
 
 @mixin post-icon(
--- a/packages/styles/src/mixins/_icons.scss
+++ b/packages/styles/src/mixins/_icons.scss
@@ -9,6 +9,40 @@
 @use './../mixins/utilities';
 @use './../utilities/env-variables' as icon-version;
 
+$loaded-icons: () !default;
+
+@mixin load-icon-css($name, $path: '@swisspost/design-system-icons/dist/custom-properties') {
+  @if list.index($loaded-icons, $name) == null {
+    $loaded-icons: list.append($loaded-icons, $name) !global;
+    @at-root {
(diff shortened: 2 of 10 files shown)


packages/documentation/src/stories/foundations/icons/icon.styles.scss
```diff
@@ -28,13 +28,13 @@ $resizer-max-width: 64px;
...
 
 .my-accessibility-icon {
-  @include post.post-icon($name: 'accessibility', $base: '/post-icons');
+  @include post.post-icon($name: 'accessibility');
 }
 
 .my-handvictory-icon {
-  @include post.post-icon($name: 'handvictory', $base: '/post-icons', $color: #666666);
+  @include post.post-icon($name: 'handvictory', $color: #666666);
 }
```
oliverschuerch: Where is that even used?
If it's not used in any example, you could remove the styles.

packages/styles/gulpfile.js
```diff
@@ -30,6 +30,21 @@ gulp.task('temporarily-copy-token-files', () => {
...
+gulp.task('temporarily-copy-icon-files', () => {
+  return gulp.src(['../icons/dist/custom-properties/**/*.css']).pipe(gulp.dest('./src/icons/temp'));
+});
+
+/**
+ * Copy icon CSS files to dist folder
+ */
+gulp.task('copy-icon-files-to-dist', () => {
+  return gulp.src(['./src/icons/temp/**/*.css']).pipe(gulp.dest(`${options.outputDir}/icons/temp`));
+});
```
oliverschuerch, line 36: This should be executed, whenever a the copy task is executed.

Somehow like this:
```js
gulp.task('copy', () => {
  gulp.series('copy-icon-files-to-dist', () => {
    return gulp
      .src(['./LICENSE', './README.md', './package.json', './src/**/*.scss'])
      .pipe(newer(options.outputDir))
      .pipe(gulp.dest(options.outputDir));
  });
});
```

packages/styles/gulpfile.js
```diff
@@ -194,7 +209,8 @@ exports.default = gulp.task(
     'prebuild-env-vars',
     gulp.parallel(
       gulp.series('map-icons', 'copy', 'autoprefixer', 'transform-package-json'),
-      gulp.series('temporarily-copy-token-files', 'sass'),
+      gulp.series('temporarily-copy-token-files', 'temporarily-copy-icon-files', 'sass'),
     ),
+    'copy-icon-files-to-dist',
```
oliverschuerch: Only try this, if you run `copy-icon-files-to-dist` as a subtask of the `copy` task (mentioned above).
It would be good to ensure, we first copy the temp files (tokens/icons) to the ./src folder and then copy things from ./src to ./dist, before we finally build our sass files.

```js
exports.default = gulp.task(
  'build',
  gulp.series(
    'prebuild-env-vars',
    gulp.parallel(
      gulp.series('map-icons', 'autoprefixer', 'transform-package-json'),
      gulp.series('temporarily-copy-token-files', 'temporarily-copy-icon-files', 'copy', 'sass'),
    ),
  ),
);
```

packages/styles/src/mixins/_icons.scss
```diff
@@ -16,32 +17,37 @@
...
   /* stylelint-disable-next-line property-no-vendor-prefix */
   -webkit-mask-position: center center;
-  mask-position: center center;
+  mask-repeat: no-repeat;
   /* stylelint-disable-next-line property-no-vendor-prefix */
   -webkit-mask-repeat: no-repeat;
-  mask-repeat: no-repeat;
+  mask-size: 100%;
   /* stylelint-disable-next-line property-no-vendor-prefix */
   -webkit-mask-size: 100%;
```
oliverschuerch: Please read the answer, marked as correct: https://stackoverflow.com/questions/7080605/ordering-of-vendor-specific-css-declarations, then make sure for every property the one without vendor prefix comes last.

```scss
.class {
  -webkit-anyprop: value;
  -moz-anyprop: value;
  -ms-anyprop: value;
  anyprop: value;
}
```

oliverschuerch (changes requested):
There are some more icons we need to implement with the new/updated mixin in the styles package. Can you do that for all of them (if possible) or add information/issues, so we know how to proceed.

packages/styles/gulpfile.js
alionazherdetska: Made the changes you suggested! 
One thing though - I had to put `transform-package-json` and `autoprefixer` after the copy step instead of in parallel. 
Turns out they both depend on the dist folder being there first (`transform-package-json` writes to dist/package.json, `autoprefixer` reads from dist). 
Running them in parallel with `copy` just gave me  "file not found" errors because they'd fire off before the directory was even created.

oliverschuerch (approved):

PR #6215, opened 2025-09-15
fix(styles): dialog padding and grid areas

oliverschuerch:
## 📄 Description

This PR fixes the dialog padding, if no action buttons are present and also fixes the the the rendering of the dialog-header and dialog-body, if no icon is present.

---

## 📝 Checklist


.changeset/tangy-meals-make.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Fixed dialog grid-area implementation. `.dialog-header` and `.dialog-body` do now also span over the icon area, if none is present.
```
alizedebray: Good catch!
Can you update this further so that the content spans the whole width when there is no close button? Currently it is like this:
image

alizedebray (approved):

PR #6216, opened 2025-09-15
chore(components): update logic to show correct header active state

### LINKED ISSUE #6107: Header has no active state
When I navigate to a page using the header, there is no visual clue that help me know which link I clicked.
https://www.swisspost-cargo.com/ch-de

## Tasks
- [x] Document the selected state in the post-header documentation
- [x] Make the `post-megadropdown-trigger`'s selected state dependent on the presence of a `.selected` class within its megadropdown's sublink
- [x] Fix the `megadropdown-trigger` chevron's position to point down when it's selected but not opened

myrtasakellariou:
## 📄 Description

This PR:

1. Documents the `.selected` class in the `<post-header>` page in storybook.
2. Adds logic to automatically apply the `selected` class to the parent `post-megadropdown-trigger`when one of the items inside the corresponding `post-megadropdown` is marked as selected.
3. Removes the styles that affected the chevron transform(rotate) when item is on `.selected` state.

## 📝 Checklist


.changeset/clever-fans-travel.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': patch
+'@swisspost/design-system-components': patch
+---
+
+Added logic to automatically apply the `selected` class to the respective `post-megadropdown-trigger` when one of the items inside the corresponding `post-megadropdown` is marked as selected.
```
leagrdv, line 5: I think it'd be good to have a separate changeset for the documentation stating the added documentation of the header's selected state. What do you think?

packages/documentation/src/stories/components/header/header.docs.mdx
```diff
@@ -18,6 +20,15 @@ import StylesPackageImport from '@/shared/styles-package-import.mdx';
   <Controls of={HeaderStories.Default} />
 </div>
 
+## `.selected` Class
```
leagrdv: I think it'd be more clear for users to have something like "Selected navigation item" or "Active navigation item" as a title here. 
Also, do you think we should set a "selected" item in the first header preview (maybe one in the megadropdown)? That way, users can see what it looks like and if they copy/paste the HTML, they will see this `selected` state and remember that they need to handle it (in case they don't read the whole doc).

leagrdv (changes requested):

leagrdv (approved):

Issue #5998, opened 2025-08-14
Replace Post Icons with UI Icons in Styles & Components


alizedebray:
All components should consistently use **UI** icons instead of **Post** icons.  
This task involves searching for and replacing all 4-digit Post icons with their corresponding UI icon equivalents within the **`styles`** and **`components`** packages.

### 🔍 How to Perform the Replacement:

1. **Search for Post icons in the codebase**  
   Use the following regular expressions:
   - In HTML files: `name="\d{4}"`
   - In SCSS files: `icon(\d{4})`

2. **Identify the current icon**  
   When you find something like `name="2045"` or `icon(1023)`:
   - Go to the **"Post Icons"** [this reference page](https://next.design-system.post.ch/?path=/docs/0dcfe3c0-bfc0-4107-b43b-7e9d825b805f--docs)
   - Enter the 4-digit code to preview the current Post icon

3. **Find the UI icon replacement**  
   - On the same page, switch to the **"UI Icons (line)"**
   - Search by keyword to find the closest matching UI icon

4. **Update the code**  
   - Replace the 4-digit Post icon code with the **name** of the selected UI icon

Issue #5960, opened 2025-08-07
Remove `post-icon` mixin from styles package


oliverschuerch:
**UserStory**
As a Developer of the `@swisspost/design-system-styles` package, I want to get rid of the unneccessary mixin `post-icon`, so we end up having again only one icon mixin, which can be used for both our icon types (simplel Post icons and UI icons).

**Acceptance Criteria**
- [ ] Replace all of it's usages in all the internal packages with the `icon` mixin.
- [ ] Remove the mixin.
- [ ] Update the [icon documentation](https://next.design-system.post.ch/?path=/docs/0dcfe3c0-bfc0-4107-b43b-7e9d825b805f--docs).
- [ ] Create a breaking change changeset, where you also explain, that instead the `icon` mixin can be used instead.

PR #6217, opened 2025-09-15
chore(components, styles): replace post-icons with ui-icons in styles and components packages

alionazherdetska:
## 📄 Description

## 🚀 Demo

---

## 📝 Checklist

### DIFF
--- a/packages/styles/src/components/card.scss
+++ b/packages/styles/src/components/card.scss
@@ -108,11 +108,8 @@ post-linkarea:has(> .card) {
     -ms-flex-align: center;
 
     &::before {
-      @include icons-mx.icon(3020);
       content: '';
-      display: inline-block;
-      width: icons.$icon-size-default;
-      height: icons.$icon-size-default;
+      @include icons-mx.post-icon-load('arrowright', $width: icons.$icon-size-default, $height: icons.$icon-size-default);
     }
   }
 }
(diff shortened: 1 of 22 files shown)


alizedebray (changes requested):
Before this PR can be merged it looks like the icon mixins need further refactoring to avoid unnecessary changes: 
- The `post-icon` mixin should be **completely removed** in favor of the `icon` mixin
- The `icon` mixin should use the `load-icon-css` mixin instead of the `get-svg-url` function [here](https://github.com/swisspost/design-system/blob/3b1de9d41c8df4fcfcaeb1f64f6f515be1a5f8aa/packages/styles/src/mixins/_icons.scss#L13)
- Optionally, the `load-icon-css` should be renamed to `load-icon-property` since it's what it actually does

packages/styles/src/components/card.scss
```diff
@@ -108,11 +108,12 @@ post-linkarea:has(> .card) {
...
       content: '';
-      display: inline-block;
-      width: icons.$icon-size-default;
-      height: icons.$icon-size-default;
+      @include icons-mx.post-icon(
+        'arrowright',
+        $width: icons.$icon-size-default,
+        $height: icons.$icon-size-default
+      );
     }
```
alizedebray, line 112: The `icon` mixin has no `$color`, `$height`, or `$width` arguments nor it is setting a `display` property. 
Therefore when the `post-icon` is removed in favor of the `icon` mixin with #5960, this changes will have to be reverted back to: 
```scss
    &::before {
      @include icons-mx.icon('arrowright');
      content: '';
      display: inline-block;
      width: icons.$icon-size-default;
      height: icons.$icon-size-default;
    }
```
Thus my comment above. I did not know #5960 existed, but it would be good to tackle it before #5998  otherwise you will have to open a new PR to change everything back.

alizedebray (approved):

PR #6220, opened 2025-09-15
chore(components): update icon path logic

myrta2302:
## 📄 Description

This PR updates the logic of the icon path construction according to the following path prioritization:

1. `@base` (absolute URL) → use directly.  
2. `@base` (relative URL) → resolve with `base href` and/or `origin`  
   - If `base href` is absolute → use just that.  
   - If `base href` is relative → prepend with `origin`.  
   - If `base href` does not exist → use only `origin`.  
3. `@meta` (absolute URL) → use directly.  
4. `@meta` (relative URL) → resolve with `base href` and/or `origin`. (same as above)
5. `CDN_URL` fallback → `https://unpkg.com/...`.

## 📝 Checklist

### DIFF
--- a/.changeset/clever-fans-travel.md
+++ b/.changeset/clever-fans-travel.md
@@ -1,5 +1,6 @@
 ---
 '@swisspost/design-system-components': patch
+'@swisspost/design-system-styles': patch
 ---
 
-Added logic to automatically apply the `selected` class to the respective `post-megadropdown-trigger` when one of the items inside the corresponding `post-megadropdown` is marked as selected.
+Added styles to highlight the current navigation item marked with `aria-current="page"` and its potential parent item in the header main navigation.
--- a/.changeset/good-rocks-double.md
+++ b/.changeset/good-rocks-double.md
@@ -2,4 +2,4 @@
 '@swisspost/design-system-documentation': patch
 ---
 
-Added documentation on styling active navigation items in `<post-header>` using the `.selected` class.
+Added documentation on styling active navigation items in `<post-header>` using the `[aria-current="page"]` attribute.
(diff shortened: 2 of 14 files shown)


packages/components/cypress/e2e/icon.cy.ts
```diff
@@ -192,30 +215,36 @@ describe('Icon', () => {
       );
     });
 
-    it('should use CDN fallback if no paths are specified', () => {
+    // When not @base prop or meta icon path are specified, cdn fallback should be used
```
leagrdv: I feel like comments on top of the tests are sometimes redundant, as the test description are describing the content of the tests already. I think I'd go with updating the test description instead when needed.

packages/components/src/components/post-icon/post-icon.tsx
```diff
@@ -64,7 +64,7 @@ export class PostIcon {
 
   @Watch('name')
   validateName() {
-    checkRequiredAndType(this, 'name', 'string');
+    if (this.name) checkRequiredAndType(this, 'name', 'string');
```
leagrdv: Doesn't adding that if statement prevent to actually check and throw the error if the name is not there?

myrtasakellariou:
> From what I see, the way the icon path was calculated was actually already correct (the issue was therefore only in v9) and you simply refactored the code to make it easier to understand, right? Then I would update the changeset accordingly stating that you refactored the function rather than changed the logic.

Should i maybe not add a changeset at all?

myrtasakellariou:
> > From what I see, the way the icon path was calculated was actually already correct (the issue was therefore only in v9) and you simply refactored the code to make it easier to understand, right? Then I would update the changeset accordingly stating that you refactored the function rather than changed the logic.
> 
> Should i maybe not add a changeset at all?

ok i updated it.

PR #6225, opened 2025-09-16
feat(eslint): automigration rules to update sizing utilities and common test creator

leagrdv:
## 📄 Description

Added automigration rules for the sizing utilities


alizedebray (approved):

packages/eslint/src/rules/html/migrations/index.ts
```diff
@@ -26,6 +33,8 @@ export const htmlMigrationRules = {
   [noUnnumberedBorderRadiusRuleName]: noUnnumberedBorderRadiusRule,
   [noDeprecatedSpacingUtilitiesRulePhase1Name]: noDeprecatedSpacingUtilitiesRulePhase1,
   [noDeprecatedSpacingUtilitiesRulePhase2Name]: noDeprecatedSpacingUtilitiesRulePhase2,
+  [noDeprecatedSizingUtilitiesRulePhase1Name]: noDeprecatedSizingUtilitiesRulePhase1,
+  [noDeprecatedSizingUtilitiesRulePhase2Name]: noDeprecatedSizingUtilitiesRulePhase2,
```
alizedebray, line 37: Did you check that the rules are applied in the expected order?

leagrdv: Yes, I've tried to add some old classes on the components-angular consumer-app project and running the migration file works as expected!

PR #6226, opened 2025-09-16
chore(docs): replace post-icons with ui-icons in docs

alionazherdetska:
## 📄 Description
This PR replaces all usages of `Post Icons` with the corresponding `UI Icons` across the documentation package (`designers request`).

## Changes

1. Updated Storybook preview styles to reference `UI Icon` names instead of `Post Icon` numbers.
2. Replaced `Post Icons` with `UI Icons` across all component stories.
3. Adjusted argTypes and control options in stories to use named `UI Icons`.

**Exceptions:**
Post icons are retained in `mission.blocks.tsx`, `design-principles.data.ts`, and `introduction.docs.md` because they offer superior semantic clarity for the intended context, with no equivalent alternatives available in the current UI icon set.

**ℹ️ Note on social media icon sizes**
Since this PR replaces Post Icons with UI Icons, and UI Icons are designed with larger internal padding, the icons in the composable footer appeared too small when swapped directly.

To address this, I manually increased the icon sizes for the social media icons inside the footer (agreed with Rouven and Oli):
- These values are not tokenized, because they are footer-specific.
- The adjustment only applies to the icons rendered in the slot="socialmedia" area of the composable footer.

---

## 📝 Checklist


packages/documentation/src/stories/foundations/icons/icon.stories.ts
```diff
@@ -254,7 +254,7 @@ export const CSS_Color: Story = {
 
 export const CSS_Size: Story = {
   render: () => {
-    return html`<div class="my-1022-icon" style="font-size: 3rem"></div>`;
+    return html`<div class="letter-icon" style="font-size: 3rem"></div>`;
```
leagrdv, line 257: With the new icons, there is no more space between the three stamp size (in usage as web components). Could you add back some spacing?

packages/styles/src/components/footer/_post-footer.scss
```diff
@@ -83,6 +83,21 @@ post-footer {
 
     &[slot='socialmedia'] > [role='list'] {
       gap: var(--post-footer-socialmedia-list-item-gap);
+
+      post-list-item a post-icon:not([name='letter']) {
```
leagrdv, line 87: Why all icons but not the letter?

alionazherdetska: image
image
The reason is that the `letter` does not have the big padding around that is why we do not need to increase the size of it.

PR #6237, opened 2025-09-17
feat(eslint): automigration rules to update gutter utilities

leagrdv:
## 📄 Description

Automigration rules for the gutter utility + two phase rules refactoring


packages/eslint/src/utils/two-phases-classes-update.ts
```diff
@@ -73,3 +77,36 @@ export function setUpClassesMutations(
 
   return returnData;
 }
+
+export function createTwoPhasesRules(
+  data: TwoPhasesData,
+  name: string,
+  descriptionPhase1: string,
+  descriptionPhase2: string,
+): {
```
alizedebray: Can you change this function signature to something like this: 
```suggestion
type RuleType = Rule.RuleMetaData['type'];
type RuleMessages = Rule.RuleMetaData['messages'];

interface RuleConfigBase {
  name: string;
  type?: RuleType;
}

interface PhaseConfig<T>{
  description: string;
  messages: T;
  mutations: Record<keyof T, [string] | [string, string]>;
}

interface TwoPhaseRuleConfig<T, U> extends RuleConfigBase {
  phases: [PhaseConfig<T>, PhaseConfig<U>];
}

const createTwoPhasesClassUpdateRule = <T extends RuleMessages, U extends RuleMessages>(
  config: TwoPhaseRuleConfig<T, U>
): {
```

This way, descriptions are included into the phase data and we ensure message keys and mutation keys are always matching. 

Then the `RuleConfig` interface could be turned into: 
```ts
type SinglePhaseRuleConfig<T> = RuleConfigBase & PhaseConfig<T>;
```

alizedebray (approved):

PR #6244, opened 2025-09-18
feat(styles): update one-pager header to work without a title

alizedebray:
## 📄 Description

This request aims to completely hide the local header in the one-pager header configuration.

## 🚀 Demo

https://preview-6244--swisspost-design-system-next.netlify.app/?path=/story/27a2e64d-55ba-492d-ab79-5f7c5e818498--default&args=title:;mainNavigation:!false;metaNavigation:!false;customControls:!false

---

## 📝 Checklist


packages/styles/src/components/header/_post-header.scss
```diff
@@ -67,6 +67,10 @@ tokens.$default-map: elements.$post-link;
       --post-global-header-expanded-height: 4.5rem;
       --post-local-header-expanded-min-height: 4rem;
     }
+
+    &:not(:has(post-header [slot='title'])) {
```
myrtasakellariou: Not sure if it is possible, but it would be nice to keep the spacing for the case of no title present but of local controls being set e.g.  `&:not(:has(post-header [slot='title'])):not(:has(post-header [slot='local-controls'])) ... `. (Not sure though how/if this can be done for the default slot.)

alizedebray: The expected behavior is to prevent having local controls without a local title.
This is how it currently works with the micro-site header as well: https://swisspost-design-system-next.netlify.app/?path=/docs/27a2e64d-55ba-492d-ab79-5f7c5e818498--docs&story=Default&args=title:#27a2e64d-55ba-492d-ab79-5f7c5e818498--default

PR #6249, opened 2025-09-19
feat(eslint): automigration rules for the loader

leagrdv:
## 📄 Description

Updated automigration for the loader to add the renaming from `loader` to `spinner`

### DIFF
--- a/packages/eslint/src/rules/html/migrations/no-deprecated-loader.ts
+++ b/packages/eslint/src/rules/html/migrations/no-deprecated-loader.ts
@@ -1,56 +1,29 @@
-import { createRule } from '../../../utils/create-rule';
-import { HtmlNode } from '../../../parsers/html/html-node';
+import { createClassUpdateRule } from '../../../utils/create-class-update-rule';
+import { generateReplacedClassMutations } from '../../../utils/generate-mutations';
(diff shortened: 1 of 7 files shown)


packages/eslint/src/rules/html/migrations/no-deprecated-loader.ts
```diff
@@ -1,56 +1,29 @@
...
-              const fixedNode = $node.removeClass('loader-xs').addClass('loader-16');
-              return fixer.replaceTextRange(node.range, fixedNode.toString());
-            },
-          });
-        }
+const classesMap = [
+  { old: 'loader', new: 'spinner' },
+  { old: 'loading-modal', new: 'spinner-modal' },
+  { old: 'loader-xs', new: 'spinner-16' },
+  { old: 'loader-sm', new: 'spinner-40' },
```
alizedebray, line 11: There is already a rule that turns `loader-[xs|sm]` into `spinner-16|40` : https://github.com/swisspost/design-system/blob/main/packages/eslint/src/rules/html/migrations/no-deprecated-loader.ts
It would act as a "Phase 1".

alizedebray (changes requested):

alizedebray (approved):

PR #6251, opened 2025-09-19
feat(docs): add auto migration filter on migration guide

leagrdv:
## 📄 Description

Add a filter to hide all lines that are handled by auto migration rules


packages/documentation/src/stories/misc/migration-guide/setup.component.ts
```diff
@@ -51,6 +52,22 @@ export class SetupComponent extends LitElement {
               your application.
             </div>
           </div>
+          <div class="col-12">
```
alizedebray: This checkbox would fit better under the "Component Migration" of the "Migration from v9 to v10" section. The main filters are more about project setup, and at this point it’s not clear what "automatic migration" means.

Also, if the checkbox stays here, the filter should apply to all migration sections (e.g. "Migration from v4 to v5").

alizedebray (changes requested):

packages/documentation/src/stories/misc/migration-guide/setup.component.ts
leagrdv: I've moved the checkbox under the Component migration and I also updated a bit the style and placement of the tag following Sandra's inputs.

alizedebray (approved):

PR #6270, opened 2025-09-22
chore(docs): add header usage examples

alizedebray:
## 📄 Description

Added the same usage example for the header than the ones in Figma.

## 🚀 Demo

https://preview-6270--swisspost-design-system-next.netlify.app/?path=/docs/27a2e64d-55ba-492d-ab79-5f7c5e818498--docs

---

## 📝 Checklist

### DIFF
--- a/packages/documentation/src/stories/components/header/header.stories.ts
+++ b/packages/documentation/src/stories/components/header/header.stories.ts
@@ -46,7 +46,8 @@ const meta: MetaComponent = {
     },
     metaNavigation: {
       name: 'Meta navigation',
-      description: 'Whether or not the meta navigation is displayed ("about us" and "jobs").',
+      description:
+        'Whether or not the meta navigation is displayed ("jobs" and "create an account").',
       control: {
         type: 'boolean',
       },
@@ -94,7 +95,7 @@ function getHeaderRenderer(mainnavigation = renderMainnavigation()) {
           <!-- Meta navigation -->
           <ul class="list-inline" slot="meta-navigation">
             <li><a href="">Jobs</a></li>
-            <li><a href="">About us</a></li>
+            <li><a href="">Create an account</a></li>
           </ul>
         `
       : ''}
@@ -134,9 +135,6 @@ function getHeaderRenderer(mainnavigation = renderMainnavigation()) {
             </li>
             <li>
               <a href="#">Business customers</a>
-            </li>
-            <li>
-              <a href="#">Authorities</a>
             </li>
           </ul>
         `


packages/documentation/src/stories/components/header/header.docs.mdx
```diff
@@ -51,3 +60,25 @@ The `<post-header>` element is part of the `@swisspost/design-system-components`
...
 
 <StylesPackageImport components={['header', 'list', 'button']} />
+
+## Examples
+
+### Post Portal
+
+The Header for the Post Portal (post.ch) is a consistently available interface element that provides a consistent navigation experience across the whole Post Portal and other portals.
+
+It not only contains globally available key elements (such as Search, Login or Language), it also provides a first content selection via the target group buttons. They define the further navigation.
```
leagrdv: The text here states that this header contains search, login and language but the example only shows the language switch

packages/documentation/src/stories/components/header/header.stories.ts
```diff
@@ -15,13 +15,19 @@ const meta: MetaComponent = {
       type: 'figma',
       url: 'https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations-%26-Components-Next-Level?node-id=558-7012&t=ywmfJhyvd2euoiGI-1',
     },
+    docs: {
+      story: {
+        inline: false,
+        iframeHeight: 300
```
leagrdv: I think blocking the height to 300px might not be sufficient as it results in two scrollbars (even 3 with the current bug) and you can barely see the megadropdown when it's opened. With all the new headers, the page might look a bit long if we increase that but maybe we could at least have the first main story bigger and then keep it fixed at 300 for the others. What do you think?
Capture d'écran 2025-09-24 105014

leagrdv (changes requested):

packages/documentation/src/stories/components/header/header.stories.ts
```diff
@@ -163,120 +159,187 @@ const Template = {
...
+          <!-- The active link must have an aria-current="page" attribute to ensure correct accessibility and styling. -->
+          <a href="/packages" class="selected">Packages</a>
+        </post-list-item>
+      </post-list>
+    </post-mainnavigation>
+  `
+};
+
+export const Portal: Story = {
+  ...getIframeParameters(500),
```
leagrdv:
```suggestion
  ...getIframeParameters(550),
```

Adding 50 pixels prevents the scrolbar in the megadropdown

PR #6402, opened 2025-10-07
feat(components, docs): add `user menu` to the `post-header`

leagrdv:
## 📄 Description

- Added the user menu to the `post-header` component
- Added the option in storybook to have the login within the custom controls for the future but disabled it for now

## 🚀 Demo

https://preview-6402--swisspost-design-system-next.netlify.app/iframe.html?args=isLoggedIn%3A!true&globals=&viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--default

---

## 🔮 Design review

- [x] Design review done
- [ ] No design review needed

## 📝 Checklist


alizedebray (changes requested):
The menu does not fully match the [design](https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-Next-Level?node-id=7476-66742&m=dev): 
- [x] It header height must be 72px/64px/56px
- [x] The header font weight must be 700
- [x] The avatar must be 36px
- [x] The padding of the menu link are incorrect (probably wrong on the post-menu itself)
- [x] The gap in the menu links must be 8px on mobile
- [x] There is not hover styles (probably wrong on the post-menu itself)
- [x] The logout button is round so it looks strange when focused

In the global header , the avatar and login buttons are also missing some styles: 
- [x] Hover and focus styles are not correct for the avatar
- [ ] The login button must be 32px on desktop and tablet and 44px on mobile
- [ ] The login button label must be hidden on mobile (we previously used the `.visually-hidden-sm` class but it would be better to actually add this behavior to the header styles so that it impact all global header links)

leagrdv:
_The login button must be 32px on desktop and tablet and 44px on mobile:_
On desktop, the padding-block is 3px and padding-inline is 10px, as defined on Figma. It's the same for all the links in the global header. Somehow, they've added a transparent 1px border around it which is why the height is incorrect. Should I align all elements to switch to `4px 10px` or do we want to keep the padding accurate? I feel like adding an unnecessary 1px border will just add more code for no real added value.

 _The avatar must be 36px:_
It currently is 36px, it shows as 40px because of the 2px border (and calculates the content as 36.8px because Windows decided that 1.6px border is better than 2px 🥲). 

_The login button label must be hidden on mobile (we previously used the .visually-hidden-sm class but it would be better to actually add this behavior to the header styles so that it impact all global header links):_
I've removed the `visuallyhidden-sm` class but had to keep the text within a span to be able to target it.

The rest of the comments have been fixed!

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -398,6 +399,7 @@ export class PostHeader {
             )}
             <slot name="global-controls"></slot>
             {!this.hasMobileMenu && <slot name="post-language-switch"></slot>}
+            <slot name="user"></slot>
```
alizedebray: To align this with the changes from #6422:
```suggestion
            <slot name="global-login"></slot>
```

packages/components/src/components/post-menu-item/post-menu-item.scss
alizedebray, line 1: Can you please add a changeset for this, the `post-menu` itself is not documented but you changes will impact multiple other components like the language switch and breadcrumbs.

packages/components/src/components/post-menu/post-menu.scss
```diff
@@ -1,11 +1,30 @@
...
+  padding-block: 0.5rem;
+}
+
+::slotted([slot='header']) {
+  margin-top: -8px;
+  padding: 16px;
+  display: flex;
+  align-items: center;
+  gap: 8px;
+  border-bottom: 1px solid #050400;
```
alizedebray:
```suggestion
  border-block-end: 1px solid #050400;
```

packages/documentation/src/stories/components/header/header.stories.ts
```diff
@@ -340,3 +418,30 @@ export const WithTargetGroup: Story = {
...
 };
+
+// User is logged in
+export const LoggedIn: Story = {
+  ...getIframeParameters(400),
+  args: {
+    isLoggedIn: true,
+    customControls: true,
+    mainNavigation: true,
+    metaNavigation: false,
```
alizedebray:
```suggestion
    isLoggedIn: true,
    loginInGlobalHeader: true,
```

packages/nextjs-integration/src/app/ssr/layout.tsx
alizedebray, line 1: So many headers!

packages/styles/src/components/avatar.scss
alizedebray, line 1: Can you add a changeset for this change as well?

packages/documentation/src/stories/components/header/header.stories.ts
```diff
@@ -75,6 +77,27 @@ const meta: MetaComponent = {
...
     },
+    isLoggedIn: {
+      name: 'Is logged in',
+      description: 'Whether the user is logged in or not.',
+      control: {
+        type: 'boolean',
+      },
+      table: { category: 'state' },
+    },
+    userMenuLocation: {
```
alizedebray: This is probably not needed:
- If there is no title the login is in the global header
- If there is a title and custom controls the login is in the local header
- If there is a title and no custom control there is no login button

packages/components/src/components/post-menu-item/post-menu-item.scss
leagrdv: I've added the changeset but I did not mention the language switch as it has style that overrides it anyway.

packages/nextjs-integration/src/app/ssr/layout.tsx
leagrdv: 🤯

packages/documentation/src/stories/components/header/header.stories.ts
leagrdv: I removed the customControls, mainNavigation and metaNavigation but I did not add the `loginInGlobalHeader` as it's not an argument, but the login does appear in the global header as there is no title

PR #6404, opened 2025-10-07
chore(nextjs-integration): add react tests for `<post-popover>`

myrta2302:
## 📄 Description

This PR adds Playwright tests for the `<post-popover>` component in the `nextjs-integration` package.

## 📝 Checklist


oliverschuerch:
In general, we should not add react/angular tests to specific components. At least not if they are generated one-to-one from the stencil component, without any framework specific specialities.

**Where to test what...**

**Test in the `components` package**
1. It the general component input/output/functionality/accessibility (props, events, methods, user interactions, etc.) works as expected.

**Test in the consumer app in the `components-angular` and/or the `nextjs-integration` packages**
1. If there is a code implementation for every component (defined in the `components` package file `component-names.json`).
2. If every component is properly attached to the DOM and renders without console errors (hydration errors are tested already in some general implemented tests).
⚠️ Not all of our components are visible by default, so we can't test `isVisible`.
3. If custom implementations for specific components (e.g. the valueAccessors for the `card-control`, which are defined in the `components-angular` package only) work as expected. 

ℹ️ Why should we not test functionality in the framework consumer apps again?
Because we test it already in the `components` package and since the angular and react components are just wrapper components around our web-components, it's stencils job to ensure/test that a framework specific wrapper component works properly with the underlaying web-component.

Btw. @alionazherdetska is working on something similar. Please contact her to not do anything twice!

Issue #6411, opened 2025-10-08
[dev] Add slots on the right to main navigation


alizedebray:
The portal header must be adapted for the Jobs application.  

<img width="2912" height="288" alt="Image" src="https://github.com/user-attachments/assets/3e0829bc-11ab-481c-aa5e-9ba8f668f31d" />

In this version of the header:  
- The global header only contains the **Jobs** link (active) and the language switch.  
- The main navigation includes a **Search** and **Login** link positioned on the far right.  

### ✅ What to do
- Create a new example in the header documentation for the **Jobs header**.  
- Add a slot to place content on the right side of the main navigation (may be inside or outside the navigation container).  
- Ensure links in this new slot are displayed at the very top of the [mobile navigation](https://www.figma.com/design/VBcGcYKSzNjt1mSMy2LTUO/PostOneWeb-2025?node-id=4548-7913&t=CdQw0AbqOR2S35mg-4).  
- Verify that the main navigation overflow logic still works correctly and that styles remain consistent (e2e tests for overflow already exist).

PR #6421, opened 2025-10-10
feat(components, docs): add slot and document header variant for jobs pages

myrta2302:
## 📄 Description

This PR:

1. Adds a `slot` for auxiliary navigation links on the right side of `post-mainnavigation`.
2. Adds a documentation example for the Jobs portal header.

## 🚀 Demo

https://preview-6421--swisspost-design-system-next.netlify.app/?path=/docs/27a2e64d-55ba-492d-ab79-5f7c5e818498--docs#jobs

## 🔮 Design review

- [x] Design review done
- [ ] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/nextjs-integration/package.json
+++ b/packages/nextjs-integration/package.json
@@ -27,7 +27,7 @@
   "dependencies": {
     "@swisspost/design-system-components-react": "workspace:10.0.0-next.49",
     "@swisspost/design-system-styles": "workspace:10.0.0-next.49",
-    "next": "15.3.3",
+    "next": "15.4.7",
     "react": "19.1.1",
     "react-dom": "19.1.1"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -309,7 +309,7 @@ importers:
         version: 0.8.2(@stencil/core@4.35.0)(@types/react@18.3.23)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
       '@stencil/ssr':
         specifier: 0.1.1
-        version: 0.1.1(next@15.3.3(@playwright/test@1.55.0)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)(sass@1.89.2))(vite@7.0.7(@types/node@22.10.5)(jiti@2.4.2)(less@4.3.0)(lightningcss@1.30.1)(sass-embedded@1.89.2)(sass@1.89.2)(terser@5.43.1)(yaml@2.7.0))(webpack@5.99.9)
+        version: 0.1.1(next@15.4.7(@playwright/test@1.55.0)(react-dom@18.3.1(react@18.3.1))(react@18.3.1)(sass@1.89.2))(vite@7.0.7(@types/node@22.10.5)(jiti@2.4.2)(less@4.3.0)(lightningcss@1.30.1)(sass-embedded@1.89.2)(sass@1.89.2)(terser@5.43.1)(yaml@2.7.0))(webpack@5.99.9)
       '@swisspost/design-system-components':
         specifier: workspace:10.0.0-next.49
         version: link:../components
@@ -777,8 +777,8 @@ importers:
         specifier: workspace:10.0.0-next.49
         version: link:../styles/dist
       next:
-        specifier: 15.3.3
-        version: 15.3.3(@playwright/test@1.55.0)(react-dom@19.1.1(react@19.1.1))(react@19.1.1)(sass@1.89.2)
+        specifier: 15.4.7
+        version: 15.4.7(@playwright/test@1.55.0)(react-dom@19.1.1(react@19.1.1))(react@19.1.1)(sass@1.89.2)
       react:
         specifier: 19.1.1
         version: 19.1.1
@@ -2452,8 +2452,8 @@ packages:
   '@dual-bundle/import-meta-resolve@4.1.0':
     resolution: {integrity: sha512-+nxncfwHM5SgAtrVzgpzJOI1ol0PkumhVo469KCf9lUi21IGcY90G98VuHm9VRrUypmAzawAHO9bs6hqeADaVg==}
 
-  '@emnapi/runtime@1.4.3':
-    resolution: {integrity: sha512-pBPWdu6MLKROBX05wSNKcNb++m5Er+KQ9QkB+WVM+pW2Kx9hoSrVTnu3BdkI5eBLZoKu/J6mW/B6i6bJB2ytXQ==}
+  '@emnapi/runtime@1.5.0':
+    resolution: {integrity: sha512-97/BJ3iXHww3djw6hYIfErCZFee7qCtrneuLa20UXFCOTCfBM2cvQHjWJ2EG0s0MtdNwInarqCTz35i4wWXHsQ==}
 
   '@esbuild/aix-ppc64@0.25.2':
     resolution: {integrity: sha512-wCIboOL2yXZym2cgm6mlA742s9QeJ8DjGVaL39dLN4rRwrOgOyYSnOaFPhKZGLb2ngj4EyfAFjsNJwPXZvseag==}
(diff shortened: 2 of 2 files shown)


packages/components/src/components/post-header/post-header.tsx
```diff
@@ -358,12 +387,34 @@ export class PostHeader {
         style={{ '--post-header-navigation-current-inset': `${mobileMenuScrollTop}px` }}
       >
         <div class="mobile-menu" ref={el => (this.mobileMenu = el)}>
+          {(this.device === 'mobile' || this.device === 'tablet') && this.auxNavLinks && (
+            <div class="mobile-aux-nav-links">
```
alizedebray: Can you align naming and use `sub` instead of `aux`?

packages/documentation/src/stories/components/header/header.stories.ts
```diff
@@ -86,96 +99,128 @@ const meta: MetaComponent = {
...
+            <ul class="list-inline" slot="meta-navigation">
+              <li>
+                <a class="selected" href="">
+                  Jobs
+                  <post-icon name="jobs" aria-hidden="true"></post-icon>
+                </a>
+              </li>
+            </ul>
+          `
+        : nothing}
```
alizedebray: Can you please remove this? 
The meta navigation should remain, only the search link must be removed.

alizedebray:
There are now two "Jobs" buttons and the "Jobs Search" and "Jobs Login" are gone:
image

packages/components/cypress/e2e/header.cy.ts
oliverschuerch, line 1: This change was not requested in the [issue ticket](https://github.com/swisspost/design-system/issues/6411) and should (if even) not be part of this PR.
Everything related to this change should be reverted or splitted into a different PR.

If we really want to do that, we need to be aware, that this is a breaking change (at least within our NEXT version)!

packages/components/src/components/post-header/post-header.scss
```diff
@@ -243,9 +246,9 @@
     padding-inline: 40px;
   }
 
+  .navigation-header,
   ::slotted(post-mainnavigation),
-  .navigation-footer,
-  .navigation-target-group {
+  .navigation-footer {
```
oliverschuerch, line 251: Can you explain, why you removed the `.navigation-target-group` selector?

I'm asking because the target-group styles are not working correctly in the jobs header example:
image

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -342,39 +343,34 @@ export class PostHeader {
   }
 
   private renderNavigation() {
```
oliverschuerch, line 345: I needed a moment to realize what you just did here.
But now I have to say: nice change!

I have one thing I would like you to think about:
You changed the logic regarding the `navigationClasses`.
Before they where added or removed, no matter how big the viewport was at that time.
Now, you only add or remove them in specific viewports, which is technically fine, but should be documented as a change (e.g. in the changeset).

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
oliverschuerch, line 1: As explained above, please revert or extract that into a different PR.

packages/components/src/components/post-mainnavigation/post-mainnavigation.scss
```diff
@@ -137,7 +139,7 @@ post-mainnavigation {
           inset-inline-start: 0;
         }
 
-        &.selected,
+        &.active,
```
oliverschuerch, line 142: As explained above, please revert or extract that into a different PR.

packages/documentation/src/stories/components/header/renderers/meta-navigation.ts
```diff
@@ -1,11 +1,12 @@
...
+import { html, nothing } from 'lit';
 
-export function renderMetaNavigation() {
+export function renderMetaNavigation(args: Args) {
   return html`
     <!-- Meta navigation -->
     <ul class="list-inline" slot="meta-navigation">
       <li>
-        <a href="">
+        <a href="" class=${args.jobs ? 'active' : nothing}>
```
oliverschuerch, line 9: Use `selected` not `active`.

packages/documentation/src/stories/components/header/renderers/target-group.ts
```diff
@@ -1,11 +1,12 @@
...
+import { html, nothing } from 'lit';
 
-export function renderTargetGroup() {
+export function renderTargetGroup(args: Args) {
   return html`
     <!-- Target Group -->
     <ul slot="target-group" class="target-group">
       <li>
-        <a href="#" class="active">Private customers</a>
+        <a href="#" class=${args.jobs ? nothing : 'active'}>Private customers</a>
```
oliverschuerch, line 9: Use selected not active.

packages/styles/src/components/header/_mixins.scss
```diff
@@ -31,7 +31,7 @@
     left: 0;
   }
 
-  &.selected,
+  &.active,
```
oliverschuerch, line 34: Nope

packages/styles/src/components/header/_post-header.scss
```diff
@@ -252,11 +258,12 @@ post-header {
     }
   }
 
-  a.selected,
+  a.active,
```
oliverschuerch, line 261: Use select not active.

packages/styles/src/components/header/_post-header.scss
```diff
@@ -171,12 +171,14 @@ post-header {
...
-      .local-login post-menu-trigger *
+  button:where(
+      :not(
+          post-mainnavigation *,
+          post-language-switch *,
+          [slot='global-login'] *,
+          .local-login post-menu *,
+          .local-login post-menu-trigger *
+        )
     ) {
```
oliverschuerch, line 182: Have these changes anything to to with the new slot?

packages/components/src/components/post-header/post-header.scss
alizedebray: I did not remove it, just renamed it to `.navigation-header` because it no longer contains only the target group but also the job links. I had the same reaction as you but the target group are working as per design: https://www.figma.com/design/VBcGcYKSzNjt1mSMy2LTUO/PostOneWeb-2025?node-id=4548-8189&t=igAyHk2avLrthhU6-4

packages/components/src/components/post-header/post-header.tsx
alizedebray: This change is internal and only touches the shadow DOM of the post-header component, as such this is not a "change relevant to users of a package" and does not require a changeset: https://github.com/swisspost/design-system/blob/main/CONTRIBUTING.md#when-to-add-a-changeset

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
alizedebray: This change is needed so that the "jobs" link is correctly styled when active, if I extract these changes into another PR this one is incomplete.

packages/styles/src/components/header/_post-header.scss
alizedebray: Can you clarify your question?

packages/components/cypress/e2e/header.cy.ts
alizedebray: As you can see in [our documentation](https://swisspost-design-system-next.netlify.app/?path=/docs/27a2e64d-55ba-492d-ab79-5f7c5e818498--docs), the `.selected` classes never documented whereas the `.active` class is (uses to set the current target group). 

I needed the classes to be aligned so that the "Jobs" link gets the same styles as the selected target group. 
I therefore decided to align everything with the documented class to avoid a breaking change, as you correctly wrote this is not the point of this ticket. 

If you think we should use `selected` instead of `active`, can you please open a ticket for the breaking change?

alizedebray:
> The jobs header example in the docs also shows some anwanted elements in the global-header section: image

I removed the login button again. You're writing about "anwanted element**s**", is there something else?

packages/components/cypress/e2e/header.cy.ts
oliverschuerch: No, I'm fine with the `active` class in this case.

Imo, technically this is either a bug in the documentation or a breaking change in the components package. Therefore it must be separated from this pr and properly documented through a changeset.

packages/components/src/components/post-header/post-header.scss
oliverschuerch: To me, this looks like an error in the design.
Did you talk about that with UX? Because the only reason, why none of these two links is active, could be because we're neighter in "private" nor in "business" section.

packages/components/src/components/post-header/post-header.tsx
oliverschuerch: I see your point. By adding a changeset, I wanted to make sure, if this is a regression (since it's not a refactoring), we would now where it's coming from.

But for non, I'm ok with keeping it as is.

packages/components/src/components/post-header/post-header.scss
oliverschuerch: I got the approval from @sandra-post.
We can keep it as is for the moment.
https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-Next-Level?node-id=21091-31043&t=hWxdoPZ7C6RDecb7-4

packages/components/cypress/e2e/header.cy.ts
alizedebray: As discussed, I checked we only have `.selected` classes for header components: https://github.com/search?q=repo%3Aswisspost%2Fdesign-system+%22.selected%22+language%3ASCSS&type=code&l=SCSS.
For other components (tabs, rating, button group) we are using `.active`: https://github.com/search?q=repo%3Aswisspost%2Fdesign-system+%22.active%22+language%3ASCSS&type=code 

I'll therefore go with `.active` in the header for now

oliverschuerch (approved):

PR #6435, opened 2025-10-13
feat(component): add raw slider

### LINKED ISSUE #5854: [web component func] Slider
### 📝 Description

Select numeric values by dragging one or more thumbs along a track.

### ♿ Accessibility

- Usable with mouse, keyboard, or touch.  
- The slider has `role="slider"` and the following ARIA attributes: `aria-valuenow`, `aria-valuemin`, `aria-valuemax`.  
- The slider must have a label, either hidden via `aria-label` or visible via `aria-labelledby`.  
- If the numbers represent meaningful values (e.g., days of the week), use `aria-valuetext` to expose the current value in a human-understandable way (e.g., `"Monday"`).  
- If the slider has multiple thumbs, the role and ARIA attributes are set on each thumb instead of the slider element itself.  

### ⚙️ Requirements

- The slider is composed of a fixed track and one or more sliding thumbs.  
- The track is horizontal by default but can be configured as vertical using the `vertical` property.  
- A `range` property adds a second thumb to allow selection of a numeric range.  
- Two thumbs cannot cross each other.  
- The minimum value of the track defaults to `0` but can be configured with a `min` property.  
- The maximum value of the track defaults to `100` but can be configured with a `max` property.  
- The step size defaults to `1` but can be configured with a `step` property.  
- The step is based on `min` if specified, otherwise on `value`, or `0` if neither is provided (e.g., with `min="3"` and `step="2"`, thumbs snap to 3, 5, 7, 9, …).  
- `min`, `max`, and `step` must be numbers. `min` must be lower than `max`. `step` must be lower than `max - min`.  
- The `value` property represents the position of the thumb(s):  
  - If `range` is false → a single number.  
  - If `range` is true → a tuple `[number, number]`.  
- Default placement of thumbs:  
  - Single thumb → centered between `min` and `max` (rounded).  
  - Range → placed at one-third and two-thirds of the track.  
- Thumbs always snap to the closest step.  
  - Example: if `step="3"`, a value of 1 snaps to 0, a value of 2 snaps to 3.  
- Thumbs never move past the `min` or `max`.  
- Values can optionally be formatted.  
- Values are always shown in ascending order.  
- Ticks can be added via a slotted `<datalist>` with `<option>` elements.  
- If the options have labels, they are shown alongside the ticks.  
- Ticks and labels are always displayed above the track (or to the left for vertical sliders) so they remain visible when using touch. (It is recommended to implement this layout using CSS grid to ensure consistent alignment between the track, ticks, and labels.)
- The slider emits:  
  - an `input` event whenever a thumb moves (multiple emissions while sliding).  
  - a `change` event whenever a thumb position is committed (emitted once when the user releases the thumb).  

### 🧪 Testing

 Single, range, and vertical slider scenarios.  
- Correct snapping to steps with different min/max/step configurations.  
- Default thumb placement logic.  
- Value formatting behavior.  
- Ticks and labels rendering correctly via `<datalist>`.  
- Event emissions (`input` vs `change`).  
- Preventing thumbs from crossing or exceeding min/max.  
- Accessibility: ARIA attributes, keyboard navigation, screen reader value announcements.  
- Responsiveness on touch devices and across browsers.  

### 💻 Proposed implementation

```html
// default slider min is 0, max is 100, step is 1 and value is 50
<post-slider />

// range slider min is 0, max is 100, step is 1 and value is [33, 67]
<post-slider range="true" />

// slider with ticks
<post-slider>
  <datalist id="values">
    <option value="0"></option>
    <option value="25"></option>
    <option value="50"></option>
    <option value="75"></option>
    <option value="100"></option>
  </datalist>
</post-slider>

// slider with ticks
<post-slider max="6" value="1" >
  <datalist>
    <option value="0" label="Sunday"></option>
    <option value="1" label="Monday"></option>
    <option value="2" label="Tuesday"></option>
    <option value="3" label="Wednesday"></option>
    <option value="4" label="Thursday"></option>
    <option value="5" label="Friday"></option>
    <option value="6" label="Saturday"></option>
  </datalist>
</post-slider>
```

### 🔗 References

- https://www.w3.org/WAI/ARIA/apg/patterns/slider/
- https://www.w3.org/WAI/ARIA/apg/patterns/slider-multithumb/
- https://ant.design/components/slider
- https://mui.com/material-ui/react-slider
- https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range

### 📃 Tasks

- [x] Implement component functionality
- [x] Create developer-only documentation in Storybook
- [ ] Add functional tests

alizedebray:
## 📄 Description

This PR add a slider component to our raw components.

## 🚀 Demo

https://preview-6435--swisspost-design-system.netlify.app/?path=/docs/d6bc3b88-050b-4ed7-af0d-c01d22a2605a--docs

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist


packages/components/src/components/post-slider/active-thumb.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  get neighborValues(): { previous: number | null; next: number | null } {
+    const previousValue = this.neighbors.previous ? getValueNow(this.neighbors.previous) : null;
+    const nextValue = this.neighbors.next ? getValueNow(this.neighbors.next) : null;
+    return { previous: previousValue, next: nextValue };
+  }
+
+  constructor(node: Node | EventTarget, host: HTMLElement, orientation: Orientation) {
+    if (!isThumb(node)) throw Error('An active thumb must be an HTML element with a slider role.');
```
leagrdv: The constructor is only called `onPointerDown` of the div with the `role="slider"`, therefore `isThumb(node)` should always be true so I'm not sure it's worth checking here.

packages/components/src/components/post-slider/post-slider.tsx
```diff
@@ -0,0 +1,288 @@
...
+  @Prop() range = false;
+  @Watch('range')
+  validateRange() {
+    checkRequiredAndType(this, 'range', 'boolean');
+  }
+
+  /**
+   * The granularity that the value must adhere to.
+   */
+  @Prop() step: number | 'any' = 1;
```
leagrdv: Why can the step be the string "any"?

packages/components/src/components/post-slider/post-slider.tsx
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  /**
+   * The number or range initially selected.
+   */
+  @Prop() value?: number | [number, number];
+  @Watch('value')
+  validateValue() {
+    if (this.range) {
+      checkEmptyOrArrayOf(this, 'value', 'number');
```
leagrdv: Should this be a more precise check that there is 2 elements within the array?

packages/components/src/components/post-slider/post-slider.tsx
```diff
@@ -0,0 +1,288 @@
...
+export class PostSlider {
+  @Element() host: HTMLPostSliderElement;
+
+  private activeThumb: ActiveThumb | null = null;
+
+  /**
+   * The greatest value in the range of permitted values.
+   */
+  @Prop() max = 100;
+  @Watch('max')
```
leagrdv: This part of the specs is not verified:
_min, max, and step must be numbers. min must be lower than max. step must be lower than max - min._
Same goes for the value: it is not checked whether the defined value is between the min and max

The result is a broken slider if I try to set wrong data but I get no errors.

packages/components/src/components/post-slider/post-slider.tsx
```diff
@@ -0,0 +1,288 @@
...
+      checkEmptyOrArrayOf(this, 'value', 'number');
+    } else {
+      checkEmptyOrType(this, 'value', 'number');
+    }
+  }
+
+  /**
+   * Event dispatched while a thumb is sliding, payload is the current value.
+   */
+  @Event() postInput: EventEmitter<number | [number, number]>;
```
leagrdv: What would be the use for that event? It's fired a lot and I don't see the reason for this one. I feel like the `postChange` one would suffise.

apps/documentation/src/stories/raw-components/slider/slider.stories.ts
leagrdv, line 1: If I try to fill in some fields to test it, it breaks the slider in the demo. 
I think defining default `args` for the different values should fix it.
https://github.com/user-attachments/assets/ff0f5525-5996-4362-93ac-64e1ea7729ef

leagrdv (changes requested):
I saw in the specs of the ticket mentions of ticks within a `datalist`. Do you plan on adding the examples and CSS in another PR?

leagrdv:
Also, last bug I found. Sometimes I can't move the slider and I get the 🚫 icon that appears. It seems to be random and happens pretty rarely though and I always manage to slide it again when I select the slider again.
https://github.com/user-attachments/assets/9f4c4d0d-5e8f-4938-962e-cc2d4ef2a550

packages/components/src/components/post-slider/post-slider.tsx
alizedebray: This is to match the behavior of the `input[type="range"]`: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range#setting_step_to_any

packages/components/src/components/post-slider/post-slider.tsx
alizedebray: This is also to match the `input[type="range"]`: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range#technical_summary

alizedebray:
> I saw in the specs of the ticket mentions of ticks within a `datalist`. Do you plan on adding the examples and CSS in another PR?

I tried the `datalist` and it might not be the right approach in the end. I'll split the ticket and create another one for this specific feature!

packages/components/src/components/post-slider/active-thumb.ts
alizedebray: I've added this check mostly for types, the `isThumb` function has `node is HTMLElement` return type and helps me securely turn the `EventTarget` into an `HTMLElement`.

alizedebray:
> Also, last bug I found. Sometimes I can't move the slider and I get the 🚫 icon that appears. It seems to be random and happens pretty rarely though and I always manage to slide it again when I select the slider again. https://github.com/user-attachments/assets/9f4c4d0d-5e8f-4938-962e-cc2d4ef2a550

I was not able to reproduce this issue but I've added a pointer capture that I hope would fix it.

gfellerph:
Provided that this component does not have any research nor design, it's a bit early to go ahead with the functional implementation. Depending on the design, the functionality could change.

PR #6436, opened 2025-10-13
chore(component): update utils to prepare for slider

alizedebray:
## 📄 Description

Update component utilities that will be used in the slider component.

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist


packages/components/src/utils/is-value-empty.ts
```diff
@@ -1,5 +1,3 @@
-import { EMPTY_VALUES } from './property-checkers/constants';
-
 export function isValueEmpty(value: unknown): boolean {
-  return EMPTY_VALUES.some(v => v === value);
```
oliverschuerch, line 4: I'd suggest to keep the ìsValueEmpty` function as is but change the way we compare the values.

The following would allow us to compare NaN with NaN, beside all the other values:
`return EMPTY_VALUES.some(v => Object.is(v, value));`

Test:
- `NaN === NaN` -> false
- `Object.is(NaN, NaN)` -> true 

The idea behind this is, to allow editing the tested values ​​without touching the test logic itself, thus minimizing the risk of regressions.

packages/components/src/utils/property-checkers/tests/check-array-of.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+  primitiveTypes.forEach(type => {
+    describe(type, () => {
+      const error = `The prop \`${propName}\` of the \`${componentName}\` component must be an \`${type}\` array.`;
+
+      const runCheckForValue = (value: unknown) => {
+        const component = { host: { localName: componentName } as HTMLElement, [propName]: value };
+        checkArrayOf(component, propName, type);
+      };
+
+      it('should log an error is the value is not an array', () => {
```
oliverschuerch:
```suggestion
      it('should log an error if the value is not an array', () => {
```

packages/components/src/utils/property-checkers/tests/check-array-of.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+        mockValues
+          .filter(value => !Array.isArray(value))
+          .forEach(value => {
+            runCheckForValue(value);
+            expect(consoleErrorSpy).toHaveBeenCalledWith(expect.stringContaining(error));
+          });
+      });
+
+      it('should log an error is the array contains some values that don\'t have the expected type', () => {
```
oliverschuerch:
```suggestion
      it('should log an error if the array contains some values that don\'t have the expected type', () => {
```

packages/components/src/utils/property-checkers/tests/check-array-of.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+        runCheckForValue(mockValues);
+        expect(consoleErrorSpy).toHaveBeenCalledWith(expect.stringContaining(error));
+      });
+
+      it('should not log an error if hte array contains only value with the expected type', () => {
```
oliverschuerch:
```suggestion
      it('should not log an error if the array contains only values with the expected type', () => {
```

packages/components/src/utils/property-checkers/tests/required-and.spec.ts
```diff
@@ -1,12 +1,11 @@
 import { requiredAnd } from '../required-and';
-import { EMPTY_VALUES } from '../constants';
 describe('requiredAnd', () => {
   const mockCheck = jest.fn();
 
   const mockRequiredAndCheck = requiredAnd(mockCheck);
 
   it('should throw error if the provided value is empty', () => {
-    EMPTY_VALUES.forEach(emptyValue => {
+    [undefined, null, '', NaN].forEach(emptyValue => {
```
oliverschuerch: If we have to update the values which are considered as "empty", we also have to update the test.
If on the other side we can keep the constant `EMPTY_VALUES` on both side, this is not the case and the maintenance effort can be kept at a minimum.

packages/components/src/utils/is-value-empty.ts
alizedebray: We didn’t notice the function wasn’t working before actually because both the function and its tests were using the same `EMPTY_VALUES` array.

As a result, we had `NaN === NaN` -> `true`, but only because both values referred to the exact same `NaN` instance from that shared array. 

For the test to work as expected we need two different `NaN` instances.

alizedebray:
> You've requested my re-review yesterday, but I can't see any changes?

Please have a look at [this comment](https://github.com/swisspost/design-system/pull/6436/changes#r2435570762).

packages/components/src/utils/is-value-empty.ts
oliverschuerch: Running some quick tests in the browsers console, leaves a bad feeling in my gut...

Comparing `NaN` with itself (no matter if it's stored in a variable, in an array index or elsewhere), returns `false` in the console:

image


| Chrome | Firefox |
| :- | :- |
| image | image |

The only reliable solution that I could find was to compare these objects with `Object.is(Object, Object)`.

image

<hr>

I don't want to block this PR, but I would like to have another opinion and therefore discuss this in a daily or the dev-roundtable.

As mentioned before, the two options we have are:

1. Replacing the test logic with something completely new and hard-coded. And thus making extensibility and maintainability more risky, because we have to touch the actual test logic over and over again, with every new value we want to test in the future.
2. Fixing the current test logic, while keeping extensibility and maintainability simple. New test values can easily be added to the `EMPTY_VALUES` array.

oliverschuerch: @alizedebray As soon as the team comes to a conclusion, you're totally free to dismiss my review, if it's blocking 😉

PR #6437, opened 2025-10-13
chore: enable trusted publishing for v9

oliverschuerch:
## 📄 Description

This updates the v9...
- ... node@22 and pnpm@10.6.5 versions.
- ... github devcontainer to use the same node version as specified in the package.json file.
- ... release workflow, so it should be possible to release it through trusted publishing.

I've updated the node and pnpm version, because v9 will be around for a while. V7 was EOL mid 2024 and v8 will probably be EOL end 2025 or mid 2026. But v9 could be maintained until end of 2027.

---

## 📝 Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,6 +7,9 @@ on:
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
+permissions:
+  id-token: write
+
 jobs:
   changeset-magic:
     name: Set up Changeset PR or Release Packages
@@ -19,6 +22,12 @@ jobs:
 
       - name: Setup Node & pnpm
         uses: ./.github/actions/setup-pnpm
+
+      # Trusted publishing requires npm 11.5.1 or later.
+      - name: Update NPM for trusted publishing
+        run: |
+          npm update -g npm
+          echo "npm: $(npm -v)"
 
       - name: Install & build Dependencies
         run: pnpm bootstrap


package.json
```diff
@@ -89,9 +89,9 @@
     "@web-types/lit": "2.0.0-3"
   },
   "engines": {
-    "node": "20"
+    "node": "22"
```
oliverschuerch, line 92: This updates the used NODE version to 22 (same as on main).

package.json
```diff
@@ -89,9 +89,9 @@
...
   },
-  "packageManager": "pnpm@9.6.0",
+  "packageManager": "pnpm@10.6.5+sha512.cdf928fca20832cd59ec53826492b7dc25dc524d4370b6b4adbf65803d32efaa6c1c88147c0ae4e8d579a6c9eec715757b50d4fa35eea179d868eada4ed043af",
```
oliverschuerch, line 94: This updates the used pnpm version to 10.6.5 (same as on main).

.github/workflows/release.yaml
```diff
@@ -7,6 +7,9 @@ on:
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
+permissions:
+  id-token: write
```
oliverschuerch, line 11: Needed for trusted publishing.
Without that permission, the OIDC object can not be created during the workflow run.
https://docs.npmjs.com/trusted-publishers

.github/workflows/release.yaml
```diff
@@ -20,6 +23,12 @@ jobs:
       - name: Setup Node & pnpm
         uses: ./.github/actions/setup-pnpm
 
+      # Trusted publishing requires npm 11.5.1 or later.
+      - name: Update NPM for trusted publishing
+        run: |
+          npm update -g npm
+          echo "npm: $(npm -v)"
```
oliverschuerch, line 30: NODE 22 comes with NPM 10.x
Trusted publishing requires NPM version 11.5.1 or later. Therefore we need to update NPM to the latest version.
The `echo` statment allows us to check in the CI if the installed version matches the requirements.

.devcontainer/devcontainer.json
```diff
@@ -3,12 +3,7 @@
 {
   "name": "Node.js & TypeScript",
   // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-  "image": "mcr.microsoft.com/devcontainers/typescript-node:1-20",
-  "features": {
-    "ghcr.io/devcontainers-contrib/features/pnpm:2": {
-      "version": "9"
-    }
-  },
+  "image": "mcr.microsoft.com/devcontainers/typescript-node:22",
```
oliverschuerch, line 6: This configurates the docker container to install NODE version 22.0.0.
PNPM is not needed, since it will be installed by COREPACK automatically (defined in our package.json packageManager field), as soon as it is used the first time.

alizedebray (approved):

PR #6440, opened 2025-10-13
chore: enable trusted publishing for v8

oliverschuerch:
## 📄 Description

This updates the v8...
- ... github devcontainer to use the same node version as specified in the package.json file.
- ... release workflow, so it should be possible to release it through trusted publishing.

---

## 📝 Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,6 +7,9 @@ on:
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
+permissions:
+  id-token: write
+
 jobs:
   changeset-magic:
     name: Set up Changeset PR or Release Packages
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -20,6 +23,12 @@ jobs:
       - name: Setup Node & pnpm
         uses: ./.github/actions/setup-pnpm
 
+      # Trusted publishing requires npm 11.5.1 or later.
+      - name: Update NPM for trusted publishing
+        run: |
+          npm update -g npm
+          echo "npm: $(npm -v)"
```
alizedebray, line 30:
```suggestion
```

.devcontainer/devcontainer.json
```diff
@@ -3,12 +3,7 @@
 {
   "name": "Node.js & TypeScript",
   // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-  "image": "mcr.microsoft.com/devcontainers/typescript-node:1-20",
```
alizedebray, line 6: What is this change for?

.github/workflows/release.yaml
oliverschuerch: I would suggest to keep this as an information for us, in case something does not work as expected during the release process.

.devcontainer/devcontainer.json
oliverschuerch: It installs node@20.0.0, without any minor/patch updates.
In our package.json we define the needed engine is node@20, so one must be able to run our dev env with node ^20.0.0.

alizedebray (approved):

PR #6459, opened 2025-10-16
chore: remove playwright continue-on-error flag

oliverschuerch:
## 📄 Description

This changes the e2e playwright workflow, so it errors if tests fail.
To enable that, I needed to change the hydration error tests, so we're only warned about occuring hydration errors, but do not resolve as an error.

## 📝 Checklist

### DIFF
--- a/.github/workflows/playwright-tests.yml
+++ b/.github/workflows/playwright-tests.yml
@@ -1,4 +1,4 @@
-name: Playwright e2e tests
+name: E2E tests (playwright)
 on:
   pull_request:
     paths:
@@ -32,5 +32,3 @@ jobs:
 
       - name: Run tests
         run: pnpm --filter design-system-nextjs-integration e2e
-        # Remove "continue-on-error: true" as soon as the components are fully ssr compatible!
-        continue-on-error: true


packages/nextjs-integration/playwright/tests/ssr.spec.ts
```diff
@@ -19,14 +19,27 @@ test.describe('SSR compatibility', () => {
...
-    expect(hydrationErrors.length).toBe(0);
+    const hydratedComponents = await page.locator('[data-hydrated]').all();
+
+    // wait for page hydration before checking for errors
+    await Promise.all(
+      hydratedComponents.map(component => component.waitFor({ state: 'attached' })),
+    );
+    await page.waitForLoadState('load');
+
+    // expect(hydrationErrors.length).toBe(0);
```
leagrdv, line 37:
```suggestion
```

leagrdv (approved):

packages/nextjs-integration/playwright/tests/ssr.spec.ts
oliverschuerch: You're like Sonar Cloud 😁 which suggested the same change as well.
I wanted to keep it, since the current implementation is only an interim solution.

PR #6460, opened 2025-10-16
fix(styles): remove user-menu need for a slot attribute

alizedebray:
## 📄 Description

This PR update the user menu styles so that it does not need to be wrapped in a element with a `slot="global-login"` attribute when it's not in the global header (i.e. when it is in the custom controls).

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist

### DIFF
--- a/.changeset/moody-squids-lay.md
+++ b/.changeset/moody-squids-lay.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': minor
+---
+
+Improved the `dialog` element documentation by adding the control-table entries for core methods and events, updating examples with aria-labelledby/aria-describedby, and exposing the closedby option.
--- a/packages/documentation/src/stories/components/dialog/dialog.docs.mdx
+++ b/packages/documentation/src/stories/components/dialog/dialog.docs.mdx
@@ -8,7 +8,9 @@ import StylesPackageImport from '@/shared/styles-package-import.mdx';
 # Dialog
 
 <p className="lead">Communicate crucial information and request user action.</p>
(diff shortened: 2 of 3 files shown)


packages/documentation/src/stories/components/header/renderers/custom-controls.ts
```diff
@@ -19,7 +19,7 @@ export function renderCustomControls(args: Args) {
           <post-icon aria-hidden="true" name="search"></post-icon>
         </a>
       </li>
-      <li>${args.isLoggedIn ? renderUserMenu() : loginButton}</li>
+      <li class="local-login">${args.isLoggedIn ? renderUserMenu() : loginButton}</li>
```
oliverschuerch, line 22: The need for that class is a breaking change.
I'm fine with it, but you have to add a changeset 😉

packages/styles/src/components/header/_post-header.scss
oliverschuerch, line 1: Same here, the `.local-login` class creates a breaking change.

packages/documentation/src/stories/components/header/renderers/custom-controls.ts
alizedebray: The `.local-login` class is only used for the placement of the user menu which has not been published, therefore this is not yet a breaking change. I added a [comment](https://github.com/swisspost/design-system/pull/6418#issuecomment-3415605893) in the release PR so that we do not merge it without this fix.

PR #6471, opened 2025-10-17
chore(components-angular, nextjs-integration): add general component e2e test suite

alionazherdetska:
Depends on
- [#6459](https://github.com/swisspost/design-system/pull/6685) **This PR fixes the console.error for `post-stepper` component in Nextjs-integration package**

## 📄 Description

This PR adds E2E tests for component error detection in Angular (Cypress) and Next.js (Playwright) integration packages, plus fixes for issues discovered during testing. 

- Added component error monitoring utilities that filter component-specific errors
- Implemented hydration error filtering for Next.js tests
- Created tests which validate components render, attach to DOM, and hydrate without errors

## Fixes for Nextjs Integration
- Added missing required label="Demo menu" attribute to <PostMenu>
- Removed second"Ligilo teksto" link from card component

## Notes

Tests intentionally exclude hydration errors (handled separately in ssr.spec.ts)
Error monitoring focuses on runtime component issues, not SSR mismatches

## 📝 Checklist


packages/components-angular/projects/consumer-app/cypress/support/component-error-filter.ts
```diff
@@ -0,0 +1,51 @@
...
+  
+  const extractMessage = (arg: any): string => {
+    if (typeof arg === 'string') return arg;
+    if (arg?.message) return arg.message;
+    if (arg?.stack) return arg.stack;
+    try { return JSON.stringify(arg); } catch { return String(arg); }
+  };
+  
+  const isRelevant = (message: string): boolean => 
+    componentNames.some(name => message.toLowerCase().includes(name.toLowerCase()));
```
oliverschuerch: Improve performance a little bit, by converting every string to lowercase only once.
Even if you in your implementation you already get lowercase (or kebab-case) componentNames...

```suggestion
  const errors: string[] = [];
  const lowerCaseComponentNames = componentNames.map(n => n.toLowerCase());
  
  const extractMessage = (arg: any): string => {
    if (typeof arg === 'string') return arg;
    if (arg?.message) return arg.message;
    if (arg?.stack) return arg.stack;
    try { return JSON.stringify(arg); } catch { return String(arg); }
  };
  
  const isRelevant = (message: string): boolean => {
    const lowerCaseMessage = message.toLowerCase();
    return lowerCaseComponentNames.some(n => lowerCaseMessage.includes(n));
  }
```

packages/nextjs-integration/playwright/support/component-error-filter.ts
```diff
@@ -0,0 +1,40 @@
...
+      }
+    }
+  });
+
+  page.on('pageerror', (error) => {
+    const message = error.message;
+    if (componentNames.some(name => message.toLowerCase().includes(name.toLowerCase()))) {
+      errors.push(message);
+    }
+  });
```
oliverschuerch: Same here, please lowercase strings before, not everytime while you iterate them.

```suggestion
  const errors: string[] = [];
  const lowerCaseComponentNames = componentNames.map(n => n.toLowerCase());

  page.on('console', (msg) => {
    if (msg.type() === 'error') {
      const text = msg.text();
      const lowerCaseText = text.toLowerCase();
     
      if (lowerCaseComponentNames.some(n => lowerCaseText.includes(n))) {
        errors.push(text);
      }
    }
  });

  page.on('pageerror', (error) => {
    const message = error.message;
    const lowerCaseMessage = message.toLowerCase();
    
    if (lowerCaseComponentNames.some(n => lowerCaseMessage.includes(n))) {
      errors.push(message);
    }
  });
```

packages/nextjs-integration/playwright/tests/post-tabs.spec.ts
oliverschuerch, line 1: Nice, but...

These tests are very general and we could use almost all of them to test every component.
If you'd set it up similar to how I did it in the SSR/CSR test files:

```js
import { expect, test } from '@playwright/test';
import { componentNames } from '@swisspost/design-system-components/dist/component-names.json';

test.describe('Components', () => {
  test.beforeEach(async ({ page }) => {
    await page.goto('/ssr');
  });

  test('your test', async ({ page }) => {
    for (const componentName of componentNames) {
      const component = page.locator(componentName).first();
      await expect(component).toBeAttached();
      await expect(component).toBeVisible();
    }
  });
});
```

packages/components-angular/projects/consumer-app/cypress/e2e/components.cy.ts
```diff
@@ -1,24 +1,40 @@
...
   beforeEach(() => {
     cy.visit('/');
-    cy.window().then(win => {
-      cy.wrap(cy.spy(win.console, 'error')).as('consoleError');
+  });
+
+  componentNames.forEach(componentName => {
+    it(`should contain <${componentName}>`, () => {
+      cy.get('body').then($body => {
+        if ($body.find(componentName).length > 0) {
```
oliverschuerch: For me this reads like `if (it exists) return it exists`.
Shouldn't it be something like:

```ts
it(`should contain <${componentName}>`, () => {
  cy.get(componentName).first().should('exist');
});
```

packages/components-angular/projects/consumer-app/cypress/e2e/components.cy.ts
```diff
@@ -1,24 +1,40 @@
...
-    it(`should contain <${tagName}>`, () => {
-      cy.get(tagName).should('exist');
+  it('should not have console errors from components', () => {
+    const errorCapture = setupComponentErrorCapture(componentNames);
+    
+    cy.visit('/', {
+      onBeforeLoad: errorCapture.onBeforeLoad
     });
+    
+    cy.wait(500);
```
oliverschuerch: Can you add a comment which explains, why this is necessary?

packages/components-angular/projects/consumer-app/cypress/support/component-error-filter.ts
oliverschuerch, line 1: The comments on the top are nice, but I think it would be helpful to also have some comments for some specific lines.
e.g. captures logged errors from console or checks if component name appears in message, etc.

packages/components/src/components/post-menu/post-menu.tsx
```diff
@@ -94,7 +94,11 @@ export class PostMenu {
 
   componentDidLoad() {
     this.validatePlacement();
-    this.validateLabel();
+    // Temporary workaround: defer validation to avoid timing issues with parent components.
```
oliverschuerch: If that's the case, it's also true for the `validatePlacement` method.
Did you already create a ticket to investigate about that topic? If not, you should.

packages/nextjs-integration/playwright/support/component-error-filter.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+export function captureComponentErrors(page: Page, componentNames: string[]): string[] {
+  const errors: string[] = [];
+  const lowerCaseComponentNames = componentNames.map(n => n.toLowerCase());
+
+  // Patterns to identify hydration errors that should be ignored
+  const hydrationErrorPatterns = [
+    'hydration failed because',
+    'server rendered html didn\'t match',
+  ];
```
oliverschuerch: Imo, this should be a constant outside of the function.
And since you declare it's a pattern by name, let's make it a pattern (a RegExp).

```ts
const IGNORE_ERROR_PATTERNS = [
  /^hydration failed because/,
  /server renderd html didn\'t match/,
];

export function captureComponentErrors(...): string[] {
  ...
  
  function isIgnoredError (text: string): boolean {
    return IGNORE_ERROR_PATTERNS.some(p => new RegExp(p.source, p.flags + 'gmi').test(text);
  }

  ...
```

I leave the decision to you if you want to do all of that, but at least move the constant out of the function, so other DEVs recoginze immediately how they can add new exeptions.

packages/nextjs-integration/playwright/tests/components.spec.ts
```diff
@@ -0,0 +1,36 @@
+import { test, expect } from '@playwright/test';
+import { componentNames } from '@swisspost/design-system-components/dist/component-names.json';
+import { captureComponentErrors, assertNoComponentErrors } from '../support/component-error-filter';
+
+test.describe('Components', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto('/ssr');
+  });
+
```
oliverschuerch: I know technically you already test that, but could you include something similar to the test `should contain <${componentName}>` you added in angular?
Or rename the first test if that's the equivalent to the mentioned test in angular.

packages/nextjs-integration/playwright/tests/components.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+  test('should render and be attached', async ({ page }) => {
+    for (const componentName of componentNames) {
+      const component = page.locator(componentName).first();
+      if (await component.count() > 0) {
+        await expect(component).toBeAttached();
+      }
```
oliverschuerch: This is again similar to `if (it exists) return it exists`, isn't it?
Shouldn't it be written similar to that:
```ts
const component = await page.locator(componentName).first();
await expect(component).toHaveCount(1);
```

packages/components-angular/projects/consumer-app/cypress/e2e/components.cy.ts
```diff
@@ -1,24 +1,40 @@
...
-  it('should not log any error', () => {
-    cy.get('@consoleError').should('not.have.been.called');
+  componentNames.forEach(componentName => {
+    it(`should render and be attached: <${componentName}>`, () => {
+      cy.get('body').then($body => {
+        if ($body.find(componentName).length > 0) {
+          cy.get(componentName).first().should('exist');
+        }
+      });
+    });
```
oliverschuerch: This is basically the same as the test before.
If you want to test attachement, you could maybe test if `${componentName}[data-hydrated]` exists.

packages/nextjs-integration/playwright/tests/components.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+  });
+
+  test('should be hydrated', async ({ page }) => {
+    for (const componentName of componentNames) {
+      const component = page.locator(componentName).first();
+      if ((await component.count()) > 0) {
+        await expect(component).toHaveAttribute('data-hydrated', { timeout: 5000 });
+      }
```
oliverschuerch: Same here, you can't do that only if the component exists. The existance of the component should always be true, or we have forgotten to add it to the consumer app.

oliverschuerch (changes requested):

packages/nextjs-integration/playwright/tests/components.spec.ts
alionazherdetska: This test already exists:
https://github.com/swisspost/design-system/blob/main/packages/nextjs-integration/playwright/tests/ssr.spec.ts#L9C3-L14C6

packages/components/src/components/post-menu/post-menu.tsx
alionazherdetska: Here is the ticket addressing it: 
https://github.com/orgs/swisspost/projects/3/views/11?filterQuery=props&pane=issue&itemId=135159287&issue=swisspost%7Cdesign-system%7C6507

oliverschuerch:
Depends on:
- #6635

packages/nextjs-integration/playwright/support/component-error-filter.ts
oliverschuerch, line 1: Imo it would help a lot for readybility, if you could define all helper functions as named functions (instead of mixing between named functions and anonymous functions stored in variables) and move them below the return value. This way one can read the function from top to bottom, instead of watching at a dozens lines of code before you see where the "program" really starts its execution.

e.g.
````ts
export function setupComponentErrorCapture(page: Page, componentNames: string[]) {
  const errors: string[] = []; // live array used by tests
  const captured: CapturedError[] = [];
  const seen = new Set<string>(); // dedupe key: message||stack

  // Pre-compute lowercase names to avoid repeated lowercasing in the hot path
  const lowerCaseComponentNames = componentNames.map(n => n.toLowerCase());
  
  page.on('console', onConsole);
  page.on('pageerror', onPageError);

  return { errors, dispose, captured } as const;
  
  function onConsole (msg: any) {
    ...
  };

  function pushError(message: string, source: CapturedError['source'], stack?: string) {
    ...
  }
  ...
}
```

I know it is working as it is, so this change is optional.
At least decide for one or the other way (currently `pushError` is defined as named function, while all the others are anonymous functions stored in `const`)!

packages/nextjs-integration/playwright/tests/csr.spec.ts
```diff
@@ -1,17 +1,34 @@
...
+
+    test(`should be hydrated: ${componentName}`, async ({ page }) => {
+      const component = page.locator(`${componentName}[data-hydrated]`).first();
+      await expect(component).toBeAttached();
+    });
+
+    test(`should not have console errors: ${componentName}`, async ({ page }) => {
+      const errorCapture = setupComponentErrorCapture(page, [componentName]);
+
+      await page.goto('/csr');
```
oliverschuerch: If really needed, use `pate.reload()`.
https://playwright.dev/docs/api/class-page#page-reload
Because if we ever change the route, this will just work out-of-the-box.

```suggestion
      await page.reload();
```

packages/components-angular/projects/consumer-app/cypress/e2e/components.cy.ts
```diff
@@ -1,24 +1,35 @@
...
+import { componentNames } from '@swisspost/design-system-components/dist/component-names.json';
 
 describe('Components', () => {
   beforeEach(() => {
     cy.visit('/');
-    cy.window().then(win => {
-      cy.wrap(cy.spy(win.console, 'error')).as('consoleError');
+  });
+
+  componentNames.forEach(componentName => {
```
oliverschuerch: Can't we wrap all the tests below in the same `forEach` function?

packages/components-angular/projects/consumer-app/cypress/e2e/components.cy.ts
```diff
@@ -1,24 +1,35 @@
...
-  COMPONENT_TAG_NAMES.forEach(tagName => {
-    it(`should contain <${tagName}>`, () => {
-      cy.get(tagName).should('exist');
+  componentNames.forEach(componentName => {
+    it(`should not have console errors from component: <${componentName}>`, () => {
+      const errorCapture = setupComponentErrorCapture([componentName]);
+
+      cy.visit('/', {
+        onBeforeLoad: errorCapture.onBeforeLoad
+      });
```
oliverschuerch: Imo, we should move this to the `beforeEach` function above (I don't expect it will affect the other tests above in a negative way) and only perform a `cy.reload()` here.

packages/components-angular/projects/consumer-app/cypress/support/component-error-filter.ts
oliverschuerch, line 1: See the topic mentioned on the `playwright/support/component-error-filter.ts` fils.

packages/nextjs-integration/playwright/tests/ssr.spec.ts
```diff
@@ -1,17 +1,34 @@
...
+
+    test(`should be hydrated: ${componentName}`, async ({ page }) => {
+      const component = page.locator(`${componentName}[data-hydrated]`).first();
+      await expect(component).toBeAttached();
+    });
+
+    test(`should not have console errors: ${componentName}`, async ({ page }) => {
+      const errorCapture = setupComponentErrorCapture(page, [componentName]);
+
+      await page.goto('/ssr');
```
oliverschuerch:
```suggestion
      await page.reload();
```

packages/components-angular/projects/consumer-app/cypress/e2e/components.cy.ts
alionazherdetska, line 1: Since the CI runs both **components and components-angular** suites in the same pipeline, test suite and test names are now explicit (prefixed with `components-angular`) to avoid ambiguous logs when both suites appear in CI.

PR #6502, opened 2025-10-23
feat(components): `post-stepper` and `post-stepper-item` components

leagrdv:
## 📄 Description

- Removed the HTML stepper
- Created the `post-stepper` and `post-stepper-item` components

## 🚀 Demo

https://preview-6502--swisspost-design-system-next.netlify.app/?path=/docs/7dc546d9-e248-4d06-befe-3ad62fcd310f--docs

---

## 🔮 Design review

- [x] Design review done
- [ ] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-stepper/post-stepper.tsx
+++ b/packages/components/src/components/post-stepper/post-stepper.tsx
@@ -68,18 +68,20 @@ export class PostStepper {
   componentDidLoad() {
     this.validateCompletedLabel();
     this.validateCurrentLabel();
+    this.validateActiveStepLabel();
 
     // Wait for slotchange
     setTimeout(() => {
-      this.validateActiveStepLabel();
       this.validateCurrentIndex();
     });
   }
 
   private updateActiveStepLabel() {
-    const labelTemplate = this.activeStepLabel;
-    this.mobileActiveStepLabel = labelTemplate.replace(/#index/g, `${this.currentIndex + 1}`);
-    this.updateMobileActiveStepVisibility();
+    if (this.activeStepLabel) {
+      const labelTemplate = this.activeStepLabel;
+      this.mobileActiveStepLabel = labelTemplate.replace(/#index/g, `${this.currentIndex + 1}`);
+      this.updateMobileActiveStepVisibility();
+    }
   }
 
   private updateSteps() {
--- a/packages/components/cypress/e2e/stepper.cy.ts
+++ b/packages/components/cypress/e2e/stepper.cy.ts
@@ -33,7 +33,7 @@ describe('stepper', { baseUrl: null }, () => {
     cy.window().then(win => {
       cy.spy(win.console, 'error').as('consoleError');
     });
-    cy.get('post-stepper').invoke('attr', 'step-label', 'Step:');
+    cy.get('post-stepper').invoke('attr', 'active-step-label', 'Step:');
     cy.get('@consoleError').should('be.called');
   });
 
@@ -41,7 +41,7 @@ describe('stepper', { baseUrl: null }, () => {
     cy.window().then(win => {
       cy.spy(win.console, 'error').as('consoleError');
     });
-    cy.get('post-stepper').invoke('attr', 'step-label', null);
+    cy.get('post-stepper').invoke('attr', 'active-step-label', null);
     cy.get('@consoleError').should('be.called');
   });
 
@@ -153,14 +153,12 @@ describe('stepper', { baseUrl: null }, () => {
       .should('have.text', 'Completed step:');
   });
 
-  it('should set correct mobile label on current step', () => {
+  it('should set correct mobile label on stepper', () => {
     cy.get('post-stepper')
       .invoke('attr', 'current-index', 1)
       .wait(100)
-      .find('post-stepper-item')
-      .eq(1)
-      .find('.step-mobile-label')
-      .should('have.text', 'Step 2:');
+      .find('.active-step')
+      .should('have.text', 'Step 2 : Step 2');
   });
 
   // Dynamically added/removed steps
--- a/packages/components/cypress/fixtures/post-stepper.test.html
+++ b/packages/components/cypress/fixtures/post-stepper.test.html
@@ -14,7 +14,7 @@
     <post-stepper
       completed-label="Completed step"
       current-label="Current step"
-      active-step-label="Step #index:"
+      active-step-label="Step #index :"
     >
       <post-stepper-item> Step 1</post-stepper-item>
       <post-stepper-item> Step 2</post-stepper-item>


sandra-post:
@leagrdv  Design review done, all good.

packages/documentation/src/stories/components/stepper/stepper.stories.ts
```diff
@@ -18,47 +14,79 @@ const meta: MetaComponent = {
...
+    nextLabel: 'Next step',
+    currentLabel: 'Current step',
+    stepLabel: 'Step',
+    stepsAmount: 5,
   },
   argTypes: {
-    navigatableSteps: {
-      name: 'Navigatable Steps',
-      description: 'Defines which steps in the current process the user can navigate to.',
+    currentStepNumber: {
```
alizedebray: Can you use the `argTypes` generated from Stencil?

packages/documentation/src/stories/components/stepper/stepper.stories.ts
```diff
@@ -18,47 +14,79 @@ const meta: MetaComponent = {
...
-      name: 'Current Step Number',
-      description: 'The number of the step the user is currently at in the process.',
+    stepsAmount: {
+      name: 'Steps amount',
+      description: 'The amount of steps in the stepper.',
       control: {
         type: 'select',
       },
-      options: Object.keys(defaultSteps).map(key => parseInt(key, 10) + 1),
+      options: [2, 3, 4, 5, 6, 7, 8, 9, 10],
```
alizedebray: As per the [stepper do's and don'ts](https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-Next-Level?node-id=2415-20694&t=LumrU1Y7u1zn41Dv-4), there should be a minimum of 3 steps.

packages/components/src/components/post-stepper-item/post-stepper-item.tsx
```diff
@@ -0,0 +1,25 @@
...
+  shadow: false,
+})
+export class PostStepperItem {
+  @Element() host: HTMLPostStepperItemElement;
+
+  render() {
+    return (
+      <Host data-version={version} role="listitem">
+        <span class="stepper-item-content">
+          <span>
```
alizedebray: Is this span needed?

packages/components/src/components/post-stepper-item/post-stepper-item.tsx
```diff
@@ -0,0 +1,25 @@
+import { Component, Element, h, Host } from '@stencil/core';
+import { version } from '@root/package.json';
+
+@Component({
+  tag: 'post-stepper-item',
+  styleUrl: 'post-stepper-item.scss',
+  shadow: false,
+})
+export class PostStepperItem {
+  @Element() host: HTMLPostStepperItemElement;
```
alizedebray:
```suggestion
```

packages/components/src/components/post-stepper/post-stepper.tsx
```diff
@@ -0,0 +1,125 @@
...
+    setTimeout(() => {
+      this.validateCurrentIndex();
+    });
+  }
+
+  private updateSteps() {
+    this.stepItems = this.host.querySelectorAll('post-stepper-item');
+
+    if (!this.stepItems || this.stepItems.length < 2) {
+      console.error('The post-stepper component should have at least two steps.');
```
alizedebray:
```suggestion
      console.error('The post-stepper component should have at least three steps.');
```

packages/components/src/components/post-stepper/post-stepper.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+      // Update "post-stepper-item" classes to show correct status
+      el.classList.toggle('stepper-item-completed', this.currentIndex > i);
+      el.classList.toggle('stepper-item-current', this.currentIndex === i);
+      el.classList.toggle('stepper-item-inactive', this.currentIndex < i);
+
+      // Update mobile label to show "Step N: ..." on mobile
+      const mobileLabel = el.querySelector('.step-mobile-label');
+      if (mobileLabel && this.stepLabel) {
+        mobileLabel.textContent = `${this.stepLabel} ${i + 1}:`;
```
alizedebray: Since is can differ from language to language, I would let users give the full label instead of constructing it. For example in French the rule is to have a space before the colon while in other languages there is none.

leagrdv: If user needs to put the full label, that means they would have to add it individually on each step because we'd need to have the step's number, so instead of:
```
<post-stepper
  completed-label="Completed step"
  current-label="Current step"
  step-label="Step"
  current-index="2"
>
  <post-stepper-item> Step 1 label</post-stepper-item>
  <post-stepper-item> Step 2 label</post-stepper-item>
  <post-stepper-item> Step 3 label</post-stepper-item>
  <post-stepper-item> Step 4 label</post-stepper-item>
  <post-stepper-item> Step 5 label</post-stepper-item>
</post-stepper>
```

we'd need:
```
<post-stepper
  completed-label="Completed step"
  current-label="Current step"
  current-index="2"
>
  <post-stepper-item step-label="Step 1:"> Step 1 label</post-stepper-item>
  <post-stepper-item step-label="Step 2:"> Step 2 label</post-stepper-item>
  <post-stepper-item step-label="Step 3:"> Step 3 label</post-stepper-item>
  <post-stepper-item step-label="Step 4:"> Step 4 label</post-stepper-item>
  <post-stepper-item step-label="Step 5:"> Step 5 label</post-stepper-item>
</post-stepper>
```

Do you think it's worth changing to this? From what I see, french is the only language out of DE/IT/EN/FR that have that extra space before the colon.

alizedebray: It could potentially be okay if the label was for screen readers only but since it is actually displayed on the screen I still think following the punctuation rules (or rather letting project follow them) is better for quality. 

But you can still have it as a single property on the `post-stepper`, then it would be to the users to update it with the current index. Usually, this is not very complicated when you are using an internationalization tool.
For example, here is how it would work with `ngx-translate`, used by most Post projects: 

```html
<post-stepper
  [completedLabel]="'stepper.completed.label' | translate"
  [currentLabel]="'stepper.current.label' | translate"
  [stepLabel]="'stepper.step.label' | translate: {currentIndex}"
  [currentIndex]="currentIndex"
>
  <post-stepper-item> Step 1 label</post-stepper-item>
  <post-stepper-item> Step 2 label</post-stepper-item>
  <post-stepper-item> Step 3 label</post-stepper-item>
  <post-stepper-item> Step 4 label</post-stepper-item>
  <post-stepper-item> Step 5 label</post-stepper-item>
</post-stepper>
```

The translation being defined as: 
```json
{
  "stepper.step.label": "Étape {{currentIndex}} :"
}
```

packages/components/src/components/post-stepper/post-stepper.tsx
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    this.stepItems.forEach((el, i) => {
+      if (this.stepLabel) {
+        el.querySelector('.step-mobile-label').textContent = `${this.stepLabel} ${i + 1}:`;
+      }
+
+      if (this.currentIndex === i) {
+        this.activeStepLabel = `${this.stepLabel} ${i + 1}: ${
+          el.querySelector('.label').textContent
```
alizedebray: The label contains slotted elements, therefore it could potentially have more than text (`span`, `strong`, `em`, `post-icon`, ...). Can you rather show/hide the `post-stepper-item` depending on the `currentIndex`?

alizedebray (changes requested):

packages/components/src/components/post-stepper/post-stepper.tsx
gfellerph: One thing we could do is to define a placeholder for the index that we replace during the rendering, then projects can define their own format for step-label: `step-label="Étape #index :"` and we'll replace #index with the current index.

packages/components/src/components/post-stepper/post-stepper.tsx
leagrdv: I've updated it using `innerHTML` to render all of the possible content. 
I've tried doing it with showing/hiding the `post-stepper-item` depending on the `currentIndex` (as it was done before) but I could not get a good looking solution with a small and equal spacing between each element. 
The way it was previously implemented was that you had a scrollbar when there were too many items, even though there was enough space between each step circle to show them all.
Capture d&#39;écran 2025-11-13 084701

I tried tweaking the values for `grid-template-columns` and  `grid-auto-columns` to find a good-looking solution but because each element is not of equal size (first and last should take half the space), I could not find a way to have 1/ no overlap and 2/ equal space between each element (here: spacing is smaller but the first line is too small and the last one too big):
Capture d&#39;écran 2025-11-13 085202

packages/components/src/components/post-stepper/post-stepper.tsx
leagrdv: @alizedebray I went with Philipp's solution to be framework agnostic

alizedebray (approved):

PR #6566, opened 2025-10-31
chore(components, styles): update active header links to use aria-current="location"

myrta2302:
## 📄 Description

This PR replaces the active class with aria-current="location" for active header links outside the main navigation to enhance accessibility and semantic clarity. This involves the active target-group and Jobs links.

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist


packages/styles/src/components/header/_post-header.scss
oliverschuerch, line 1: This looks like a breaking change to me, since this is something the user have to change on the project side.
@gfellerph do you confirm or deny?

packages/styles/src/components/target-group.scss
oliverschuerch, line 1: Same here.

packages/styles/src/components/header/_post-header.scss
gfellerph: Like this it is a breaking change. We could do to make it a feature, remove the `a.active` documentation and make it breaking in v11. There is not too much precedent where we write selectors for aria attributes, but it makes sense to me to use this so people automatically do the right thing.

```css
a.active, a[aria-current="location"] { /* ... */ }
```

oliverschuerch (approved):

PR #6579, opened 2025-11-03
feat(components): update post-popovercontainer animation and fix menu trigger and focus delays

myrta2302:
## 📄 Description

This PR follows #6606:
- Updates `post-popovercontainer` to open with animation only when animation is set to 'pop-in'.
- Updates `post-menu` to handle focus and aria-expanded on the trigger in `postBeforeToggle` instead of `postToggled`, fixing related delays.
- Updates documentation examples accordingly
- Enables by default the `pop-in` animation for the popovercontainer and removes all animation props.

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-menu/post-menu.tsx
+++ b/packages/components/src/components/post-menu/post-menu.tsx
@@ -164,7 +164,7 @@ export class PostMenu {
   }
 
   @EventFrom('post-popovercontainer')
-  private readonly handlePostBeforeToggle(event: CustomEvent<{ willOpen: boolean }>) {
+  private handlePostBeforeToggle(event: CustomEvent<{ willOpen: boolean }>) {
     this.isVisible = event.detail.willOpen;
     this.toggleMenu.emit(this.isVisible);
 


packages/components/src/components/post-menu/post-menu.tsx
oliverschuerch, line 1: Please update also the `<post-menu>` raw story in the docs.

packages/components/src/components/post-menu/post-menu.tsx
```diff
@@ -247,8 +253,9 @@ export class PostMenu {
       <Host data-version={version}>
         <post-popovercontainer
           onPostShow={this.handlePostShown}
-          onPostToggle={this.handlePostToggled}
+          onPostBeforeToggle={this.handlePostBeforeToggle}
           placement={this.placement}
+          animation={this.animationOff ? undefined : 'pop-in'}
```
oliverschuerch: For now, the `post-popovercontainer` component only has one specific animation (`pop-in`) or no animation at all.
But the `animation` property is implemented in a way, so we can add more animations as a feature.

If somewhen in the future, we decide to use a different animation for let's say the `post-breadcrumbs` then we use for the `post-language-switch`, this would not be possible with the given implementation, since the `animationOff` prop on the `post-menu` component is a boolean. Instead, we would need to implement a breaking change to update it.

Therefore, my suggestion is the following:

**Post Menu**
```ts
// Implement a similar, but less type specific prop as we have in post-popovercontainer
// Prop validation can happen in post-popovercontainer, so there is no redundancy
@Prop() readonly animation?: string | null = null;

...

// Simply pass the given animation prop to the post-popovercontainer
// The decision which animation we want, can now happen internally or externally, on the public component (e.g. post-breadcrumbs, post-language-switch, etc.), depending on what kind of property we provide there to configure the animation.
<post-popovercontainer
  onPostShow={this.handlePostShown}
  onPostBeforeToggle={this.handlePostBeforeToggle}
  placement={this.placement}
  animation={this.animation}
  ref={e => (this.popoverRef = e)}
>
```

**Internally defined animation (e.g. post-breadcrums or post-language-switch)**
```ts
@Prop() readonly menuAnimationOff?: boolean = false;

...

// Define a specific animation internally in our component logic
// only we can define it, the user can only turn it on/off
<post-menu
  animation={this.menuAnimationOff ? undefined : 'pop-in'}
>

```

**Externally defined animation**
```ts
@Prop() readonly menuAnimation?: 'pop-in' | 'pop-out' | null = null;

...

// Let the user decide which animation he wants
<post-menu
  animation={this.menuAnimation}
>

```

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -197,19 +197,29 @@ export class PostPopovercontainer {
...
+        if (animation?.playState === 'running') {
+          this.postBeforeToggle.emit({ willOpen: true });
+          this.postBeforeShow.emit({ first: this.hasOpenedOnce });
+        }
+
+        await animation.finished;
+
+        this.postToggle.emit({ isOpen: true });
+        this.postShow.emit({ first: this.hasOpenedOnce });
+      } else {
```
oliverschuerch: Better, but not future proof.
And since no animation is the default state, I would turn it around...

```ts
if (this.animation === null) {

 ...

} else {
  // if the component should continue to work, even the user passes an invalid, not-existing animation name, the following COULD be a good idea!
  // fallback() should return an object, with a immediately resolved promise in the finished property and a fake playState
  // (e.g. { playState: 'finished', finished: Promise })

  let animation = fallback(content);

  if (this.animation === 'pop-in') {
    animation = popIn(content);
  }
  /* in case we want to add more animations, simply add something like...
  else if (this.animation === 'pop-out') {
    animation = popOut(content);
  } */

  ...
}
```

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -197,19 +197,29 @@ export class PostPopovercontainer {
...
+        // No animation case
         this.postBeforeToggle.emit({ willOpen: true });
         this.postBeforeShow.emit({ first: this.hasOpenedOnce });
```
oliverschuerch, line 245: Before, these events only were emitted, if an animation was defined and the `animation?.playState === 'running'` statement was true.

oliverschuerch (changes requested):

packages/components/src/components/post-menu/post-menu.tsx
myrtasakellariou: After discussing this with @gfellerph we will move forward with the 'Externally defined animation'.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
myrtasakellariou: It was wrong before. It was just working because there was no check for the animation and the events were always emitted.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
myrtasakellariou: Updated the implementation to support multiple animations.

packages/documentation/src/stories/raw-components/menu-button/menu-button.stories.ts
```diff
@@ -44,6 +45,7 @@ const meta: MetaComponent = {
 };
 
 function render(args: Args) {
+  console.log(args.animation);
```
oliverschuerch:
```suggestion
```

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -117,7 +123,7 @@ export class PostPopovercontainer {
   /**
    * Animation style
    */
-  @Prop() readonly animation?: 'pop-in' | null = null;
+  @Prop() readonly animation?: AnimationName | 'none' | null = null;
```
oliverschuerch: We already have `null` as the value if no animation should be applied, why do we need to add `none`? This seems like an unwanted workaround for the validation.

packages/components/src/components/post-breadcrumbs/post-breadcrumbs.tsx
```diff
@@ -30,6 +30,12 @@ export class PostBreadcrumbs {
    */
   @Prop() menuLabel!: string;
 
+  /**
+   *
```
oliverschuerch: Where is the description?

packages/components/src/components/post-breadcrumbs/post-breadcrumbs.tsx
```diff
@@ -30,6 +30,12 @@ export class PostBreadcrumbs {
...
+   */
+
+  @Prop() readonly menuAnimation?: 'pop-in' | 'none';
```
oliverschuerch: Should be a string or `undefined`.

```suggestion
  @Prop() readonly menuAnimation?: string;
```

packages/components/src/components/post-tooltip/post-tooltip.tsx
```diff
@@ -71,7 +71,7 @@ export class PostTooltip {
     this.popoverRef.hide();
   }
 
-  /**
+  /**s
```
oliverschuerch:
```suggestion
  /**
```

oliverschuerch (changes requested):
Your PR adds an unwanted side effect on the tooltip.
It's not opening anymore, because:
image


Please fix the issue and test the following components before re-request a review:
- post-breadcrumbs
- post-language-switch
- post-popover
- post-tooltip

packages/documentation/src/stories/raw-components/menu-button/menu-button.stories.ts
oliverschuerch, line 1: There is an wrong animation property control type set in the story.

Image

https://preview-6579--swisspost-design-system-next.netlify.app/?path=/docs/8ca2bd70-56e6-4da9-b1fd-4e55388dca88--docs&devModeEnabled=true

myrtasakellariou: This ticket #6617 is about controls missing for the Popover and also about the controls being present but not effective for the tooltip.

myrtasakellariou:
Should be merged together with #6617 .

packages/documentation/src/stories/raw-components/menu-button/menu-button.stories.ts
```diff
@@ -62,6 +62,7 @@ function render(args: Args) {
       id="${args.id}"
       placement="${args.placement !== 'bottom' ? args.placement : nothing}"
       label="Example menu"
+      animation="${args.animation === 'none' ? null : nothing}"
```
alizedebray: The `post-menu` has no `animation` property:
```suggestion
```

.changeset/seven-breads-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': minor
+---
+
+Enabled a 'pop-in' animation for the `post-breadcrumbs` concatenated version and the `post-language-switch` menu variant components.
```
alizedebray:
```suggestion
Enabled pop-in animations for the concatenated `post-breadcrumbs`, the `post-language-menu`, and the header's user menu components.
```

PR #6584, opened 2025-11-04
fix(components): update outline offset of focused elements

leagrdv:
## 📄 Description

Updated the outline offset of focused elements in the header.
Updated the main navigation overflow to allow for the focus ring to be visible.

---

## 🔮 Design review

- [ ] Design review done
- [X] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-megadropdown/post-megadropdown.scss
+++ b/packages/components/src/components/post-megadropdown/post-megadropdown.scss
@@ -121,6 +121,10 @@
     align-items: center;
     min-height: 56px;
 
+    &:focus-visible {
+      border-radius: 4px;
+    }
+
     @include header-mx.mobile-header-interactive() {
       height: unset;
     }


packages/components/src/components/post-language-option/post-language-option.scss
```diff
@@ -27,7 +27,7 @@ post-language-option {
     text-transform: uppercase;
 
     @include utilities-mx.focus-style {
-      outline-offset: -2px !important;
+      outline-offset: 2px !important;
```
gfellerph: Why is the important necessary here?

leagrdv: It previously was to override the one set in the `focus-style` mixin which had an `!important` but actually since it's the same value as the original, I just removed the line

leagrdv:
@gfellerph Regarding the focus, I checked with Sandra and she confirmed it was option 1!

PR #6607, opened 2025-11-07
chore(components): fix popovercontainer animation bug affecting tooltip functionality

myrta2302:
## 📄 Description

This PR updates the popovercontainer animation logic to fix a timing issue that happened when users hovered in and out quickly, causing the tooltip to get out of sync.

## 📝 Checklist

### DIFF
--- a/.changeset/young-pears-guess.md
+++ b/.changeset/young-pears-guess.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Updated `post-popovercontainer` to fix a tooltip desynchronization issue when hovering quickly in and out.


packages/components/src/components/post-menu/post-menu.tsx
```diff
@@ -246,6 +248,7 @@ export class PostMenu {
     return (
       <Host data-version={version}>
         <post-popovercontainer
+
```
oliverschuerch:
```suggestion
```

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -117,7 +117,7 @@ export class PostPopovercontainer {
   /**
    * Animation style
    */
-  @Prop() readonly animation?: 'pop-in' | null = null;
+  @Prop() readonly animation?: 'pop-in' | null = 'pop-in';
```
oliverschuerch: Why is this necessary, to fix the bug related to the animation?
Don't we handle that with your other PR?

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -193,23 +194,21 @@ export class PostPopovercontainer {
...
+    const popoverContentEl: HTMLElement = this.host.querySelector('.popover-content');
     this.startAutoupdates();
 
-    if (content) {
-      const animation = popIn(content);
-
-      if (animation?.playState === 'running') {
+    if (popoverContentEl) {
+      // Only run animation and emit related events if animation is defined
+      if (this.animation == null) {
```
oliverschuerch:
```suggestion
      if (this.animation === null) {
```

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -193,23 +194,21 @@ export class PostPopovercontainer {
    */
   @Method()
   async open() {
-    const content = this.host.querySelector('.popover-content');
+    const popoverContentEl: HTMLElement = this.host.querySelector('.popover-content');
```
oliverschuerch: Please reset the naming change here.
Since the variable is immediately assigned a value, it should be very clear what value it contains and therefore a type and a more specific variable name is not necessary.

My personal stance for this topic:
1. Respect the work of the developers who worked on a piece of code before you 😉 
2. If all of us constantly rename variables, which we don't like, we should rather introduce some naming conventions for them or we will end up renaming variables in each and every PR.
3. As long as we have no such naming convention, please rename existing variables only if the naming, typing or the initial value assignment does not clearly indicate its purpose.

oliverschuerch (changes requested):

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
myrtasakellariou: That was left there by mistake. I ll revert it.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -193,23 +194,21 @@ export class PostPopovercontainer {
...
+    const content: HTMLElement = this.host.querySelector('.popover-content');
     this.startAutoupdates();
 
     if (content) {
-      const animation = popIn(content);
-
-      if (animation?.playState === 'running') {
+      // Only run animation and emit related events if animation is defined
+      if (this.animation === null) {
+        // No animation case
```
oliverschuerch, line 203: Just to be super secure...
These events have not been emitted before, when no animation was defined.
But now they are emitted with or without a defined animation.
If this is what you want, I will approve your PR.

oliverschuerch (changes requested):
`postBeforeToggle` and `postBeforeShow` are actually the same lifecycle hooks, just with different payloads.
This is not to be fixed here, but it could be an improvement in the future.
Same is true for `postToggle` and `postShow`.

It would be possible to combine the payload of them and end up having only two events, instead of four.

PR #6611, opened 2025-11-10
fix(component): fix `post-collapsible-trigger` event binding

alionazherdetska:
## Problem
Toggling the collapsible threw a TypeError: Cannot read properties of undefined (reading 'localName'). The handlePostToggle event handler was losing its this context when the event fired.

**Root Cause**
Passing the method directly to addEventListener without binding loses the `this` context:
```js // this becomes undefined when the event fires
this.root.addEventListener('postToggle', this.handlePostToggle);
```

## Solution
Added private boundHandlePostToggle!: (e: CustomEvent) => void;
Bound handler in connectedCallback()
Updated cleanup in disconnectedCallback()

## 🚀 Demo

[Collapsible docs](https://preview-6611--swisspost-design-system-next.netlify.app/?path=/docs/6a91848c-16ec-4a23-bc45-51c797b5b2c3--docs&devModeEnabled=true)

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
+++ b/packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
@@ -10,7 +10,6 @@ export class PostCollapsibleTrigger {
   private trigger?: HTMLButtonElement;
   private readonly observer = new MutationObserver(() => this.setTrigger());
   private root?: Document | ShadowRoot;
-  private boundHandlePostToggle!: (e: CustomEvent) => void;
 
   @Element() host: HTMLPostCollapsibleTriggerElement;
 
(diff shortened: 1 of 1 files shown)


packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
```diff
@@ -31,7 +32,8 @@ export class PostCollapsibleTrigger {
    */
   connectedCallback() {
     this.root = getRoot(this.host);
-    this.root.addEventListener('postToggle', this.handlePostToggle);
+    this.boundHandlePostToggle = (e: CustomEvent) => this.handlePostToggle(e);
```
alizedebray: In other components, we do it in the constructor with:
```suggestion
    this.handlePostToggle= this.handlePostToggle.bind(this);
```

That way you do not have to add an additional property. You can find examples in the `post-header`: https://github.com/swisspost/design-system/blob/main/packages/components/src/components/post-header/post-header.tsx#L92.

PR #6509, opened 2025-10-23
fix(styles): fix dialog header when content is scrollable

alionazherdetska:
## 📄 Description
This PR fixes the header position inside the `<dialog>` element when content becomes scrollable. The header, footer and close button now remain visible while only the body scrolls.

## 🚀 Preview link:

[Dialog docs](https://preview-6509--swisspost-design-system-next.netlify.app/?path=/docs/562eac2b-6dc1-4007-ba8e-4e981cef0cbc--docs)

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 📝 Checklist

- ✅ My code follows the style guidelines of this project
- 🛠️ I have performed a self-review of my own code
- 📄 I have made corresponding changes to the documentation
- ⚠️ My changes generate no new warnings or errors
- 🧪 I have added tests that prove my fix is effective or that my feature works
- ✔️ New and existing unit tests pass locally with my changes

PR #6625, opened 2025-11-11
feat(styles): implement signal dialogs

alizedebray:
## 📄 Description
Updated the dialog to have a `data-type` attribute, it can now be style with signal icons.

## 🚀 Demo

https://preview-6625--swisspost-design-system-next.netlify.app/?path=/docs/562eac2b-6dc1-4007-ba8e-4e981cef0cbc--docs

---

## 🔮 Design review

- [x] Design review done
- [ ] No design review needed

## 📝 Checklist


packages/styles/src/components/dialog.scss
```diff
@@ -10,19 +12,22 @@
...
 
+$dialog-types: 'info', 'error', 'success', 'warning';
+
 dialog {
   box-shadow: elevation.$elevation-500;
   padding: 0;
   min-width: min(389px, 90vw);
   max-width: 590px;
   max-height: 90vh;
-  overflow: visible;
```
alizedebray, line 19: @alionazherdetska you added the overflow visible in #6509, do you remember why it is need and if it is safe to change it?

packages/documentation/src/stories/components/dialog/dialog.stories.ts
```diff
@@ -14,10 +14,9 @@ const meta: Meta = {
   args: {
     title: 'Dialog',
     content: 'This is a dialog',
+    type: 'default',
     size: 'medium',
     position: 'center',
-    icon: 'none',
-    palette: 'palette-default',
```
leagrdv, line 20: Why did you remove the palette argument?

.changeset/honest-grapes-bow.md
leagrdv, line 1: Since it's a breaking change, could you add the information regarding the change in how the icon is added to the dialog in the migration guide?

leagrdv (changes requested):

packages/documentation/src/stories/components/dialog/dialog.stories.ts
alizedebray: I removed them to match the Figma design where this concept no longer exist: https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-Next-Level?node-id=14783-80658&t=T70AspqsMPRrFBfW-4

packages/styles/src/components/dialog.scss
alionazherdetska: In my PR, I changed `overflow: auto` to `overflow: visible` because overflow: auto was creating an extra scroll:
Screenshot 2025-11-13 145804

I also tried changing it to overflow: hidden and the behavior is exactly the same, so changing shouldn't break anything.

.changeset/honest-grapes-bow.md
alizedebray: I added a change set but not major since you can technically still use the palette class on the dialog.

packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -354,6 +355,23 @@ export class MigrationV99Component extends LitElement {
                       </label>
                     </div>
                   </li>
+                  <li class="mb-16">
```
leagrdv: The "Components" section is only for the components package, can you move that to the "Styles > Other styles" section instead?

packages/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -354,6 +355,23 @@ export class MigrationV99Component extends LitElement {
...
+                    <div class="form-check">
+                      <input
+                        id="components-dialog_icon"
+                        class="form-check-input"
+                        type="checkbox"
+                        ?checked="${this.state.components.dialog_icon}"
```
leagrdv: Then it should be in the `others` section there as well
```suggestion
                        ?checked="${this.state.others.dialog_icon}"
```

PR #6626, opened 2025-11-11
feat(components): updated the post-popover-trigger to enable the option of wrapping the popover

myrta2302:
## 📄 Description

This PR updates the `post-popover-trigger` to enable the option of wrapping the `post-popover` inside it and omitting the id reference.

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist


packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
leagrdv, line 1: Since there is no "for", the aria-controls is set to `undefined`
Capture d&#39;écran 2025-11-14 111606

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
```diff
@@ -40,7 +40,7 @@ export class PostPopoverTrigger {
    */
   @Watch('for')
   validateFor() {
-    checkRequiredAndType(this, 'for', 'string');
+    checkEmptyOrType(this, 'for', 'string');
```
leagrdv, line 43: Using a `post-popovertrigger` without a `for` or a wrapped `post-popover` shows no error (just a warning when clicking on the trigger). There should be a check that there is either one or the other and show an error if that's not the case.
Could you also add some tests for this?

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
myrtasakellariou: Correct! I ve updated that to either use the Elements API ariaControlsElements property, or fallback to a random ID generation.

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
myrtasakellariou: ok!

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -3,34 +3,58 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
...
-      cy.get('post-popover[data-hydrated]');
+      // For/id relationship case
+      cy.get('post-popover[data-hydrated][id="popover-one"]').as('post-popover');
 
-      cy.get('post-popover-trigger[data-hydrated][for="popover-one"]')
-        .children()
-        .first()
-        .as('trigger');
-      cy.get('#testtext').as('popover');
+      cy.get('post-popover-trigger[data-hydrated][for="popover-one"]').as('post-popover-trigger');
```
oliverschuerch: Since the context is pretty clear in an e2e test, I would suggest to keep aliases as short as possible.


```suggestion
      cy.get('post-popover[data-hydrated][id="popover-one"]').as('popover');

      cy.get('post-popover-trigger[data-hydrated][for="popover-one"]').as('trigger');
```

packages/components/cypress/fixtures/post-popover.test.html
```diff
@@ -17,10 +17,11 @@
     <post-popover id="popover-one" close-button-caption="Close Popover">
       <p id="testtext">This is a <a href="">test</a></p>
     </post-popover>
-
-    <!-- Alternative trigger case with no interactive element inside the trigger -->
-    <post-popover-trigger for="popover-two">
-      <div>Popover Trigger</div>
+    <post-popover-trigger id="popover-two">
```
oliverschuerch: For me it was a little confusing, that the `post-popover-trigger` here has an `id` attribute.
Imo, it would be less confusing, if you'd use something like `data-cy-id="popover-two"` instead.

packages/components/cypress/e2e/popover.cy.ts
oliverschuerch, line 1: Just a hint...

When the scope of a test is very limited (e.g. we're testing one specific component), it is often easier, to keep the names of the aliases short, because you have to repeat them a lot afterwards.

Try to be consisting with the alias naming. 

```ts
beforeEach(() => {
  cy.get('post-popover[data-hydrated][id="popover-one"]').as('popoverOne');
  cy.get('@popoverOne').find('p').as('contentOne');
  cy.get('post-popover-trigger[data-hydrated][for="popover-one"]').as('triggerOne');
  cy.get('@triggerOne').find('button').as('buttonOne');
});
```

.changeset/late-planes-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Updated `<post-popover-trigger>` to provide the option to wrap the popover, as alternative to using an id reference.
```
oliverschuerch:
```suggestion
Updated the `<post-popover-trigger>` component, to provide the ability to wrap it around the `<post-popover>` element, instead of using the `id` and `for` attribute references.
```

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -3,34 +3,58 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
...
+      );
     });
 
     it('if the element inside the trigger is not interactive, it should at least have a set tabindex="0" and role="button"', () => {
-      cy.get('@trigger2').then($child => {
-        const child = $child[0];
-        cy.wrap(child).should('have.attr', 'tabindex', '0');
-        cy.wrap(child).should('have.attr', 'role', 'button');
+      cy.window().then(win => {
+        cy.stub(win.console, 'warn').as('consoleWarn');
```
oliverschuerch: Remove this unused code snippet.

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -3,34 +3,58 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
...
       });
+
+      cy.get('@post-popover-trigger')
+        .find('button')
```
oliverschuerch: Why don't you use the before created alias `trigger`?

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -3,34 +3,58 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
...
+        });
+    });
+
+    it('if the trigger is empty', () => {
+      cy.window().then(win => {
+        cy.spy(win.console, 'error').as('consoleError');
+      });
+
+      cy.get('@post-popover-trigger')
+        .find('button')
```
oliverschuerch: Why don't you use the before created alias `trigger`?

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -3,34 +3,58 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
...
+        cy.stub(win.console, 'warn').as('consoleWarn');
       });
+
+      cy.get('@post-popover-trigger')
+        .find('button')
+        .invoke('replaceWith', '<div id="not-interactive">No interactive content.</div>')
+        .then(() => {
+          cy.get('@post-popover-trigger').children().first().should('have.attr', 'tabindex', '0');
+          cy.get('@post-popover-trigger').children().first().should('have.attr', 'role', 'button');
+        });
```
oliverschuerch:
```suggestion
        .then($children => {
          cy.wrap($children)
            .filter('#not-interactive')
            .should('have.attr', 'tabindex', '0')
            .and('have.attr', 'role', 'button');
        });
```

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -43,13 +67,13 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
...
-      cy.get('@popover').should('not.be.visible');
+      cy.get('@popover-content').should('not.be.visible');
       cy.get('@trigger').should('have.attr', 'aria-expanded', 'false');
       cy.get('.nested-element').click();
-      cy.get('@popover').should('be.visible');
+      cy.get('@popover-content').should('be.visible');
       cy.get('@trigger').should('have.attr', 'aria-expanded', 'true');
-      cy.get('post-closebutton').click();
-      cy.get('@popover').should('not.be.visible');
+      cy.get('@post-popover').find('post-closebutton').click();
```
oliverschuerch: Instead of doing that multiple times, you could define the post-closebutton as an alias in the `beforeEach` function above and then directly use it in the tests when needed.

`cy.get('@post-popover').find('post-closebutton').as('closeButton');`

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -58,75 +82,117 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
...
+      cy.get('@post-popover-trigger')
+        .invoke('attr', 'for')
+        .then(forId => {
+          cy.get(`post-popover#${forId}`).should('exist');
+
+          cy.get('@trigger').click();
+          cy.get('@consoleWarn').should('not.be.called');
+
+          cy.get('@trigger').click();
+          cy.get('@consoleWarn').should('not.be.called');
```
oliverschuerch:
```suggestion
          cy.get('@trigger').click();
          cy.get('@consoleWarn').should('not.be.called');
```

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -58,75 +82,117 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
...
+    });
+
+    it('should warn if there is no internal popover wrapped', () => {
+      cy.window().then(win => {
+        cy.spy(win.console, 'warn').as('consoleWarning');
+      });
+
+      cy.get('@post-popover-trigger-wrapped').find('post-popover').invoke('remove');
+
+      cy.get('post-popover-trigger[id="popover-two"][data-hydrated]').find('button').click();
```
oliverschuerch: We don't you use the alias `post-popover-trigger-wrapped`?

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
```diff
@@ -52,45 +52,61 @@ export class PostPopoverTrigger {
...
   private readonly boundHandlePostToggle: (event: CustomEvent<{ isOpen: boolean }>) => void;
 
-  // Gets the associated popover element to the trigger based on 'for'
+  // Gets the associated to the trigger popover element
   private get popover(): HTMLPostPopoverElement | null {
-    const ref = document.getElementById(this.for);
+    const ref = this.host.querySelector('post-popover') ?? document.getElementById(this.for);
+
+    if (!ref) {
+      console.log(this.for);
```
oliverschuerch:
```suggestion
```

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
```diff
@@ -52,45 +52,61 @@ export class PostPopoverTrigger {
...
+      const target = this.for ? `with ID: ${this.for}` : 'inside the <post-popover-trigger>';
+      console.warn(`No post-popover found ${target}.`);
```
oliverschuerch:
```suggestion
      console.error(`No post-popover found ${target}.`);
```

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
```diff
@@ -52,45 +52,61 @@ export class PostPopoverTrigger {
...
+    // check if its not focusable and add aria role and tabindex
+    if (!isFocusable(this.trigger)) {
+      this.trigger.setAttribute('tabindex', '0');
+      this.trigger.setAttribute('role', 'button');
+    }
 
-      // Set aria attributes
-      this.trigger.setAttribute('aria-haspopup', 'true');
-      this.trigger.setAttribute('aria-controls', this.for);
+    const popover = this.popover;
```
oliverschuerch: As a general rule, declare a `const` as early as possible in its scope!
And no worry, it has no effect to the memory usage even if you don't use it afterwards, as `const popover` is a simple reference to the same object (memory address) as `this.popover` (as long as `this.popover` does not return a primitive like `null` or a string).

This is considered good practice.

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
```diff
@@ -52,45 +52,61 @@ export class PostPopoverTrigger {
...
+    } else if ('ariaControlsElements' in this.trigger) {
+      // Use Elements Api ariaControlsElements property
+      this.trigger.ariaControlsElements = [popover];
     } else {
-      console.error(
-        'No content found in the post-popover-trigger slot. Please insert a focusable element or content that can receive focus.',
-      );
+      // Alternatively connect with a random ID
+      popover.id ||= `popover-${crypto.randomUUID()}`;
+      (this.trigger as HTMLElement).setAttribute('aria-controls', popover.id);
```
oliverschuerch: Why is the type assertion necessary here, but not on line 93?

myrtasakellariou: It happens because of the `else if ('ariaControlsElements' in this.trigger)` check, ts sees it as `never`. I updated with a helper function.

packages/components/src/components/post-popover-trigger/post-popover-trigger.tsx
myrtasakellariou: Good to know, thanx!

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -1,144 +1,220 @@
...
+  //   });
 
-    it('Has no detectable a11y violations on load', () => {
-      cy.checkA11y('post-popover');
-    });
-  });
+  //   it('Has no detectable a11y violations on load', () => {
+  //     cy.checkA11y('post-popover');
+  //   });
+  // });
```
oliverschuerch: Why?

oliverschuerch (changes requested):

packages/components/cypress/e2e/popover.cy.ts
myrtasakellariou: I left it commented out by mistake.

oliverschuerch (approved):

leagrdv (approved):

PR #6635, opened 2025-11-12
fix(components, eslint): missing required properties in child components

oliverschuerch:
## 📄 Description

This updates all properties of our stencil component that are defined as `required`, so that they are also mapped as attributes in the DOM, and adds a custom ESLint rule to ensure compliance.

## 📝 Checklist

### DIFF
--- a/packages/eslint/src/rules/stencil-no-unreflected-required-props.ts
+++ b/packages/eslint/src/rules/stencil-no-unreflected-required-props.ts
@@ -11,7 +11,7 @@ export default createRule({
     docs: {
       dir: 'ts',
       description:
-        'Reports any required Stencil component property, which is not reflected on its host element, since this would lead to errors in our property validations, when used in a SSR environment.',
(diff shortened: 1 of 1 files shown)


packages/eslint/src/rules/stencil-no-unreflected-required-props.ts
```diff
@@ -0,0 +1,66 @@
...
+export const name = 'stencil-no-unreflected-required-props';
+export const messageId = 'stencilNoUnreflectedRequiredProps';
+
+export default createRule({
+  name,
+  meta: {
+    docs: {
+      dir: 'ts',
+      description:
+        'Reports any required Stencil component property, which is not reflected on its host element, since this would lead to errors in our property validations, when used in a SSR environment.',
```
gfellerph: Just explaining the source of the error better so it's easier to understand why it's leading to errors.

```suggestion
        'Reports any required Stencil component property, which is not reflected on its host element, since this does not render the props on the server, thus leading to missing required prop errors in our property validations when used in a SSR environment.',
```

PR #6644, opened 2025-11-13
chore: remove deprecated package

gfellerph:
## 📄 Description

`@types/prettier` has been deprecated. It can be removed safely as it's only providing typing support for the prettier package.

### DIFF
--- a/packages/icons/package.json
+++ b/packages/icons/package.json
@@ -40,7 +40,6 @@
     "@types/mock-fs": "4.13.4",
     "@types/node": "22.10.5",
     "@types/node-fetch": "2.6.13",
-    "@types/prettier": "2.7.3",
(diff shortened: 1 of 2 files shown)


oliverschuerch (changes requested):
The deprecation message for the package has been added for version `3.x.x` only. We still use prettier `2.8.8` in the `@swisspost/design-system-icons` package and if we uninstall `@types/prettier@2.7.3`, typescript complains about missing types.

So no, we can't safely uninstall the `@types/prettier` package. First we need to migrate to `prettier@3.x`.
I've created the ticket https://github.com/swisspost/design-system/issues/6662 for that.

Message on npmjs.com:
Image

gfellerph:
Closing because depends on major prettier update.

PR #6661, opened 2025-11-13
feat(header): update header spacing and sizing

alizedebray:
## 📄 Description

Update header styles to match the Figma design. 
This PR focuses on the `post-header`, the `post-mainnavigation` and `post-megadropdown` will be finalized in another PR.

## 🚀 Demo

Updated version:

- [Header docs](https://preview-6661--swisspost-design-system-next.netlify.app/?path=/docs/27a2e64d-55ba-492d-ab79-5f7c5e818498--docs)
- Portal header: [user logged out](https://preview-6661--swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--portal) | [user logged in](https://preview-6661--swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--portal&args=isLoggedIn:true)
- Microsite header: [user logged out](https://preview-6661--swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--microsite) | [user logged in](https://preview-6661--swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--microsite&args=isLoggedIn:true)
- [Jobs header](https://preview-6661--swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--jobs)
- [One-Pager header](https://preview-6661--swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--one-pager)

Current version to compare to:

- [Header docs](https://swisspost-design-system-next.netlify.app/?path=/docs/27a2e64d-55ba-492d-ab79-5f7c5e818498--docs)
- Portal header: [user logged out](https://swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--portal) | [user logged in](https://swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--portal&args=isLoggedIn:true)
- Microsite header: [user logged out](https://swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--microsite) | [user logged in](https://swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--microsite&args=isLoggedIn:true)
- [Jobs header](https://swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--jobs)
- [One-Pager header](https://swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--one-pager)

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,7 +1,7 @@
 # Basic dependabot configuration file
 version: 2
 updates:
-  - package-ecosystem: 'pnpm'
+  - package-ecosystem: 'npm'
     directory: '/'
     schedule:
       interval: 'daily'


packages/components/src/components/post-mainnavigation/post-mainnavigation.scss
```diff
@@ -50,8 +50,16 @@ post-mainnavigation {
     &::after {
       content: '';
       @include icons.mask-image('chevrondown');
-      height: 24px;
-      width: 24px;
+      height: 1em;
+      width: 1em;
```
oliverschuerch, line 54: This means, we could actually use the `post-icon` mixin.

.changeset/hungry-bugs-grin.md
```diff
@@ -0,0 +1,8 @@
+---
+'@swisspost/design-system-styles': major
+'@swisspost/design-system-documentation': patch
+---
+
+Updated styles for elements slotted in the `post-header` component.
+Slotted lists should now omit the `.list-inline` class and will be automatically styled as part of the `post-header`.
+Keeping the `.inline-list` class will cause incorrect spacing between header elements.
```
oliverschuerch:
```suggestion
Keeping the `.list-inline` class will cause incorrect spacing between header elements.
```

.changeset/hungry-bugs-grin.md
alizedebray, line 1: To simplify the review, I only updated the header docs in this PR. All other implementations of the `post-header` have been updated in this other PR: https://github.com/swisspost/design-system/pull/6699.

packages/styles/src/components/header/_mixins.scss
```diff
@@ -1,16 +1,147 @@
...
+@use '../../mixins/button';
+@use '../../mixins/media';
+@use '../../tokens/components';
+
+/**
+ selects all slotted controls
+ */
+@mixin slotted-controls {
+  &:where(a, button, [role='button']),
+  &:where(ul) > li > :where(a, button, [role='button']) {
```
oliverschuerch: What do you mean, should we include `[role="link"]` and/or `[tabindex="0"]` as well, just to make sure they are styled correctly if developers insert such an element into a slot?

packages/styles/src/components/header/_mixins.scss
```diff
@@ -1,16 +1,147 @@
...
+  padding: 0.25rem 0.625rem;
+  gap: 0.375rem;
+  line-height: 1.5;
+  font-size: 1rem;
+  font-weight: 400;
+
+  > post-icon {
+    width: 1em;
+    height: 1em;
+  }
```
oliverschuerch: This does add properties, that are already defined on the `post-icon` by default. It only changes the specificity user would need to override the `width` and `height` properties.

packages/styles/src/components/header/_mixins.scss
```diff
@@ -1,16 +1,147 @@
...
+/**
+ default styles for header controls
+ */
+@mixin header-control-base {
+  @include button.button-base;
+  display: flex;
+  border: unset;
+  padding: 0.25rem 0.625rem;
+  gap: 0.375rem;
+  line-height: 1.5;
```
oliverschuerch: `line-height: 1.5` is the default on the `body` element and is never changed in between. Defininig it does not change anything.

I see that you probably added it, to ensure it is defined correctly, in case the `button-base` mixin should ever define it differently.
However, if this is the only reason for its presence, we could use `line-height: unset` instead, so it falls back to the `line-height` defined on the `body`.

packages/styles/src/components/header/_mixins.scss
```diff
@@ -1,16 +1,147 @@
...
+  font-size: 1rem;
```
oliverschuerch, line 35: While `font-size: 1rem` might be ok for most of the header links/buttons in the global header, it is currently not working or there is a potential risk for the folloging ones:

- Links/buttons in `post-megadropdown`.
  `font-size` should change between breakpoints.
  [Figma Design](https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-Next-Level?node-id=7421-60513&t=aSjitBotJoacCLbu-0)
- Links/buttons slotted in flyouts such as the user-menu, etc.
- Links/buttons slotted in the custom `navigation-controls` slot.
  Image

packages/styles/src/components/header/_mixins.scss
```diff
@@ -1,16 +1,147 @@
...
+  font-weight: 400;
```
oliverschuerch: `font-weight: 400` is the default on the `body` element and the only reason we need to re-define it here is, because the is set to `700` by the `button-base` mixin.

I would suggest to do the same thing as with the `line-height` property. Set `font-weight: unset` and let it fall back to the default `font-weight`, defined on the `body` element, instead of defining an explicit value.

packages/styles/src/mixins/_button.scss
```diff
@@ -10,8 +10,6 @@
 @use '../functions/tokens';
 @use '../tokens/components';
 
-tokens.$default-map: components.$post-button;
```
oliverschuerch, line 13: Did you had to do this because of a technical limitation?
e.g. using the mixin in a file which defines the `tokens.$default-map` variable on its own and therefore overrides the one defined here?

If so, we should do the following things:
- Create a ticket to remove the `tokens.$default-map` variable in all mixin files and instead implement the used map in the same way you did it here.
- Add an ADR on github, which explaines the decision and its necessity.

alizedebray: Mixins are using the `tokens.$default-map` that is declared in the file they're included in.
Therefore, you have to include the `components.$post-button` in all files where you want to use a button mixin but that is not always the map you would like to use. 
I'll open a ticket from your comment.

packages/styles/src/components/header/_mixins.scss
alizedebray: With this PR, only the links and buttons slotted in the `post-header` are using the mixins.
I am still working on updating the `post-mainnavigation` (PR https://github.com/swisspost/design-system/pull/6706), and other header components such as the `post-language-switch` and `post-megadropdown` to use the mixin there as well.
Yet, I'll update the `navigation-controls` in this PR.

packages/styles/src/components/header/_mixins.scss
alizedebray: I added the `[role="link"]` but I’m not sure about using `[tabindex="0"]`, it feels a bit too much of an edge case. What do you think?

packages/styles/src/components/header/_mixins.scss
oliverschuerch: What I meant is, the styles added with the mixin do leak also to some elements that should not receive these styles at all. We should implement more specific selectors (there were we use the mixin) to ensure, this does not happen.

oliverschuerch (changes requested):

packages/styles/src/components/header/_mixins.scss
oliverschuerch: If you can ensure this is not the case, I'm fine with it.
I think it looks better already, than when I checked it on the preview link the first time, but I'm not sure if it may leak to an element like the one I showed above.

Tell me if you're sure, then I will accept the PR ;)

alizedebray:
> If you can ensure this is not the case, I'm fine with it.
> I think it looks better already, than when I checked it on the preview link the first time, but I'm not sure if it may leak to an element like the one I showed above.
> 
> Tell me if you're sure, then I will accept the PR ;)


I checked the output and these styles are now only applied to links and buttons that are direct children of the `post-header` (either standalone or in a list):
image

They should therefore not leak into menus or drop-downs anymore.

PR #6663, opened 2025-11-14
feat(component): add `post-pagination` component

alionazherdetska:
## 📄 Description

Adds a new `post-pagination` web component (the initial styles were added just to be able to check the responsive behaviour works) for navigating through paginated content. The component features:

- Responsive design that adapts page button visibility based on available width
- Smart ellipsis handling for large page counts
- Full keyboard navigation support
- Accessibility with proper ARIA labels
- Disabled state support
- Auto-clamping of out-of-range page values

The component uses a hidden measurement system (similar to breadcrumbs) to calculate how many page buttons can fit in the available space, automatically collapsing intermediate pages into ellipses when needed.

**Note:** This is currently a raw component (only visible in dev mode). Final design system styling will be implemented in a follow-up ticket, that is why the changeset has not been added yet.

## 🚀 Preview link

[Pagination docs](https://preview-6663--swisspost-design-system-next.netlify.app/?path=/docs/d6f8b5c7-4e2a-4f3a-9d3a-1a2b3c4d5e6f--docs&devModeEnabled=true)

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 📝 Checklist


packages/documentation/src/stories/components/pagination/pagination.docs.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+### Many Pages
+
+When there are many pages, the component collapses intermediate pages into ellipses.
+
+<Canvas of={PaginationStories.ManyPages} />
+
+### Page Out Of Range
+
+If the provided `page` is outside the valid range, the component clamps it to the valid page.
```
myrtasakellariou: the nearest valid

packages/components/src/components/post-pagination/post-pagination.scss
```diff
@@ -0,0 +1,230 @@
+@use '@swisspost/design-system-styles/functions/breakpoint' as breakpoint;
```
myrtasakellariou, line 1: Minor observation: selected item has a hover state when the component is disabled.

packages/documentation/src/stories/components/pagination/pagination.stories.ts
```diff
@@ -0,0 +1,94 @@
...
+    labelFirst: 'First page',
+    labelLast: 'Last page',
+  },
+};
+
+export default meta;
+
+type Story = StoryObj;
+
+export const Default: Story = {
```
myrtasakellariou, line 99: It would be great if interacting with the code example updated args.page in real time.  e.g. 

```
    const [, updateArgs] = useArgs();
    return html`
      <post-pagination
        ${spread(attributes)}
        @postChange=${(e: CustomEvent) => {
          const newPage = e.detail;
          updateArgs({ page: newPage });
       ...
```

packages/components/src/components/post-pagination/post-pagination.scss
oliverschuerch: I didn't see that in chrome.
But what I saw is that horrible `not-allowed` cursor flickering around, when I hover the disabled elements.
I see why you added it, but since we don't use it on other disabled elements and I would be glad, if you could remove it for now and instead create a ticket to ask the team whether to add it on all disabled elements or not.

packages/components/src/components/post-pagination/post-pagination.scss
```diff
@@ -0,0 +1,228 @@
...
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border: 0;
+}
```
oliverschuerch, line 221: There is a mixin for that.

```scss

.visually-hidden {
  @include post.visuallyhidden;
}
```

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+
+/**
+ * Valid prop names for validation
+ */
+type ValidatableProp = 'page' | 'pageSize' | 'collectionSize' | 'label' | 'labelPrevious' | 'labelNext' | 'labelPage' | 'labelFirst' | 'labelLast' | 'disabled';
+
+@Component({
+  tag: 'post-pagination',
+  styleUrl: './post-pagination.scss',
+  shadow: false,
```
oliverschuerch: Please no!
Or in other words: Why is this even necessary?

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+   */
+  private getAvailableWidth(): number {
+    if (!this.navRef) return 0;
+    
+    let parent = this.host.parentNode;
+    while (parent && !(parent instanceof HTMLElement)) {
+      parent = parent.parentNode;
+    }
+    
+    return parent instanceof HTMLElement ? parent.clientWidth : window.innerWidth;
```
oliverschuerch: Check if you can use `parentElement` instead of `parentNode` and get rid of all the typing noise.

But thinking about it again, the whole function does not make sense to me? `parentNode` will always return a HTML element, because your host can't be nested in a text node. This will continue until you reach the `document` and if the while loop runs then again the parent will become `null`.
This means, below you'll always get the `window.innerWidth` as a return value and the whole function becomes unnecessary.

Please check that as well.

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+  private clampPageToValidRange(totalPages: number): number {
+    if (totalPages === 0 || this.collectionSize === 0 || isNaN(totalPages)) {
+      return 1;
+    }
+    if (!this.page || this.page < 1 || isNaN(this.page)) {
+      return 1;
+    }
+    if (this.page > totalPages) {
+      return totalPages;
+    }
```
oliverschuerch: Can you create a constant per if statement above with appropriate names (so other DEVs can understand which statement means what) and then put the if statemens in an if-else structure, for readability?

Also the first and second statements return the same value. Can we combine them?

e.g. `if(invalidTotalPages || invalidPage) { ... }`

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+    const totalPages = this.getTotalPages();
+    this.page = this.clampPageToValidRange(totalPages);
+    this.generatePages(totalPages);
+  }
+
+  /**
+   * Calculates the total number of pages.
+   */
+  private getTotalPages(): number {
+    if (this.collectionSize === 0 || this.pageSize === 0) {
```
oliverschuerch: I saw this statements mutliple times.
Think about creating a getter function to store it in one value.

e.g.
```ts
get invalidSize(): boolean {
  return this.collectionSize === 0 || this.pageSize === 0;
}
```

You can still combine it with other values.
e.g. `this.invalidSize || isNaN(totalPages)`

Only do that where it makes sense 😉

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+  }
+
+  componentWillLoad() {
+    this.paginationId = `pagination-${this.host.id || nanoid(6)}`;
+
+    if (this.page == null) {
+      this.page = 1;
+    }
+
+    this.runAllValidations();
```
oliverschuerch: Didn't we say `componentWillLoad` is to early for calling the prop validations?
`componentDidLoad` would be better.

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+  }
+
+  componentDidLoad() {
+    this.connected = true;
```
oliverschuerch: I would rename this var to `loaded` or `renderd`, because one could think it refers to the `connectedCallback` lifecycle hook.

optional 😉

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+    
+    this.debouncedResize = debounce(this.handleResizeInternal.bind(this), RESIZE_DEBOUNCE_MS);
```
oliverschuerch: Can't we define that where you initializes the class member?
I don't think you need to assign a dummy value first and only add this afterwards.
As long as the eventbinding is not done, it won't be called and you have also the `this.connected` guard within the callback function.

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+    
+    window.addEventListener('resize', this.debouncedResize);
+    this.lastWindowWidth = window.innerWidth;
```
oliverschuerch: `window` is available from the very beginning. So you can initialize the class member immediately, not only in a lifecycle hook.

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+
+    this.runAllValidations();
+  }
+
+  componentDidLoad() {
+    this.connected = true;
+    
+    this.debouncedResize = debounce(this.handleResizeInternal.bind(this), RESIZE_DEBOUNCE_MS);
+    
+    window.addEventListener('resize', this.debouncedResize);
```
oliverschuerch: This event binding on the other hand should be added as soon as possible, so you don't miss a callback.
Can you try what happens if you add it in the `connectedCallback`?

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+    }
+    
+    const canMeasure = this.navRef?.clientWidth > 0 && this.hiddenItemsRef;
+    
+    if (canMeasure) {
+      this.measureAndCalculateVisiblePages();
+    } else {
+      this.scheduleMeasurement();
+    }
+  }
```
oliverschuerch: Since you only call this method in the one above, you could nest it within the one above, which would automatically create a clearer scope and allow you to remove the redundant guard in the inner function.

```ts
private scheduleMeasurement() {
  if (!this.connected) {
    return;
  }
  
  this.measurementTimeoutId = window.setTimeout(attemptMeasurement, MEASUREMENT_DEBOUNCE_MS);
  
  // since this function lifes in the scope of its parent, you still have access to `this` within ;)
  function attemptMeasurement() {
    const canMeasure = ...
    ...
  }
}
```

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+
+  /**
+   * Internal resize handler
+   */
+  private handleResizeInternal() {
+    if (!this.connected) {
+      return;
+    }
+    
+    const newWidth = window.innerWidth;
```
oliverschuerch: `window.innerWith` is already a variable (or a reference to a value in the RAM), so there is no need to save it in another one.

e.g. `if (window.innerWidth === this.lastWindowWidth) {`
or `this.lastWindowWidth = window.innerWidth;`

is totally fine.

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+    
+    this.maxVisiblePages = clampedMaxPages;
+    this.updatePagesWithValidation();
+  }
+
+  /**
+   * Gets total width of control buttons (prev/next)
+   */
+  private getControlButtonsWidth(): number {
+    if (!this.hiddenItemsRef) return 0;
```
oliverschuerch, line 294: Try to be consistent with your if statments.
Some are one-liners some not. And I couldn't see how they differ from each other.

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+    
+    const controlButtons = Array.from(
+      this.hiddenItemsRef.querySelectorAll('.hidden-control-button')
+    );
+    
+    const totalWidth = controlButtons.reduce((sum, el) => {
+      return sum + el.getBoundingClientRect().width;
+    }, 0);
```
oliverschuerch:
```suggestion
    const totalWidth = controlButtons
      .reduce((sum, el) => sum + el.getBoundingClientRect().width, 0);
```

optional

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+
+  /**
+   * Compute total items that would be rendered for a given range
+   */
+  private computeTotalItems(startPage: number, endPage: number, totalPages: number): number {
+    const { leftSection, rightSection } = this.getSections(startPage, endPage, totalPages);
+    const middle = Math.max(0, endPage - startPage + 1);
+    const leftCount = leftSection === 'none' ? 0 : 1;
+    const rightCount = rightSection === 'none' ? 0 : 1;
+    return 2 + leftCount + rightCount + middle; // first + last + sections + middle
```
oliverschuerch: Why is `first` equal to 2?
Can you add a comment for future DEVs, please.

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,844 @@
...
+    
+    // Fix leftGap=2 to maintain symmetry
+    const leftGap = adjusted.startPage - 1;
+    if (leftGap === 2) {
+      adjusted.startPage = 2;
+      adjusted.endPage = Math.min(totalPages - 1, adjusted.startPage + middleSlots - 1);
+    }
+    
+    return adjusted;
+  }
```
oliverschuerch, line 436: The number `2` seems to be a constant, would be nice to add one (for example above the component class), with a clear name so one does understand what it stands for.
Then use it whereever it makes sense.

e.g. `const EDGE_ITEM_COUNT = 2;` 

optional and only if it makes sense.

packages/documentation/src/stories/components/pagination/pagination.docs.mdx
```diff
@@ -0,0 +1,42 @@
...
+  # Pagination
+
+  <link-design of={JSON.stringify(PaginationStories)}></link-design>
+</div>
+
+<PackageTag meta={meta} />
+
+<div className="lead">A pagination splits content into multiple pages and contains controls to navigate between these pages.</div>
+
+<Canvas sourceState="shown" of={PaginationStories.Default} />
```
oliverschuerch, line 18: The source code alwas showes the `disabled` property, no matter it is enabled or not.
Instead `disabled="false"` should not be rendered at all and the prop should only appear if the component is really disabled.

packages/documentation/src/stories/components/pagination/pagination.docs.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+When there are many pages, the component collapses intermediate pages into ellipses.
+
+<Canvas of={PaginationStories.ManyPages} />
+
+### Page Out Of Range
+
+If the provided `page` is outside the nearest valid range, the component clamps it to the valid page.
+
+<Canvas of={PaginationStories.PageOutOfRange} />
```
oliverschuerch: Without the source-code your explanation above is hard to understand.
```suggestion
<Canvas of={PaginationStories.PageOutOfRange} sourceState="shown" />
```

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,865 @@
...
+import { checkEmptyOrType, checkRequiredAndType, debounce } from '@/utils';
+
+const ELLIPSIS = '...';
+const MEASUREMENT_DEBOUNCE_MS = 50;
+const RESIZE_DEBOUNCE_MS = 150;
+const MIN_VISIBLE_PAGES = 3;
+const EDGE_ITEM_COUNT = 2; // Always show first and last page
+const MAX_ELLIPSIS_COUNT = 2; // Maximum ellipsis elements
+const GAP_THRESHOLD_FOR_PAGE = 2; // Show page number instead of ellipsis when gap is this value
+const MIDDLE_RANGE_START = 2; // Middle range starts from page 2 (page 1 is always shown separately)
```
oliverschuerch, line 13: 😍

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -0,0 +1,837 @@
...
+    const adjusted = { startPage, endPage };
+    let totalItems = this.computeTotalItems(adjusted.startPage, adjusted.endPage, totalPages);
+
+    // Trim if too many items
+    while (totalItems > maxVisible) {
+      const distLeft = currentPage - adjusted.startPage;
+      const distRight = adjusted.endPage - currentPage;
+      
+      if (distRight >= distLeft && adjusted.endPage > adjusted.startPage) {
+        adjusted.endPage = Math.max(adjusted.startPage - 1, adjusted.endPage - 1);
```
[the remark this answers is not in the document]
oliverschuerch: Would that work as well?
`adjusted.endPage = Math.max(adjusted.startPage, adjusted.endPage) - 1;`

oliverschuerch (approved):

oliverschuerch:
@alionazherdetska this is (imo) a very well coded component, congrats!

myrta2302 (approved):

PR #6670, opened 2025-11-17
chore: remove internet header package

alizedebray:
## 📄 Description

This PR removes the `@swisspost/internet-header` package from the `release/v7` branch. The package is now only maintained from the `release/v9` (current) and `main` (next) branches.

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist

### DIFF
--- a/.changeset/loud-falcons-dance.md
+++ b/.changeset/loud-falcons-dance.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': patch
+---
+
+Updated the `@swisspost/internet-header` documentation to reflect the latest package version.


.changeset/loud-falcons-dance.md
oliverschuerch, line 1: Maybe it is good to have it for ourself, even when for the visitors of the documentation, nothing really changes.

packages/documentation/src/stories/getting-started/changelogs/internet-header/changelog.stories.ts
oliverschuerch, line 1: After a short discussion with @gfellerph, he brought up the following, and I actually agree: "We could simply set `private: true` in the `package.json` of the `@swisspost/internet-header` package.
This way we would make sure, it can't be released by accident, but we could also keep the documentation and everything as is."

PR #6696, opened 2025-11-19
fix(components): prevent animation functions from being executed on server side in `post-header` and `post-collapsible`

oliverschuerch:
## 📄 Description

- Implements an optional chaining operator in the `post-header`, so we don't call `animation.finish` by accident, when `animation` is not defined (e.g. on the server).
- Implements SSR specific guards in the `post-collapsible` component.
- Refactors the `post-collapsible` toggle function, by improving the behavior when user perfers reduced motion.

## 🚀 Demo

---

## 🔮 Design review

- [x] No design review needed

## 📝 Checklist


packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -48,40 +46,39 @@ export class PostCollapsible {
...
-    this.isLoaded = true;
-
     this.updateTriggers();
   }
 
   /**
    * Triggers the collapse programmatically.
-   *
    * If there is a collapsing transition running already, it will be reversed.
+   * If is called accidentally on the server or if the requested state is already set, it will do nothing.
```
alizedebray: The `toggle` method was not called accidentally, it was called from line 37. If you have `<post-collapsible collapsed="true"/>` the toggle method is called when the component loads to have it collapsed by default (the accordion also sets this property). If the `toggle` does nothing at this stage, the component does not work as expected.

packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -48,40 +46,39 @@ export class PostCollapsible {
...
-    this.isOpen = open;
-    this.collapsed = !open;
-    if (this.isLoaded) this.postToggle.emit(open);
-
     const animation = open ? expand(this.host) : collapse(this.host);
 
-    if (!this.isLoaded || isMotionReduced()) animation.finish();
-
-    await animation.finished;
+    if (animation && animation.finished) {
```
alizedebray: `animation` is now always defined

packages/components/src/components/post-collapsible/post-collapsible.tsx
oliverschuerch: Sure, but the components lifecycle hooks (e.g. `componentDidLoad`) should not be called on the server. Or that's at least what I have thought. Since I'm scrolling through stencils hydrate app output files from time to time, I think it actually calls the lifecycle hooks, if they are defined. That explains a lot.

I'll remove the comment.

packages/components/src/components/post-collapsible/post-collapsible.tsx
oliverschuerch: Sadly it's not during server side rendering.

Stencils hydrate app is mocking a lot of browser specific things (e.g. `window`, etc.), so it is able to pre-render the components on the server.
But it is not mocking the `HTMLElement` [Animation API](https://developer.mozilla.org/en-US/docs/Web/API/Animation).
And adding a `IS_SERVER` guard did not help at all.

I'm still trying to figure out how we can detect that, because `const IS_SERVER = typeof window === 'undefined';` does not work, since we updated some packages to their latest versions.

oliverschuerch: Instead of checking the animation, I fixed the IS_BROWSER and IS_SERVER utility flags (it only took me 2 days 😢) and used them in the component. This way it's clearer for devs, why this check is necessary.

Reason: The hydrate app mocks a lot of things (e.g. `window`, `document`, etc.) and calls every lifecycle hook since we updated it to the latest version. Sadly it seems not to mock everything (e.g. the Animation API, etc.). Using functionality like that, without a proper `if(IS_BROWSER) { ... }` guard, can lead to fatal errors on the server.

packages/components/src/utils/tests/environment/post-env-test.tsx
```diff
@@ -0,0 +1,12 @@
+import { Component, h } from '@stencil/core';
+import { IS_BROWSER, IS_SERVER } from '../../environment';
+
+@Component({
+  tag: 'post-env-test',
+  shadow: true,
+})
+export class PostEnvTest {
```
alizedebray, line 8: It this component necessary?

packages/components/src/components/post-collapsible/post-collapsible.tsx
```diff
@@ -93,7 +87,11 @@ export class PostCollapsible {
 
   render() {
     return (
-      <Host data-version={version} tabindex={this.collapsed ? -1 : undefined}>
+      <Host
+        data-version={version}
+        tabindex={this.collapsed ? -1 : undefined}
+        style={IS_SERVER && this.collapsed ? (collapsedKeyframe as InlineStyles) : undefined}
```
alizedebray: We could probably use the same approach on browser side: 
- always initialize the component using `collapsedKeyframe` in the `style` attribute
- use the `toggle` method only to toggle the collapsible once it's loaded

alizedebray (changes requested):

packages/components/src/utils/tests/environment/post-env-test.tsx
oliverschuerch: Yes, the component is used during the unit tests of `IS_BROWSER` and `IS_SERVER`.

alizedebray (approved):

Issue #6473, opened 2025-10-20
Improve header controls


alizedebray:
In the header docs we are currently restricted some blocks from being displayed together. 
For example, if a title is set, the header is considered as a micro-site header and we are restricting having the global login. However the "global login" control can still be set to true and nothing happens.

There are therefore two solutions: 
1. we update the controls to conditionally hide some depending on what is selected for others 
2. we update the renderer and allow full configuration of the header by users

alionazherdetska:
**Proposal after DevRound table discussion:** 
- we update the renderer and allow full configuration of the header by users.

gfellerph:
@alizedebray organise a naming session to bring clarity to the different header implementation variants and configurations. The goal is to define common names that we can communicate to designers and devs to make it clear how they can get the header they need in the end.

gfellerph:
**Resolution:**

### Self managed header/breadcrumbs/footer

Projects assemble the header components themselves and thus have fine grained control over all aspects like URLs, translations, direct access to HTML elements for tracking. This is the Ikea type header where understanding the structure and reading the docs is required for success.

### Pre-configured header/breadcrumbs/footer

Projects have a very easy implementation path. This is the "just drop one line and you're done", worry-free, all-inclusive header. The configuration (URLs and translations) are managed by POW.

PR #6747, opened 2025-11-26
chore(components): merge local navigation slots

alizedebray:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/documentation/src/stories/components/header/header.stories.ts
+++ b/packages/documentation/src/stories/components/header/header.stories.ts
@@ -5,8 +5,8 @@ import { fakeContent } from '@/utils';
 import { renderMainnavigation } from '@/stories/components/header/renderers/main-navigation';
 import { renderMetaNavigation } from '@/stories/components/header/renderers/meta-navigation';
 import { renderTargetGroup } from '@/stories/components/header/renderers/target-group';
-import { renderLocalControls } from '@/stories/components/header/renderers/local-controls';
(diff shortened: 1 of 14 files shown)


packages/documentation/src/stories/components/header/header.stories.ts
```diff
@@ -31,7 +31,7 @@ const meta: MetaComponent = {
     globalControls: true,
     targetGroup: true,
     globalLogin: true,
-    localControls: false,
+    localNav: false,
     isLoggedIn: false,
     jobs: false,
```
leagrdv, line 36: I'm wondering if we should keep the **Jobs** argument here. I feel like it's not really a state but more of a variant of the header layout (like Post portal, Microsite, etc). The jobs arg was previously defining whether the controls were at the top or bottom of the local header but now that we base it on the presence of the title, I feel like we could remove it. What do you think?

leagrdv (approved):

packages/documentation/src/stories/components/header/header.stories.ts
alizedebray: For the moment it is still used to set the `aria-current="location"` attribute on the "Jobs" link instead of a target group, and then hide the global login. But there is a ticket to improve the header controls altogether: [#6473](https://github.com/swisspost/design-system/issues/6473), maybe we can then remove or hide the "Jobs" control if it is not useful

PR #6805, opened 2025-12-05
chore(docs, styles): add `.form-check-sm` class to htmlcss radio button and add size variant on stories

bucknatt:
## 📄 Description

The CSS  `.form-check-sm` class is added to the Form Radio Button  component. The variant `Size` has been added on the component's documentation.

## 🚀 Demo

For Radio Button
image

For Radio Group
image

---

## 🔮 Design review

- [x] Design review done
- [ ] No design review needed

## 📝 Checklist

### DIFF
--- a/.changeset/cyan-months-float.md
+++ b/.changeset/cyan-months-float.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': patch
+---
+
+Addressed inconsistencies across the form elements documentation and updated the validation guidelines for radio button elements.
--- a/packages/documentation/package.json
+++ b/packages/documentation/package.json
@@ -77,7 +77,7 @@
     "storybook": "9.1.19",
     "typescript": "5.8.3",
     "typescript-eslint": "8.20.0",
-    "vite": "7.1.12"
+    "vite": "7.3.2"
   },
   "keywords": [
     "storybook",
--- a/packages/documentation/src/stories/components/forms/validation/validation.docs.mdx
+++ b/packages/documentation/src/stories/components/forms/validation/validation.docs.mdx
@@ -42,12 +42,22 @@ These classes control the display of corresponding feedback messages that should
 
 ### Radio Button
 
+Radio buttons are always used as part of a group. Validation is applied to the group as a whole, not to individual radio buttons.
+
+This example demonstrates how a validation message can be programmatically associated with a single radio input. However, in real-world usage, validation should be implemented at the group level.
+
 <Canvas sourceState="shown" of={ValidationStories.RadioButton} />
 <div className="hide-col-default">
   <Controls of={ValidationStories.RadioButton} />
 </div>
 
-### Radio Button Inline Group
+### Radio Button Group
+
+Radio buttons are validated as a group (all inputs sharing the same `name`), not individually. Validation feedback must be linked to every radio input using `aria-describedby`.
+
+When the group is `required` and no option is selected, `aria-invalid="true"` needs to be applied to all radio inputs.
+
+The `required` attribute only needs to be set on one radio input in the group.
 
 <Canvas sourceState="shown" of={ValidationStories.RadioButtonGroup} />
 <div className="hide-col-default">
--- a/packages/documentation/src/stories/components/forms/validation/validation.stories.ts
+++ b/packages/documentation/src/stories/components/forms/validation/validation.stories.ts
@@ -45,12 +45,16 @@ export function getValidationProps(component: string, args: Args) {
   };
 }
 
-function renderFeedback(validFeedbackId: string | symbol, invalidFeedbackId: string | symbol) {
+function renderFeedback(
+  validFeedbackId: string | symbol,
+  invalidFeedbackId: string | symbol,
+  validationState: string,
+) {
   return html`
-    ${validFeedbackId
+    ${validFeedbackId && validationState === 'is-valid'
       ? html`<p id="${validFeedbackId}" class="valid-feedback">Valid message.</p>`
       : nothing}
-    ${invalidFeedbackId
+    ${invalidFeedbackId && validationState === 'is-invalid'
       ? html`<p id="${invalidFeedbackId}" class="invalid-feedback">Invalid message.</p>`
       : nothing}
   `;
@@ -84,16 +88,20 @@ export const Checkbox: Story = {
   render: (args: Args, context: StoryContext) => {
     const component = context.name.replace(/\s+/g, '');
     const props = getValidationProps(component, args);
-    const feedbackTemplate = renderFeedback(props.validFeedbackId, props.invalidFeedbackId);
+    const feedbackTemplate = renderFeedback(
+      props.validFeedbackId,
+      props.invalidFeedbackId,
+      props.validationState,
+    );
     return html`<div class="form-check">
       <input
         type="checkbox"
-        id="Checkbox_1${props.validationState}${props.scheme}"
+        id="checkbox_1${props.validationState}${props.scheme}"
         class="form-check-input ${props.isValidationSet ? props.validationState : ''}"
         aria-invalid=${props.ariaInvalid}
         aria-describedby="${props.ariaDescribedBy}"
       />
-      <label class="form-check-label" for="Checkbox_1${props.validationState}${props.scheme}"
+      <label class="form-check-label" for="checkbox_1${props.validationState}${props.scheme}"
         >Label</label
       >
       ${feedbackTemplate}
@@ -111,19 +119,23 @@ export const Input: Story = {
   render: (args: Args, context: StoryContext) => {
     const component = context.name.replace(/\s+/g, '');
     const props = getValidationProps(component, args);
-    const feedbackTemplate = renderFeedback(props.validFeedbackId, props.invalidFeedbackId);
+    const feedbackTemplate = renderFeedback(
+      props.validFeedbackId,
+      props.invalidFeedbackId,
+      props.validationState,
+    );
     return html`<div class="form-floating">
       <input
-        id="Input_1${props.validationState}${props.scheme}"
+        id="input_1${props.validationState}${props.scheme}"
         class="form-control form-control-lg ${props.isValidationSet ? props.validationState : ''}"
         aria-invalid=${props.ariaInvalid}
-        aria-describedby="Input_1-form-hint${props.validationState}${props.scheme} ${props.ariaDescribedBy}"
+        aria-describedby="input_1-form-hint${props.validationState}${props.scheme} ${props.ariaDescribedBy}"
         type="text"
         placeholder="Placeholder"
       />
-      <label class="form-label" for="Input_1${props.validationState}${props.scheme}">Label</label>
+      <label class="form-label" for="input_1${props.validationState}${props.scheme}">Label</label>
       ${feedbackTemplate}
-      <p class="form-hint" id="Input_1-form-hint${props.validationState}${props.scheme}">
+      <p class="form-hint" id="input_1-form-hint${props.validationState}${props.scheme}">
         This is helpful text that provides guidance or additional information to assist the user in
         filling out this field correctly.
       </p>
@@ -141,18 +153,22 @@ export const RadioButton: Story = {
   render: (args: Args, context: StoryContext) => {
     const component = context.name.replace(/\s+/g, '');
     const props = getValidationProps(component, args);
-    const feedbackTemplate = renderFeedback(props.validFeedbackId, props.invalidFeedbackId);
+    const feedbackTemplate = renderFeedback(
+      props.validFeedbackId,
+      props.invalidFeedbackId,
+      props.validationState,
+    );
     return html` <legend>Legend</legend>
       <div class="form-check form-check-inline">
         <input
           type="radio"
-          id="Radio_1${props.validationState}${props.scheme}"
-          name="radio_group"
+          id="radio-single${props.validationState}${props.scheme}"
+          name="radio-single"
           class="form-check-input ${props.isValidationSet ? props.validationState : ''}"
           aria-invalid=${props.ariaInvalid}
           aria-describedby="${props.ariaDescribedBy}"
         />
-        <label class="form-check-label" for="Radio_1${props.validationState}${props.scheme}"
+        <label class="form-check-label" for="radio-single${props.validationState}${props.scheme}"
           >Option 1</label
         >
         ${feedbackTemplate}
@@ -170,45 +186,35 @@ export const RadioButtonGroup: Story = {
   render: (args: Args, context: StoryContext) => {
     const component = context.name.replace(/\s+/g, '');
     const props = getValidationProps(component, args);
-    const feedbackTemplate = renderFeedback(props.validFeedbackId, props.invalidFeedbackId);
+    const feedbackTemplate = renderFeedback(
+      props.validFeedbackId,
+      props.invalidFeedbackId,
+      props.validationState,
+    );
+
     return html`<fieldset>
       <legend>Legend</legend>
-      <div class="form-check form-check-inline">
-        <input
-          type="radio"
-          id="Radio_10${props.validationState}${props.scheme}"
-          name="radio_group"
-          class="form-check-input ${props.isValidationSet ? props.validationState : ''}"
-          aria-invalid=${props.ariaInvalid}
-          aria-describedby="${props.ariaDescribedBy}"
-          required
-        />
-        <label class="form-check-label" for="Radio_10${props.validationState}${props.scheme}"
-          >Option 1</label
-        >
-      </div>
-      <div class="form-check form-check-inline">
-        <input
-          type="radio"
-          id="Radio_11${props.validationState}${props.scheme}"
-          name="radio_group"
-          class="form-check-input"
-        />
-        <label class="form-check-label" for="Radio_11${props.validationState}${props.scheme}"
-          >Option 2</label
-        >
-      </div>
-      <div class="form-check form-check-inline">
-        <input
-          type="radio"
-          id="Radio_12${props.validationState}${props.scheme}"
-          name="radio_group"
-          class="form-check-input"
-        />
-        <label class="form-check-label" for="Radio_12${props.validationState}${props.scheme}"
-          >Option 3</label
-        >
-      </div>
+      ${Array.from({ length: 3 }, (_, i) => {
+        return html`
+          <div class="form-check">
+            <input
+              type="radio"
+              id="radio_group_${i + 1}${props.validationState}${props.scheme}"
+              name="radio_group"
+              class="form-check-input ${props.isValidationSet ? props.validationState : ''}"
+              aria-invalid=${props.ariaInvalid}
+              aria-describedby="${props.ariaDescribedBy}"
+              ?required="${i === 0}"
+            />
+            <label
+              class="form-check-label"
+              for="radio_group_${i + 1}${props.validationState}${props.scheme}"
+            >
+              Option ${i + 1}
+            </label>
+          </div>
+        `;
+      })}
       ${feedbackTemplate}
     </fieldset>`;
   },
@@ -224,13 +230,17 @@ export const Select: Story = {
   render: (args: Args, context: StoryContext) => {
     const component = context.name.replace(/\s+/g, '');
     const props = getValidationProps(component, args);
-    const feedbackTemplate = renderFeedback(props.validFeedbackId, props.invalidFeedbackId);
+    const feedbackTemplate = renderFeedback(
+      props.validFeedbackId,
+      props.invalidFeedbackId,
+      props.validationState,
+    );
     return html`<div class="form-floating">
       <select
-        id="Select_1${props.validationState}${props.scheme}"
+        id="select_1${props.validationState}${props.scheme}"
         class="form-select form-select-lg ${props.isValidationSet ? props.validationState : ''}"
         aria-invalid=${props.ariaInvalid}
-        aria-describedby="Select_1-form-hint${props.validationState}${props.scheme} ${props.ariaDescribedBy}"
+        aria-describedby="select_1-form-hint${props.validationState}${props.scheme} ${props.ariaDescribedBy}"
       >
         <option>Select option...</option>
         <option value="value_1">Option 1</option>
@@ -238,11 +248,11 @@ export const Select: Story = {
         <option value="value_3">Option 3</option>
         <option value="value_4">Option 4</option>
       </select>
-      <label class="form-label" for="Select_1${props.validationState}${props.scheme}">
+      <label class="form-label" for="select_1${props.validationState}${props.scheme}">
         <span>Label</span>
       </label>
       ${feedbackTemplate}
-      <p class="form-hint" id="Select_1-form-hint${props.validationState}${props.scheme}">
+      <p class="form-hint" id="select_1-form-hint${props.validationState}${props.scheme}">
         This is helpful text that provides guidance or additional information to assist the user in
         filling out this field correctly.
       </p>
@@ -260,19 +270,23 @@ export const Switch: Story = {
   render: (args: Args, context: StoryContext) => {
     const component = context.name.replace(/\s+/g, '');
     const props = getValidationProps(component, args);
-    const feedbackTemplate = renderFeedback(props.validFeedbackId, props.invalidFeedbackId);
+    const feedbackTemplate = renderFeedback(
+      props.validFeedbackId,
+      props.invalidFeedbackId,
+      props.validationState,
+    );
     return html`<div class="form-check form-switch">
       <input
         type="checkbox"
         role="switch"
-        id="Switch_1${props.validationState}${props.scheme}"
+        id="switch_1${props.validationState}${props.scheme}"
         class="form-check-input ${props.isValidationSet ? props.validationState : ''}"
         aria-invalid=${props.ariaInvalid}
         aria-describedby="${props.ariaDescribedBy}"
       />
       <label
         class="form-check-label order-first"
-        for="Switch_1${props.validationState}${props.scheme}"
+        for="switch_1${props.validationState}${props.scheme}"
         >Notifications</label
       >
       ${feedbackTemplate}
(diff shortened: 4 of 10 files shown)


packages/styles/src/components/radio-button.scss
```diff
@@ -123,3 +123,25 @@ tokens.$default-map: components.$post-radio-button;
     }
   }
 }
+
+.form-check-sm:not(.form-switch):has(input[type='radio']) {
```
leagrdv: When small radios are in a group (e.g. the grouped example in the docs), if radios have the `.form-check-sm` class, the spacing between each should be different, which is not the case right now.

leagrdv (dismissed):
You can see in the screenshot (desktop) that the small version has the radio misaligned with the text.

You can go through the checklist of the ticket to make sure you've not forgotten any task. The last element "Design review" should be done by a designer: Once you've reviewed the design yourself to make sure what you've implemented looks the same as on figma, you can contact Sandra and ask her to review the small version of the component.
Capture d&#39;écran 2025-12-08 103958

sandra-post:
**Design Review**
All looks good, except font size:
- IS: 16 / 14 / 14
- SHOULD: 14 / 14 / 14

I see that this has been missed to tokenize. I will therefore add the missing token.

.changeset/funny-rabbits-think.md
```diff
@@ -0,0 +1,6 @@
+---
+"@swisspost/design-system-documentation": minor
+"@swisspost/design-system-styles": minor
+---
+
+The CSS `.form-check-sm` class has been added to the Form Radio Button component and the variant size is added to the component's documunetation.
```
alizedebray: Can you rephrase the message? Changeset entries must start with a verb in past tense, like "Added [...]."

packages/documentation/src/stories/components/radio/radio.snapshot.stories.ts
```diff
@@ -44,10 +44,52 @@ export const Radio: Story = {
...
+              const id = `${scheme}-${crypto.randomUUID()}`;
+              return meta.render?.(
+                { ...context.args, ...args, name: `${scheme}-snapshot-group` },
+                { ...context, id },
+              );
             })}
         </div>
+
+        <!-- Radio Group/Vertical (Grouped) -->
+        <div class="mt-16">
```
leagrdv: Only the large radio group vertical and horizontal are being rendered here. Could you add the small ones as well?

packages/documentation/src/stories/components/radio/radio.snapshot.stories.ts
```diff
@@ -44,10 +44,52 @@ export const Radio: Story = {
...
+                checkedRadio: null,
+              },
+              groupContext,
+            );
+          })()}
+        </div>
+
+        <!-- Radio Group/Horizontal (Inline) -->
+        <div class="mt-16">
+          ${(() => {
```
leagrdv: You could put those two renders together and define it only once using `bombArgs` to have them both horizontal and vertical, as well as large and small

packages/styles/src/components/radio-button.scss
```diff
@@ -123,3 +123,61 @@ tokens.$default-map: components.$post-radio-button;
...
+.form-check-sm:not(.form-switch) {
+  display: flex;
+  align-items: center;
+}
+
+/* Radio Group/Vertical small */
+fieldset .form-check-sm:not(.form-switch, .form-check-inline, :last-of-type) {
+  display: flex;
+  align-items: center;
+  margin-block-end: tokens.get('radio-button-small-group-items-gap-block');
```
leagrdv: If you inspect the CSS, you can see that the `margin-bottom` set on the `.form-check` in the file `_form-check.scss` is overriding the `margin-block-end` you've defined for the `.form-check-sm`, meaning the gap between small elements within a fieldset is not correct.

When applying a property on an element, it's always good to use the inspector to make sure it is correctly applied, especially when the difference it brings might be hard to see.
Capture d&#39;écran 2025-12-12 154050

packages/documentation/src/stories/components/radio/radio.stories.ts
```diff
@@ -166,11 +183,38 @@ export const Default: Story = {};
 
 export function renderGroup(args: Args, context: Partial<StoryContext>) {
   const [_, updateArgs] = useArgs();
-  const baseId = `${context.viewMode}_${context.name?.replace(/\s/g, '-')}_ExampleRadio`;
-  const id1 = baseId + '1';
-  const id2 = baseId + '2';
-  const id3 = baseId + '3';
-  const id4 = baseId + '4';
+  // Ensure a unique suffix for ids: prefer provided context.id, otherwise try crypto.randomUUID(),
```
leagrdv: Is all of this necessary? Simply using `crypto.randomUUID()` on other stories seems to be working

leagrdv (changes requested):

packages/documentation/src/stories/components/radio/radio.stories.ts
bucknatt: **Before**: Each input needs a unique `id` so that the label’s `for` attribute points to the correct input. Generating a suffix (via `randomUUID` or `context.id`) prevents collisions when multiple stories/components are rendered on the same page. This is why we needed to generate a unique ID.

**New**: `baseId` serves as a common root for `id1..id4` and for the group’s `name` attribute. This way, all buttons within the same group share the same `name` (correct radio-button behavior) while remaining isolated from other groups.
As the e2e tests failed (by run the command `pnpm run docs:e2e:watch`) on the `accessibility violations`, I had to change the unique ID generation for grouping. To ensure stability for the snapshots tests: `context.id` (when provided by Storybook) gives a deterministic identifier, and `randomUUID()` provides a unique fallback when `context.id` does not exist.

packages/documentation/src/stories/components/radio/radio.snapshot.stories.ts
```diff
@@ -11,6 +11,13 @@ export default {
   title: 'Snapshots',
 };
 
+// Simple id generator for snapshots: timestamp + counter
+let __snapshotCounter = 0;
+function snapshotId(prefix = '') {
+  __snapshotCounter += 1;
+  return `${prefix}${Date.now().toString(36)}-${__snapshotCounter.toString(36)}`;
+}
+
```
alizedebray: We are usually using `nanoid`, you can see it used in other file if you need examples on how to use it.

packages/documentation/src/stories/components/radio/radio.stories.ts
```diff
@@ -27,127 +27,100 @@ const meta: MetaComponent = {
...
       name: 'Hidden Legend',
       description: 'Render the group with or without a visible legend.',
-      control: {
-        type: 'boolean',
-      },
-      table: {
-        category: 'General',
-      },
+      control: { type: 'boolean' },
+      table: { category: 'General' },
```
alizedebray: Why was this refactored?

.changeset/funny-rabbits-think.md
leagrdv:
```suggestion
Added the small variant class `.form-check-sm` to the form radio button.
```

packages/documentation/src/stories/components/radio/radio.snapshot.stories.ts
```diff
@@ -11,6 +11,13 @@ export default {
   title: 'Snapshots',
 };
 
+// Simple id generator for snapshots: timestamp + counter
+let __snapshotCounter = 0;
+function snapshotId(prefix = '') {
+  __snapshotCounter += 1;
+  return `${prefix}${Date.now().toString(36)}-${__snapshotCounter.toString(36)}`;
+}
```
leagrdv: This can be replaced with the crypto UUID generation

packages/documentation/src/stories/components/radio/radio.snapshot.stories.ts
```diff
@@ -44,10 +51,36 @@ export const Radio: Story = {
                 !(args.requiredOptional === 'required' && args.disabled === true),
             )
             .map((args: Args) => {
-              context.id = `${scheme}-${crypto.randomUUID()}`;
-              return meta.render?.({ ...context.args, ...args }, context);
+              const id = `${scheme}-${snapshotId()}`;
+              return meta.render?.(
+                { ...context.args, ...args, name: `${scheme}-snapshot-group` },
```
leagrdv: Is the name needed here?

packages/documentation/src/stories/components/radio/radio.snapshot.stories.ts
```diff
@@ -27,13 +34,13 @@ export const Radio: Story = {
               checked: [false, true],
               disabled: [false, true],
               size: ['null', 'form-check-sm'],
-              validation: context.argTypes.validation.options,
+              validation: Array.from(context.argTypes.validation.options ?? []),
```
leagrdv: Is this modification needed?

packages/documentation/src/stories/components/radio/radio.snapshot.stories.ts
```diff
@@ -44,10 +51,36 @@ export const Radio: Story = {
                 !(args.requiredOptional === 'required' && args.disabled === true),
             )
             .map((args: Args) => {
-              context.id = `${scheme}-${crypto.randomUUID()}`;
-              return meta.render?.({ ...context.args, ...args }, context);
+              const id = `${scheme}-${snapshotId()}`;
+              return meta.render?.(
+                { ...context.args, ...args, name: `${scheme}-snapshot-group` },
+                { ...context, id },
+              );
```
leagrdv: You've made some modifications that complicates the logic here that don't seem to change the output in the snapshot. Was there a reason for the changes?

packages/documentation/src/stories/components/radio/radio.stories.ts
```diff
@@ -27,127 +27,100 @@ const meta: MetaComponent = {
...
-        category: 'States',
-      },
+      control: { type: 'boolean' },
+      table: { category: 'States' },
     },
     validation: {
       name: 'Validation',
       description:
-        'Defines the validation state of the radio button and controls the display of the corresponding return message. <post-banner data-size="sm"><p>Please read our <a href="/?path=/docs/1aa900d9-aa65-4ae0-b8cd-e6cca6cc3472--docs#radio-button">validation guidelines here</a>.</p></post-banner> ',
+        'Defines the validation state of the radio button and controls the display of the corresponding return message. <post-banner data-size="sm"><p>Please read our <a href="/?path=/docs/1aa900d9-aa65-4ae0-b8cd-e6cca6cc3472--docs">validation guidelines here</a>.</p></post-banner> ',
```
leagrdv: Why did you change the link here?

packages/documentation/src/stories/components/radio/radio.stories.ts
```diff
@@ -27,127 +27,100 @@ const meta: MetaComponent = {
...
       type="radio"
       ?checked="${args.checked}"
       .checked="${args.checked}"
       ?disabled="${args.disabled}"
       aria-label="${useAriaLabel ? args.label : nothing}"
       ?aria-invalid="${VALIDATION_STATE_MAP[args.validation]}"
-      aria-describedby="${args.validation != 'null'
-        ? `${args.validation}-id-${context.id}`
-        : nothing}"
+      aria-describedby="${args.validation != 'null' ? `${args.validation}-desc-${id}` : nothing}"
```
leagrdv: You've changed the ID here but not on the actual description, which means this `aria-describedby` leads to a non-existing element. Please put it back to what it was before

packages/documentation/src/stories/components/radio/radio.stories.ts
```diff
@@ -184,46 +158,54 @@ export function renderGroup(args: Args, context: Partial<StoryContext>) {
     }
   }
 
+  const itemClass = [
+    'form-check',
+    args.size && args.size !== 'null' ? args.size : undefined,
```
leagrdv: Can you please add examples on the docs for the grouped small and inline grouped small as well for the radio?

packages/documentation/src/stories/components/radio/radio.stories.ts
```diff
@@ -166,16 +139,17 @@ export const Default: Story = {};
...
 export function renderGroup(args: Args, context: Partial<StoryContext>) {
   const [_, updateArgs] = useArgs();
-  const baseId = `${context.viewMode}_${context.name?.replace(/\s/g, '-')}_ExampleRadio`;
-  const id1 = baseId + '1';
-  const id2 = baseId + '2';
-  const id3 = baseId + '3';
-  const id4 = baseId + '4';
+  const generatedUuid: string = randomUUID();
+  const uniqueSuffix: string = context.id ?? generatedUuid;
+  const baseId = `${context.viewMode ?? 'view'}_${(context.name ?? '').replace(/\s/g, '-')}_${uniqueSuffix}_ExampleRadio`;
```
leagrdv: Why did you modify the way the IDs are generated here?

packages/documentation/src/stories/components/radio/radio.stories.ts
```diff
@@ -237,50 +219,30 @@ export function renderGroup(args: Args, context: Partial<StoryContext>) {
...
+// To generate secure Id, prefer crypto.randomUUID,
+// else a simple timestamp+counter.
+export function randomUUID(): string {
+  const maybeCrypto = (globalThis as unknown as { crypto?: { randomUUID?: () => string } }).crypto;
+  if (maybeCrypto && typeof maybeCrypto.randomUUID === 'function') return maybeCrypto.randomUUID();
+  // fallback: timestamp + counter as short id
+  const g = globalThis as unknown as { __randomUUIDFallbackCounter?: number };
+  g.__randomUUIDFallbackCounter = (g.__randomUUIDFallbackCounter || 0) + 1;
+  return `${Date.now().toString(36)}-${g.__randomUUIDFallbackCounter.toString(36)}`;
+}
```
leagrdv: Why all of this instead of `crypto.randomUUID()`?

packages/styles/src/components/_form-check.scss
```diff
@@ -164,3 +169,155 @@
     }
   }
 }
+
+// Checkbox Group
+fieldset {
+  &:has(.form-check-inline) {
+    display: flex;
+    flex-wrap: wrap;
+    align-items: flex-start;
```
leagrdv: Removing those lines doesn't seem to change anything.

The changes in the CSS are pretty much the same on this PR than the checkbox one, so please refer to the other PR to update this one.

Also this PR seems to break the checkboxes:

Image

packages/styles/src/components/radio-button.scss
```diff
@@ -13,113 +13,138 @@ tokens.$default-map: components.$post-radio-button;
...
 
-.form-check > input[type='radio'] {
-  appearance: none;
-  flex: none;
-  display: inline-flex;
-  justify-content: center;
+// Default Radio
+.form-check:not(.form-switch):has(input[type='radio']) {
+  display: flex;
   align-items: center;
```
leagrdv: This is already defined in the main form-check file, please remove it

packages/documentation/src/stories/components/radio/radio.snapshot.stories.ts
bucknatt: The second line converts objects into and array and returns an empty array in case that options is null oder undefined: `options = undefined  → []`
I don't that this is really needed. 

Before it was: `options = undefined → validation = undefined`. The same value for validation. 

I will revert the change and keep it how it was before!

packages/styles/src/components/radio-button.scss
```diff
@@ -13,113 +13,131 @@ tokens.$default-map: components.$post-radio-button;
...
-  &::after {
-    display: block;
-    content: '';
-    width: tokens.get('radio-button-icon-dot');
-    height: tokens.get('radio-button-icon-dot');
-    border-radius: inherit;
-  }
+// Default Radio
+.form-check:not(.form-switch):has(input[type='radio']) {
+  gap: tokens.get('radio-button-gap-inline-text-start');
```
leagrdv: This adds extra margin between a radio button element and the validation message underneath:

Now:
Image
Before:
Image

packages/styles/src/components/_form-check.scss
```diff
@@ -20,7 +20,7 @@
 
   &-inline {
     display: inline-flex;
-    vertical-align: top;
+    vertical-align: middle;
```
leagrdv: No matter the value here, it seems to not affect the label at all. Therefore, you could remove this line.

packages/styles/src/components/_form-check.scss
```diff
@@ -183,7 +189,6 @@ fieldset {
 
   .form-check {
     &:not(.form-check-inline):not(.form-check-sm):not(:last-of-type) {
-      padding-block: tokens.get('checkbox-padding-block-single', components.$post-checkbox);
```
leagrdv, line 185: You're modifying things that are affecting the checkboxes here, not just the radio button. Did you check that it doesn't break it?

packages/styles/src/components/_form-check.scss
```diff
@@ -217,5 +223,81 @@ fieldset {
         components.$post-checkbox
       );
     }
+
+    &.form-check-inline:has(input[type='checkbox']):not(.form-switch) {
```
leagrdv: You're adding code for the checkbox here, not the radio button. Is this necessary?

packages/styles/src/components/_form-check.scss
```diff
@@ -217,5 +223,81 @@ fieldset {
...
+  &:has(.form-check-inline) {
+    display: flex;
+    flex-wrap: wrap;
+    align-items: flex-start;
+  }
+
+  legend {
+    margin-block-end: tokens.get('radio-button-gap-block-group', components.$post-radio-button);
+  }
+
```
leagrdv: All of this code has been added but inline radio buttons were existing before. All you're adding here is the small version of them, technically there shouldn't be new code for the existing radio.

packages/styles/src/components/_form-check.scss
leagrdv:
```suggestion
        margin-block-end: tokens.get('radio-button-gap-block-group', components.$post-radio-button);
```

It's not your code here but can you change this?

packages/styles/src/components/_form-check.scss
```diff
@@ -217,5 +223,81 @@ fieldset {
...
+      );
+    }
+  }
+
+  .form-check {
+    // Grouped / Vertical
+    &:not(.form-check-inline):not(.form-check-sm):not(:last-of-type) {
+      padding-block: tokens.get('radio-button-padding-block-single', components.$post-radio-button);
+      margin-block-end: tokens.get('radio-button-gap-block-group', components.$post-radio-button);
+    }
```
leagrdv: Those tokens are already used on lines 100 and 104 , if needed, adjust the selectors there but don't duplicate it.

packages/styles/src/components/_form-check.scss
```diff
@@ -217,5 +223,81 @@ fieldset {
...
+    &.form-check-inline:has(input[type='radio']):not(.form-switch) {
+      padding-block-start: tokens.get(
+        'radio-button-small-padding-block',
+        components.$post-radio-button
+      );
+      padding-block-end: tokens.get(
+        'radio-button-small-padding-block',
+        components.$post-radio-button
+      );
+    }
```
leagrdv: All of this could be done in a cleaner way, for each of those selectors, you look for `:has(input[type='radio'])`, therefore you could do: 
```
&:has(input[type="radio"]) {
    &:not(...) {}
    &.form-check-inline {}
   ...
}
```
It's more understandable this way than having selectors that are very long

leagrdv (changes requested):
Please always make sure you're not breaking anything from other components when doing changes on a branch (in this case here: always check the checkbox and switch components)

packages/styles/src/components/_form-check.scss
bucknatt: If this line is not added, the padding would be missing (0px)

Image

But, it has to be 1px for `padding-block-end` and `padding-block-start`. That is what I have noticed, when checking that nothing breaks in Checkbox component, so I have added this line to correct it.

Image

packages/styles/src/components/_form-check.scss
bucknatt: The idea was to refactor the code, to be similar to that one that already is existing on the branch `6291-htmlcss-checkbox-small`. As the code in the checkbox-branch is functioning and well-structured code, I decided to refactor it in the radiobutton-branch. Also, it is to keep merge conflicts minimal. This is the reason why there are changes on the checkbox-code in this PR. 

Goal: Adopt checkbox-code from `6291-htmlcss-checkbox-small` branch.

packages/styles/src/components/_form-check.scss
bucknatt: The deleted line have been moved to line 33-35

 ```
// Checkbox
  &:has(input[type='checkbox']):not(.form-switch):not(.form-check-sm) {
    padding-block: tokens.get('checkbox-padding-block-single', components.$post-checkbox);
```

File: packages/styles/src/components/_form-check.scss (hunk 23 above)
Comment (leagrdv) [reply packages/styles/src/components/_form-check.scss]: The PR for the checkbox has been merged on main already, there is no reason to add code for the checkbox here. Refactoring is moving things around to have less code at the end, but that's not the case here. You're defining something that is already being defined.
Looking for the token in this file results in 3 usages:
Image
Can you make sure this one here is necessary and explain why if that is the case.

File: packages/styles/src/components/_form-check.scss (hunk 24 above)
Comment (leagrdv) [reply packages/styles/src/components/_form-check.scss]: You're resolved the comment but testing it a bit, I see that even just removing the lines from 235 to 239 doesn't change anything on the layout result, meaning those properties are probably not useful. Please make sure that all of those added lines have an effect on the component.

**Review by leagrdv — CHANGES_REQUESTED** [review]
Your changes have broken the checkbox again. 
Please always check before asking for a review that components that depend on the form check are not broken (form-switch, checkbox and card-control).

Image

File: packages/styles/src/components/_form-check.scss [hunk 27]
```diff
@@ -74,6 +73,7 @@
 
   // Small variant
   &.form-check-sm:has(input[type='checkbox']):not(.form-switch) {
+    gap: tokens.get('checkbox-gap-inline-text-start', components.$post-checkbox);
```
leagrdv: This should only be inline, adding it this way adds a horizontal and vertical gap:

Image

packages/styles/src/components/_form-check.scss
```diff
@@ -158,6 +155,28 @@
     }
   }
 
+  // Small variant (Checkbox & Radio Button)
+  &.form-check-sm:not(.form-switch) {
+    &:has(input[type='checkbox']) {
+      padding-block: tokens.get('checkbox-small-padding-block', components.$post-checkbox);
```
leagrdv: This is already defined on line 77, why are you duplicating things that were already defined?
Please check for each token added here if it wasn't already defined previously

packages/styles/src/components/_form-check.scss
```diff
@@ -158,6 +155,28 @@
...
+
+      > input[type='checkbox'] ~ label {
+        font-size: tokens.get('checkbox-small-label-font-size', components.$post-checkbox);
```
leagrdv: Same thing here. Again, this is the small radio button PR, there shouldn't be new definitions of CSS properties for the checkbox (except for the inline gap one that was missing on the checkbox).

packages/styles/src/components/_form-check.scss
```diff
@@ -168,53 +187,91 @@
...
 fieldset {
   legend {
+    // Checkbox Group (default)
     margin-block-end: tokens.get('checkbox-gap-block-group', components.$post-checkbox);
-  }
 
-  &:has(.form-check-sm) {
-    legend {
+    // Checkbox small
+    &:has(+ .form-check-sm:has(input[type='checkbox'])) {
```
leagrdv: This selector is not working, if you inspect the code, the legend does not have the `margin-block-end` defined here.
I don't think nesting a `:has` within an other works.

leagrdv (changes requested):

sandra-post:
@leagrdv Design Review:

**Group default & small**
- Padding top/bottom missing (post.radio-button.padding.block.group)
- Margin left/right on fieldset:
  - IS: 2 px
  - SHOULD: 0 px

packages/styles/src/components/switch.scss
```diff
@@ -210,4 +210,18 @@ $switch-handle-icon: url("data:image/svg+xml,<svg viewBox='0 0 16 16' xmlns='htt
...
     }
   }
+
+  > input[type='checkbox'] {
+    ~ label {
+      color: tokens.get('switch-text-enabled-fg');
+    }
+
+    &:not([disabled]):hover ~ label {
+      color: tokens.get('switch-text-enabled-fg');
```
alizedebray, line 220: Is this correct?
I would expect this instead:
```suggestion
      color: tokens.get('switch-text-hovered-fg');
```

packages/styles/src/components/_form-check.scss
alizedebray, line 1: Can this be moved to a `form-check-base` mixin that is then included into `checkbox.scss` and `radio-button.scss`? This would probably make the code easier to maintain.

packages/documentation/src/stories/components/radio/radio.stories.ts
```diff
@@ -154,89 +125,70 @@ function render(args: Args, context: StoryContext) {
...
+      })}
     </fieldset>
   `;
 }
 
 export const Grouped: Story = {
   render: renderGroup,
+  parameters: {
+    controls: {
+      include: ['Size', 'Hidden Legend'],
```
alizedebray, line 189: The 'Hidden Legend' control is not shown in the docs.

packages/documentation/src/stories/components/radio/radio.stories.ts
```diff
@@ -154,89 +125,70 @@ function render(args: Args, context: StoryContext) {
   `;
 
   return html`
-    <div class="${groupClasses}">${[control, label, ...contextual].filter(el => el !== null)}</div>
+    <div class="${groupClasses}">${[control, label].filter(el => el !== null)}</div>
+    ${contextual}
```
alizedebray, line 129: The validation looks a bit off: 

Image

I guess the validation icon should be centered with the checkbox (tokens are probably needed).

sandra-post:
> @leagrdv Design Review:
> 
> **Group default & small**
> 
> * Padding top/bottom missing (post.radio-button.padding.block.group)
> * Margin left/right on fieldset:
>   
>   * IS: 2 px
>   * SHOULD: 0 px

@leagrdv Design review of mentioned bugs, they are now fixed.

alizedebray (approved):

PR #6813, opened 2025-12-08
feat(components): update megadropdown slots

alizedebray:
## 📄 Description

This PR simplifies the `post-megadropdown` component to remove unnecessary slots.
This prepares the component to become a shadow component.

The content of the megadropdown should not be impacted by these changes and remain the same as on the `main` branch. 

## 🚀 Demo

[Current state on the main branch](https://preview-6672--swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--portal)

[After this PR changes](https://preview-6813--swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--portal) (no visual changes expected)

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-megadropdown/post-megadropdown.tsx
+++ b/packages/components/src/components/post-megadropdown/post-megadropdown.tsx
@@ -15,6 +15,9 @@ import { version } from '@root/package.json';
 import { breakpoint, Device } from '@/utils/breakpoints';
 import { checkRequiredAndType } from '@/utils';
 
+/**
+ * @slot default - Slot for placing content.
+ */
 @Component({
   tag: 'post-megadropdown',
   styleUrl: 'post-megadropdown.scss',


packages/components-angular/projects/consumer-app/src/app/app.component.html
```diff
@@ -80,36 +80,32 @@
...
-            @if (section.titleLink) {
-            <p>
-              <a [href]="section.titleLink">{{ section.title }}</a>
+        <post-list-item slot="post-list-item">
+          <post-megadropdown-trigger [for]="dropdown.id">{{
+            dropdown.trigger
+          }}</post-megadropdown-trigger>
+          <post-megadropdown [id]="dropdown.id">
+            <p class="megadropdown-overview-link" close-label="Close" back-label="Back">
+              <a [href]="dropdown.overviewLink">Overview {{ dropdown.trigger }}</a>
```
oliverschuerch: Since the PR is still a draft, I leave this as a comment, so you can do with the information, whatever you want 😉

The `a` tag represents [phrasing content](https://html.spec.whatwg.org/#phrasing-content) and the user should not add any other text content aside it, right?
If so, you could omit the surrounding `p` tag.

Also shouldn't the labels go to the `post-megadropdown` tag?

alizedebray: Right, I removed the wrapping p-tags!
For the label I'm not sure, the overview link is not mandatory so I would rather full flexibility to the users.

.changeset/orange-beers-knock.md
```diff
@@ -0,0 +1,29 @@
+---
+'@swisspost/design-system-components': major
+'@swisspost/design-system-documentation': patch
+---
+
+Updated the `post-megadropdown` component to no longer uses named slots, it now provides only a default slot for its content.
```
oliverschuerch:
```suggestion
Updated the `post-megadropdown` component, so it no longer uses named slots, it now provides only a default slot for its content.
```

packages/documentation/src/stories/raw-components/megadropdown/megadropdown.stories.ts
```diff
@@ -51,8 +51,12 @@ export function megadropdownDecorator(story: StoryFn, context: StoryContext) {
...
         <post-language-menu-item active="false" code="fr" name="French">FR</post-language-menu-item>
-        <post-language-menu-item active="false" code="it" name="Italian">IT</post-language-menu-item>
-        <post-language-menu-item active="false" code="en" name="English">EN</post-language-menu-item>
+        <post-language-menu-item active="false" code="it" name="Italian"
+          >IT</post-language-menu-item
+        >
+        <post-language-menu-item active="false" code="en" name="English"
+          >EN</post-language-menu-item
+        >
       </post-language-menu>
```
oliverschuerch, line 63: Why did that formatting happen? 
The code is the same as before and has the same length than the `de` and `fr` tags above, or do I miss something?

packages/components/src/index.html
```diff
@@ -14,6 +14,12 @@
     <link rel="stylesheet" href="/assets/css/index.css" />
     <script type="module" src="/build/post-components.esm.js"></script>
     <script nomodule src="/build/post-components.js"></script>
+
+    <script>
+      setTimeout(() => {
+        document.querySelector('a[href="/sch"]').setAttribute('aria-current', 'page');
+      }, 3000);
+    </script>
```
oliverschuerch: Is that a left over or something that need to stay?
If the later, can you please add a comment to explain it.

packages/documentation/src/stories/raw-components/megadropdown/megadropdown.stories.ts
```diff
@@ -92,13 +96,8 @@ export function megadropdownDecorator(story: StoryFn, context: StoryContext) {
...
     <post-megadropdown-trigger for="packages">Packages</post-megadropdown-trigger>
-    <post-megadropdown id="packages">
-      <button slot="back-button" class="btn btn-tertiary px-0">
-        <post-icon name="arrowleft"></post-icon>
-        Back
-      </button>
-      <post-closebutton slot="close-button">Close</post-closebutton>
-      <a slot="megadropdown-overview-link" href="/packages">Overview Packages</a>
+    <post-megadropdown id="packages" label-close="Close" label-back="Back">
+      <a class="megadropdown-overview-link" href="/packages">Overview Packages</a>
```
oliverschuerch: Witout a specific slot, exclusively for the overview link, the link always must be added as the very first content of the megadropdown. Otherwhise it would be missplaced, right?

With that in mind, I'm pretty sure we could get rid of the `.megadropdown-overview-link` class.

What do you think?

packages/documentation/src/stories/raw-components/megadropdown/megadropdown.stories.ts
oliverschuerch: `labelBack` and `labelClose` args should be defiend by default.

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
```diff
@@ -92,6 +128,11 @@ export class PostMegadropdown {
    */
   @Method()
   async show() {
+    if (this.device !== 'desktop') {
+      const triggerLabel = this.megadropdownTrigger?.querySelector('.nav-el-active');
+      if (triggerLabel) this.megadropdownTitle = triggerLabel.innerHTML;
```
oliverschuerch, line 136: Since you use it in a `p` tag afterwards, I'd expect it's the [innerText](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText) you want not the [innerHTML](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML).
If this is wrong, be aware only other [phrasing content](https://html.spec.whatwg.org/#phrasing-content-2) elements are allowed as its decendants.

```suggestion
      if (triggerLabel) this.megadropdownTitle = triggerLabel.innerText;
```

alizedebray: The `triggerLabel` is the content of the `post-magadropdown-trigger`.
Therefore, the `innerHTML` would only contain elements that are valid children for a button which are also valid children for a paragraph.

packages/documentation/src/stories/raw-components/megadropdown/megadropdown.stories.ts
alizedebray: The content of the megadropdown needs to be fully customizable, therefore I cannot target element solely according to their position. If I go with `post-megadropdown > a:first-child` then the overview link becomes mandatory if you want to have any other link in the megadropdown.

packages/documentation/src/stories/raw-components/megadropdown/megadropdown.stories.ts
alizedebray: IT and EN are 101 character long and it wraps at 100.

oliverschuerch: Sorry, I just found the location of the linebreak strange, but ok...

packages/documentation/src/stories/raw-components/megadropdown/megadropdown.stories.ts
oliverschuerch: I disagree. The overview link does not become mandatory, but if there is a link as the first element, it is styled like the overview link.
If the content of the megadropdown needs to be fully customizable, then the styles for its content would need to be defined by the content itself. Similar to how it works in the tabs component for example.

Imo, there are already to many specific styles around for the megadropdown content (flexbox, etc.) and adding `post-megadrpdown > a:first-child` would not be an issue. 
I saw it more as an optimization, but I'm ok if you want to keep the class.

---

However, the following image shows how the megadropdown looks, when there is no such class on the overview link and this is actually the current state on the header docs page.

Image

oliverschuerch (changes requested):

packages/documentation/src/stories/raw-components/megadropdown/megadropdown.stories.ts
oliverschuerch: I didn't get it's a pr-train. So I'll review the other pull-requests first.

packages/components/src/components/post-megadropdown/post-megadropdown.tsx
```diff
@@ -1,7 +1,19 @@
...
+  Method,
+  Prop,
+  State,
+  Watch,
+} from '@stencil/core';
 import { version } from '@root/package.json';
 import { breakpoint, Device } from '@/utils/breakpoints';
+import { checkRequiredAndType } from '@/utils';
 
 @Component({
```
oliverschuerch, line 21: Please add a jsdoc comment to document the default slot.
e.g.

```ts
/**
 * @slot default - Slot for placing content.
 */
@Component({
  ...
```

packages/components/src/components/post-megadropdown/post-megadropdown.scss
```diff
@@ -112,11 +112,12 @@
     padding-inline: 16px;
   }
 
-  [slot='megadropdown-overview-link'] {
+  .post-megadropdown-overview {
+    flex: 1 1 100%;
```
oliverschuerch: Only change that if the link must always take 100% of the available width. If there are cases when it must shrink and/or grow, keep what you have alreay.
```suggestion
    flex: 0 0 100%;
```

Issue #6538, opened 2025-10-28
[bug] fix post-menu animation 


alizedebray:
The `post-menu` opening animation should be fixed, the delay appears on all components using it.
___
There is a slight delay in the language switch: on click, it opens the dropdown and a second later it gets the active state.
https://github.com/user-attachments/assets/2f920eb6-fdc3-4268-8712-09376e56dbea

_Originally posted by @leagrdv in https://github.com/swisspost/design-system/pull/6512#discussion_r2470269107_

alionazherdetska:
Resolution:  Enable animation by default in post-menu do not expose props for post-breadcrumbs/post-language-switch.

alionazherdetska:
Resolution: remove the `prop` from the `post-menu`.

myrtasakellariou:
Updated resolution: Remove animation prop from the `<post-popovercontainer>` and enable by default the 'pop-in' animation.

PR #6821, opened 2025-12-09
feat(components): enable open animation on `<post tooltip>`

myrta2302:
## 📄 Description

This PR updates `<post-tooltip>` so that it now correctly uses the open animation provided by `<post-popovercontainer>`, which was previously not visible. 

Additionally, but not directly connected to the scope of the current ticket it:
- Removes the animation cancellation logic from the popovercontainer, that seems to not work as expected. 
- Includes a small improvement on the `<post-popover>` component. (Popover animation has already been enabled in the #6538 )

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 📝 Checklist


packages/components/src/components/post-popover/post-popover.tsx
oliverschuerch, line 1: What are you doing in post-popover and why is it not mentioned in the changeset?

myrtasakellariou: When the popover opens, the first interactive element should be focused. This is working right now because the trigger uses the toggle() method, but as far as i understand this should also work when calling the show() method. I considered this an internal change, thats why i did not mention in the changeset.

packages/components/src/components/post-popover/post-popover.tsx
```diff
@@ -75,10 +76,13 @@ export class PostPopover {
   @Method()
   async toggle(target: HTMLElement, force?: boolean) {
     await this.popoverRef.toggle(target, force);
+    this.focusFirstEl();
```
oliverschuerch: Can we avoid calling this function when we toggle to false?

myrtasakellariou: I 'm not sure. I had already thought about it, but currently post-menu toggle() calls it like this - without the force: ` this.popoverRef.toggle(target)`. Would you prefer updating the post-menu?

oliverschuerch: No, that was not my intention, but it would be good if it does the exact same thing.

I just had a look into your PR and I think we can do the following to optimize this functionality:

```ts
async toggle(target: HTMLElement, force?: boolean) {
  // you can expect the popoverRef.toggle function to return the current state (after toggling), no matter if we call it with a `force` parameter or not ;) 
  const isOpen = await this.popoverRef.toggle(target, force);
  if (isOpen) this.focusFirstEl();
}
```

This is only a micro improvement reguarding performance, but it could save us from unwanted side-effects.

oliverschuerch (changes requested):

PR #6823, opened 2025-12-09
test: implement matrix e2e tests

oliverschuerch:


oliverschuerch:
Dividing the tests into a matrix did not lead to the hoped-for increase in performance.

PR #6826, opened 2025-12-09
chore(tests): visual regression tests for `post-header`

alionazherdetska:
## 📄 Description

This PR implements visual regression testing for the `post-header` component using Playwright. This is a temporary solution until a better approach for storing and managing baseline snapshots is implemented.

### What's included:
- GitHub Actions workflow for automated visual testing
- Playwright configuration with multi-browser support (Chromium, Firefox, WebKit)
- Test helpers and utilities for header interactions
- Comprehensive test matrix covering header variants and states

### How it works:

- This is the first PR with visual tests so there are no baseline screenshots yet to compare against. The tests are looking for screenshots that don't exist yet, that's why they fail. This is totally expected and normal.
- Once this PR is merged to main the workflow will automatically notice there are no baseline screenshots, generate them from the current code and save them in the GitHub Actions cache. After that every new PR will compare its screenshots against those baselines from main and if something looks different the tests will catch it.
- If a future PR changes how things look the tests will fail and upload diff images so you can see exactly what changed. Then you decide, is this change intentional or a bug? If it's intentional you can manually update the baselines by going to Actions tab, selecting the Visual Regression Tests workflow, clicking Run workflow and checking the Force update visual snapshots checkbox. If it's a bug you fix it until the screenshots match again.

### 📋 Header Variants Tested:
image (14)

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -119,7 +119,7 @@
     "cypress": "14.3.2",
     "cypress-axe": "1.6.0",
     "cypress-storybook": "1.0.0",
-    "eslint": "9.18.0",
+    "eslint": "9.26.0",
     "eslint-plugin-cypress": "4.2.0",
     "eslint-plugin-react": "7.37.5",
     "globals": "16.0.0",
--- a/.changeset/whole-deer-burn.md
+++ b/.changeset/whole-deer-burn.md
@@ -1,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': patch
+'@swisspost/design-system-styles': patch
+---
+
+Aligned `ngx-toastr` styling with the existing toast component and documented the integration.
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -30,7 +30,7 @@ jobs:
           use_cache: false
 
       - name: Cache cypress
-        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830
+        uses: actions/cache@b7e8d49f17405cc70c1c120101943203c98d3a4b
         with:
           path: /home/runner/.cache/Cypress
           key: ${{ runner.os }}-cypress-${{ hashFiles('./pnpm-lock.yaml') }}
--- a/packages/changelog-github/package.json
+++ b/packages/changelog-github/package.json
@@ -31,7 +31,7 @@
   "devDependencies": {
     "@changesets/parse": "0.4.1",
     "@eslint/js": "9.18.0",
-    "eslint": "9.18.0",
+    "eslint": "9.26.0",
     "globals": "16.0.0",
     "typescript": "5.8.3",
     "typescript-eslint": "8.20.0",
--- a/packages/components-angular/package.json
+++ b/packages/components-angular/package.json
@@ -50,7 +50,7 @@
     "angular-eslint": "19.8.1",
     "copyfiles": "2.4.1",
     "cypress": "14.3.2",
-    "eslint": "9.18.0",
+    "eslint": "9.26.0",
     "globals": "16.0.0",
     "karma": "6.4.4",
     "karma-chrome-launcher": "3.2.0",
--- a/packages/components-react/package.json
+++ b/packages/components-react/package.json
@@ -57,7 +57,7 @@
   },
   "dependencies": {
     "@stencil/react-output-target": "1.2.0",
-    "@stencil/ssr": "0.1.1",
+    "@stencil/ssr": "0.3.0",
     "@swisspost/design-system-components": "workspace:10.0.0-next.60"
   },
   "devDependencies": {
@@ -67,7 +67,7 @@
     "@types/react": "18.3.26",
     "@types/react-dom": "18.3.7",
     "copyfiles": "2.4.1",
-    "eslint": "9.18.0",
+    "eslint": "9.26.0",
     "eslint-plugin-react": "7.37.5",
     "globals": "16.0.0",
     "react": "18.3.1",
--- a/packages/documentation/package.json
+++ b/packages/documentation/package.json
@@ -56,7 +56,7 @@
     "@typescript-eslint/parser": "7.18.0",
     "cypress": "14.3.2",
     "cypress-axe": "1.6.0",
-    "eslint": "9.18.0",
+    "eslint": "9.26.0",
     "eslint-plugin-cypress": "4.2.0",
     "eslint-plugin-mdx": "3.6.2",
     "eslint-plugin-react": "7.37.5",
--- a/packages/documentation/src/stories/components/toast/ngx-toastr-dismissible.sample.ts
+++ b/packages/documentation/src/stories/components/toast/ngx-toastr-dismissible.sample.ts
@@ -1,0 +1,4 @@
+this.toastr.success('The task has been completed.', 'Success!', {
+  toastClass: 'toast toast-success toast-dismissible',
+  closeButton: true,
+});
--- a/packages/documentation/src/stories/components/toast/ngx-toastr.sample.ts
+++ b/packages/documentation/src/stories/components/toast/ngx-toastr.sample.ts
@@ -1,0 +1,3 @@
+this.toastr.success('The task has been completed.', 'Success!', {
+  toastClass: 'toast toast-success',
+});
--- a/packages/documentation/src/stories/components/toast/toast.docs.mdx
+++ b/packages/documentation/src/stories/components/toast/toast.docs.mdx
@@ -1,7 +1,9 @@
-import { Canvas, Controls, Meta } from '@storybook/addon-docs/blocks';
+import { Canvas, Controls, Meta, Source } from '@storybook/addon-docs/blocks';
 import meta, * as ToastStories from './toast.stories';
 import StylesPackageImport from '@/shared/styles-package-import';
 import PackageTag from '@/shared/package-tag';
+import NgxToastr from './ngx-toastr.sample?raw';
+import NgxToastrDismissible from './ngx-toastr-dismissible.sample?raw';
 
 <Meta of={ToastStories} />
 
@@ -55,3 +57,26 @@ You can stack toasts by wrapping them in a `.toast-container`, which will vertic
 Toasts are intended to be small interruptions to your visitors or users, so to help those with screen readers and similar assistive technologies, you should wrap your toasts in an Live region.
 
 See how to implement it on our [live regions documentation](/?path=/docs/548d973c-679c-4a02-96cd-70f82aec5f65--docs).
+
+## Integration with `ngx-toastr`
+
+When using `ngx-toastr` in an Angular project, Design System styles can be applied to toast notifications by overriding the default `ngx-toastr` classes.
+
+To do this, set the appropriate classes via the `toastClass` option:
+
+- Information toast: `toast toast-info`
+- Success toast: `toast toast-success`
+- Warning toast: `toast toast-warning`
+- Error toast: `toast toast-error`
+
+<Source code={NgxToastr} language="ts" />
+
+#### Dismissible toasts
+
+For dismissible toasts (i.e. `closeButton` is set to `true`), the `toast-dismissible` class must also be included in the `toastClass` configuration.
+
+<Source code={NgxToastrDismissible} language="ts" />
+
+#### CSS import order
+
+To ensure that Design System styles override the default `ngx-toastr` styles, the Design System CSS files must be imported <b>after</b> the `ngx-toastr` CSS files.
--- a/packages/eslint/package.json
+++ b/packages/eslint/package.json
@@ -48,7 +48,7 @@
     "@typescript-eslint/rule-tester": "8.49.0",
     "concurrently": "9.2.1",
     "domhandler": "5.0.3",
-    "eslint": "9.19.0",
+    "eslint": "9.26.0",
     "globals": "16.0.0",
     "jest": "29.7.0",
     "parse5": "7.2.1",
--- a/packages/icons/package.json
+++ b/packages/icons/package.json
@@ -43,7 +43,7 @@
     "@typescript-eslint/eslint-plugin": "7.18.0",
     "@typescript-eslint/parser": "7.18.0",
     "dotenv": "17.2.3",
-    "eslint": "9.18.0",
+    "eslint": "9.26.0",
     "eslint-plugin-jest": "28.11.0",
     "globals": "16.0.0",
     "http-server": "14.1.1",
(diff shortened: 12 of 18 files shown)


packages/components/package.json
```diff
@@ -80,6 +80,13 @@
     "e2e:ci": "xvfb-run -a cypress run --browser chrome",
     "lint": "eslint",
     "lint:fix": "eslint --fix",
+    "generate:header-html": "node visual-tests/generate-header-html.js",
+    "test:visual": "npm run generate:header-html && playwright test",
+    "test:visual:update": "npm run generate:header-html && playwright test --update-snapshots",
+    "test:visual:ui": "npm run generate:header-html && playwright test --ui",
+    "test:visual:chrome": "npm run generate:header-html && playwright test --project=chromium",
+    "test:visual:firefox": "npm run generate:header-html && playwright test --project=firefox",
+    "test:visual:safari": "npm run generate:header-html && playwright test --project=webkit",
```
oliverschuerch: I would generally prefer a short form for the script names, to keep the naming concept we used for `e2e`, `unit`, `lint`, etc.

My personal suggestion: `vrt` (visual regression testing) instead of `test:visual`.
- `generate:header-html` => `vrt:generate:header` (would be nice if you could get rid of that completely, can you maybe explain, why this is necessary?)
- `test:visual` => `vrt`
- `test:visual:update` => `vrt:update`
- ...


But you can also ask @gfellerph and/or @alizedebray or the whole DEV team what they think about the topic 😉

⚠️ If you rename the scripts, do the same with the folder names, etc.

packages/components/tsconfig.json
```diff
@@ -28,6 +28,6 @@
     "resolveJsonModule": true,
     "skipLibCheck": true
   },
-  "include": [".config/*", "src", "types", "prebuild.spec.ts"],
+  "include": ["prebuild.ts", "prebuild.spec.ts", ".config/*", "src", "types", "visual-tests/**/*", "playwright.config.ts"],
```
oliverschuerch: As you can see, the `prebuild.ts` file is handled by the ts-node config at the very beginning of this file. Actually I don't even know, why the `prebuild.spec.ts` file needs to be included here.

Did you encounter any problems, or why is it now included here as well?

On the other side, if we can get rid of the ts-node config, this would be even better.

package.json
```diff
@@ -48,6 +48,12 @@
     "components:lint": "pnpm --filter design-system-components lint",
     "components:lint:fix": "pnpm --filter design-system-components lint:fix",
     "components:generate": "pnpm --filter design-system-components generate",
+    "components:visual": "pnpm --filter design-system-components build && pnpm --filter design-system-components test:visual",
+    "components:visual:update": "pnpm --filter design-system-components build && pnpm --filter design-system-components test:visual:update",
+    "components:visual:ui": "pnpm --filter design-system-components build && pnpm --filter design-system-components test:visual:ui",
+    "components:visual:chrome": "pnpm --filter design-system-components build && pnpm --filter design-system-components test:visual:chrome",
+    "components:visual:firefox": "pnpm --filter design-system-components build && pnpm --filter design-system-components test:visual:firefox",
+    "components:visual:safari": "pnpm --filter design-system-components build && pnpm --filter design-system-components test:visual:safari",
```
oliverschuerch: Rename those as well, in case you rename the scripts in the components package.json file.
Otherwise ignore this message.

packages/components/.gitignore
```diff
@@ -28,3 +28,17 @@ $RECYCLE.BIN/
...
+
+test-results/
+test-results-*
+playwright-report/
+
+**/visual-tests/__screenshots__/**/*-actual.png
+**/visual-tests/__screenshots__/**/*-diff.png
+**/visual-tests/__screenshots__/**/*-expected.png
+**/visual-tests/__screenshots__/
+**/visual-tests/components/post-header-*.html
```
oliverschuerch: Do we need something to be checked in in the `visual-tests` folder?
If not, `**/visual-tests` would be enough to exclude the whole folder and all it's descendants.

packages/components/.gitignore
```diff
@@ -28,3 +28,17 @@ $RECYCLE.BIN/
 Thumbs.db
 UserInterfaceState.xcuserstate
 .env
+
+test-results/
+test-results-*
```
oliverschuerch: Be aware, this could also exclude files starting with `test-results-`.

packages/components/.gitignore
```diff
@@ -28,3 +28,17 @@ $RECYCLE.BIN/
...
+playwright-report/
+
+**/visual-tests/__screenshots__/**/*-actual.png
+**/visual-tests/__screenshots__/**/*-diff.png
+**/visual-tests/__screenshots__/**/*-expected.png
+**/visual-tests/__screenshots__/
+**/visual-tests/components/post-header-*.html
+
+/playwright/.cache/
+playwright/.auth/
```
oliverschuerch: Can we align both of these lines, or is the second `playwright` folder nested somewhere, while the other only exists in the package root?

.github/workflows/visual-tests.yaml
```diff
@@ -0,0 +1,104 @@
...
+            ${{ runner.os }}-playwright-
+
+      - name: Install Playwright browsers
+        working-directory: packages/components
+        run: npx playwright install --with-deps
+
+      - name: Get baseline snapshots from cache
+        id: snapshot-cache
+        uses: actions/cache@v4
+        with:
```
oliverschuerch, line 57: Should we add `fail-on-cache-miss: true`?
[Source](https://github.com/actions/cache/tree/v4.3.0?tab=readme-ov-file#inputs)

.github/workflows/visual-tests.yaml
```diff
@@ -0,0 +1,104 @@
...
+          path: packages/components/visual-tests/__screenshots__
+          key: visual-snapshots-main-${{ hashFiles('packages/components/src/**/*.{tsx,ts,scss}') }}
```
oliverschuerch: Is it mandatory to create a hash by all src files in the components package? I'm asking because it seems time consuming to me.
Can't we just use the lock file or the package.json to create the hash?

Also I'm not absolutely sure if we should include the `runner.os` as well, similar to how you create the Playwright browser cache above.

And why do we need the `main` keyword in the key?
As far as I understand, this is not needed.
See [Restrictions for accessing a cache](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache)

.github/workflows/visual-tests.yaml
```diff
@@ -0,0 +1,104 @@
...
+          restore-keys: |
+            visual-snapshots-main-
+      
+      - name: Check if baseline update needed
+        id: should-update
+        run: |
+          if [[ "${{ github.ref }}" == "refs/heads/main" && "${{ steps.snapshot-cache.outputs.cache-hit }}" != "true" ]] || [[ "${{ github.event.inputs.update-snapshots }}" == "true" ]]; then
```
oliverschuerch: `github.ref_name == 'main'` could be a shorter variant of your ref check.

What if this must run on a future release branch (e.g. `release/v10`, `release/v11`, etc.) instead of on `main`?

In the release.yaml file we have to update that manually, if I remember correctly.
See [release.yaml on release/v9](https://github.com/swisspost/design-system/blob/release/v9/.github/workflows/release.yaml).
Maybe we can find a better way here.

e.g. `(github.ref_name == 'main' || github.ref_name == release/v*) && ...`
or `(github.ref == 'refs/heads/main' || github.ref == release/v*) && ...`

.github/workflows/visual-tests.yaml
```diff
@@ -0,0 +1,104 @@
...
+      - name: Generate baseline snapshots
+        if: steps.should-update.outputs.update == 'true'
+        working-directory: packages/components
+        run: pnpm test:visual:update
+
+      - name: Optimize PNG screenshots
+        if: steps.should-update.outputs.update == 'true'
+        run: |
+          sudo apt-get install -y pngquant
+          find packages/components/visual-tests/__screenshots__ -name "*.png" -exec pngquant --force --ext .png --quality 80-95 {} \;
```
oliverschuerch: Did you see some results from this quality reduction?
Compressing jpgs and pngs in the past with Photoshop, it was usually ok to go down to a quality of 60-80%.
If the size is not already down to 1/5 - 1/3, I'd suggest to change the quality setting to `60-80` and maybe start playing with the `--speed` option.
See pngquant [docs](https://pngquant.org/).

packages/components/vrt/helpers/header-test-config.ts
```diff
@@ -0,0 +1,193 @@
...
+  { name: 'tablet', width: 780, height: 1024 },
+  { name: 'mobile', width: 375, height: 667 },
+] as const;
+
+export type BreakpointName = typeof TEST_BREAKPOINTS[number]['name'];
+
+/**
+ * Handler type definitions
+ */
+export type HandlerFunction = (page: Page) => Promise<void>;
```
oliverschuerch: Do we need this type definition somewhere outside of this file?
If not, you could merge it with the `HandlerMap`.

packages/components/vrt/helpers/header-test-config.ts
oliverschuerch, line 1: I really like this file, good job!

packages/components/vrt/helpers/header-test-config.ts
```diff
@@ -0,0 +1,193 @@
...
+  'scrolled-first-level-opened',
+] as const;
+
+/**
+ * Helper for tablet + mobile duplication
+ */
+const tm = (states: readonly string[]) => ({
+  tablet: { states },
+  mobile: { states },
+});
```
oliverschuerch, line 154: This helper could be smarter and more future ready...
As your solution seems to work, it's ok if you keep it if the example below does not work out of the box or if you have no time to try it out😉 

```ts
const stateComposeHelper = (
  desktop: BreakpointConfig['states'],
  tablet?: BreakpointConfig['states'],
  mobile?: BreakpointConfig['states']
) => ({
  desktop: { states: desktop },
  tablet: { states: tablet ?? desktop },
  mobile: { states: mobile ?? tablet ?? desktop },
});

export const TEST_MATRIX: Record<string, VariantConfig> = {
  //                                     desktop                    tablet & mobile
  'portal-loggedout': stateComposeHelper(DESKTOP_PORTAL_LOGGED_OUT, BURGER_FLOW),
  //                                    desktop, tablet & mobile
  'portal-loggedin': stateComposeHelper(LOGGED_IN),
  //                                    desktop         tablet         mobile
  'portal-loggedin': stateComposeHelper(DESKTOP_STATES, TABLET_STATES, MOBILE_STATES),
  //                                    desktop & tablet         mobile
  'portal-loggedin': stateComposeHelper(SOME_OTHER_STATES, null, MOBILE_STATES),
  ...
};
```

packages/components/vrt/helpers/header-variants.config.ts
```diff
@@ -0,0 +1,127 @@
+export interface VariantConfig {
+  slots: string[];
```
oliverschuerch: I'm not sure if we should be more specific here, so only specific values are allowed in the slots property.

e.g. `slots: Array<'post-logo' | 'language-menu' | 'title' | ...>`

What do you think?

packages/components/vrt/generate-header-html.js
```diff
@@ -0,0 +1,473 @@
...
+    .replace('{{HEADER_CONTENT}}', headerContent);
+
+  // Apply variant-specific replacements
+  Object.entries(config.replacements || {}).forEach(([placeholder, value]) => {
+    html = html.replaceAll(placeholder, value);
+  });
+
+  // Clean up any remaining placeholders
+  // Use a linear-time scanner instead of a regex to avoid super-linear
+  // backtracking risks when processing untrusted or very large input
```
oliverschuerch, line 423: Is this something coming from sonar?
Because `html = html.replaceAll(/{{[^}]+}}/gm, '')` is really much simpler to understand and easier to maintain.

packages/components/vrt/post-header.consolidated.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+      test.describe(`${breakpoint} (${width}x${height})`, () => {
+        test.beforeEach(async ({ page }) => {
+          await page.setViewportSize({ width, height });
+          await page.goto(`/visual-tests/post-header-${variant}.html`);
+          await waitForHeaderReady(page);
+        });
+
+        const isLoggedInVariant = variant.includes('loggedin');
+        const isBreakpointToSkipDefault =
+          breakpoint === 'mobile' || breakpoint === 'tablet' || breakpoint === 'desktop';
```
oliverschuerch: Programming such things like that would (imo) increase maintainability.
But I'm also ok if you keep what you have as it is totally valid code.

```ts
const BREAKPOINT_KEYS = ['desktop', 'tablet', 'mobile'];

TEST_BREAKPOINTS.forEach(...
  ...
  const isBreakpointToSkipDefault = BREAKPOINT_KEYS.includes(breakpoint);
  ...
);
```

oliverschuerch (changes requested):
This is a huge and insane PR (even if there are only 14 files)!
I really want to see that in action and hope it works as expected from the very beginning.

Bravo @alionazherdetska!!!

packages/components/tsconfig.json
alionazherdetska: I removed it as a redundant.

packages/components/vrt/generate-header-html.js
alionazherdetska: Yes, this is coming from Sonar. I originally had the simple regex `html.replaceAll(/{{[^}]+}}/gm, '')` but Sonar flagged it.

packages/components/vrt/helpers/header-test-config.ts
alionazherdetska: Thank you

.github/workflows/visual-tests.yaml
oliverschuerch: Can you explain, why you changed the hashFiles function from `*.{tsx,ts,scss}` to `*.scss`?
This is not what I asked for and your first solution might be even better, but I'm really only asking, because I would like to understand, the thoughts behind your decision.

.github/workflows/visual-tests.yaml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Generate baseline snapshots
+        if: steps.should-update.outputs.update == 'true'
+        working-directory: packages/components
+        run: pnpm vrt:update
+
+      - name: Optimize PNG screenshots
+        if: steps.should-update.outputs.update == 'true'
+        run: |
+          sudo apt-get install -y pngquant
+          find packages/components/vrt/__screenshots__ -name "*.png" -exec pngquant --force --ext .png --quality 60-80 --speed 1 {} \;
```
oliverschuerch: Again, did you see some output files, so you can visually decide if the settings are nice?

If not I would keep the speed (for now) on the default value of `3`.

Issue #6687, opened 2025-11-19
Clarify Post Header Semantics


alizedebray:
The `post-header` currently does not include an HTML `<header>` element.

We need to evaluate the correct semantic structure and implement the chosen approach.

## Questions to resolve

- Should we use a `<header>` element, or is `role="banner"` sufficient?
- Is it correct for the `<nav>` element from the `post-mainnavigation` to be placed inside a `<header>` element?

## Tasks

- [ ] Decide on the correct semantic structure for `post-header` and implement it

alizedebray:
@gfellerph Can you please add links to good resources on that topic?

gfellerph:
Trusted sources include:
- https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/general-principles.html
- https://www.w3.org/WAI/ARIA/apg/
- https://developer.mozilla.org/de/
- https://www.accessibility-developer-guide.com/

PR #6841, opened 2025-12-11
chore(components, accessibility): improve `post-header` semantics

alionazherdetska:
## 📋 Summary

This PR improves the semantic structure and accessibility of the `post-header` and `post-mainnavigation` components.

---

## Key Question

**Original Question from [[#6687](https://github.com/swisspost/design-system/issues/6687)]**

> "Is it correct for the `<nav>` element from the `post-mainnavigation` to be placed inside a `<header>` element?"

###  Decision: 

**Yes**, placing `<nav>` inside `<header>` is correct and follows W3C best practices.

### Why this is valid:

Both W3C and MDN provide examples showing navigation elements nested within the header. According to **MDN**:

> "Even if it's not mandatory, it's common practice to put the main navigation menu within the main header."

**Source:** [[MDN - WAI-ARIA Basics](https://developer.mozilla.org/de/docs/Learn_web_development/Core/Accessibility/WAI-ARIA_basics?utm_source=chatgpt.com#wegweiserlandmarken)](https://developer.mozilla.org/de/docs/Learn_web_development/Core/Accessibility/WAI-ARIA_basics?utm_source=chatgpt.com#wegweiserlandmarken)

### Trusted Source Examples:

1. [[MDN - WAI-ARIA Basics: Landmark Navigation](https://developer.mozilla.org/de/docs/Learn_web_development/Core/Accessibility/WAI-ARIA_basics?utm_source=chatgpt.com#wegweiserlandmarken)](https://developer.mozilla.org/de/docs/Learn_web_development/Core/Accessibility/WAI-ARIA_basics?utm_source=chatgpt.com#wegweiserlandmarken)
2. [[W3C Design System - Header with Navigation](https://design-system.w3.org/code/header-nav.html)](https://design-system.w3.org/code/header-nav.html)

---

#### Accessibility Tree Comparison:

**BEFORE:**

Accessibility tree before changes

**AFTER:**

Accessibility tree after changes

#### Why `role="banner"` is needed:

W3C guidance on role banner

---

### 2. Added Required `caption` Prop to `post-mainnavigation`

---

### 3. Moved Event Listener to Shadow Root

**Problem with host listener:**

When the event listener is attached to the host element, screen readers detect the entire component as interactive and announce it as **"clickable banner landmark"**, which is incorrect and confusing for users with assistive technology.

**Solution:**

Changed from `this.host.addEventListener()` to `this.host.shadowRoot.addEventListener()`.

### DIFF
--- a/packages/components/src/components.d.ts
+++ b/packages/components/src/components.d.ts
@@ -308,6 +308,9 @@ export namespace Components {
         "url"?: string | URL;
     }
     interface PostMainnavigation {
+        /**
+          * Defines the accessible label for the navigation element. This text is used as the `aria-label` attribute to provide screen reader users with a description of the navigation's purpose.
+         */
         "caption": string;
     }
     interface PostMegadropdown {
@@ -1327,6 +1330,9 @@ declare namespace LocalJSX {
         "url"?: string | URL;
     }
     interface PostMainnavigation {
+        /**
+          * Defines the accessible label for the navigation element. This text is used as the `aria-label` attribute to provide screen reader users with a description of the navigation's purpose.
+         */
         "caption": string;
     }
     interface PostMegadropdown {
--- a/packages/components/src/components/post-mainnavigation/readme.md
+++ b/packages/components/src/components/post-mainnavigation/readme.md
@@ -7,9 +7,9 @@
 
 ## Properties
 
-| Property               | Attribute | Description | Type     | Default     |
-| ---------------------- | --------- | ----------- | -------- | ----------- |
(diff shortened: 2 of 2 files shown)


packages/components/src/components/post-header/post-header.tsx
alionazherdetska, line 1: Moved the click event listener from the **host element** to the **shadow root** 
to fix an accessibility issue where screen readers were incorrectly announcing 
the banner landmark as **"clickable banner landmark"**.

oliverschuerch (approved):

Issue #6304, opened 2025-09-24
[web component styles] Datepicker


alizedebray:
### 📃 Tasks

- [ ] https://github.com/swisspost/design-system/issues/6303
- [ ] Implement styles according to design
- [ ] Complete documentation in Storybook and make it accessible to all users
- [ ] Add visual regression tests
- [ ] Design review

sandra-post:
@myrta2302  Implemented following tokens adjustments:

- Added token for vertical gap: `post.datepicker.grid.gap.block`
- Adjust token for min-height
  - OLD:  `post.datepicker.grid.min-height: post.core.dimension.176`
  - NEW:  `post.datepicker.grid.min-height: post.core.dimension.182`
- Added token for disabled cell size (38 px): `post.datepicker.grid.day.cell.element.disabled.size`
- Deleted token for disabled cell padding: `post.datepicker.grid.day.cell.element.disabled.padding`
- Added token for day-range disabled-bg: `post.datepicker.grid.day-range.cell.element.color.disabled-bg`
- Corrected token name for weekdays:
  - OLD: `post-datepicker-header-weekdays-color`
  - NEW: `post-datepicker-header-weekdays-color-fg`

Token adjustments also included in [[base] Datepicker](https://github.com/swisspost/design-system-planner/issues/91).

PR #6243, opened 2025-09-18
chore(components): investigate cally calendar for the date picker

myrta2302:
## 📄 Description

This PR showcases an example of a datepicker component using the `air-datepicker` . It is added in the **components playground** as `<post-datepicker2>`.

There is also  a `<post-datepicker>` within the same PR. This was used to test Cally, but it does not meet the functionality requirements, so it can be ignored.

PR #6843, opened 2025-12-11
feat(components): datepicker

leagrdv:
## 📄 Description

Datepicker functionality implementation

The CSS is temporary and is not 100% accurate to what is expected. It was mainly copied from [Myrta's test PR](https://github.com/swisspost/design-system/pull/6243). The actual styling of the calendar will be done in #6304 

## 🚀 Demo

https://preview-6843--swisspost-design-system-next.netlify.app/?path=/docs/eb77cd02-48b2-42e1-a3e4-cd8a973d431e--docs&devModeEnabled=true

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist


packages/components/src/components/post-datepicker/post-datepicker.tsx
```diff
@@ -0,0 +1,850 @@
...
+
+  /**
+   * Whether the calendar is inline in the page (not showing in a popover when input clicked)
+   */
+  @Prop() inline = false;
+
+  /**
+   * Label for "Next month" button
+   */
+  @Prop({ reflect: true }) labelNextMonth!: string;
```
oliverschuerch: All the label props need to be prefixed by `text`, instead of `label`.

packages/documentation/src/stories/components/datepicker/datepicker.stories.ts
```diff
@@ -0,0 +1,207 @@
...
+  exclude: ['id'],
+  parameters: {
+    badges: [],
+    design: {
+      type: 'figma',
+      url: 'https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-Next-Level?node-id=21044-101376&t=FNgdl36MZPMbc1Bs-4',
+    },
+  },
+  args: {
+    id: 'main',
```
oliverschuerch, line 23: Is this argument necessary in production?
Because for the DEVs the more important attribute will be the `name` or `id` attribute on the input fields themself.

packages/components/src/components/post-datepicker/post-datepicker.tsx
oliverschuerch, line 1: The component shows the native datepicker table (tested in chrome):

Image

---

**To reproduce**
1. Focus the input field
2. Press <kbd>space</kbd>

packages/components/src/components/post-datepicker/post-datepicker.scss
oliverschuerch, line 1: Hovering a disabled date by mouse causes a wrong visual state (tested in chrome):

Image

packages/components/src/components/post-datepicker/post-datepicker.tsx
oliverschuerch, line 1: Changing the locale profile, does not update the dateFormat as expected.
I would expect the dateFormat (shown in the input itself) must change when switching from
 
- `mm/dd/yyyy` (en-GB, en) to
- `mm/dd.yyy` (en-US, not implemented) to
- `dd.mm.yyyy` (de-DE, de) or to
- `yyyy/mm/dd` (zh-CN, not available)
- ...

See [example](https://mui.com/x/react-date-pickers/adapters-locale/#set-a-custom-locale)

Instead it always uses the german format (`tt.mm.jjjj`) in my browser.

packages/components/src/components/post-datepicker/post-datepicker.scss
```diff
@@ -0,0 +1,320 @@
...
+          height: 40px; // post.datepicker.year.grid.height
+        }
+      }
+
+      &.-months- {
+        grid-template-columns: repeat(4, 1fr);
+        padding: 8px; // post.datepicker.month.grid.padding
+        column-gap: 8px; // post.datepicker.month.grid.column-gap
+        height: 216px; // post.datepicker.year.grid.height
+        text-transform: uppercase;
```
oliverschuerch, line 139: For both we could set a fix `height`, since the amount of years showing at once will probably not change dynamically.
Or we could keep `height` dynamic (content related), since we can't match the design anyway.

**Figma**

Image

**Code**
| Month | Year |
| - | - |
| Image | Image |

packages/components/src/components/post-datepicker/post-datepicker.scss
oliverschuerch, line 1: The "Switch to view" button needs a background and border reset and probably shows the wrong chevron down icon (tested in chrome).

Image

Image

packages/components/src/components/post-datepicker/post-datepicker.tsx
oliverschuerch, line 1: It seems we focus the wrong year/month/date when switching between the views.

Image

Image

Image

**To reproduce**
1. Open a datepicker with no date set.
2. Switch to the `year` view (the current year is focussed).
3. Choose the year.
4. The calendar switches automatically to the month view (the first month is focussed and the current is highlighted).
5. Choose the month.
6. The calendar switches automatically to the day view (the first day of the month is focussed and the current is highlighted).

If it is possible to change that technically, I would ask the team to define a rule how to do so.

packages/components/src/components/post-datepicker/post-datepicker.tsx
```diff
@@ -0,0 +1,850 @@
...
+    this.dpContainer = this.host.shadowRoot.querySelector('.datepicker-container');
+
+    if (this.dpContainer) {
+      const options: AirDatepickerCustomOptions = {
+        navTitles: {
+          days: `<button aria-label="${this.labelSwitchYear}"><strong>MMMM yyyy</strong><post-icon size="small" name="2052"></post-icon></button>`,
+          months: `<button aria-label="${this.labelSwitchYear}"><strong>yyyy</strong><post-icon size="small" name="2052"></post-icon></button>`,
+        },
+        prevHtml: '<button><post-icon size="small" name="2049" ></post-icon></button>',
+        nextHtml: '<button><post-icon size="small" name="2050" ></post-icon></button>',
```
oliverschuerch: I think we should use the `chevronleft` and `chevronright` UI icons, no?

packages/components/src/components/post-datepicker/post-datepicker.tsx
```diff
@@ -0,0 +1,850 @@
...
+    this.dpInputs = assignedNodes?.filter(el => el.tagName === 'INPUT') as HTMLInputElement[];
+    this.setMinAndMaxToInputs();
+
+    this.dpContainer = this.host.shadowRoot.querySelector('.datepicker-container');
+
+    if (this.dpContainer) {
+      const options: AirDatepickerCustomOptions = {
+        navTitles: {
+          days: `<button aria-label="${this.labelSwitchYear}"><strong>MMMM yyyy</strong><post-icon size="small" name="2052"></post-icon></button>`,
+          months: `<button aria-label="${this.labelSwitchYear}"><strong>yyyy</strong><post-icon size="small" name="2052"></post-icon></button>`,
```
oliverschuerch: We should use the `chevrondown` UI icon here, right?

packages/components/src/components/post-datepicker/post-datepicker.scss
oliverschuerch, line 1: The first and last year in the year view are outgreyed.

Image

oliverschuerch, line 1: There is some wrong visual states when you select a date range:

**After selecting only the start date, before mouse leaves the start date cell**

Image

**When selecting the end date, the last date has a wrong fg-color**

Image

**When the selected range overlaps the current date**

Image

**When the selected range overlaps the focused element, after month change**

Image

packages/components/src/components/post-datepicker/post-datepicker.tsx
oliverschuerch, line 1: Disabled dates are still focussable.
Ignore this message, if this is wanted.

Image

oliverschuerch:
@leagrdv and @myrta2302 
Since this is a very big and complex component, I would like to have another reviewer involved, if this is ok?

packages/components/src/components/post-datepicker/post-datepicker.tsx
gfellerph: Change your browser or OS locale settings to see an updated format, this is the native HTML `<input type="date">`.

packages/components/src/components/post-datepicker/post-datepicker.scss
myrtasakellariou: This should be probably handled in the styles implementation ticket.

myrtasakellariou: Also in the scope of the styles implementation ticket.

packages/components/src/components/post-datepicker/post-datepicker.tsx
myrtasakellariou: Also in the scope of the styles implementation ticket.

packages/components/src/components/post-datepicker/post-datepicker.tsx
myrtasakellariou: Updated with `<post-icon name="chevronleft" ></post-icon> / <post-icon name="chevroright" ></post-icon>`

packages/components/src/components/post-datepicker/post-datepicker.tsx
myrtasakellariou: Updated with `<post-icon name="chevrondown" ></post-icon>`

packages/components/src/components/post-datepicker/post-datepicker.scss
leagrdv: Will be done in the style ticket

leagrdv: Will be done in the style ticket

packages/components/src/components/post-datepicker/post-datepicker.scss
leagrdv: Will be done in the style ticket

packages/documentation/src/stories/components/datepicker/datepicker.stories.ts
leagrdv: I excluded it from the args

packages/components/src/components/post-datepicker/post-datepicker.tsx
leagrdv: Updated the input type to "text", what I've implemented it the following:
By default it's "eu" format, meaning: dd.mm.yyyy
If html lang is `en-US`, then the format changes to mm/dd/yyyy

Because it's not a date input anymore, this is not done automatically. What has been implemented makes the datepicker work with english, french, italian and german.

leagrdv:
@alionazherdetska I've assigned this review to you as Oli won't be able to do it. Note that the design of the datepicker isn't the same as on Figma, this will be handled in another PR done by Myrta

packages/nextjs-integration/src/app/ssr/page.tsx
alizedebray, line 1: Have you checked SSR already?

packages/documentation/src/stories/components/datepicker/datepicker.stories.ts
```diff
@@ -0,0 +1,219 @@
...
+  return html`
+    <post-datepicker id=${args.id} range="true" ?min="${args.min}" ?max="${args.max}"
+    text-toggle-calendar="${args.textToggleCalendar}"
+    text-next-decade="${args.textNextDecade}"
+    text-next-month="${args.textNextMonth}"
+    text-next-year="${args.textNextYear}"
+    text-previous-decade="${args.textPreviousDecade}"
+    text-previous-month="${args.textPreviousMonth}"
+    text-previous-year="${args.textPreviousYear}"
+    text-switch-year="${args.textSwitchYear}">
```
alizedebray: This could be refactored to avoid duplication:
```ts
const textArgs = {
   textToggleCalendar: 'Open calendar',
   textNextDecade: 'Next decade',
   ...
};

const meta = {
    args: {
        ...textArgs
    }
};

const renderer = () => html`
    <post-datpicker ${spreadArgs(textArgs)}></post-datepicker>
`;
```

packages/documentation/src/stories/components/datepicker/datepicker.stories.ts
```diff
@@ -0,0 +1,219 @@
...
+      table: {
+        category: 'Labels',
+      },
+    },
+    textSwitchYear: {
+      table: {
+        category: 'Labels',
+      },
+    },
+  },
```
alizedebray: Not sure this is necessary, we are not doing this for other components.

packages/documentation/src/stories/components/datepicker/datepicker.stories.ts
```diff
@@ -0,0 +1,219 @@
+import { Args, StoryObj, StoryContext, StoryFn } from '@storybook/web-components-vite';
+import { html } from 'lit';
+import { MetaComponent } from '@root/types';
+
+const meta: MetaComponent = {
+  id: 'eb77cd02-48b2-42e1-a3e4-cd8a973d431e',
+  title: 'Components/Datepicker',
```
alizedebray: Since the styles are missing:
```suggestion
  title: 'Raw Components/Datepicker',
```

packages/components/src/components/post-datepicker/post-datepicker.tsx
```diff
@@ -0,0 +1,1016 @@
...
+export class PostDatepicker {
+  @Element() host: HTMLPostDatepickerElement;
+
+  @State() today = new Date();
+
+  @State() inputDisabled = false;
+
+  /**
+   * Selected date
+   * If range datepicker: Selected start date
```
alizedebray: Punctuation is missing in the JSdoc resulting in incorrectly formatted sentence in the docs: 

Image

packages/documentation/src/stories/components/datepicker/datepicker-disable.sample.ts
```diff
@@ -0,0 +1,8 @@
+window.addEventListener('DOMContentLoaded', () => {
+  const dp = document.querySelector('post-datepicker');
```
alizedebray:
```suggestion
  const dp: HTMLPostDatepickerElement = document.querySelector('post-datepicker');
```

packages/documentation/src/stories/components/datepicker/datepicker.docs.mdx
```diff
@@ -0,0 +1,53 @@
...
+<div className="docs-title">
+  # Datepicker
+
+  <link-design of={JSON.stringify(DatepickerStories)}></link-design>
+</div>
+
+<PackageTag meta={meta}/>
+
+<div className="lead mb-16">
+The Datepicker component allows users to select a single date or a range of dates either by typing into an input field or by interacting with a fly-out calendar.
```
alizedebray: The Figma description is:
```suggestion
A date pickers enables the user to input a single date or a date range via regular text input or selection within a fly-out calendar. 
```

packages/components/src/components/post-datepicker/post-datepicker.tsx
```diff
@@ -0,0 +1,1016 @@
...
+
+    if (this.selectedStartDate) {
+      const date = new Date(this.selectedStartDate);
+
+      if (isNaN(date.getTime())) {
+        console.error(
+          'Selected start date is not valid. Format should either be DD/MM/YYYY or YYYY/MM/DD',
+        );
+      }
+    }
```
alizedebray: Can you please create a `checkEmptyOrDate` function to avoid code duplication?

packages/components/src/components/post-datepicker/post-datepicker.tsx
```diff
@@ -0,0 +1,1016 @@
...
+   */
+  @Prop() renderCellCallback?: AirDatepickerCustomOptions['onRenderCell'];
+
+  /**
+   * Whether the calendar is inline in the page (not showing in a popover when input clicked)
+   */
+  @Prop() inline = false;
+  @Watch('inline')
+  validateInline() {
+    console.log('is inline', this.inline, this.dpInput);
```
alizedebray:
```suggestion
```

packages/nextjs-integration/src/app/ssr/page.tsx
leagrdv: I've opened a ticket: #7193 
I'll refine it when this one's merged

packages/components/src/components/post-datepicker/post-datepicker.tsx
```diff
@@ -0,0 +1,994 @@
...
+  validateMin() {
+    checkEmptyOrType(this, 'min', 'string');
+
+    if (this.min) {
+      const date = new Date(this.min);
+
+      if (isNaN(date.getTime())) {
+        console.error('Min date is not valid. Format should either be DD/MM/YYYY or YYYY/MM/DD');
+      }
+    }
```
alizedebray: This can be replaced by the `checkEmptyOrDate`, same for the `max` property.

alizedebray (changes requested):
A few more things I've noticed:
- [x] If you select a date in the dropdown, then blur the input, you get an error in the console
- [ ] If you select a date in the dropdown, then change the date manually and open the dropdown again, the selected date is incorrect
- [ ] The `input` and `change` events are not emitted when a date is selected from the dropdown
- [ ] The `input` event is emitted whenever a character changes when is should only be emitted when the value is a correct date
- [ ] The input value currently includes the mask like "25.02.___", it should be an empty string if the value is not a valid date and a "yyyy-mm-dd" date otherwise (native date input behavior)

leagrdv:
Adressed the changes except for those two:
- The input event is emitted whenever a character changes when is should only be emitted when the value is a correct date -> This is the native input event so I can't really prevent it from emitting.
-  The input value currently includes the mask like "25.02.___", it should be an empty string if the value is not a valid date and a "yyyy-mm-dd" date otherwise (native date input behavior) -> Same here, I don't think I can really have two different values for the input (the visual one and the one we want to give the user (empty string if invalid and yyyy-mm-dd if not)). One solution would be to have a second hidden input who would get the correct value (the non-visual one) but that would be on our side and the user input would still have the wrong value.

The `postUpdateDates` event of the `post-datepicker` should be used to get the actual date.

Issue #6869, opened 2025-12-15
Set font-size for buttons


alizedebray:
Currently the font-size for buttons is not tokenized leading to buttons using user agent styles.

- [ ] Create tokens for the font-size of default, small and large buttons
- [ ] Use these tokens in the styles package

sandra-post:
@alizedebray Since the font sizes for Desktop / Tablet / Mobile is 20 / 16 / 14 for all 3 sizes (small, medium, large), following token proposal:


```yml

# FONT-SIZE
post.button.font-size: '20 / 16 / 14' # post.device.font-size.6 

```

Does that work, or would you have a counter-proposal?

alizedebray:
@sandra-post perfect, thanks!

sandra-post:
@alizedebray Token is created and pushed to GitHub.

PR #6874, opened 2025-12-15
chore(components): remove post list from megadropdown

alizedebray:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 📝 Checklist

### DIFF
--- a/.changeset/orange-beers-knock.md
+++ b/.changeset/orange-beers-knock.md
@@ -23,7 +23,7 @@ BEFORE:
 AFTER:
 ```html
 <post-megadropdown label-close="Close" label-back="Back">
-  <a class="megadropdown-overview-link" href="/letters">Overview Letters</a>
+  <a class="post-megadropdown-overview" href="/letters">Overview Letters</a>
   <!-- Mega drop-down links -->
 </post-megadropdown>
 ```
--- a/packages/components-angular/projects/consumer-app/src/app/app.component.html
+++ b/packages/components-angular/projects/consumer-app/src/app/app.component.html
@@ -84,25 +84,34 @@
             <button>{{ dropdown.trigger }}</button>
           </post-megadropdown-trigger>
           <post-megadropdown [id]="dropdown.id" labelClose="Close" labelBack="Back">
-            <a class="post-megadropdown-overview" [href]="dropdown.overviewLink"
+            <a slot="post-megadropdown-overview" [href]="dropdown.overviewLink"
             >Overview {{ dropdown.trigger }}</a
             >
-            @for (section of dropdown.sections; track section.title) {
-              <post-list>
-                @if (section.titleLink) {
-                  <p>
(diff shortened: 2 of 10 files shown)


packages/styles/src/components/header/_post-megadropdown.scss
```diff
@@ -0,0 +1,49 @@
+@use '../../mixins/media';
+
+@use './mixins' as *;
```
oliverschuerch, line 3: Should we namespace that to something like `header-mx`, so it is clearer where the `$interactive-elements-selector` variables comes from? What do you think?

packages/styles/src/mixins/_button.scss
```diff
@@ -44,9 +44,11 @@
   background-color: transparent;
   box-shadow: none;
   font-family: inherit;
+  font-size: 1rem;
```
oliverschuerch, line 47: The buttons have different font-sizes per breakpoint and per button size variant. See the current [design](https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-Next-Level?node-id=302-9575&t=Iij7dA0hQ6xMwh9h-0).
This change adds the same font-size on all breakpoinst, for all button size variants.

| Small | Middle | Large |
| :-: | :-: | :-: |
| Image | Image | Image |

[Source](https://preview-6874--swisspost-design-system-next.netlify.app/?path=/docs/eb78afcb-ce92-4990-94b6-6536d5ec6af4--docs) of the taken printscreens

If adding font-size to buttons is aboslutely unavoidable, you could maybe mock (the soon to be added) font-size tokens, somehow like this:

```scss
/* components/button.scss */
:root {
  @include media.max(600px) {
    --post-button-sm-font-size: {number}px;
    --post-button-md-font-size: {number}px;
    --post-button-lg-font-size: {number}px;
  }
  
  @include media.between(600px, 1024px) {
    --post-button-sm-font-size: {number}px;
    --post-button-md-font-size: {number}px;
    --post-button-lg-font-size: {number}px;
  }
  
  @include media.min(1024px) {
    --post-button-sm-font-size: {number}px;
    --post-button-md-font-size: {number}px;
    --post-button-lg-font-size: {number}px;
  }
}

/* mixins/button.scss */
@mixin button-size($size: md) {
  ...
  font-size: var(--post-button-#{$size}-font-size);
  ...
}
```

alizedebray: Yes I know, I have checked the design and this is why I open the ticket #6869 we triaged together this morning.

packages/styles/src/components/header/_post-megadropdown.scss
alizedebray: It's not the only place where I did it that way (utilities use the same technique for example). 
I'm using `*` because it's the mixin of the current folder and I didn't see the added value of a namespace here, but we can discuss about this during a Dev Roundtable if you like.

PR #6911, opened 2025-12-22
fix(components): fix wrong data-expanded state after window resize

oliverschuerch:
## 📄 Description

This fixes an issue (currently only happening in FF), that caused the header component to end up in a wrong visual state, after receiving focus on mobile and a viewport resize afterwards (crossing mobile/tablet and desktop breakpoint boundaries).

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist


packages/components/src/components/post-header/post-header.tsx
```diff
@@ -367,10 +367,14 @@ export class PostHeader {
...
   @Listen('focusout')
-  onFocusChange() {
+  onFocusChange(e: FocusEvent) {
     const fixedElements =
       this.device === 'desktop' ? '.logo, .navigation' : '.global-header, .burger-menu';
     const isHeaderExpanded =
+      // ensure the expanded state stays accurate during focus changes,
+      // e.g., when the focused element is removed from the DOM
+      // during a window resize
+      e.target === document.activeElement &&
```
gfellerph, line 365: When focus is within a shadow DOM tree, `document.activeElement` will always be the root of the upper most shadow host. Don't know if that's an issue here, just be aware of that. It does not represent the element focused within the shadow root. You can get that via `root.activeElement`. 

-> https://stackoverflow.com/questions/78712866/document-activeelement-not-displaying-focused-elements-within-shadow-root

oliverschuerch: I'm going to check that, before I merge this PR, because there are some elements that may trigger the event, which completely live in the shadowDOM now (e.g. the logo, etc.).

oliverschuerch: If I get it right, the listeners `@Listen('focusin')` and `@Listen('focusout')` are registered on the host (e.g. `post-header`).
Therefore `e.target` will always return the `post-header` element if the element itself, or one of its shadowDOM children is focussed. This making a comparison with `document.activeElement` completely secure. This is also true if a slotted element is focused, because this lives in lightDOM anyway.

PR #6929, opened 2025-12-29
chore(styles): update focus ring in the main navigation

alizedebray:
## 📄 Description

Updates the styles of the main navigation items so that the focus ring is always visible. 

## 🚀 Demo

image
image
image
image

https://preview-6929--swisspost-design-system-next.netlify.app/iframe.html?viewMode=story&id=27a2e64d-55ba-492d-ab79-5f7c5e818498--active-navigation-item

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/styles/src/mixins/_utilities.scss
+++ b/packages/styles/src/mixins/_utilities.scss
@@ -156,7 +156,9 @@
 // this mixin should only be used if the default focus ring is not fully visible
 @mixin focus-inset($additional-selector: '') {
   &#{$additional-selector} {
-    outline-offset: calc(-2px - #{tokens.get('focus-outline-width', helpers.$post-focus)}) !important;
+    outline-offset: calc(
+      -2px - #{tokens.get('focus-outline-width', helpers.$post-focus)}
+    ) !important;
   }
 }
 


packages/components/src/components/post-logo/post-logo.scss
```diff
@@ -19,10 +19,8 @@
...
   @include utilities.focus-style;
-  @include utilities.focus-inset;
-}
-
-.logo-link :focus,
-:focus-visible {
-  border-radius: .5rem;
+  @include utilities.focus-style-custom {
+    border-radius: 0.5rem;
+    outline-offset: calc(-1 * #{tokens.get('focus-outline-width', helpers.$post-focus)}) !important;
```
gfellerph: So it does not touch the edges of the global header.

```suggestion
    outline-offset: calc(-2px - #{tokens.get('focus-outline-width', helpers.$post-focus)}) !important;
```

packages/styles/src/components/header/_mixins.scss
```diff
@@ -95,7 +96,8 @@ $interactive-elements-selector: 'a, button, [role="link"], [role="button"]';
   }
 
   @include utilities.focus-style-custom {
-    border-radius: 0.5rem;
+    outline-offset: calc(-2 * #{tokens.get('focus-outline-width', helpers.$post-focus)}) !important;
```
gfellerph: Multiplying is probably not the thing we actually want to do here. It's also a good idea to specify the unit.

```suggestion
    outline-offset: calc(-2px - #{tokens.get('focus-outline-width', helpers.$post-focus)}) !important;
```

packages/components/src/components/post-logo/post-logo.scss
alizedebray: With -4px it looks like this when the header is reduced:
image

Instead of: 
image

packages/styles/src/components/header/_mixins.scss
alizedebray: The `focus-outline-width` is 2px so both lines give the same result (-2 * 2px = -2px - 2px = -4px).
I did it this way so that it look even no matter the focus-outline-width but I can also change to `-3px - focus-outline-width = -5px` to match your example: 
image

packages/components/src/components/post-logo/post-logo.scss
gfellerph: I think the header should expand already when the focus is on the logo in minified mode. What do you think?

gfellerph (approved):

PR #6935, opened 2025-12-30
chore: remove old unused button mixin

gfellerph:
## 📄 Description

Removes a mixin from the past.


alizedebray (commented):
Can you please add a changeset for the removal?

PR #6946, opened 2026-01-05
chore(docs): deprecate header properties

alizedebray:
## 📄 Description

Deprecate header properties that are removed in v10.

## 🚀 Demo

https://preview-6946--swisspost-design-system.netlify.app/?path=/docs/ebb11274-091b-4cb7-9a3f-3e0451c9a865--docs

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist


gfellerph (changes requested):
Can you re-write the deprecation messages to say "this property WILL no longer be available in the future and be replaced by this other property". If I'm reading the notes, it sounds like the new header is already released and that the migration should happen immediately.

PR #7011, opened 2026-01-16
feat(components-react): export v2 icons from components react

gfellerph:
### New feature
Adds generated React icon components built from @swisspost/design-system-icons. It's now possible to import react- and SSR-ready icons directly like `import { PostIconLetter } from @swisspost/design-system-components-react/icons` and use it like `<PostIconLetter size="1.5em"></PostIconLetter`>. With this change, it's no longer necessary to list the `@swisspost/design-system-icons` as a dependency of the project.

### Fix for icons package
This change includes a fix for the icons package that makes all used ids unique (more unique than before at any rate) to make the svgs usable in light-dom where symbol id's need to be unique across icons. This was not an issue so far because the `<post-icon>` component uses shadow-dom, encapsulating the id scope.

---

## 🔮 Design review

- [ ] Design review done
- [X] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/components-react/scripts/generate-icons.mjs
+++ b/packages/components-react/scripts/generate-icons.mjs
@@ -13,8 +13,7 @@ const OUTPUT_DIR = path.resolve(process.cwd(), 'src/icons-generated');
  * @param {string} svgInnerContent
  * @returns
  */
-const getComponentTemplate = (iconName, base64) => `/* eslint-disable */
-/* Auto-generated file. Do not edit directly. */
+const getComponentTemplate = (iconName, base64) => `/* Auto-generated file. Do not edit directly. */
 import * as React from "react";
 import { PostIcon } from "../index.server";
 import type { StencilReactComponent } from "@stencil/react-output-target/runtime";
@@ -35,8 +34,7 @@ export default ${iconName};
  * @param {string[]} iconNames
  * @returns
  */
-const getIndexFileTemplate = iconNames => `/* eslint-disable */
-/* Auto-generated file. Do not edit directly. */
+const getIndexFileTemplate = iconNames => `/* Auto-generated file. Do not edit directly. */
 ${iconNames.map(name => `export { default as ${name} } from './${name}';`).join('\n')}
 `;
 
--- a/packages/nextjs-integration/src/app/ssr/page.tsx
+++ b/packages/nextjs-integration/src/app/ssr/page.tsx
@@ -26,13 +26,9 @@ import {
   PostStepperItem,
 } from '@swisspost/design-system-components-react/server';
 import {
-  PostIconAccessblocked,
-  PostIconAccessblockedSolid,
-  PostIconCarbatteryleakSolid,
   PostIconExplosives,
   PostIconLetter,
   PostIconLetterSolid,
-  PostIconPrint,
 } from '@swisspost/design-system-components-react/icons';
 
 export default function Home() {


packages/components-react/scripts/generate-icons.mjs
oliverschuerch, line 1: Imo, this should be a typescript file at the end. But I'd be ok with adding a ticket to do that later.

packages/components-react/scripts/generate-icons.mjs
```diff
@@ -0,0 +1,150 @@
...
+ * @param {string} svgInnerContent
+ * @returns
+ */
+const getComponentTemplate = (iconName, base64) => `/* eslint-disable */
+/* Auto-generated file. Do not edit directly. */
+import * as React from "react";
+import { PostIcon } from "../index.server";
+import type { StencilReactComponent } from "@stencil/react-output-target/runtime";
+
+const ${iconName}: StencilReactComponent<HTMLPostIconElement & {}> = props => (
```
oliverschuerch: For what is the `& {}`?

Deleting it in the build-script does not seem to make a difference.

packages/components-react/scripts/generate-icons.mjs
```diff
@@ -0,0 +1,150 @@
...
+  <PostIcon
+    name="${iconName}"
+    url="${`data:image/svg+xml;base64,${base64}`}"
+    {...props}
```
oliverschuerch, line 25: While the `<post-icon>` properties `flip-h`, `flip-v`, `scale` and `rotate` already can be applied to the react components and work, the `animation` property does not work. It is somehow not passed from the react to the web-component.

Unrelated, but similar important...
Props need to be transformed from camelCase (used in react) to kebab-case (used in html). Or else props like `flipH` and `flipV` can't be used on the components. This is something usually stencil is doing for us, but in this special case, we have to do it on our own.

packages/documentation/src/stories/packages/components-react/icon.sample.tsx
```diff
@@ -0,0 +1,5 @@
+import { PostIconLetter } from '@swisspost/design-system-components-react/icons';
+
+export default function Page() {
+  return <PostIconLetter style={{ fontSize: '1.5em' }}></PostIconLetter>;
```
oliverschuerch: Since you added an example to change the icon size, shouldn't we add also the information about how to change color, or add a predefined transformation (scale, rorate, flip-h, etc.) or an animation?

packages/nextjs-integration/src/app/ssr/page.tsx
```diff
@@ -224,6 +230,14 @@ export default function Home() {
         textFirst="First page"
         textLast="Last page"
       />
+
+      <h2>Icons</h2>
+      <div className="d-flex gap-16">
+        <PostIconAccessblocked></PostIconAccessblocked>
+        <PostIconLetter></PostIconLetter>
+        <PostIconCarbatteryleakSolid></PostIconCarbatteryleakSolid>
```
oliverschuerch: Add one with the props `flipV` and `flipH`, one with `scale` and `rotate` and one with `animation`, to see if they work ;)

packages/components-react/scripts/generate-icons.mjs
gfellerph: Update the reflect settings for the post-icon to support these use-cases.

packages/components-react/scripts/generate-icons.mjs
gfellerph: Removed it.

packages/nextjs-integration/src/app/ssr/page.tsx
gfellerph: Added.

packages/documentation/src/stories/packages/components-react/icon.sample.tsx
gfellerph: Updated the code example.

PR #7014, opened 2026-01-16
fix(component): resolve burger-menu button state sync issue on first click in `post-header` component

alionazherdetska:
## 📄 Description

Fixes the burger menu button not displaying the correct icon on the first click. The button now properly shows the close icon when the menu opens and the burger icon when it closes.

## Problem

image

When clicking the burger menu button for the first time, the menu would open correctly but the button icon would not update to show the close icon. This was caused by a race condition where both the button's internal click handler and the parent header's click handler were updating the button state simultaneously, resulting in a double-toggle that left the button in the wrong state.

**Sequence of the bug:**
1. User clicks button
2. Button's internal handler toggles: `toggled: false → true` 
3. Header's handler also sets: `toggled = true` (based on old state)
4. Button's internal handler completes and toggles again: `toggled: true → false`
5. Result: Menu is open but button shows wrong icon

## Solution

Implemented event handler phase separation using the capture phase:

1. **post-togglebutton**: Added capture phase listener so the button updates its state first before parent handlers run
2. **post-header**: Modified to receive the button's already-updated state and only update the button programmatically when not triggered by a button click

## 🚀 Preview link:

[Header docs](https://preview-7014--swisspost-design-system-next.netlify.app/?path=/story/27a2e64d-55ba-492d-ab79-5f7c5e818498--default)
---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-header/post-header.tsx
+++ b/packages/components/src/components/post-header/post-header.tsx
@@ -230,7 +230,7 @@ export class PostHeader {
       : fade(this.burgerMenu, 'in', this.animationOptions);
 
     // Update the state of the toggle button
-    if (this.burgerMenuButton) {
+    if (this.burgerMenuButton && this.burgerMenuButton.toggled !== nextExtendedState) {
       this.burgerMenuButton.toggled = nextExtendedState;
     }
 


packages/components/src/components/post-header/post-header.tsx
```diff
@@ -219,12 +219,20 @@ export class PostHeader {
   @Method()
   async toggleBurgerMenu(force?: boolean) {
```
alizedebray: To avoid repeating `force ?? !this.burgerMenuExtended` in the function body:
```suggestion
  async toggleBurgerMenu(nextExtendedState = !this.burgerMenuExtended) {
```

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -476,7 +484,7 @@ export class PostHeader {
                 {this.hasNavigation && this.device !== 'desktop' && (
                   <post-togglebutton
                     ref={el => (this.burgerMenuButton = el)}
-                    onClick={() => this.toggleBurgerMenu()}
+                    onClick={() => this.toggleBurgerMenu(this.burgerMenuButton?.toggled)}
```
alizedebray: This change is not needed.
```suggestion
                    onClick={() => this.toggleBurgerMenu()}
```

packages/components/src/components/post-togglebutton/post-togglebutton.tsx
```diff
@@ -19,8 +19,9 @@ export class PostTogglebutton {
   @Prop({ mutable: true }) toggled: boolean = false;
 
   componentWillLoad() {
-    // add event listener to not override listener that might be set on the host
-    this.host.addEventListener('click', () => this.handleClick());
+    // add event listener in capture phase so the toggle state is updated
+    // before any parent bubble listeners react to the click.
+    this.host.addEventListener('click', () => this.handleClick(), true);
```
alizedebray: To make it more explicit: 
```suggestion
    this.host.addEventListener('click', () => this.handleClick(), { capture: true });
```

packages/components/src/components/post-togglebutton/post-togglebutton.tsx
```diff
@@ -19,8 +19,9 @@ export class PostTogglebutton {
...
     this.host.addEventListener('keydown', (e: KeyboardEvent) => this.handleKeydown(e));
```
alizedebray: To ensure we don't have the same issue with keyboard navigation:
```suggestion
    this.host.addEventListener('keydown', (e: KeyboardEvent) => this.handleKeydown(e), { capture: true });
```

.changeset/light-buses-create.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Fixed the burger menu button in the `post-header` component not updating its icon correctly on the first click.
```
alizedebray:
```suggestion
Fixed the burger menu button in the `post-header` component not updating its state correctly on the first click.
```

packages/components/src/components/post-header/post-header.tsx
```diff
@@ -217,26 +217,34 @@ export class PostHeader {
...
+      return;
+    }
+
     this.burgerMenuAnimation = this.burgerMenuExtended
       ? fade(this.burgerMenu, 'out', this.animationOptions)
       : fade(this.burgerMenu, 'in', this.animationOptions);
 
     // Update the state of the toggle button
-    if (this.burgerMenuButton) this.burgerMenuButton.toggled = force ?? !this.burgerMenuExtended;
+    if (this.burgerMenuButton) {
```
alizedebray: You've removed this check: 
```suggestion
    if (this.burgerMenuButton && this.burgerMenuButton.toggled !== nextExtendedState) {
```

PR #7075, opened 2026-01-26
fix(components): update getFocusableChildren not to return hidden elements

alizedebray:
## 📄 Description

This PR fixes the focus issues for the language menu.

## 🚀 Demo

https://preview-7075--swisspost-design-system-next.netlify.app/?path=/docs/27a2e64d-55ba-492d-ab79-5f7c5e818498--docs

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/components/src/utils/get-focusable-children.ts
+++ b/packages/components/src/utils/get-focusable-children.ts
@@ -32,7 +32,7 @@ export function getFocusableChildren(element: Element | Document | ShadowRoot): 
     `${focusableSelector}:not(${focusDisablingSelector})`,
   );
 
-  return Array.from(focusableChildren).filter(isVisible);
+  return Array.from(focusableChildren).filter(el => !isFocusBlockedByCSS(el));
 }
 
 // Searches deeper across shadow DOM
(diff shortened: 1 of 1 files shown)


packages/components/src/utils/get-focusable-children.ts
oliverschuerch, line 1: Should we use the browser Element [`checkVisibility()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility) method and only implement the fallback below for older browsers?

With the current implementation I see some unhandled edge-cases:
- A CSS [`visibility: collapse`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/visibility) statement could potentially also cause the element to be hidden.
- Same with a CSS `opacity: 0` statement.

alizedebray: I added the `checkVisibility` method. 

Concerning the `visibility: collapse`, it is already taken into account with `style.visibility !== 'visible'`.
And `opacity: 0` does not prevent focus, I updated the name of the function to make it clear that it is not exclude invisible elements but on elements that are not focusable.

PR #7097, opened 2026-01-29
fix(components): null check on `post-stepper`

leagrdv:
## 📄 Description

Fixed an error in the console that occured because the label was being read even though it was not yet present in the DOM.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [X] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-stepper/post-stepper.tsx
+++ b/packages/components/src/components/post-stepper/post-stepper.tsx
@@ -99,8 +99,10 @@ export class PostStepper {
     this.updateActiveStepNumber();
 
     this.stepItems.forEach((el, i) => {
-      if (this.currentIndex === i) {
-        this.mobileActiveStepName = el.querySelector('.label').innerHTML;
+      const labelEl = el.querySelector('.label');
+
+      if (this.currentIndex === i && labelEl) {
+        this.mobileActiveStepName = labelEl.innerHTML;
       }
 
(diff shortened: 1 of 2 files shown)


packages/components/src/components/post-stepper/post-stepper.tsx
```diff
@@ -99,8 +99,10 @@ export class PostStepper {
     this.updateActiveStepNumber();
 
     this.stepItems.forEach((el, i) => {
-      if (this.currentIndex === i) {
-        this.mobileActiveStepName = el.querySelector('.label').innerHTML;
+      const labelEl = el.querySelector('.label');
```
alizedebray, line 102: If the label does not exist, it means that not all stepper items have been loaded yet and that updateSteps will be called again.

To avoid performing the same logic twice, you can exit the function earlier with:
```js
const currentStep = this.stepItems[this.currentIndex];
if (!currentStep || !currentStep.querySelector('.label')) return;
```

leagrdv: That doesn't actually work as currentIndex might be negative in the case that the stepper hasn't yet started (currentIndex = -1) or has been completed (currentIndex higher than the last step), and in that case, the steps would never be updated

alizedebray (approved):

PR #7142, opened 2026-02-04
chore: enable shadow root for the post-menu

alizedebray:
## 📄 Description

Enable shadow root for the `post-menu-trigger` and simplify code.

## 🚀 Demo

https://preview-7142--swisspost-design-system-next.netlify.app/iframe?viewMode=story&id=8ca2bd70-56e6-4da9-b1fd-4e55388dca88--default

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-menu-trigger/post-menu-trigger.tsx
+++ b/packages/components/src/components/post-menu-trigger/post-menu-trigger.tsx
@@ -1,4 +1,4 @@
-import { Component, Element, Prop, h, Host, Watch } from '@stencil/core';
+import { Component, Element, Prop, h, Host, Watch, Listen } from '@stencil/core';
 import { version } from '@root/package.json';
 import { getRoot, checkRequiredAndType } from '@/utils';
 
@@ -9,6 +9,8 @@ import { getRoot, checkRequiredAndType } from '@/utils';
 })
 export class PostMenuTrigger {
   @Element() host: HTMLPostMenuTriggerElement;
+
+  private slottedButton: HTMLButtonElement;
 
   /**
    * ID of the menu element that this trigger is linked to. Used to open and close the specified menu.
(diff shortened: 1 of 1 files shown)


packages/components/src/components/post-menu-trigger/post-menu-trigger.tsx
```diff
@@ -1,104 +1,81 @@
...
+    button.setAttribute('aria-haspopup', 'menu');
+    button.setAttribute('aria-expanded', 'false');
+
+    button.addEventListener('click', this.handleToggle.bind(this));
+    button.addEventListener('keydown', this.handleKeyDown.bind(this));
+
+    // Listen to the `toggleMenu` event emitted by the `post-menu` component
+    const menu = this.getMenu();
+    if (menu) {
+      menu.addEventListener('toggleMenu', (event: CustomEvent<boolean>) => {
```
alionazherdetska: Since the menu element could potentially outlive the trigger component, do you think it would be a good idea to also remove this listener in `disconnectedCallback()` to prevent potential memory leaks? Or is there something about the lifecycle here that makes cleanup unnecessary?

packages/components/src/components/post-menu-trigger/post-menu-trigger.tsx
```diff
@@ -1,104 +1,81 @@
...
+    if (ref && ref.localName === 'post-menu') return ref as HTMLPostMenuElement;
 
-      this.slottedButton.addEventListener('click', () => {
-        this.handleToggle();
-      });
-      this.slottedButton.addEventListener('keydown', this.handleKeyDown);
-    } else {
-      console.warn('No button found within post-menu-trigger');
-    }
+    console.warn(`No post-menu found with ID: ${this.for}`);
```
alionazherdetska, line 75: I noticed that `getMenu()` is called on every click through `handleToggle()`. If the menu isn't found, would this result in repetitive console warnings? Would it make sense to validate the menu existence earlier in `componentDidLoad()`?

alizedebray: Doing it this way allows the `post-menu` to be added to the DOM asynchronously while still being triggered by the `post-menu-trigger`. The first time you try to open it, the menu may not exist yet, so a warning is logged. By the second attempt, the menu may have been added and no warning appears.

packages/components/src/components/post-menu-trigger/post-menu-trigger.tsx
alizedebray: I actually changes this part to use a `@Listen` decorator to handle cases were the `post-menu` is added to the DOM asynchronously.

PR #7149, opened 2026-02-06
fix(react-icons): removed default exports from icons

gfellerph:
## 📄 Description

The default exports were blocking tree shaking, leading bundlers to bundle 6MB of icons when a single one was imported.


alionazherdetska (approved):
☺️

PR #7156, opened 2026-02-06
chore(components): update utils to prepare for number input

alizedebray:
## 📄 Description

Update component utilities that will be used in the number input component.

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist

### DIFF
--- a/packages/components/src/utils/repeat-on-long-press.ts
+++ b/packages/components/src/utils/repeat-on-long-press.ts
@@ -16,8 +16,6 @@ export function repeatOnLongPress(callback: () => void, options?: LongPressOptio
   window.addEventListener('pointerup', handler);
   window.addEventListener('pointercancel', handler);
   window.addEventListener('pointerleave', handler);
-
-  return stop;
 }
 
 function repeat(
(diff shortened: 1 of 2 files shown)


packages/components/src/utils/repeat-on-long-press.ts
```diff
@@ -0,0 +1,39 @@
...
+    window.removeEventListener('pointerup', handler);
+    window.removeEventListener('pointercancel', handler);
+    window.removeEventListener('pointerleave', handler);
+  };
+
+  window.addEventListener('pointerup', handler);
+  window.addEventListener('pointercancel', handler);
+  window.addEventListener('pointerleave', handler);
+
+  return stop;
```
alionazherdetska: I noticed that when `stop()` is called manually, the event listeners aren't removed from window.
The current `post-number-input` doesn't call stop() manually, so there's no memory leak right now. But the utility should probably handle both cases.
What do you think?

alionazherdetska (changes requested):
The SonarCloud analysis is failing. Could you please take a look when you get a chance☺️

packages/components/src/utils/repeat-on-long-press.ts
alizedebray: Good catch! That’s a leftover from my first attempt to stop the function manually from the component itself. It adds a lot of unnecessary code though, so the automatic approach alone is enough for now in my opinion. We can always add a manual option later if we need it.

PR #7157, opened 2026-02-06
feat(components): add post-number-input component

alizedebray:
## 📄 Description

This PR adds the `post-number-input` component as a raw component.

## 🚀 Demo

https://preview-7157--swisspost-design-system-next.netlify.app/?path=/docs/d5f43fa8-42ba-4cb9-98c7-9386d4c939bb--docs

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 📝 Checklist

### DIFF
--- a/.changeset/ripe-glasses-invent.md
+++ b/.changeset/ripe-glasses-invent.md
@@ -1,0 +1,6 @@
+---
+'@swisspost/design-system-components': minor
+'@swisspost/design-system-documentation': minor
+---
+
+Documented the `post-pagination` component and added design-token–based styles aligned with the design.
--- a/packages/components/cypress/e2e/pagination.cy.ts
+++ b/packages/components/cypress/e2e/pagination.cy.ts
@@ -4,7 +4,6 @@ interface PaginationEl {
   page: number;
   pageSize: number;
   collectionSize: number;
-  disabled?: boolean;
 }
 
 describe('pagination', () => {
@@ -15,14 +14,7 @@ describe('pagination', () => {
         .find('.pagination-link')
         .not('.pagination-control-button')
         .as('pageButtons');
-      cy.get('@pagination')
-        .find('.pagination-control .pagination-control-button')
-        .first()
-        .as('prevButton');
-      cy.get('@pagination')
-        .find('.pagination-control .pagination-control-button')
-        .last()
-        .as('nextButton');
+      cy.get('@pagination').find('.pagination-control .next-button').as('nextButton');
     });
 
     it('should render', () => {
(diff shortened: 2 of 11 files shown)


packages/components/src/components/post-number-input/post-number-input.tsx
```diff
@@ -0,0 +1,117 @@
...
+    });
+  }
+
+  private updateStepButtonState() {
+    const min = parseFloat(this.input.min);
+    const max = parseFloat(this.input.max);
+    const value = parseFloat(this.input.value);
+
+    this.isIncrementDisabled = !isNaN(max) && value >= max;
+    this.isDecrementDisabled = !isNaN(min) && value <= min;
```
myrtasakellariou: 1. Should n't these also be disabled in the case of a disabled input?

2. Few edge cases: When **min > max** (e.g. min = 10 and max = -10) and the input has no value, the increment/decrement buttons don’t update (which is expected), but should nt the whole component be disabled?
Image

3. When **min = max**:

**and  value  > max:**

Image

**and value < min:**

Image

packages/components/src/components/post-number-input/post-number-input.tsx
```diff
@@ -0,0 +1,117 @@
...
+        this.input.stepUp();
+      } else {
+        this.input.stepDown();
+      }
+
+      this.updateStepButtonState();
+    });
+  }
+
+  private updateStepButtonState() {
```
myrtasakellariou, line 106: Not sure if this is required, but in the absence of the correct input element, there are console.errors because of `this.input` being undefined.

myrta2302 (dismissed):

packages/documentation/src/stories/components/number-input/number-input.stories.ts
```diff
@@ -0,0 +1,140 @@
...
+
+const meta: MetaComponent = {
+  id: 'd5f43fa8-42ba-4cb9-98c7-9386d4c939bb',
+  title: 'Raw Components/Form Number Input',
+  component: 'post-number-input',
+  tags: ['package:WebComponents', 'status:InProgress'],
+  parameters: {
+    design: {
+      type: 'figma',
+      url: 'https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-Next-Level?node-id=14016-18313&t=xQdJ3KtfCT99hmhK-4',
```
leagrdv: The component is going to be moved when it will be released so this link won't be working anymore. It might be best not to put a link at the moment, that way we can see it more clearly that this needs to be added later.

packages/components/src/components/post-number-input/post-number-input.tsx
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  get isAboveMax(): boolean {
+    return compare(this.value, this.max, (v, m) => v > m);
+  }
+
+  /**
+   * The icon to be used in the control that increases the number in the input.
+   */
+  @Prop() incrementIcon = 'plus';
```
leagrdv: Do we want the user to be able to choose any icon?
We don't allow it for other components (accordion toggle icon, close button icon, back to top button, stepper valid step icon)

alizedebray: I agree, but it was in the ticket so added the feature anyways.
@gfellerph @sandra-post Do you see use cases where this might be needed?

alizedebray: I've extracted the icon properties into the PR https://github.com/swisspost/design-system/pull/7203 to unblock this one until a decision is taken.

PR #7177, opened 2026-02-12
chore(components): replace IS_BROWSER/IS_SERVER utils with Build.isBrowser/isServer

myrta2302:
## 📄 Description

This tickets replaces the previous implementation of `IS_BROWSER`, `IS_SERVER` with the Stencil Build Constants `isBrowser` and `IsServer`.

Initial Testing (before util update)
------------------------
In order to test both, the `IS_BROWSER` and `Build.isBrowser` flags were added to all relevant component renderers as below:

`{IS_BROWSER.toString()}
{Build.isBrowser.toString()}`

The Next.js integration app was run in and opened in two tabs with JavaScript enabled and disabled. In both cases the previous `IS_BROWSER` implementation resolved to `true`, while `Build.isBrowser` was `true` only for the JS-enabled version and `false` for the JS-disabled version.

Test Environment behaviour of updated util
----------------------------------
In Jest the `Build.isBrowser` defaults to `false` and the `Build.isServer` to `true` (possibly because Jest runs on node.js), thus the `isBrowser` has to be mocked to true.

In Cypress and Playwright, `Build.isBrowser` defaults to `true` and the `Build.isServer` to `false`, as expected since they run in a real Browser.

Previous Implementation files affected:
------------------------
hydrate-app-helper.ts // removed
post-env-test.tsx // removed
environment.spec.ts // removed
environment.ts // updated

---------------------------------

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -45,7 +45,7 @@ importers:
     devDependencies:
       '@changesets/cli':
         specifier: 2.29.8
-        version: 2.29.8(@types/node@25.5.2)
+        version: 2.29.8(@types/node@25.6.0)
       start-server-and-test:
         specifier: 2.1.3
         version: 2.1.3
@@ -271,16 +271,16 @@ importers:
     devDependencies:
       '@angular/build':
         specifier: 20.3.14
-        version: 20.3.14(@angular/compiler-cli@20.3.16(@angular/compiler@20.3.18)(typescript@5.8.3))(@angular/compiler@20.3.18)(@angular/core@20.3.18(@angular/compiler@20.3.18)(rxjs@7.8.2)(zone.js@0.15.1))(@angular/platform-browser@20.3.16(@angular/animations@20.3.16(@angular/core@20.3.18(@angular/compiler@20.3.18)(rxjs@7.8.2)(zone.js@0.15.1)))(@angular/common@20.3.16(@angular/core@20.3.18(@angular/compiler@20.3.18)(rxjs@7.8.2)(zone.js@0.15.1))(rxjs@7.8.2))(@angular/core@20.3.18(@angular/compiler@20.3.18)(rxjs@7.8.2)(zone.js@0.15.1)))(@types/node@25.5.2)(chokidar@4.0.3)(jiti@2.6.1)(karma@6.4.4)(less@4.6.4)(lightningcss@1.30.2)(ng-packagr@20.1.0(@angular/compiler-cli@20.3.16(@angular/compiler@20.3.18)(typescript@5.8.3))(tailwindcss@4.1.18)(tslib@2.8.1)(typescript@5.8.3))(postcss@8.5.6)(sass-embedded@1.78.0)(tailwindcss@4.1.18)(terser@5.46.1)(tslib@2.8.1)(typescript@5.8.3)(yaml@2.8.3)
+        version: 20.3.14(@angular/compiler-cli@20.3.16(@angular/compiler@20.3.18)(typescript@5.8.3))(@angular/compiler@20.3.18)(@angular/core@20.3.18(@angular/compiler@20.3.18)(rxjs@7.8.2)(zone.js@0.15.1))(@angular/platform-browser@20.3.16(@angular/animations@20.3.16(@angular/core@20.3.18(@angular/compiler@20.3.18)(rxjs@7.8.2)(zone.js@0.15.1)))(@angular/common@20.3.16(@angular/core@20.3.18(@angular/compiler@20.3.18)(rxjs@7.8.2)(zone.js@0.15.1))(rxjs@7.8.2))(@angular/core@20.3.18(@angular/compiler@20.3.18)(rxjs@7.8.2)(zone.js@0.15.1)))(@types/node@25.6.0)(chokidar@4.0.3)(jiti@2.6.1)(karma@6.4.4)(less@4.6.4)(lightningcss@1.30.2)(ng-packagr@20.1.0(@angular/compiler-cli@20.3.16(@angular/compiler@20.3.18)(typescript@5.8.3))(tailwindcss@4.1.18)(tslib@2.8.1)(typescript@5.8.3))(postcss@8.5.6)(sass-embedded@1.78.0)(tailwindcss@4.1.18)(terser@5.46.1)(tslib@2.8.1)(typescript@5.8.3)(yaml@2.8.3)
       '@angular/cli':
         specifier: 20.3.14
-        version: 20.3.14(@types/node@25.5.2)(chokidar@4.0.3)(hono@4.12.10)
+        version: 20.3.14(@types/node@25.6.0)(chokidar@4.0.3)(hono@4.12.12)
       '@angular/compiler-cli':
         specifier: 20.3.16
         version: 20.3.16(@angular/compiler@20.3.18)(typescript@5.8.3)
       '@cypress/schematic':
         specifier: 3.0.0
-        version: 3.0.0(@angular/cli@20.3.14(@types/node@25.5.2)(chokidar@4.0.3)(hono@4.12.10))(@angular/core@20.3.18(@angular/compiler@20.3.18)(rxjs@7.8.2)(zone.js@0.15.1))
+        version: 3.0.0(@angular/cli@20.3.14(@types/node@25.6.0)(chokidar@4.0.3)(hono@4.12.12))(@angular/core@20.3.18(@angular/compiler@20.3.18)(rxjs@7.8.2)(zone.js@0.15.1))
       '@eslint/js':
         specifier: 9.18.0
         version: 9.18.0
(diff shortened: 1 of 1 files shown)


packages/components/src/utils/environment.ts
```diff
@@ -1,53 +1,4 @@
...
-/**
- * This is the final hydrate app flag
- * if it`s `true`, IS_BROWSER equals `false` and IS_SERVER equals `true
- */
-const isHydrateApp = isHydrateAppEnv || isTestAppServerTestEnv;
-
-export const IS_BROWSER: boolean = hasWindow && !isHydrateApp;
-export const IS_SERVER: boolean = !IS_BROWSER;
+export const IS_BROWSER: boolean = hasWindow && Build.isBrowser;
+export const IS_SERVER: boolean = !hasWindow && Build.isServer;
```
alizedebray: The two are not equivalent, what you want is probably this: 
```suggestion
export const IS_BROWSER: boolean = hasWindow && Build.isBrowser;
export const IS_SERVER: boolean = !hasWindow || Build.isServer;
```

Or even:
```suggestion
export const IS_BROWSER: boolean = hasWindow && Build.isBrowser;
export const IS_SERVER: boolean = !IS_BROWSER;
```

packages/components/src/components/post-icon/post-icon.tsx
```diff
@@ -97,7 +103,7 @@ export class PostIcon {
   private getUrl(): string {
     const fileName = `${this.name}.svg`;
 
-    if (!IS_BROWSER && !this.base) {
+    if (IS_SERVER && !this.base) {
```
alizedebray: Then this change can be reverted.

packages/components/src/components/post-icon/post-icon.tsx
```diff
@@ -97,9 +103,7 @@ export class PostIcon {
   private getUrl(): string {
     const fileName = `${this.name}.svg`;
 
-    if (!IS_BROWSER && !this.base) {
-      return `${CDN_URL}${fileName}`;
```
alizedebray, line 101: Why did you remove this logic?

packages/components/src/utils/tests/is-motion-reduced.spec.ts
alizedebray, line 1: Isn't it possible to simply mock the import?
 
```ts
jest.mock('./environment', () => ({
  IS_BROWSER: true,
}));
```

myrtasakellariou: I ve tried to apply this```
 jest.mock('../environment', () => ({
  IS_BROWSER: true,
}));``` but unfortunately the IS_BROWSER still logs `false` and the test fails.

packages/components/src/utils/index.ts
oliverschuerch, line 1: Please reset this unnecessary change.

packages/components/src/components/post-icon/post-icon.tsx
oliverschuerch, line 1: Please reset this unnecessary change.

packages/components/src/animations/collapse.ts
```diff
@@ -1,6 +1,7 @@
-import { IS_BROWSER, isMotionReduced } from '@/utils';
-
+import { isMotionReduced } from '@/utils';
+import { Build } from '@stencil/core';
```
oliverschuerch:
```suggestion
import { Build } from '@stencil/core';
import { isMotionReduced } from '@/utils';

```

packages/components/src/animations/collapse.ts
```diff
@@ -1,6 +1,7 @@
...
 const collapseDuration = 350;
+
```
oliverschuerch:
```suggestion
```

packages/components/src/utils/tests/is-motion-reduced.spec.ts
```diff
@@ -1,19 +1,24 @@
...
 
   beforeEach(() => {
+    jest.resetModules();
     matchMedia = window.matchMedia = jest.fn();
   });
 
-  it('should return true if reduced motion is requested', () => {
+  it('should return true if reduced motion is requested', async () => {
     matchMedia.mockReturnValue({ matches: true });
+    const { Build } = await import('@stencil/core');
```
oliverschuerch: Can we move that to the `beforeEach` function?

packages/components/src/components.d.ts
oliverschuerch, line 1: I don't think we can delete this file!

packages/nextjs-integration/src/app/ssr/page.tsx
oliverschuerch, line 1: Why are these changes part of this PR?

packages/components/src/components.d.ts
myrtasakellariou: it was a merge mistake

packages/nextjs-integration/src/app/ssr/page.tsx
myrtasakellariou: These are added to bypass the CORS error in the tests.

oliverschuerch: Ah I see.
Then please use one of our general ways to specify the base path, instead of adding it to every icon manually:
https://next.design-system.post.ch/?path=/docs/40ed323b-9c1a-42ab-91ed-15f97f214608--docs#configure-the-icons-base-path

packages/components/src/utils/tests/is-motion-reduced.spec.ts
```diff
@@ -1,18 +1,21 @@
...
   let matchMedia;
+  let isMotionReduced;
 
-  beforeEach(() => {
+  beforeEach(async () => {
+    jest.resetModules();
     matchMedia = window.matchMedia = jest.fn();
+    const { Build } = await import('@stencil/core');
+    Build.isBrowser = true; // Mocking the Browser, as Stencil's Build.isBrowser = false and Build.isServer = true in Jest.
+    ({ isMotionReduced } = await import('../is-motion-reduced'));
```
oliverschuerch: Now that I see your current implementation I start thinking, that the approach from @alizedebray actually is the right way to go.

Try the following?
It it works, we can make the stencil/core mock a utility to be used in other tests.

```ts
// mock first
jest.mock('@stencil/core', () => {
  const module = jest.requireActual('@stencil/core');
  return {
    ...module,
    Build: {
      ...module.Build,
      isBrowser: true,
      isServer: false,
    },
  };
});

let mockMatchMedia: boolean | undefined;

Object.defineProperty(window, 'matchMedia', {
  value: jest.fn(() => ({
    matches: mockMatchMedia ?? false,
    addListener: jest.fn(),
    removeListener: jest.fn(),
  })),
});

// then import the modules which need the mocked objects
// note: this is not allowed in esmodules but works fine in jest ;)
import { isMotionReduced } from '../is-motion-reduced';

describe('isMotionReduced', () => {
  it('should return true if reduced motion is requested', () => {
    mockMatchMedia = true;
    expect(isMotionReduced()).toBe(true);
  });

  it('should return false if reduced motion is not requested', () => {
    mockMatchMedia = false;
    expect(isMotionReduced()).toBe(false);
  });
});

```

oliverschuerch (changes requested):
test

packages/nextjs-integration/src/app/ssr/page.tsx
myrtasakellariou: Not needed anymore after post-icon.tsx update.

packages/components/src/components/post-icon/post-icon.tsx
```diff
@@ -96,26 +96,29 @@ export class PostIcon {
...
-
-    if (!IS_BROWSER && !this.base) {
-      return `${CDN_URL}${fileName}`;
-    }
+    const isAbsolute = (url: string) => /^https?:\/\//.test(url) || url.startsWith('data:');
 
     if (this.url) {
+      if (Build.isServer && !isAbsolute(this.url)) {
+        return `${CDN_URL}${fileName}`;
+      }
```
oliverschuerch: This is NOT necessary, if `this.url` contains a data image!
But it is necessary, if it contains something else. No matter if it's an absolute or a relative URL.

And that's why the correct order must be:

1. Use `this.url`, if it contains a data image (server | client).
2. Use CDN fallback (server only).
3. Use `this.url`, if it IS defined (server | client).
4. Or, if none of the above cases is true, use something else...

packages/components/src/components/post-icon/post-icon.tsx
```diff
@@ -96,26 +96,29 @@ export class PostIcon {
 
   private getUrl(): string {
     const fileName = `${this.name}.svg`;
-
-    if (!IS_BROWSER && !this.base) {
-      return `${CDN_URL}${fileName}`;
-    }
+    const isAbsolute = (url: string) => /^https?:\/\//.test(url) || url.startsWith('data:');
 
     if (this.url) {
```
oliverschuerch:
```suggestion
    const isDataUrl = this.url?.startsWith('data:image/');

    if (isDataUrl) {
```

packages/components/src/utils/tests/helpers/mock-match-media.ts
```diff
@@ -0,0 +1,12 @@
+// Mock the browser's matchMedia API on the global object
+export const mockConfig = { matchMedia: false as boolean | undefined };
```
oliverschuerch: Should we namespace the exported object, to prevent naming conflicts?
And also the option one can set should match the property which is mocked.

Also the value of the property will never be something else than a boolean, since `globalThis.matchMedia()` can't return something else!

```suggestion
export type mockConfigMatchMediaType = {
  matches: boolean;
};

export const mockConfigMatchMedia: mockConfigMatchMediaType = { matches: false };
```

With this, you could also remove the `?? false` fallback below.

oliverschuerch (approved):

PR #7223, opened 2026-02-26
feat(component): added a klp-login-widget component

### LINKED ISSUE #7175: [web component func] KLP Login Widget
### 📝 Description

Implement a minimal version of the KLP Login Widget.

### ♿ Accessibility

Re-use the [user menu example](https://swisspost-design-system-next.netlify.app/?path=/docs/27a2e64d-55ba-492d-ab79-5f7c5e818498--docs#user-menu) from the header's docs.

### ⚙️ Requirements

Makes an API request to https://n.account.post.ch/v1/session/subscribe initially and uses the result to render the widget or display the login link. Available links in the menu are user-defined.

Mocked data can be found at https://github.com/swisspost/design-system/blob/main/packages/internet-header/cypress/fixtures/internet-header/auth.json#L0-L1

### 💻 Proposed implementation

```html
<!-- logged out -->
<a href="/login">Login <post-icon name="arrow-right"/></a>

<!-- logged in -->
<post-menu-trigger for="menu--default">
  <button class="btn btn-tertiary">
    <post-avatar ...></post-avatar>
  </button>
</post-menu-trigger>
<post-menu id="menu--default" label="Menu description">
  <div slot="header">
    <post-avatar ...></post-avatar>
    <p>Name of the user </p>
  </div>
  <post-menu-item><a href="/first">First menu item</a></post-menu-item>
  <post-menu-item><a href="/second">Second menu item</a></post-menu-item>
  <post-menu-item><a href="/third">Third menu item</a></post-menu-item>
</post-menu>
```

### 📃 Tasks

- [ ] Add functional tests

alionazherdetska:
## 📄 Description

This PR introduces a new post-login-widget web component that connects to the KLP session API and renders either a login link or a user menu depending on the current session state. The logout button emits a postLogout event with the logoutUrl as payload, leaving redirect handling to the consumer.

### Notes:
1. Named `post-login-widget` to avoid conflicts with the existing `post-klp-login-widget` in @swisspost/internet-header. Will be renamed once the internet-header components are fully migrated.

⚠️ SonarQube
Sonar fails on the TODO comment in post-login-widget.tsx tracking the planned rename to post-klp-login-widget. Will be resolved as part after the completion of the `internet-header` package components migration.

2. I render the avatar and user menu UI directly in the LoggedIn example in the docs rather than mocking the fetch. When we mock the fetch in Storybook, the mock persists across both the Default and LoggedIn stories, causing the Default story to also show the logged-in state.

## 🚀 Demo

[KLP Login Widget](https://preview-7223--swisspost-design-system-next.netlify.app/?path=/docs/a1b2c3d4-e5f6-7890-abcd-ef1234567890--docs&devModeEnabled=true)

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/components/post-login-widget/post-login-widget.scss
```diff
@@ -0,0 +1,71 @@
...
+  }
+}
+
+post-avatar {
+  border: none;
+
+  &:hover {
+    --post-avatar-bg: #{tokens.get('avatar-hover-bg', components.$post-avatar)};
+  }
+}
```
oliverschuerch: Why is this necessary?

The post-avatar component has global styles defined (in the styles package), which handle its look-and-feel, when nested in an anchor or button tag.
Do they not work?

See:
- [avatar global styles](https://github.com/swisspost/design-system/blob/main/packages/styles/src/components/avatar.scss)
- [docs](https://preview-7223--swisspost-design-system-next.netlify.app/?path=/docs/09aac03d-220e-4885-8fb8-1cfa01add188--docs#button)

packages/components/src/components/post-login-widget/post-login-widget.scss
```diff
@@ -0,0 +1,71 @@
...
+
+tokens.$default-map: components.$post-button;
+
+:host {
+  display: contents;
+}
+
+.user-menu-wrapper {
+  display: contents;
+}
```
oliverschuerch: Can you add a comment in the SCSS file why this is necessary, please.

packages/components/src/components/post-login-widget/post-login-widget.scss
```diff
@@ -0,0 +1,71 @@
...
+
+  @include media.only(mobile) {
+    @include header-mx.nav-item-icon-only;
+    padding: 0.625rem;
+
+    span {
+      @include utilities.visuallyhidden;
+    }
+  }
+}
```
oliverschuerch: If I'm not wrong, you're recreating anchor styles already defined in the header component.
If we nest the component in the header, are these styles here still necessary?

packages/components/src/components/post-login-widget/post-login-widget.tsx
```diff
@@ -0,0 +1,201 @@
...
+  validateTextLogout() {
+    checkRequiredAndType(this, 'textLogout', 'string');
+  }
+
+  @Watch('textMenuLabel')
+  validateTextMenuLabel() {
+    checkRequiredAndType(this, 'textMenuLabel', 'string');
+  }
+
+  async componentDidLoad() {
```
oliverschuerch, line 107: Would it be possible to execute the fetch on `connectedCallback`?

packages/components/src/components/post-login-widget/post-login-widget.tsx
```diff
@@ -0,0 +1,201 @@
...
+    if (!IS_BROWSER) {
+      this.isLoading = false;
+      return;
+    }
```
oliverschuerch: `this.isLoading` is not used anywhere in the view. As long as you're not planning to render a loading icon or something, we don't need it at all.

Using the `IS_SERVER` constant, would be easier to read, imo.


```suggestion
    if (IS_SERVER) return;
```

packages/components/src/components/post-login-widget/post-login-widget.tsx
```diff
@@ -0,0 +1,201 @@
...
+    } catch (e) {
+      console.warn('post-login-widget: session fetch failed', e);
+      this.sessionData = null;
+    } finally {
+      this.isLoading = false;
+    }
+  }
+
+  private handleLogout = () => {
+    this.postLogout.emit(this.logoutUrl);
```
oliverschuerch, line 121: So the DEV using the component has to give us the logoutUrl, just so we can return it to him when the logout event is emitted?

As long as there is no default `logoutUrl` defined in the component, this seems to be complete nonsense.

packages/components/src/components/post-login-widget/post-login-widget.tsx
```diff
@@ -0,0 +1,201 @@
...
+      </a>
+    );
+  }
+
+  render() {
+    return (
+      <Host data-version={version}>
+        {!this.sessionData ? (
+          // Show login link while loading or if not logged in
+          this.renderLoginLink()
```
oliverschuerch: I like how you extracted the "unauthenticated" HTML structure into a different function. It makes it much easier to read the render function.

I would suggest to do the same with the "authenticated" HTML structure 😉

packages/components/src/components/post-login-widget/post-login-widget.tsx
```diff
@@ -0,0 +1,201 @@
...
+                  {this.textMessages}
+                </a>
+              </post-menu-item>
+
+              <post-menu-item>
+                <a href="https://www.post.ch/selfadmin/settings/">
+                  <post-icon aria-hidden="true" name="gear"></post-icon>
+                  {this.textSettings}
+                </a>
+              </post-menu-item>
```
oliverschuerch: I see this as a problem.
Except from the text, it is completely hardcoded.

No one else than us can add/remove/reorder/change any of these links, nor define custom destinations.
Even if this is set up and discussed with POW, I would prefer to have a more sustainable solution with slots or props to provide at least minimal flexibility.

packages/components/src/components/post-login-widget/post-login-widget.tsx
```diff
@@ -0,0 +1,201 @@
...
+// TODO: 'post-login-widget' is a temporary name to avoid conflict with 'post-klp-login-widget' 
+// which is part of @swisspost/internet-header. Rename to 'post-klp-login-widget' after 
+// @swisspost/internet-header components are migrated.
+@Component({
+  tag: 'post-login-widget',
+  styleUrl: 'post-login-widget.scss',
+  shadow: true,
+})
+export class PostLoginWidget {
+  private static readonly SUBSCRIBE_URL = 'https://n.account.post.ch/v1/session/subscribe';
```
oliverschuerch: Do we have a swagger URL or something, so we can check out the API?

**Personal opinion**
means, changing anything in your solution is optional

The URL is not used to login, nor to subscribe. Both of these actions would make it mandatory to send credentials or an auth-token to the server.

Instead, we use it to fetch the user session:

- If the server response contains a session, the user is authenticated (logged in).
- If the server response contains no session, the user is unauthenticated (logged out).

Because of this, I would rename the constant to `SESSION_URL`.

packages/documentation/src/stories/raw-components/login-widget/login-widget.docs.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+<ComponentsPackageInstall component="<post-login-widget>" />
+
+
+## Examples
+
+### Logged In
+
+When an active session is found, the widget renders a `<post-avatar>` button that opens a dropdown
+with the user's name, links to their profile, messages, settings, and a logout button.
```
oliverschuerch: I would be less specific about the provided links or we have to update the docs, everytime they change.

e.g.
"... opens a dropdown with the user's name and the users profile specific links and the logout button."

packages/components/src/components/post-login-widget/post-login-widget.scss
oliverschuerch, line 1: Generally this file creates a lot of cross dependencies to other components. It would be beneficial if we could avoid that.

If for example the styles for the `post-menu-item` change, we can only hope that our visual regression tests for this component will fail and inform us about the problem.

But chances are good, the VR-tests will not cover that and the login-widget menu-items will just keep their old look, while the rest of the menu-items in the header look different from this point on.

packages/documentation/src/stories/raw-components/login-widget/login-widget.stories.ts
```diff
@@ -0,0 +1,101 @@
...
+      text-messages=${args.textMessages}
+      text-settings=${args.textSettings}
+      text-logout=${args.textLogout}
+      text-menu-label=${args.textMenuLabel}
+    ></post-login-widget>
+  `,
+};
+
+export const LoggedIn: Story = {
+  render: () => renderUserMenu(),
```
oliverschuerch: So this is actually not rendering the `post-login-widget` component, due to problems mocking the fetch for one story, while keeping the other story showing as unauthenticated, right?

This means as well, we show "internal" shadow code of the component in the code panel. Even worse, it's not the actual shadow structure of the component it self, but a completely different story!

Depending on how much time you already spent to tackle the topic, I would say we should see if we can:

- encapsulate the story in an iframe
  (probably not working because of the dropdown menu)
- mock a backend endpoint for a specific story only
- find another solution so we can use the real component instead of a rebuild we have to maintain along the real one in the docs

oliverschuerch (changes requested):

packages/components/src/components/post-login-widget/post-login-widget.scss
alionazherdetska: External styles cannot penetrate the shadow DOM boundary. So even though the component is nested in the header, the styles applied to its parent or slot don't reach the anchor element inside the shadow DOM.

alionazherdetska:
Closing as a duplicate of https://github.com/swisspost/design-system/pull/7488

PR #7283, opened 2026-03-09
fix(component): visually separate active page state from open megadropdown state for `post-header` component

alionazherdetska:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [x] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


leagrdv:
I've removed myself from the reviewers since it's in draft, but feel free to add me back when it's ready!

packages/components/src/components/post-megadropdown-trigger/post-megadropdown-trigger.scss
```diff
@@ -34,7 +37,19 @@ button {
 
   @include media.only(desktop) {
     @include header-mx.active-state {
```
leagrdv, line 39: This seems to target and change both if expanded and with class `.active`, as well as not expanded with class `.active`, which is not correct.
Looking at cargo, it seems that when a megadropdown trigger has a child element that is the current page, it has the `.active` class added to it. Those changes would change all triggers to being with a light gray background

leagrdv (changes requested):

leagrdv (changes requested):
It's still not working as it should. If I set the `class="active"` on a megadropdown trigger that is not expanded, it's grey when it should be black. The grey should only be for when a megadropdown is expanded

leagrdv (approved):

PR #7287, opened 2026-03-09
feat(components): add autocomplete component

JoaoSequeira489:
## 📄 Description

- Added new post-autocomplete, post-listbox, and post-listbox-option components with styles, docs, and generated typings.
- Implemented autocomplete behavior (filtering, keyboard navigation, selection) and listbox option management, including support for detached listbox wiring.
- Added a changeset for @swisspost/design-system-components (major) announcing the new autocomplete component.
- *Design Tokens not yet created for this component so keeping styles without them for now*

## 🚀 Demo

https://github.com/user-attachments/assets/9f28b282-7cb9-4ba2-910a-a906f1863f87

---

## 🔮 Design review

- [X] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

- [X] Visual changes detected and approved _(Check this box if VRT fails and changes are intentional)_

## 📝 Checklist


packages/components/src/components/post-autocomplete/post-autocomplete.tsx
```diff
@@ -0,0 +1,222 @@
...
+  /** Number of characters to type before filtering methods are called */
+  @Prop() readonly filterThreshold: number = 0;
+
+  /** Show or hide the clear button */
+  @Prop() readonly clearable: boolean = false;
+
+  /** Optional idref to connect the autocomplete with the options dropdown if not nested */
+  @Prop() readonly options?: string;
+
+  @State() inputValue: string = '';
```
gfellerph, line 27: This should also be a getter so we can safely eliminate state de-sync issues.

```tsx
private get inputValue() {
  return this.inputElement.value;
}

File: packages/components/src/components/post-autocomplete/post-autocomplete.tsx [hunk 2]
```diff
@@ -0,0 +1,222 @@
...
+  private readonly handleOnBlur = () => {
+    if (this.inputElement) {
+      this.inputElement.style.cursor = 'pointer';
+    }
+    // Delay so mousedown on an option fires before the listbox closes
+    setTimeout(() => {
+      this.inputElement.value = this.inputValue;
+      this.listBoxElement.resetFilter();
+      this.hideListBox();
+    }, 150);
```
gfellerph: This looks like a workaround for a problem that's not immediately clear to me. What was the issue here?

packages/components/src/components/post-listbox-option/post-listbox-option.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+@Component({
+  tag: 'post-listbox-option',
+  styleUrl: 'post-listbox-option.scss',
+  shadow: true,
+})
+export class PostListboxOption {
+  private readonly id = crypto.randomUUID();
+  /** A value string, similar to <option value="val1">Value 1</option> */
+  @Prop({ reflect: true }) readonly value!: string;
```
gfellerph: Implement chrome datalist behaviour, that seems to be a reasonable fallback if both a value and a text are provided. In this case, the text acts like a description text to the value shown above.

Image

See https://codepen.io/tuelsch/pen/gbwgPEq for an example in Chrome or Safari.

packages/components/src/components/post-listbox-option/post-listbox-option.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  /** Represents option is selected . */
+  @Prop() readonly selected: boolean = false;
+
+  /** Represents option is highlighted . */
+  @Prop() readonly highlighted: boolean = false;
```
gfellerph: Should be excluded from documentation as it's only intended as an internal prop. Therefore also not reflected for SSR.

packages/components/src/components/post-listbox/post-listbox.tsx
```diff
@@ -0,0 +1,148 @@
...
+   *  Emitted option id for the active option
+   */
+  @Event() postOptionActive: EventEmitter<string | null>;
+
+  componentWillLoad() {
+    this.registerOptions();
+  }
+
+  private registerOptions() {
+    this.options = Array.from(this.host.querySelectorAll('post-listbox-option'));
```
gfellerph: This should be a getter.

packages/components/src/components/post-listbox/post-listbox.tsx
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  private clearActive() {
+    this.highlightedIndex = -1;
+    this.postOptionActive.emit(null);
+  }
+
+  /** Opens the listbox */
+  @Method()
+  async open() {
```
gfellerph:
```suggestion
  async show() {
```

packages/components/src/components/post-listbox/post-listbox.tsx
```diff
@@ -0,0 +1,148 @@
...
+    this.isOpen = true;
+  }
+
+  /** Closes the listbox */
+  @Method()
+  async close() {
```
gfellerph:
```suggestion
  async hide() {
```

packages/components/src/components/post-listbox/post-listbox.tsx
```diff
@@ -0,0 +1,148 @@
...
+    this.isOpen = false;
+  }
+
+  /** Uses the internal default filtering mode to filter the list of options.
+   * An empty string resets the filter to it's original state. */
+  @Method()
+  async filter(query: string) {
+    this.visibleOptions = this.options.filter(option => {
+      const label = (option.textContent || option.value || '').trim();
+      const isVisible = label.toLocaleLowerCase().includes(query.toLocaleLowerCase());
```
gfellerph: To improve the search heuristics, we should ignore diacritics as well:

```js
str.normalize("NFD").replace(/[\u0300-\u036f]/g, "")
```

See https://stackoverflow.com/questions/990904/remove-accents-diacritics-in-a-string-in-javascript/37511463#37511463.

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
```diff
@@ -0,0 +1,222 @@
...
+    if (this.listBoxElement && this.isDetachedListbox) {
+      this.listBoxElement.removeEventListener('postOptionSelected', this.onOptionSelected);
+      this.listBoxElement.removeEventListener('postOptionActive', this.onOptionActive);
+    }
+  }
+
+  private readonly handleOnBlur = () => {
+    if (this.inputElement) {
+      this.inputElement.style.cursor = 'pointer';
+    }
```
gfellerph: I've discussed this with Alizé and we were wondering why it's implemented like this. Since the basis of the component is the input and inputs always have the text cursor by default, we'd argue to keep the default in place.

A customisable select would have a button as basis and there the pointer would be more accurate.

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
```diff
@@ -0,0 +1,222 @@
...
+ */
+
+@Component({
+  tag: 'post-autocomplete',
+  styleUrl: 'post-autocomplete.scss',
+  shadow: true,
+})
+export class PostAutocomplete {
+  private inputElement: HTMLInputElement;
+  private listBoxElement: HTMLPostListboxElement;
```
gfellerph: For these it's good to use a getter function, since they could change anytime in DOM, especially if it's a detached listbox. The association could update or the listbox could be removed and re-attached and we'd loose the connection.

Use something like:

```tsx
private get inputElement() {
  return this.host.querySelector('input');
}
```

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
```diff
@@ -0,0 +1,222 @@
...
+  private isDetachedListbox: boolean = false;
```
gfellerph: Could we treat an inlined listbox like a detached one to get rid of some duplication?

```tsx
private get listBoxElement() {
  if (this.options) {
    this.listBoxElement = document.getElementById(this.options) as HTMLPostListboxElement;
  } else {
    this.listBoxElement = this.host.querySelector('post-listbox');
  }
}
```

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
```diff
@@ -0,0 +1,222 @@
...
+  @Element() host: HTMLPostAutocompleteElement;
+
+  /** Number of characters to type before filtering methods are called */
+  @Prop() readonly filterThreshold: number = 0;
+
+  /** Show or hide the clear button */
+  @Prop() readonly clearable: boolean = false;
+
+  /** Optional idref to connect the autocomplete with the options dropdown if not nested */
+  @Prop() readonly options?: string;
```
gfellerph: For SSR compatibility, we need to reflect the props, otherwise NextJS uses just the JS API to set them and then complains about hydration errors because the client does not know about the attribute.

```suggestion
  @Prop({ reflect: true }) readonly filterThreshold: number = 0;

  /** Show or hide the clear button */
  @Prop({ reflect: true }) readonly clearable: boolean = false;

  /** Optional idref to connect the autocomplete with the options dropdown if not nested */
  @Prop({ reflect: true }) readonly options?: string;
```

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
```diff
@@ -0,0 +1,222 @@
...
+    if (!query) {
+      this.inputValue = '';
+
+      // Hide the listbox if the input is empty, allow space key to open listbox
+      if (this.inputElement.value === '') {
+        this.hideListBox();
+      }
+    }
+    this.listBoxElement.resetFilter();
+  };
```
gfellerph, line 120: Not sure about this, space should not necessarily open the listbox, only down arrow. Can't really think of any use case, but if anyone needs a space in the value, then this would fail. I think this handler can be simplified for better predictability.

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
```diff
@@ -0,0 +1,222 @@
...
+      this.inputElement.style.cursor = 'text';
+    }
+  };
+
+  private readonly handleInput = (event: Event) => {
+    if (!this.listBoxElement) return;
+    const { value } = event.target as HTMLInputElement;
+    const query = value.trim();
+    if (query && query.length >= this.filterThreshold) {
+      this.listBoxElement.filter(query);
```
gfellerph: Filtering directly skips the mechanics for customising the filter. The autocomplete should implement a cancellable Event called `filterRequest`. If this event is being cancelled, the autocomplete should not call the filter function itself because the user is implementing a custom filter.

I think I'd implement a private function that handles this logic. See https://codepen.io/tuelsch/pen/myEaqbN for an example.

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
```diff
@@ -0,0 +1,222 @@
...
+      this.listBoxElement = this.host.querySelector('post-listbox');
+    }
+    if (!this.listBoxElement) return;
+
+    this.isDetachedListbox = !this.host.contains(this.listBoxElement);
+
+    if (!this.listBoxElement.id) this.listBoxElement.id = crypto.randomUUID();
+    this.inputElement.setAttribute('aria-controls', this.listBoxElement.id);
+    this.inputElement.setAttribute('aria-expanded', 'false');
+    this.inputElement.addEventListener('input', this.debouncedHandleInput);
```
gfellerph: Can we also add a click listener to open the listbox on click? Screen readers like voiceOver offer a keyboard shortcut to show suggestions and this triggers a click on the element with focus. See https://www.bekk.christmas/post/2023/06/how-to-detect-clicks-from-voiceover

packages/components/src/components/post-listbox-option/post-listbox-option.tsx
```diff
@@ -0,0 +1,38 @@
...
+  styleUrl: 'post-listbox-option.scss',
+  shadow: true,
+})
+export class PostListboxOption {
+  private readonly id = crypto.randomUUID();
+  /** A value string, similar to <option value="val1">Value 1</option> */
+  @Prop({ reflect: true }) readonly value!: string;
+
+  /** Represents option is selected . */
+  @Prop() readonly selected: boolean = false;
```
gfellerph:
```suggestion
  @Prop({ reflect: true }) readonly selected: boolean = false;
```

packages/components/src/components/post-listbox/post-listbox.tsx
```diff
@@ -0,0 +1,148 @@
...
+  /** Resets the filter to show all options */
+  @Method()
+  async resetFilter() {
+    this.visibleOptions = this.options;
+    this.options.forEach(option => {
+      option.hidden = false;
+      option.highlighted = false;
+    });
+    this.clearActive();
+  }
```
gfellerph: I'm wondering if the reset filter should just be a filter call with `""` as the query by convention. Then we don't need two methods and also the `filterRequest` event does not need a `resetFilterRequest` equivalent cancellable event. What do you think?

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
gfellerph: -> Race condition with onBlur event that resets the value to the previously selected value.

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
gfellerph: Watch out for the race condition when elements are initialised in random order.

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
gfellerph: Check event and prop option -> https://codepen.io/tuelsch/pen/myEaqbN?editors=1011

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
JoaoSequeira489: Unfortunately, we're not able to do this for this variable.
This value isn't `this.inputElement.value`, it's the "saved" value to track the selected option that the user has chosen so that onBlur we can default to that value.

E.g. User selects "Portugal" -> deletes 2 characters -> focuses another element -> Input defaults to "Portugal" instead of "Portug"

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
JoaoSequeira489: Found a workaround for this by canceling the onPointerDown event in the  `post-listbox-option` host

Have a look 😄

packages/components/src/components/post-listbox-option/post-listbox-option.tsx
JoaoSequeira489: Done

packages/components/src/components/post-listbox-option/post-listbox-option.tsx
JoaoSequeira489: Leaving it out prompts a plugin warning from stencil

`The @Prop decorator must be documented.eslint(@stencil-community/required-jsdoc)`

packages/components/src/components/post-listbox/post-listbox.tsx
JoaoSequeira489: Done

packages/components/src/components/post-listbox/post-listbox.tsx
JoaoSequeira489: Good point! Thanks for that 😄 

Added a helper function and applied it to query, textcontent and value before comparing

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
JoaoSequeira489: Yeah this was my bad I thought I saw it in the React Autocomplete that we based ourselves on!

Removed

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
JoaoSequeira489: Done

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
JoaoSequeira489: Refactored to use a getter and depend on addEventListener for both cases instead of using @Listen() for the nested case. This should make it more streamlined

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
JoaoSequeira489: Done

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
JoaoSequeira489: Yeah, this was in the specification but I also don't agree with it and it requires changing/deleting user input for it to work which is usually not a good approach.

Removed (specification should be updated though)

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
JoaoSequeira489: Done, used `defaultPrevented` for this

packages/components/src/components/post-listbox-option/post-listbox-option.tsx
JoaoSequeira489: Done

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
JoaoSequeira489: Done

packages/components/src/components/post-listbox/post-listbox.tsx
JoaoSequeira489: Made a small refactor to autocomplete.handleInput() and listbox.filter() for that, have a look 😄

PR #7316, opened 2026-03-11
chore(components): web component styles datepicker

myrta2302:
## 📄 Description

Implement Datepicker styles and publish the component.

## 🔮 Design review

- [x] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/documentation/src/stories/components/datepicker/date-picker.docs.mdx
alionazherdetska, line 1: When you clear the input, the calendar still shows the previously selected range highlighted. These two should probably be in sync:
 
Image

alionazherdetska, line 1: When I open the calendar menu using the Enter or Space key there are two focus rings: 1 on the element and one empty:

Image

sandra-post:
@myrta2302 Design review:

- Assistive/Hint text missing
- Position calendar popover:
  - IS: unclear
  image

  - SHOULD: Aligned to bottom right of input field (screenshot)
  image

- Border radius popover bottom left/right:
  - IS: 0px
  - SHOULD: 4px (post.datepicker.border-radius
 - Popover does not close after date is selected
 - Year view nav title color
   - IS: #4A4A4A
   - SHOULD: Enabled (#050400)
 - Year view first and last year color
   - IS: #4A4A4A
   - SHOULD: Enabled (#050400)

Rest looks good, thanks!

packages/documentation/src/stories/components/datepicker/date-picker.docs.mdx
alionazherdetska, line 1: The selected date is not reflected when using the keyboard navigation. When I press Enter on the date the selected state is reflected by color but it is not reflected in the input field itself and the popover is not getting closed (only if I press Enter again):

https://github.com/user-attachments/assets/282fda90-0b88-4d25-9aff-bb11269aec7d

packages/components/src/components/post-date-picker/post-date-picker.scss
```diff
@@ -104,209 +126,583 @@ tokens.$default-map: components.$post-breadcrumb;
...
 
   button {
-    position: absolute;
-    right: 16px;
-    top: 16px;
-    border: none;
-    background-color: transparent;
-  }
-}
+    @include form-input-mx.input-btn();
```
alionazherdetska, line 690: The focus ring on this button is currently barely visible. 
Image

As a temporary fix, could you please add `outline-offset: 4px`.

The proper fix will be addressed in the https://github.com/orgs/swisspost/projects/3/views/18?filterQuery=icon&pane=issue&itemId=164717998&issue=swisspost%7Cdesign-system%7C7323

packages/documentation/src/stories/components/datepicker/date-picker.docs.mdx
alionazherdetska, line 1: Should it be possible to enter an end date that is before the start date (e.g. `12.12.2020 - 12.12.1910`)? If not, we should probably prevent this or at least trigger a validation error.

Image

alionazherdetska (changes requested):

sandra-post:
@myrta2302 Design review done, all issues from last design review fixed, thanks!

packages/components/src/components/post-date-picker/post-date-picker.tsx
alionazherdetska, line 1: If the end date is set to a value earlier than the start date, the system  reassigns the end date as the new start date. Any end date set afterward is not reflected in the calendar range.

Image

**Steps to reproduce:**
1. Set a start date
2. Set an end date earlier than the start date → the end date becomes the new start date
3. Set a new end date → the range is not reflected in the calendar

.changeset/calm-sails-follow.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': minor
+'@swisspost/design-system-styles': patch
```
alionazherdetska: The `post-date-picker` docs page previously existed under "Raw Components" which isn't publicly accessible to users. Since this PR moves it to a proper public-facing page, shouldn't that still count as a minor?

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -187,9 +187,9 @@ export class PostDatepicker {
    * @param target - The HTML element relative to which the popover calendar should be displayed
    */
```
alionazherdetska: The @param target in the show() docstring is outdated: the parameter was removed. Please update the JSDoc to reflect the new signature.

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -209,6 +209,8 @@ export class PostDatepicker {
     }
   }
 
+  private flyoutOffset: number;
```
alionazherdetska: flyoutOffset is declared but never assigned, so it'll always be undefined.

packages/components/src/components/post-date-picker/post-date-picker.scss
alionazherdetska, line 1: I don't know if it is a valid comment but if I do not put any width constraints (e.g, padding to the body) then it is the behaviour of the datepicker in the components playground 

https://github.com/user-attachments/assets/e3883fd3-140a-4792-a857-bb494f817ba4

packages/components/src/components/post-date-picker/post-date-picker.scss
```diff
@@ -0,0 +1,717 @@
...
+              );
+              color: tokens.get('post-datepicker-grid-day-cell-element-color-hovered-fg');
+              background: tokens.get('post-datepicker-grid-day-cell-element-color-hovered-bg');
+
+              &.-disabled-,
+              &:hover {
+                &::after,
+                &::before {
+                  background: none;
+                  border: none;
```
alionazherdetska: The original nested selector used an invalid pattern where pseudo-elements (::after) appeared before pseudo-classes, causing CSS parsing issues.

**Original problem:**
```scss
&.-current-:hover::after {
  &.-disabled- { &::after { } }  // Invalid: Multiple pseudo-elements and wrong order
}

File: packages/components/src/components/post-date-picker/post-date-picker.scss [hunk 6]
```diff
@@ -0,0 +1,717 @@
...
+            }
+
+            &.-disabled-:not(.-other-month-) {
+              color: tokens.get(
+                'post-datepicker-grid-day-cell-element-color-disabled-fg'
+              ) !important;
+
+              &::after {
+                z-index: -1;
+                background: white;
```
alionazherdetska: Should we maybe tokenize it or there is the reason we should have it hardcoded?

packages/components/src/components/post-date-picker/post-date-picker.scss
```diff
@@ -0,0 +1,717 @@
...
+              &:hover {
+                color: tokens.get('post-datepicker-grid-day-cell-element-color-hovered-fg');
+
+                &:before {
+                  @include utilities.high-contrast-mode {
+                    forced-color-adjust: none;
+                  }
+                }
+
+                &:after {
```
alionazherdetska: These two blocks use CSS2 single-colon syntax (:before, :after) while the entire rest of the file uses the modern CSS3 double-colon syntax (::before, ::after). Maybe we should apply consistency.

packages/components/src/components/post-date-picker/post-date-picker.tsx
myrtasakellariou: Yes, you are right. I had left it in an in-between state. Now offset within popovercontainer is no longer assigned a value. And the actual 4px value is coming from the datepicker.

packages/components/src/components/post-date-picker/post-date-picker.scss
myrtasakellariou: Correct! i forgot to assign the token.

packages/components/src/components/post-date-picker/post-date-picker.scss
myrtasakellariou: Removed this code.

packages/components/src/components/post-date-picker/post-date-picker.scss
myrtasakellariou: This is actually also reproducible with a `<post-popover>` with a large button width and  `bottom-end`  placement, so it seems to be a general `post-popovercontainer`/floating ui setup issue. I will create a separate ticket.

packages/components/src/components/post-date-picker/post-date-picker.scss
```diff
@@ -0,0 +1,710 @@
...
+@use '@swisspost/design-system-styles/mixins/utilities';
+
+tokens.$default-map: components.$post-datepicker;
+
+/* Flyout Styles */
+
+.air-datepicker {
+  /* --------------------------------------------
+   * General
+   * ------------------------------------------ */
```
alionazherdetska: Just a suggestion:
It seems that the comment format is inconsistent with the rest of the codebase. We don't typically use this styled box-format for section comments. Maybe you should align with the format used in the rest of the components.

Feel free to ignore

packages/documentation/src/stories/components/datepicker/date-picker.snapshot.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+const { id, ...metaWithoutId } = meta;
+
+export default {
+  ...metaWithoutId,
+  title: 'Snapshots',
+};
+
+type Story = StoryObj;
+
+export const Datepicker: Story = {
```
alionazherdetska: Should Datepicker be updated to DatePicker to stay consistent with the rest of the codebase?

packages/styles/src/mixins/_form-input.scss
```diff
@@ -0,0 +1,37 @@
...
+  &:not([disabled]) :where(:hover > *) {
+    post-icon {
+      color: tokens.get('input-color-hover-fg', components.$post-text-input);
+
+      @include utilities.high-contrast-mode() {
+        color: Highlight;
+      }
+    }
+  }
+}
```
alionazherdetska, line 39: In the `form-search-input.scss` there was cursor: pointer on the button, but the new input-btn() mixin doesn't include it. The date picker button adds it back manually via &:not([disabled]) { cursor: pointer; }, but the search input no longer has it since it now uses the mixin. Was that intentional?

packages/documentation/src/stories/components/datepicker/date-picker.snapshot.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+  render: (_args: Args, context: StoryContext) => {
+    return schemes(
+      () => html`
+        <div class="row">
+          <h1>Datepicker</h1>
```
alionazherdetska: Adjust naming for consistency

alionazherdetska (changes requested):

packages/styles/src/mixins/_form-input.scss
myrtasakellariou: This was not intentional. I ve restored it inside the mixin.

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -873,33 +885,49 @@ export class PostDatepicker {
...
+  }
+
+  private isValidDate(date: Date): boolean {
+    return date instanceof Date && !isNaN(date.getTime());
+  }
+
+  private syncDatePickerState() {
     if (this.dpInput.disabled) {
       this.inputDisabled = true;
     }
```
oliverschuerch: What if `this.dpInput.disabled` becomes `false`? Is `this.inputDisabled` then still up-to-date?

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -209,6 +208,8 @@ export class PostDatepicker {
...
   }
 
+  private flyoutOffset: number = 4;
```
oliverschuerch: This appears to be a constant that does not change throughout the component's lifetime. Therefore, I consider it best practice to add the `static readonly` keywords to it and change the naming convention to uppercase snake case.

```suggestion
  private static readonly FLYOUT_OFFSET: number = 4;
```

packages/components/src/components/post-date-picker/post-date-picker.tsx
oliverschuerch, line 1: Before I dive deeper I would like to present you the following finding:

The component seems to differentiate between two date formats:

- `euFormat` => `DD.MM.YYYY`
- `usFormat` => `MM/DD/YYYY`

It also provides several functions to convert a `euFormatDateString|usFormatDateSTring` to a date and vice verca.

Now the problem that I see with that is, that there are way more than just two date formats in the world.

---

The following script can be used to log the different locale formats.

<details>
  <summary>Script</summary>
  
  **Copy & Paste the following script to your browsers console and hit `enter`**

  ```js
  // edit the input locales as you like
  const INPUT_LOCALES = ['de', 'de-DE', 'de-CH', 'fr', 'fr-FR', 'fr-CH', 'it', 'it-IT', 'it-CH', 'en', 'en-US', 'en-GB', 'nl-NL', 'ja-JP', 'zh', 'zh-CN'];
  
  (function logDateFormats() {
    const date = new Date();
    const options = {
      day: '2-digit',
      month: '2-digit',
      year: 'numeric'
    };
    
    console.clear();
  
    const localeFormats = [];
    
    for (const locale of INPUT_LOCALES) {
      localeFormats.push({
        locale,
        region: new Intl.DateTimeFormat(locale, options).format(date)
      });
    }
  
    console.table(localeFormats);
  })();
  
  // https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat
  ```
</details>

**Expected output**

Image

As you can see, a lot of date formats differ, even for EU countries.

EU formats

- `de` ≠ `fr`
- `fr-FR` ≠ `fr-CH`
- `nl-NL` ≠ `de-DE` | `fr-FR`
- ...

None EU formats
- `en-GB` ≠ `en-US`
- `de-DE` ≠ `ja-JP`
- `ja-JP` ≠ `en-US`
- ...

---

**Conslusion**

Our Date Picker may be used on a website with the following locals only:

- `de`
- `fr` or `fr-CH`
- `it` or `it-CH`
- `en` or `en-GB`

This means, we have to handle at least 3 different formats:

- `DD.MM.YYYY`: `de`, `de-DE`, `de-CH`, `fr-CH`, `it-CH`
- `DD/MM/YYYY`: `fr`, `fr-FR`, `it`, `it-IT`, `en-GB`
- `MM/DD/YYYY`: `en`, `en-US`

**Recommendation**

I strongly recommend using the ISO date format internally (international standard, e.g. `YYYY-MM-DD`) and standard functions like [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) or [date.toLocaleDateString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) to format it before sending it to the UI, rather than implementing custom functions to convert it.
Or this may become a major problem, at least when we need to add further locales (e.g., `nl-NL` or `ja-JA`) with completely different formats or when the component may be used in different countries like France, GB, etc. where it could confuse users, which are used to different date formats.

oliverschuerch: Please let's discuss this tomorrow in the daily or in the next DEV RT!

oliverschuerch: There are more event listener registered on the `body`, `this.prevBtn`, `this.nextBtn` and `this.dpInput`, which must be removed as well.

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -184,12 +184,11 @@ export class PostDatepicker {
...
-   * @param target - The HTML element relative to which the popover calendar should be displayed
    */
   @Method()
-  async show(target: HTMLElement) {
+  async show() {
     if (this.popoverRef) {
-      await this.popoverRef.show(target);
+      await this.popoverRef.show(this.dpInput);
       this.enhanceAccessibility();
       this.host.shadowRoot.addEventListener('keydown', this.handleTab, true);
```
oliverschuerch, line 207: Everytime the `show` method runs a new event listender is registered on the `shadowRoot` element, adding more and more listener, without cleaning them up.

If we would register the listener on the `host`, we would probably not need to re-register it over and over again.
But if this is not working, we (at least) need to remove the potentially existing event listener, before we add the new one.

packages/components/src/components/post-date-picker/post-date-picker.tsx
oliverschuerch: The event gets registered with the third parameter set to `true`. As far as I know, this must be reflected in the `removeEventListener` call as well.

oliverschuerch: I've created a new ticket to implement the discussed logic: #7367

packages/components/src/components/post-date-picker/post-date-picker.scss
oliverschuerch, line 1: Please try to fix the following:

- [ ] When picking the start date for a range, the yellow background immediately starts showing up.
   We should try to reset this if:
   <br>
   - There is no end date set already AND
   - the start date is currently hovered.
   <br>
   Image

- [ ] When picking a date range, just after setting the start date and hovering the next day on the same table row, the gray circle on the end date seems to be cut off on the left.
   
   Image

packages/components/src/utils/is-iso-date.ts
oliverschuerch, line 1: First of all, the utility test seems works as expected.
But, because of the wrong reason!
And this could become a problem when we use the given ISO string somewhere in the code to create a new JS Date object.

```js
isIsoDate('2026-02-28')
// y = 2026, m = 2, d = 28
// date.toISOString() = '2026-02-28T00:00:00.000Z'
// Correctly retuns true!

isIsoDate('2026-02-35')
// y = 2026, m = 2, d = 35
// date.toISOString() = '2026-03-06T00:00:00.000Z'
// Correctly returns false, but...
// While new Date('2026-02-35') returns an invalid Date object,
// new Date(2026, 2, 35) returns a valid Date object, resolved to ISO '2026-03-06'.
// This is also the case, in the following examples:
// - isIsoDate('0000-01-00')
// - isIsoDate('2026-01-00')
// - isIsoDate('2026-15-30')
// - isIsoDate('2026-00-15')
// - ...
```

Instead of the regex test, the number extraction and the value-by-value comparison, we could check if the given ISO string creates a valid Date object and then compare the ISO string, we get from the Date object with the given one:

```js
function isIsoDate(value: string): boolean {
  try {
    // fails if `value` is an invalid string to create a JS Date object
    const date = new Date(value);
    // fails if `value` is not ISO date formatted (e.g. `yyyy-mm-dd`)
    // for example if `value` is set to `March 15, 2026 00:00:00`!
    return date.toISOString().split('T')[0] === value;
  } catch {
    return false;
  } 
}
```

packages/components/src/components/post-date-picker/post-date-picker.tsx
oliverschuerch: Since the native `input[type="date"]` element returns its value always as an ISO Date string (e.g. `yyyy-mm-dd`), our component should do the same!

I have no idea if we can reflect that with typescript, but we should at least emit the values in ISO format.
e.g.

`'yyyy-mm-dd'` or `['yyyy-mm-dd', 'yyyy-mm-dd']`

Test the native elements [here](https://codepen.io/oliverschuerch/pen/myrwvWp).

packages/components/src/components/post-date-picker/post-date-picker.tsx
oliverschuerch: Any update here?

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -734,6 +768,15 @@ export class PostDatepicker {
...
   }
 
+  private isoToDate(iso: string): Date {
+    const [y, m, d] = iso.split('-').map(Number);
+    return new Date(y, m - 1, d);
+  }
+
+  private dateToIso(date: Date): string {
+    return `${date.getFullYear()}-${String(date.getMonth() + 1).padStart(2, '0')}-${String(date.getDate()).padStart(2, '0')}`;
+  }
```
oliverschuerch: Please use the native JS functions for converting these objects:

e.g.

```js
function isoDateToDate(isoDate: TDateISODate): Date {
  return new Date(isoDate);
}

function dateToIsoDate(date: Date): TDateISODate {
  return date.toISOString().split('T')[0];
}
```

To get an idea about the mentioned `TDateISODate` type, please visit https://gist.github.com/MrChocolatine/367fb2a35d02f6175cc8ccb3d3a20054

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -628,18 +637,29 @@ export class PostDatepicker {
...
+          // update props
+          if (Array.isArray(date) && date.length === 2) {
+            this.selectedStartDate = this.dateToIso(date[0]);
+            this.selectedEndDate = this.dateToIso(date[1]);
+          } else if (!Array.isArray(date)) {
+            this.selectedStartDate = this.dateToIso(date);
+            this.selectedEndDate = undefined;
+          }
+
+          this.postUpdateDates.emit(date);
```
oliverschuerch: As mentioned above, this should return ISO dates, so it works the same as the native `input[type="date"]` element!

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -701,18 +721,32 @@ export class PostDatepicker {
...
+  private warnIfNotIso(prop: 'selectedStartDate' | 'selectedEndDate' | 'min' | 'max') {
+    const value = this[prop];
+    if (!value) return;
+
+    if (!isIsoDate(value)) {
+      console.error(
+        `The prop \`${prop}\` of the \`${this.host.localName}\` component must be in ISO format (YYYY-MM-DD).`,
+      );
+    }
+  }
```
oliverschuerch: Any chance we can extract that from here and instead add as a utility?

PR #7321, opened 2026-03-11
fix(component): fix the `post-back-to-top` button being overlaid by the `post-stepper`

alionazherdetska:
## 📄 Description

The post-back-to-top button was being overlaid by the post-stepper component. Added a z-index value of **$zindex-header - 1** so it renders above page content without overlapping the header or modals.
Screenshot 2026-03-11 162535

## 🚀 Demo

https://preview-7321--swisspost-design-system-next.netlify.app/iframe.html?full=true&id=[1a1b4cab-d0a8-4b01-bd85-b70e18668cb5--default](https://preview-7321--swisspost-design-system-next.netlify.app/iframe.html?full=true&id=1a1b4cab-d0a8-4b01-bd85-b70e18668cb5--default&viewMode=story)&viewMode=story
---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/components/post-back-to-top/post-back-to-top.scss
```diff
@@ -15,6 +16,7 @@ tokens.$default-map: components.$post-floating-button;
   );
   --post-back-to-top-elevation: #{tokens.get('floating-button-elevation')};
   position: fixed;
+  z-index: commons.$zindex-header - 1;
```
myrtasakellariou, line 19: Not sure, but would it maybe be better to have a dedicated z-index variable for the back-to-top button, since it should always be the only element in that layer? (e.g. `commons.$zindex-backtotop`) or is it too much? In the latter case just ignore my comment :)

alionazherdetska: Since this value is only used once and in a single file, maybe we should keep it as a calculated value (`$zindex-header - 1`) is fine for now.

PR #7328, opened 2026-03-13
feat(components): enable ssr for post-accordion

alionazherdetska:
## 📄 Description

This PR adds a CSS-only fallback for the accordion so that when JavaScript is disabled, the first item is open and the rest are collapsed by default what is matching the JS behaviour.

Note: if the user explicitly sets collapsed on all items, all items will be collapsed without JS too. This is intentional and matches real use cases like Cargo where all items start closed.

The only edge case that doesn't work without JS is when the user sets collapsed on some items but not others, and expects the accordion to automatically close the non-collapsed items that come after the first open one. For example: item 1 no collapsed, item 2 no collapsed, item 3 collapsed: with JS, items 2 and 3 would both end up closed because registerAccordionItems() finds item 1 as the first open one and closes everything else. Without JS, item 2 stays open because CSS has no way to replicate that "find the first open item and close the rest" logic:
image

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/styles/src/components/accordion/index.scss
```diff
@@ -0,0 +1 @@
+@use 'post-accordion';
```
leagrdv: Because this is done outside of the `post-accordion` component, it should be documented in the accordion doc on storybook that user needs to import the styles for this component. Same thing for the collapsible I think.

leagrdv: Also, I'm wondering if because users now need to import styles from the styles package, is it not a breaking change?

alionazherdetska: I will raise it during the Daily on Monday then

.changeset/beige-snakes-lose.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': minor
+---
+
+Updated `post-accoridon` component to support Server-Side Rendering.
```
leagrdv:
```suggestion
Updated `post-accordion` component to support Server-Side Rendering.
```

PR #7353, opened 2026-03-16
fix(styles-primeng): correct css paths on primeng showcase watch

leagrdv:
## 📄 Description

Changed the scss imports of the `styles-primeng` on the `styles-primeng-showcase` to use the source file rather than the `dist` output to avoid timing issues.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [X] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist

### DIFF
--- a/.changeset/angry-pillows-warn.md
+++ b/.changeset/angry-pillows-warn.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': minor
+---
+
+Documented the usage of the `@swisspost/design-system-components-angular` package with standalone components.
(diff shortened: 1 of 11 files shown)


oliverschuerch (changes requested):
It feels wrong, to import source files from a different module.
Instead we should fix the timing issue as mentioned in the ticket.

Can you explain, why the approach described by @alizedebray (using the `concurrently` module instead of `npm-run-all`) did not work?

leagrdv:
@oliverschuerch It works with just this change indeed, thanks!

PR #7393, opened 2026-03-23
chore(styles): deprecate `form-select-empty`

leagrdv:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [X] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist

### DIFF
--- a/.changeset/linklist-background-fix.md
+++ b/.changeset/linklist-background-fix.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Removed background color from `.list-interactive`.
--- a/packages/components/src/index.html
+++ b/packages/components/src/index.html
@@ -164,18 +164,5 @@
       <option value="compact">Post compact</option>
     </select>
     <hr />
-    <post-stepper
-      text-completed-step="Completed step"
-      text-current-step="Current step"
-      text-step-number="Step #number:"
-      current-index="3"
-      selected-index="2"
-    >
-      <post-stepper-item> Step 1 label</post-stepper-item>
-      <post-stepper-item> Step 2 label</post-stepper-item>
-      <post-stepper-item> Step 3 label</post-stepper-item>
-      <post-stepper-item> Step 4 label</post-stepper-item>
-      <post-stepper-item> Step 5 label</post-stepper-item>
-    </post-stepper>
   </body>
 </html>
--- a/packages/styles/src/components/list-interactive.scss
+++ b/packages/styles/src/components/list-interactive.scss
@@ -9,7 +9,6 @@ tokens.$default-map: components.$post-list;
 
 // Base class
 .list-interactive {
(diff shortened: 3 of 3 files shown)


packages/documentation/src/stories/components/select/select.docs.mdx
alionazherdetska, line 1: Just my personal feeling: before, the default example showed a visible placeholder (e.g. “Choose an option”), which feels natural for many users since selects usually display an initial value. 

Image

Now the default is an empty option, which makes the label act as the placeholder.

Image

It might be worth keeping the default example with a visible placeholder, and then documenting the empty option pattern as the improved/modern approach for floating labels.

That way:

- users who expect a filled first option still feel at home
- and those who want the improved behavior can opt into it via the empty option

Again, just my feeling, since you have much more experience feel free to ignore.

.changeset/petite-beans-sell.md
```diff
@@ -0,0 +1,6 @@
+---
+'@swisspost/design-system-documentation': patch
+'@swisspost/design-system-styles': patch
+---
+
+Deprecated the `.form-select-empty` class and replaced it with a modern CSS selector.
```
alionazherdetska:
```suggestion
Deprecated `.form-select-empty` CSS class. Floating label behavior is now automatic.

BEFORE:
<select class="form-select form-select-empty">
  <option></option>
  <option value="1">Option 1</option>
</select>

AFTER:

<select class="form-select">
  <option></option>
  <option value="1">Option 1</option>
</select>

File: packages/styles/src/components/form-select.scss [hunk 2]
```diff
@@ -240,7 +240,9 @@ tokens.$default-map: components.$post-select;
       background-color: tokens.get('select-color-disabled-bg');
     }
 
-    &:empty,
+    &:has(option[value='']:checked:empty),
+    &:has(option:not([value]):checked:empty),
```
alionazherdetska, line 244: I’m not fully sure whether the ticket intends to support placeholder text inside the empty option, since the main goal seems to be replacing the workaround for selects using an empty option.

However, with the current selector we may narrow the previously supported behavior. Teams using `.form-select-empty` together with:
```html
<option value="">Choose an option</option>
```

would lose the expected behavior after running the automigration, since the selector no longer matches (due to :empty). The migration will remove the class, but won’t adapt the markup.

packages/documentation/src/stories/components/select/select.docs.mdx
leagrdv: I've updated it: main example shows a "Choose an option" while the `Floating label` example shows an empty option

packages/styles/src/components/form-select.scss
leagrdv: Good catch, I've removed the automigration and updated the line in the migration guide. That way, users can update it manually to ensure they have an empty option.

PR #7404, opened 2026-03-25
chore(docs): improve migration guide code snippets

leagrdv:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [X] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/documentation/src/stories/misc/migration-guide/code-block/code-block.component.ts
```diff
@@ -0,0 +1,43 @@
...
+    try {
+      await navigator.clipboard.writeText(this.code);
+      this.copied = true;
+
+      clearTimeout(this.timeoutId);
+      this.timeoutId = window.setTimeout(() => {
+        this.copied = false;
+      }, 2000);
+    } catch (err) {
+      console.error('Copy failed', err);
```
oliverschuerch: Can we make this a warning and add a little context?
I personally often struggle with log messages, I don't know where they're coming from.

e.g.

```suggestion
      console.warn(\`Code Block: copying "${this.code}"\ failed!\n`, err);
```

packages/documentation/src/stories/misc/migration-guide/code-block/code-block.component.scss
oliverschuerch, line 1: The scrollable area is set differently than in the `<Source>` block of storybook.
I'd prefer a visually accurate reproduction of the Storybook component.

| Storybook Source Block | Custom Code Block |
| :- | :- |
| Image | Image |

oliverschuerch (changes requested):
Is there a good reason why we can't use the official `<Source>` block from Storybook on the migration guide pages?

The custom component you just created does not offer code highlighting.
I know it's not needed for bash scripts, as they have no highlighting anyway...

https://storybook.js.org/docs/api/doc-blocks/doc-block-source

---

If the usage of the Storybook Source block is possible, I'd prefer that as a solution.

leagrdv:
Usage of the `<Source>` is not possible within the migration as the migration is built using Lit elements and not React component.
I could change the whole migration guide to React Components but it seemed to be overkill (9 different components) just to add a scrollbar and a copy button to some bash lines, hence the small custom `code-block`.
Regarding the scrollbar, the `<Source>` is using a custom scrollbar, I went with the native one to avoid adding a scrollbar lib.
If despite this, you'd still rather I switch the migration to React components, I can do it, it will just take a bit more time than originally planned.

oliverschuerch:
> Usage of the `<Source>` is not possible within the migration as the migration is built using Lit elements and not React component. I could change the whole migration guide to React Components but it seemed to be overkill (9 different components) just to add a scrollbar and a copy button to some bash lines, hence the small custom `code-block`. Regarding the scrollbar, the `<Source>` is using a custom scrollbar, I went with the native one to avoid adding a scrollbar lib. If despite this, you'd still rather I switch the migration to React components, I can do it, it will just take a bit more time than originally planned.

Thanks for the hint, I missed the fact the components are made with lit.
I would say we should convert it to react in the future, but surly not within the [v10 release](https://github.com/swisspost/design-system/milestone/47) milestone.

Can you create a ticket for that and add a https://github.com/swisspost/design-system/labels/v11 label?

Using the native scrollbar is the right way to go here.
However, please fix/resolve the previously added [comment](https://github.com/swisspost/design-system/pull/7404#discussion_r2999617133).

leagrdv:
I've styled the scrollbar and created ticket #7425

PR #7416, opened 2026-03-27
fix(styles): fix CLS for one-pager header without title

alionazherdetska:
## 📄 Description

I was able to reproduce the bug by throttling the network to 3G/Slow 4G in DevTools: the cumulative layout shift was clearly visible as the page loaded. The fix resolves it.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


.changeset/every-roses-push.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
```
leagrdv, line 2: Technically the patch is in the styles but it's affecting mostly the components package, do you think we should add it?
```suggestion
'@swisspost/design-system-styles': patch
'@swisspost/design-system-components': patch
```

PR #7419, opened 2026-03-27
chore(components, components-angular, nextjs-integration): automate syncing of component markup across integrations

### LINKED ISSUE #6603: Automate Syncing of Component Markup Across Integrations
### Problem Statement

Stencil components often receive new attributes, slots, and other updates. These changes are **not consistently reflected** in the component wrappers used in our **Next.js** and **Angular** integration packages.  
This results in outdated markup, missing props, and inconsistencies across frameworks.

### Goal

Identify and implement a method to **automatically update component markup** in the Next.js and Angular integrations whenever Stencil components change.

### Areas to Explore

Investigate potential sources of truth that could drive automated synchronization:

- **Storybook default stories**  
  Use stories as canonical component usage definitions.
- **Stencil-generated files**  
  Review available metadata (e.g., docs JSON, generated typings, output targets) that could be used for automation.

### Proposed Solution (Initial Concept)

Explore adopting a unified approach using **[Mitosis](https://mitosis.builder.io/)**:

- Maintain a private Mitosis library where component examples are written once in JSX.
- Use Mitosis to generate component examples in multiple formats (React/Next.js, Angular, etc.).
- Integration packages can then import or consume these generated examples to stay automatically in sync.

myrta2302:
## 📄 Description

This PR aims to **extract the story markup, transform it, and dynamically reuse it in the Angular and Next.js** consumer packages.

**_Since the initial approach involving Mitosis proved to be unsuitable for the scope of this ticket, an alternative solution was implemented._**

The solution extends the Cypress configuration in the components package by leveraging story rendering during E2E runs to programmatically extract markup into JSON. **This requires adding a minimal, standardized test per component.**

The added scripts (under components/scripts) **run sequentially as the last step of the E2E tests**, producing the necessary files for the Next.js and Angular consumer applications.

##  Demo
Run `pnpm components:e2e` to produce the markup and write the consumer app pages.

## Important notes
------------------------
**Not all elements showcased in the consumer apps could be mapped to a component story page.**

1. This applies to the **React Server Icons** in the Next.js package, which are therefore temporarily hardcoded in the generated templates.
2. The Angular app previously used **dynamic @for loops** in the content. With the new approach, components should be derived from the stories. For the moment only the `<header> ...</header>` part remains hardcoded.

Also, the run script in the package.json is temporary, because there should be a better approcach reagarding the order of running scripts. It is temporarily set to run during the `pnpm components:e2e`.

## How to test

1. `pnpm components:e2e`
2. `pnpm nextjs:start ` / `pnpm components-angular:start`

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


myrta2302:
automate-component-markup

alionazherdetska:
I have tried adding the `required` prop for the `<post-accordion>`, updated the default story and pushed to CI without running `pnpm e2e` tests since the developers sometimes might not need to add the extra test just guarding it with the  **required propery checker** which would anyway throw console error:
Screenshot 2026-04-28 140345

This is what happened in CI:
1. E2E tests failing 
image
image

2. Failing linting:
image
3. Failing Sonar:
image

Thus, it seems that having the setup which will run on `pnpm e2e` might even slow us down since the developer might need to handle all of these first if someone forgot to run e2e tests locally which actually also take time to finish running.

I checked locally it took **9 mins 52 seconds** to accomplish the run and get all of the files generated.

alionazherdetska:
After I ran `e2e` locally two `donwloads.htm` files got generated:
image
It seems that the user might mistakenly commit them with the bunch of the other autogenerated files. I am not sure it has to be commited (correct me If I am wrong).

packages/components/cypress/e2e/accordion.cy.ts
alionazherdetska, line 1: The Extract markup describe blocks are pretty copy-pasty at this point: there are 20+ of them and they're all basically the same shape. Would it be worth pulling these into a shared config somewhere so adding a new component doesn't mean writing another 10-line block? Happy to be wrong if there's a reason to keep them separate.

packages/components/cypress/e2e/popover.cy.ts
```diff
@@ -21,6 +39,10 @@ describe('popover', { baseUrl: null, includeShadowDom: true }, () => {
       cy.get('post-popover-trigger[data-hydrated][data-cy-id="popover-two"]').as('popoverTrigger2');
     });
 
+    it('should write component markup to file', () => {
+      cy.writeMarkup('post-popover');
+    });
```
alionazherdetska: Looks like there's a leftover test  since that just calls cy.writeMarkup('post-popover') with no setup. Pretty sure this got replaced by the new **Extract markup block** at the top and was never cleaned up. Worth double-checking and removing if so.

alionazherdetska (changes requested):

myrtasakellariou:
> I have tried adding the `required` prop for the `<post-accordion>`, updated the default story and pushed to CI without running `pnpm e2e` tests since the developers sometimes might not need to add the extra test just guarding it with the **required propery checker** which would anyway throw console error: Screenshot 2026-04-28 140345
> 
> This is what happened in CI:
> 
> 1. E2E tests failing
> 
> image image
> 2. Failing linting:
> 
> image 3. Failing Sonar: image
> Thus, it seems that having the setup which will run on `pnpm e2e` might even slow us down since the developer might need to handle all of these first if someone forgot to run e2e tests locally which actually also take time to finish running.
> 
> I checked locally it took **9 mins 52 seconds** to accomplish the run and get all of the files generated.


> I have tried adding the `required` prop for the `<post-accordion>`, updated the default story and pushed to CI without running `pnpm e2e` tests since the developers sometimes might not need to add the extra test just guarding it with the **required propery checker** which would anyway throw console error: Screenshot 2026-04-28 140345
> 
> This is what happened in CI:
> 
> 1. E2E tests failing
> 
> image image
> 2. Failing linting:
> 
> image 3. Failing Sonar: image
> Thus, it seems that having the setup which will run on `pnpm e2e` might even slow us down since the developer might need to handle all of these first if someone forgot to run e2e tests locally which actually also take time to finish running.
> 
> I checked locally it took **9 mins 52 seconds** to accomplish the run and get all of the files generated.

To address this issue the individual tests have been removed and there is now one e2e test that writes the markup for all components and runs on postbuild. This probably adds (in my case) about +20 secs to the building time.

packages/components/cypress/e2e/accordion.cy.ts
myrta2302: This has been resolved by merging the tests in a single file.

myrtasakellariou:
> After I ran `e2e` locally two `donwloads.htm` files got generated: image It seems that the user might mistakenly commit them with the bunch of the other autogenerated files. I am not sure it has to be commited (correct me If I am wrong).

This is created by cypress. It is now cleaned up and gitignored.

PR #7420, opened 2026-03-27
feat(components): add global styles to the components global styles entry

### LINKED ISSUE #7392: Add global styles to the components global styles entry
Some of our web-components require additional global styles (not scoped in the shadowRoot of the component itself), to ensure that slotted light-DOM elements are rendered correctly.

Adding these global styles to the project's CSS bundle is mandatory as soon as a developer uses the respective web component.
Since it would be very impractical to make these imports mandatory only when needed, we consider it more sensible (at least for the moment) if these styles were implemented in an output file that must be part of the CSS bundle anyway.

**Current Setup**

- Web-components with mandatory global styles ar living in the Components package.
- Global styles for the aforementioned components are living in the Styles package.

If a web-component has mandatory global styles (which is not documented anywhere at the moment), the user has to install the Styles package and import either one of our bundle output files or the specific global styles SCSS files.

**Future Setup**

- Web-components with mandatory global styles ar living in the Components package.
- Global styles for the aforementioned components live in the Components package as well. More specifically in the global styles of the Components package CSS bundle.

The requirement of, and how to import the global styles from the Components package, is already documented.
Since there is no way to import only one or the other file from these global styles, we must ensure that these styles stay as small as possible.

For now, we could move over or add new global styles for the following components:

- Header (already existing [styles/src/components/header](https://github.com/swisspost/design-system/tree/main/packages/styles/src/components/header))
- Footer (already existing [styles/src/components/footer](https://github.com/swisspost/design-system/tree/main/packages/styles/src/components/footer))
- Tabs (already existing [styles/src/components/tabs](https://github.com/swisspost/design-system/tree/main/packages/styles/src/components/tabs))

- [ ] Move the existing styles files from the Styles to the Components package and add them to the [global styles bundle](https://github.com/swisspost/design-system/blob/main/packages/components/src/styles/index.scss), by importing them in the [components file](https://github.com/swisspost/design-system/blob/main/packages/components/src/styles/components/index.scss) file.
- [ ] These changes will cause breaking changes in the Styles package, and minor updates in the Components package. Create the changesets accordingly and as [PO](@gfellerph) for permission to add them to v10.

alionazherdetska:
## 📄 Description

This PR moves the global styles for `post-header`, `post-footer`, and `post-tabs` out of `@swisspost/design-system-styles` and into `@swisspost/design-system-components`.
 
## The unexpected detour: a Sass `:is()` bug
 
This should have been a straightforward move, but porting the header SCSS into the components package immediately broke the Next.js build:
 
```
CssSyntaxError: post-components.css:4:10555: Unclosed bracket
```
 
Turned out Stencil's Sass compiler generates broken `:is(:focus-visible, ...)` output when these mixins are called on comma-separated selectors or inside `@include interactive-elements`. The styles package never hit this because its gulp build compiles each file independently with simple selectors. Stencil's compilation context is more complex and exposed the problem.
 
To fix it, `nav-item-block-layout` was split into two separate mixins: one for layout (safe to use anywhere) and one for focus styles (`nav-item-block-focus`, which must only ever be called on a single simple selector). All the call sites that used comma lists or `@include interactive-elements` were rewritten as explicit individual selectors.

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


myrta2302 (approved):
The calc() inside max() may not be necessary. (https://stackoverflow.com/questions/16617248/calc-of-max-or-max-of-calc-in-css)

alizedebray (commented):
Are we sure we want to keep all the empty files?

packages/components/src/components/post-tab-item/post-tab-item.scss
```diff
@@ -1,4 +1,4 @@
-@use '@swisspost/design-system-styles/components/tabs/tab-title';
+@use '../../styles/components/post-tabs/tab-title.scss';
```
alizedebray: Can you remove the `.scss`  on all files please?

apps/documentation/src/stories/components/header/header.stories.ts
```diff
@@ -179,10 +179,12 @@ function getHeaderRenderer(
...
-          <a href="" slot="post-login">
-            <span>Login</span>
-            <post-icon name="login"></post-icon>
-          </a>
+          <div slot="post-login">
+            <a href="">
+              <span>Login</span>
+              <post-icon name="login"></post-icon>
+            </a>
+          </div>
```
alizedebray: Can you revert these changes? They lead to incorrect styling: 
Screenshot 2026-05-04 at 17 03 49

alizedebray (changes requested):

apps/documentation/.storybook/addons/styles-switcher/StylesSwitcher.tsx
alionazherdetska, line 1: This file needed refactor because the theme stylesheet was being appended to the end of <head>, which meant it loaded after post-components.css, as you cans ee on the screenshots:

Image
Image

Since global header styles now live in post-components.css, the base `a {}` rule from the theme was overriding them. Fixed by inserting the theme stylesheet right after post-default.css instead, so post-components.css always wins.

packages/components/src/index.html
alionazherdetska, line 1: `post-components.css` needs to load after post-default.css so component styles (now bundled in post-components.css) aren't overridden by base rules like `a {}`.

alizedebray (approved):

alizedebray:
> [!CAUTION]
> I've approved the PR because the implementation works. However, after discussing it again with the POW Team, I want to re-raise the concern that not being able to choose which CSS to include will likely be an issue for them.

alionazherdetska:
> Caution
> 
> I've approved the PR because the implementation works. However, after discussing it again with the POW Team, I want to re-raise the concern that not being able to choose which CSS to include will likely be an issue for them.

@alizedebray Can you tell me please if I should raise it during the DEV Roundtable or?

gfellerph:
Not a candidate for v10

gfellerph:
Closing as won't do for v10.

PR #7445, opened 2026-04-01
fix(components): fix language menu interactions

### LINKED ISSUE #7216: [bug]: Language Menu conflicts with Local Header and Megadropdown
### 🐞 Describe the bug

There are two header issues related to interactions with the language menu:
1. Header collapse stops working – After clicking the language menu, the local header no longer collapses on scroll.
2. Megadropdown open fails and scroll jumps – If the language menu is opened while scrolled down, clicking the mega-dropdown does not open it, and the page scrolls upward unexpectedly.

### 🛠️ To Reproduce

Bug 1: Header collapse stops working

1. Go to '[Swisspost Cargo](https://www.swisspost-cargo.com/)' on Desktop
2. Open the 'Language Menu' (you can either close it again or leave it open - the behavior is the same)
3. Scroll down
4. Observe that the Local Header does not collapse

Bug 2: Megadropdown open fails and scroll jumps 

1. Go to '[Swisspost Cargo](https://www.swisspost-cargo.com/)' on Desktop
2. Open the 'Language Menu' (leave it open)
3. Scroll down
4. Click on a Megadropdown trigger
5. Observe that the Megadropdown trigger does not open and the page scrolls upward

### ✨ Expected behavior

1. The local header should collapse on scroll as intended, except when an element within the header has focus (to be confirmed).
2. The megadropdown should open on click, even if the language menu was previously left open and the current scroll position should not be affected.

### 📷 Screenshots

https://github.com/user-attachments/assets/74214232-f6c4-44a5-9d71-ae28d2d398e4

https://github.com/user-attachments/assets/e883baee-d8ce-40b3-858d-5e25ce591c36

alizedebray:
## 📄 Description

This PR fixes a bug where the header was not collapsing on scroll after opening the language menu.

## 🚀 Demo

https://preview-7445--swisspost-design-system-next.netlify.app/?path=/docs/27a2e64d-55ba-492d-ab79-5f7c5e818498--docs&devModeEnabled=true

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/components/post-language-menu/post-language-menu.tsx
```diff
@@ -108,6 +109,11 @@ export class PostLanguageMenu {
     }
   }
 
+  @Listen('scroll', {target: 'document', capture: true})
+  hideMenuOnScroll() {
+    if (this.menu) this.menu.hide();
```
myrtasakellariou, line 114: Probably an edge case, but should we exclude scroll events originating from within the menu itself? For the case of a really long language list.. (actually, not sure what the behavior of language menu should be in this case, if it would have a limited height and scroll).

alizedebray: The menu is currently not scrollable, but if it becomes scrollable in the future, this will still work correctly since the scroll events do not cross shadow DOM boundaries. Could you create a separate ticket for better handling overflow in the popover container?

myrta2302 (approved):

PR #7447, opened 2026-04-02
feat(components): add i18n to our date picker component

### LINKED ISSUE #7367: [web component func] Add support for Locale- and Language Code based date formatting
### 📝 Description

As a user of the Date Picker, I want the component to respect the date format of my actual region (usually defined by the OS, the installed browser language or the language definition in the HTML page), because writing a date in a foreign format can be difficult and is an common source for mistakes.

### ♿ Accessibility

The date format in a datepicker is critically important for a11y, as ambiguous formats like "01/02/03" cause severe confusion and cognitive load for users with disabilities. A clear, consistent format (preferably YYYY-MM-DD or spelled out) paired with an aria-describedby label is essential to inform screen readers of the expected input.

Read more on [W3C Docs](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/#:~:text=Cancel%20OK-,Accessibility%20Features,to%20understand%20the%20day%20names)

### ⚙️ Requirements

The component must be able to handling different [Locale Codes](https://simplelocalize.io/data/locales).

- [x] Add an optional property to the component, so it is possible to specify a Locale Code (e.g. `fr-FR`, `de-CH`, etc.).
- [x] If the property is undefined, fall back to the documents Language Code (e.g. `de`, `en`, etc.), by reading out the `lang` attribute of the `document.documentElement`.
- [x] If the documents Language Code is undefined, fall back to the browsers/OS Language Code, by accessing the `window.navigator.language` property.
- [x] Use the Locale Code/Language Code to format the outputted date.

### 🧪 Testing

- [x] Test that at least the following Locale Codes/Language Codes return the correct locale date format in the component:
  
  | Language Code | Locale Code | Language | Country | Format |
  | :- | :- | :- | :- | :- |
  | `de` || german | | `DD.MM.YYYY` |
  || `de-DE` | | germany | `DD.MM.YYYY` |
  || `de-CH` | | switzerland | `DD.MM.YYYY` |
  || `de-AT` | | austria | `DD.MM.YYYY` |
  || `de-LI` | | liechtenstein | `DD.MM.YYYY` |
  || `de-LU` | | luxembourg | `DD.MM.YYYY` |
  || `de-VA` | | vatican city | `DD.MM.YYYY` |
  | `fr` || french | | `DD/MM/YYYY` |
  || `fr-FR` | | france | `DD/MM/YYYY` |
  || `fr-CH` | | switcherland | `DD.MM.YYYY` |
  || `fr-BE` | | belgium | `DD/MM/YYYY` |
  || `fr-LU` | | luxembourg | `DD/MM/YYYY` |
  || `fr-MC` | | monaco | `DD/MM/YYYY` |
  || `fr-VA` | | vatican city | `DD/MM/YYYY` |
  | `it` || italien | | `DD/MM/YYYY` |
  || `it-IT` | | italy | `DD/MM/YYYY` |
  || `it-CH` | | switzerland | `DD.MM.YYYY` |
  || `it-SM` | | san marino | `DD/MM/YYYY` |
  || `it-VA` | | vatican city | `DD/MM/YYYY` |
  | `en` || english | | `MM/DD/YYYY` |
  || `en-GB` | | great britain | `DD/MM/YYYY` |
  || `en-US` | | united states | `MM/DD/YYYY` |

### 💻 Proposed implementation

```js
class PostDatePicker {
  private static readonly FORMAT_OPTIONS = {
    day: '2-digit',
    month: '2-digit',
    year: 'numeric'
  };

  @Prop() locale?: LanguageCode | LocaleCode | undefined;
  @Watch('locale') {
    this.localeCode = this.getLocaleCodeOrFallback();
  }

  private localeCode = this.getLocaleCodeOrFallback();
  
  const options: AirDatepickerCustomOptions = {
    onSelect: ({ date }) => {
      this.inputMask.value = date.toLocaleDateString(this.localeCode, this.FORMAT_OPTIONS); 
    },
  };
}
```

### 📃 Tasks

- [x] Implement component functionality
- [x] Add functional tests

oliverschuerch:
## 📄 Description

With this PR we're adding i18n to the date picker component.

## 🚀 Demo

---

## 🔮 Design review

- [x] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -39,9 +55,42 @@ export interface AirDatepickerCustomOptions extends AirDatepickerOptions<HTMLDiv
 export class PostDatePicker {
   @Element() host: HTMLPostDatePickerElement;
 
-  @State() today = new Date();
+  /**
+   * The date picker's selected locale (e.g. "it" or "it-CH").
+   * <br>Note: The Date Picker popup may not support the configured language and fall back to English, the right locale is important anyway to get the correct date format and text direction!
+   */
+  @Prop() locale?: string = this.systemLocale;
```
leagrdv, line 62: I think this should be more clearly stated in the story documentation that if not set, it will be using the default locale of the user's system. IMO, this is the use case most people will want

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -182,8 +225,10 @@ export class PostDatePicker {
     }
   }
 
+  @State() inputDisabled = false;
+  @State() today = new Date();
   @State() startDate = new Date();
-  @State() locale: string = Build.isBrowser ? document.documentElement.lang : 'en';
+  // @State() locale: string = Build.isBrowser ? document.documentElement.lang : 'en';
```
leagrdv, line 231: Can you remove this commented code?

packages/components/cypress/e2e/date-picker.cy.ts
```diff
@@ -1,146 +1,170 @@
...
+        cy.focused().should('have.prop', 'tagName', 'INPUT');
+      });
+    });
 
-      cy.get('@date-picker')
-        .shadow()
-        .find('.datepicker-container .air-datepicker-cell')
-        .first()
-        .should('have.attr', 'role', 'gridcell');
+    describe.only('i18n', () => {
```
leagrdv, line 117: Adding the `.only` disables all other tests

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -236,6 +279,152 @@ export class PostDatePicker {
...
+    const isBuddhistDate = this.languageCode === 'th';
+    const date = new Date(Object.values(DATE_FORMAT_MAP).join('-'));
+    // get the locale date format e.g. `22.11.3333` for `de-CH` or `11/22/3333` for `en-US`
+    let localeDateString = date.toLocaleDateString(this.localeCode, DATE_FORMAT_STRING_OPTIONS);
+
+    // replace the date parts (3333, 11, 22) with the corresponding format keys (y, m, d)
+    for (const [key, value] of Object.entries(DATE_FORMAT_MAP)) {
+      // for Thai locale, the year is in Buddhist calendar which is 543 years ahead of Gregorian calendar, so we need to adjust the year value accordingly
+      localeDateString = localeDateString.replace(
+        isBuddhistDate && value.length === 4 ? (Number(value) + 543).toString() : value,
```
leagrdv, line 320: I’m wondering if it makes sense for us to introduce locale-specific handling for Thai in this component, given that our primary audience is largely Swiss (at best european). It feels like we might be adding complexity for a relatively rare edge case.

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -726,7 +923,6 @@ export class PostDatePicker {
             }
 
             this.popoverRef?.hide();
-            requestAnimationFrame(() => this.dpInput.focus());
```
leagrdv, line 729: This line was putting the focus to the input element after selecting a date. Is there a reason why you removed it?

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -385,6 +574,7 @@ export class PostDatePicker {
         this.skipOnSelectCount = reversed ? 0 : 2; // don't skip if reversed
         this.dpInstance.selectDate([start, end]);
         this.dpInstance.setViewDate(start);
+        this.postUpdateDates.emit([this.dateToIso(start), this.dateToIso(end)]);
```
leagrdv, line 577: If I remember correctly, `this.dpInstance.selectDate([start, end]);` selects the date in the datepicker which then emits it with `postUpdateDates` so if I'm not wrong adding it here triggers it twice

packages/components/src/components/post-date-picker/post-date-picker.tsx
leagrdv, line 1: There is an issue with the datepicker that was not there before. If you don't write a date but navigate to the datepicker using the keyboard, the focused element by default should be today's date, but it's not the case anymore:

Image

packages/components/prebuild.ts
leagrdv, line 1: I can't seem to build the components package anymore, running `pnpm bootstrap` on your branch breaks on my side with the following error: 
```
│ [ ERROR ]  sass error: ...n-system\design-system\packages\components\src\styles\utilities\not-defined.scss:3:1
│            Can't find stylesheet to import.
│       L2:  @use 'sass:list';
│       L3:  @use '../../_generated/component-names.scss' as *;
│ [48:25.1]  build failed in 63.78 s
└─ Failed in 1m 7s at C:\work\swisspost\design-system\design-system\packages\components
C:\work\swisspost\design-system\design-system\packages\components:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @swisspost/design-system-components@10.0.0-next.67 build: `stencil build --docs-readme`
Exit status 1
 ELIFECYCLE  Command failed with exit code 1.
```

leagrdv (changes requested):

packages/components/src/components/post-date-picker/air-locales.ts
leagrdv, line 1: Looking at the e2e workflows, the tests that were added to the datepicker seem to be adding about 2 minutes of tests time. Is there a way to reduce that?

leagrdv (changes requested):

packages/components/cypress/e2e/date-picker.cy.ts
oliverschuerch: That's a leftover I used for testing. Thanks for spotting it!

packages/components/src/components/post-date-picker/post-date-picker.tsx
oliverschuerch: While I totally get your point.

I decided to implement it, because `th` (Thai) is one of the supported languages by the [air-datepicker](https://github.com/t1m0n/air-datepicker/tree/v3/src/locale) dependency. It was the only one missing, from all the languages the air-datepicker supports.
Therefore and because the implementation is already done, I'd like to keep it!

Btw, there are still plenty of locales, which are not supported.
Find out which one, by spotting the ones with a `exclude: true` property here:
https://github.com/swisspost/design-system/pull/7447/changes#diff-1b5b4816ccf3c1295632379051369f8018bb2b1d6f9b437d945f09a7e8127ed5

packages/components/src/components/post-date-picker/post-date-picker.tsx
oliverschuerch: Yes, there are two:

1. It is not how the native date input element works (in chrome).
  When selecting a date through the datepicker flyout, the focus stays on the flyout trigger button, which is where the user clicked:
   image
2. In the current implementation, the `postUpdateDates` event is only triggered, when the date is picked through the air-datepicker. If the user changes the input manually (set a new date or delete the current value), the air-datepicker instance is updated accordingly, but NO custom event is emitted.
  This is wrong, because we need to inform the user about every change!
  
If we emit the custom event...

- `onSelect` (change date through air-datepicker) and also
- `onBlur` (change date manually in the input),

we must NOT focus the input again when a date is picked through the air-datepicker (which triggers `onSelect`), or the `handleInputBlur` will emit the custom event again, as soon as the user moves along.

**Conclusion**

This code change and the one [here](https://github.com/swisspost/design-system/pull/7447#discussion_r3116538040), both have been done to ensure, that the event is always fired, when the date-picker value changes, but is never emitted twice for the same value update.

packages/components/src/components/post-date-picker/post-date-picker.tsx
oliverschuerch: You're right, we must avoid emitting the event multiple times for the same user action! That's why I removed `this.dpInput.focus()` [here](https://github.com/swisspost/design-system/pull/7447#discussion_r3116529477).

Also...

In the current implementation, the `postUpdateDates` event is only triggered, when the date is picked through the air-datepicker. If the user changes the input manually (set a new date or delete the current value), the air-datepicker instance is updated accordingly, but NO custom event is emitted.
  This is wrong, because we need to inform the user about every change!

packages/components/prebuild.ts
oliverschuerch: Have you been on a v9 branch before?
I don't think this is an issue of the branch, as it builds correctly on my machine and on the CI.

oliverschuerch: Sorry, this was a windows related issue I didn't have on my mac. It's fixed now.

packages/components/src/components/post-date-picker/air-locales.ts
oliverschuerch: My observations showed, that the date-picker e2e test went from ~22s up to 42-49s(+20-27s). Yes, this is a long time, but I've already optimized the tests to a minimum, at these tests alone can not be the only reasons we we have +2min testing time. My last [run](https://github.com/swisspost/design-system/actions/runs/24806385805/job/72601434094) took 13min 28s, which is "only" 2min 44s over the [average](https://github.com/swisspost/design-system/actions/metrics/performance).

image


One specific locale format (e.g. `dd.mm.yyyy` or `dd/mm/yyyy`, etc.), is tested only once, while the localeCodes using this format are summarized in the test description.
This means I only have to test 12 formats, instead of 385.

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -39,9 +57,42 @@ export interface AirDatepickerCustomOptions extends AirDatepickerOptions<HTMLDiv
 export class PostDatePicker {
   @Element() host: HTMLPostDatePickerElement;
 
-  @State() today = new Date();
+  /**
+   * The date pickers locale (e.g. "it", "it-CH", etc.), which specifies the date format and language.
+   * <post-banner type="info" data-size="sm"><span>If not set, it defaults to either the closest ancestor with a `lang` attribute (e.g. \<html lang="de"\>), or falls back to english.</span></post-banner>
```
alionazherdetska:
```suggestion
   * <post-banner type="info" data-size="sm"><span>If not set, it defaults to either the closest ancestor with a `lang` attribute (e.g. \<html lang="de"\>), or falls back to English.</span></post-banner>
```

packages/components/src/components/post-date-picker/post-date-picker.tsx
alionazherdetska: Tested `postUpdateDates` across all three input paths: calendar click, manual single date, and manual range. The event fires exactly once in each case. The skipOnSelectCount mechanism correctly absorbs the onSelect calls that selectDate() triggers internally in handleInputBlur, so no double-emit occurs.

packages/components/prebuild.ts
alionazherdetska: Confirmed: it works on `Window` now.

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -39,9 +57,42 @@ export interface AirDatepickerCustomOptions extends AirDatepickerOptions<HTMLDiv
...
+   */
+  @Prop() locale?: string = this.systemLocale;
+  @Watch('locale')
+  validateLocale() {
+    if (!isValidLocale(this.localeCode)) {
+      console.error(
+        'The prop `locale` of the `post-date-picker` component must be a valid localeCode (e.g. `en`, `en-GB`, etc.), based on <a href="https://www.rfc-editor.org/info/bcp47">BCP 47 (RFC 5646)</a> standard.',
+      );
+    }
+  }
```
alionazherdetska, line 72: `validateLocale` never actually fires the error when an invalid locale is passed. The issue is that it checks `this.localeCode` instead of `this.locale`. Since `localeCode` already applies the fallback to **'en'** before the check runs, the condition `!isValidLocale(this.localeCode)` is always false. 

Tested by setting `locale="xyz-INVALID"`: no error appeared in the console, and the logs confirmed localeCode resolved to **'en'** before validation ran.

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -580,70 +779,78 @@ export class PostDatePicker {
...
+        to: { ...baseMaskOptions },
+      },
+      lazy: false,
+      overwrite: true,
+    };
+
+    this.inputMask = IMask(this.dpInput, this.range ? rangeMaskOptions : singleMaskOptions);
+  }
+
+  private updateMask() {
```
alionazherdetska, line 847: There seems to be a crash in `updateMask` if you change the locale prop while the picker has no date selected, it throws **Cannot read properties of undefined (reading 'toLocaleDateString')**. Reproduced with a normal valid locale change (de-CH) on an empty picker.
## Reproduction:
```
<post-date-picker
  text-toggle-calendar="Open calendar"
  text-next-decade="Next decade"
  text-next-month="Next month"
  text-next-year="Next year"
  text-previous-decade="Previous decade"
  text-previous-month="Previous month"
  text-previous-year="Previous year"
  text-switch-year="Switch to year view"
  id="main"
>
  <input class="form-control" type="text" />
</post-date-picker>
```

```
document.querySelector('#main').setAttribute('locale', 'de-CH');
```

CONSOLE OUTPUT:
```
[date-picker] updateMask
selectedDates: []
selectedDates[0]: undefined
Uncaught TypeError: Cannot read properties of undefined (reading 'toLocaleDateString')
    at PostDatePicker.dateToString
    at PostDatePicker.updateMask
    at PostDatePicker.updateLocale`
```

packages/components/src/components/post-date-picker/air-locales.ts
alionazherdetska: I ran the tests locally and on CI on the `main` and on the `current PR branch`.

CI (time increased from 22 to 45 seconds):
image
Locally (time increased from 18 seconds to 1 minute 15 seconds):
image

Since it is the important func I think this increase is acceptable

oliverschuerch: If we drop tests, just because they take too long, we will end up with even more regressions. I would rather improve the e2e workflow on the ci, than miss some e2e tests, just because of timing reasons!

alionazherdetska (approved):

packages/components/cypress/e2e/helper/date-picker.ts
```diff
@@ -0,0 +1,492 @@
...
+  { locale: 'ar-ER', name: 'Arabic (Eritrea)', exclude: true },                        // | '٢٢‏/١١‏/٣٣٣٣' | rtl          | ⚠️ BUGGY ⚠️ |
+  { locale: 'ar-IL', name: 'Arabic (Israel)', exclude: true },                         // | '٢٢‏/١١‏/٣٣٣٣' | rtl          | ⚠️ BUGGY ⚠️ |
+  { locale: 'ar-IQ', name: 'Arabic (Iraq)', exclude: true },                           // | '٢٢‏/١١‏/٣٣٣٣' | rtl          | ⚠️ BUGGY ⚠️ |
+  { locale: 'ar-JO', name: 'Arabic (Jordan)', exclude: true },                         // | '٢٢‏/١١‏/٣٣٣٣' | rtl          | ⚠️ BUGGY ⚠️ |
+  { locale: 'ar-KM', name: 'Arabic (Comoros)', exclude: true },                        // | '٢٢‏/١١‏/٣٣٣٣' | rtl          | ⚠️ BUGGY ⚠️ |
+  { locale: 'ar-KW', name: 'Arabic (Kuwait)', exclude: true },                         // | '٢٢‏/١١‏/٣٣٣٣' | rtl          | ⚠️ BUGGY ⚠️ |
+  { locale: 'ar-LB', name: 'Arabic (Lebanon)', exclude: true },                        // | '٢٢‏/١١‏/٣٣٣٣' | rtl          | ⚠️ BUGGY ⚠️ |
+  { locale: 'ar-LY', name: 'Arabic (Libya)' },                                         // | 'dd‏/mm‏/yyyy' | rtl          ||
```
leagrdv, line 66: If they're being excluded, can't we just remove them from the list?

packages/components/src/components/post-date-picker/post-date-picker.tsx
leagrdv: I'm still not sure it's worth keeping because the implementation has been done. The datepicker is complex enough and adding code that we'll have to maintain for such a specific use case (that we will probs never encounter) is, in my opinion, not a great idea.

It would be great if we could use the findings in Nattaya's work towards such functionalities, to see if we could get infos on which countries/lang the post website is being called.

leagrdv (changes requested):

packages/components/cypress/e2e/helper/date-picker.ts
oliverschuerch: The file is only used in the e2e tests and if we remove them, nobody knows which one are missing, no?

But if there is a good reason to remove them, I‘ll do it 👍🏼

packages/components/src/components/post-date-picker/post-date-picker.tsx
oliverschuerch: I did some research and I'm pretty sure we don't see the big picture.

**But first...**

You're right, the component is complex, but mainly because the topic of date-handling is.

**And then there is this...**

Swiss Post owns companies all around the world, not only in Switzerland or Europe. And similar to Swiss Post Cargo, these companies might need a theme and/or a localized website in the future based on our Design System.

So just because there is no subsidiary and/or joint-venture in Thailand right now, does not mean there won't be one by tomorrow. Being prepared for as much countries/date-formats as possible should not be omitted in favor of maintainability. Apart from that, we're talking about 5-10 lines of code out of a total of 1200+.

**Sources**

- https://www.post.ch/en/about-us/profile/subsidiaries
- https://www.post.ch/en/about-us/profile/joint-ventures

**Summary**

As of the official Swiss Post lists dated 31.03.2026:

- We have 98 subsidiaries and 13 associates/joint ventures, for a total of 111 entities.
- Those are spread across 17 countries in total.
- Those are spread across 3 continents:
  Europe, Asia, and North America.

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -236,6 +281,170 @@ export class PostDatePicker {
...
+    const { y, m, d } = DATE_FORMAT_KEYS.reduce(
+      (parts, key) => ({
+        ...parts,
+        [key]: dateParts[formatParts.indexOf(key)],
+      }),
+      {} as Record<string, string>,
+    );
+
+    // Adjust for Thai Buddhist calendar
+    const year = this.isBuddhistCalendar
```
leagrdv, line 416: Did you see this comment in my previous review? https://github.com/swisspost/design-system/pull/7447#discussion_r3195417741

Somehow I can't link it to the review

leagrdv (changes requested):

leagrdv (approved):

PR #7471, opened 2026-04-10
feat(styles, docs): button group style update and vertical variant

### LINKED ISSUE #7402: [html/css] Button Group Horizontal / Vertical
### 📝 Description

A button group is used to visually and functionally group a series of related buttons. It ensures consistent spacing, alignment, and interaction patterns across multiple controls.

The component supports multiple types of grouped elements: **Buttons**, **Radio buttons**, **Checkboxes**, **Links**.
All variants share the same visual styling and layout behavior.

The button group supports two orientations: **Horizontal (default)**, **Vertical**

It also supports **responsive orientation switching** depending on viewport breakpoints.

### ♿ Accessibility

- The parent container must have `role="group"`.
- The group must have an accessible name: Make sure the example has an `aria-label` on the parent container.
- When grouping radio buttons, consider using `role="radiogroup"` instead of group.
- Focus styles must remain clearly visible for all items.

### ⚙️ Requirements

The component must support:
- **Horizontal and vertical layouts**
- **Responvie orientation changes**
- **All supported child element types**

#### Orientation system
To avoid breaking changes, the `.btn-group` class would handle all the default styling in horizontal mode. 
Add the following classes:
- `.btn-group-vertical` → forces vertical layout
- `.btn-group-horizontal` → forces horizontal layout

Add the following breakpoint modifiers for both orientation classes: `*-sm-*`, `*-md-*`, `*-lg-*`, `*-xl-*`.

*Example:*
`<div class="btn-group btn-group-vertical btn-group-md-horizontal">` → button group is vertical but switches to horizontal from `md` and up

### 🧪 Testing

Add to snapshot tests:
- Vertical variant
- Responsive combinations

Accessibility:
- Group is announced correctly with the accessible name being read
- Keyboard navigation works as expected

### 💻 Proposed implementation

Example for a button group of button elements:
```
<div class="btn-group btn-group-vertical btn-group-md-horizontal" role="group" aria-label="Button group example">
  <button type="button">Left</button>
  <button type="button">Middle</button>
  <button type="button">Right</button>
</div>
```

### 📃 Tasks

- [ ] Update storybook documentation
- [ ] Document the responsive behaviour and how to use the modifier classes
- [ ] Update visual regression tests
- [ ] Design review

leagrdv:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [x] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


sandra-post:
@leagrdv 
### Design review

**General**
- Icon size:
  - IS: 24 x 20 (Desktop), 16 x 14 (Mobile)
  - SHOULD: 24 x 24 (Desktop), 16 x 16 (Mobile)
- In the controls, I only see icon TRUE/FALSE, but no label TRUE/FALSE. Icon only should also be an option.

**Button group horizontal**
- All buttons within a group should have the same width and NOT their individual calculated width from their respective content (data +  padding). Meaning: Button widths = Same width as button with longest content OR Group width / number of buttons (thats how its done in Figma bc 1st option technically impossible)

sandra-post:
@leagrdv 
### Design review 2
All addressed issues are fixed now thanks.

apps/documentation/src/stories/components/button-group/button-group.stories.ts
```diff
@@ -186,46 +197,66 @@ function createButtonTemplate(args: Args, context: StoryContext, index: number)
...
-      return html` <button type="button" class="btn ${args.size} btn-secondary">${label}</button> `;
+      return html` <button type="button">${label}</button> `;
   }
 }
 
 function renderButtonGroup(args: Args, context: StoryContext) {
   return html`
-    <div class="btn-group" role="group" aria-label="Button group example">
-      ${Array.from({ length: 3 }).map((_, i) => createButtonTemplate(args, context, i))}
+    <div
```
alizedebray: For radio and checkboxes the wrapper should probably be a field set. 

Also, I've noticed the focus behavior is different than regular checkboxes: to go from a button to the other you cannot use the `Tab` key, you need to use the arrows.

alizedebray (changes requested):

apps/documentation/src/stories/components/button-group/button-group.stories.ts
leagrdv: I've updated to a fieldset for the radio and checkbox variants.

For the focus behavior, the radio differ from the other (buttons, links and checkboxes) as the native way of navigating between them is the arrows. The other types use the `Tab` as expected.
I think changing the names of the radio inputs would make it possible to use with `Tab` instead of arrows but it goes against the way radio groups are supposed to work.

alizedebray (approved):

PR #7488, opened 2026-04-14
feat(component): added a `klp-login-widget` component

### LINKED ISSUE #7175: [web component func] KLP Login Widget

alionazherdetska:
## 📄 Description

Introduces the `post-login-widget` web component, which connects to the KLP session API and conditionally renders either a login link or a user menu based on the current session state.

### Notes

1. **Naming:** The component is named `post-login-widget` to avoid conflicts with the existing `post-klp-login-widget` in `@swisspost/internet-header`. It will be renamed once the `internet-header` package migration is complete — a TODO comment is left in `post-login-widget.tsx` to track this.
2. No changeset included: This component should probably be released only after the internet-header components are migrated and the component is renamed. The changeset will be added at that point.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/integration-next/src/app/ssr/layout.tsx
```diff
@@ -83,10 +84,12 @@ export default function Layout({ children }: { readonly children: React.ReactNod
...
 
         {/* Global header login/user menu */}
-        <a href="" slot="post-login">
-          <span>Login</span>
-          <PostIcon name="login" />
-        </a>
+        <PostLoginWidget slot="post-login">
+          <a href="#login" slot="unauthenticated">
+            <span>Login</span>
+            <PostIcon name="login" />
```
leagrdv: `aria-hidden` is missing on the icon

apps/documentation/src/stories/raw-components/login-widget/post-login-widget.stories.ts
```diff
@@ -0,0 +1,119 @@
...
+    design: {},
+    controls: {
+      exclude: ['postLoginChange'],
+    },
+  },
+  args: {
+    authenticated: false,
+  },
+  argTypes: {
+    authenticated: {
```
leagrdv, line 22: Should this be an argument? This state should be defined by the API's answer, I don't think it should be a prop

packages/components/src/components/post-login-widget/post-login-widget.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * An event emitted when the authentication state changes. 
+   * The event payload is an object: `authenticated` is `true` when the user is logged in, `false` when logged out.
+   */
+  @Event() postLoginChange: EventEmitter<{ authenticated: boolean }>;
+
+  async componentWillLoad() {
+    if (Build.isBrowser) {
+      if (this.host.hasAttribute('authenticated')) {
```
leagrdv: This forces a state that should be defined by the API's response. What happens when `authenticated` is forced but there is no user?

apps/documentation/src/stories/raw-components/login-widget/post-login-widget.stories.ts
```diff
@@ -0,0 +1,119 @@
+import type { Args, StoryObj } from '@storybook/web-components-vite';
+import { html } from 'lit';
+import { MetaComponent } from '@root/types';
+
+const meta: MetaComponent = {
+  id: 'a3f8b2c1-1d4e-4f6a-9b0c-2e7d8f3a5c6b',
+  title: 'Raw Components/Login Widget',
```
leagrdv, line 7: Despite the title, the login widget isn't placed in the `Raw Components` folder

leagrdv (changes requested):

packages/components/src/components/post-login-widget/post-login-widget.tsx
```diff
@@ -0,0 +1,87 @@
...
+  tag: 'post-login-widget',
+  shadow: true,
+})
+export class PostLoginWidget {
+  @Element() host: HTMLPostLoginWidgetElement;
+
+  /**
+   * The current authentication state: `null` (loading), `true` (authenticated), `false` (not authenticated).
+   */
+  @Prop({ mutable: true, reflect: true }) authenticated: boolean | null = null;
```
leagrdv: If it's a state, it should be a state and not a prop.

leagrdv (changes requested):

leagrdv (approved):

packages/components/src/components/post-login-widget/post-login-widget.tsx
```diff
@@ -0,0 +1,81 @@
...
+  @Method()
+  async refresh(): Promise<void> {
+    await this.fetchAuthState();
+  }
+
+  private async fetchAuthState(): Promise<void> {
+    try {
+      const response = await fetch(SESSION_URL, {
+        credentials: 'include',
+      });
```
oliverschuerch, line 62: I'm pretty sure, using the `POST` method for a login request is standard practice and highly recommended, even if you are not sending data in the request body!

And also it's not clear to me, if we have to fetch a bearer token or something similar for security reasons first, or if the endpoint is publicly available for everyone. Getting a bearer token first (which indicates authentication status) and then use it to get the current user information, seems more secure to me. But I'm maybe not up-to-date with these topics anymore...

We should ask @gfellerph and someone from the Team which creates the endpoint.

packages/components/src/components/post-login-widget/post-login-widget.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+      if (!response.ok) {
+        this.setAuthState(false);
+        return;
+      }
+
+      const data = await response.json();
+      const isAuthenticated = !!(data && data.data && Object.keys(data.data).length > 0);
```
oliverschuerch: This seems like a very uncommon way to check if a user is logged in.
I'd prefer to check for a specific property.

`const isAuthenticated = data.auth; // boolean property`
or
`const isAuthenticated = data.email !== undefined;`

packages/components/src/components/post-login-widget/post-login-widget.tsx
```diff
@@ -0,0 +1,81 @@
...
+      this.setAuthState(false);
+    }
+  }
+
+  private setAuthState(next: boolean): void {
+    if (this.authenticated === next) return;
+
+    this.authenticated = next;
+    this.postLoginChange.emit({ authenticated: next });
+  }
```
oliverschuerch, line 82: I'm not sure about this event. Do we really need it?
The application should know independently if the user is logged in.
Imo, the authentication state must be clear, before a specific route can be rendered.

Be aware, I have no idea how the login mechanics worked in the old way, nor what is planned for the future!

packages/components/cypress/e2e/login-widget.cy.ts
```diff
@@ -0,0 +1,163 @@
...
+  data: {
+    name: 'First Name',
+    surname: 'Last Name',
+    email: 'Email',
+    userType: 'B2C',
+    authLevel: 'PASSWORD',
+    support: false,
+    changeUserAndProfile: 'notAvailable',
+  },
+};
```
oliverschuerch, line 16: Is this really representing the object, that will be sent back by the session endpoint, inside the response `data|body` property?

I'd expect it will send back only something like:
```json
{
  "name": "First Name",
  "surname": "Last Name",
  "email": "Email",
  "userType": "B2C",
  "authLevel": "PASSWORD",
  "support": false,
  "changeUserAndProfile": "notAvailable",
}
```

oliverschuerch (changes requested):

packages/components/src/components/post-login-widget/post-login-widget.tsx
oliverschuerch: Sorry, this was a misunderstanding of mine...

It's ok to go with `GET`, as this is not an authentication request/login attempt, but just the trial to get the current users data.
The user must have logged in earlier, after which the server sat an authentication cookie on the client, which is always sent along with the request for authentication purposes, thanks to the `credentials: 'include' flag.

packages/components/src/components/post-login-widget/post-login-widget.tsx
oliverschuerch: Btw. This is how the old object looks like:
```json
{
  "userId": "{number}",
  "firstName": "firstname",
  "lastName": "lastname",
  "email": "oli@skaos.ch",
  ...
}
```
The `userId` would be the perfect field to check for, because the field will probably never be renamed/removed.

oliverschuerch (approved):

PR #7494, opened 2026-04-15
feat(components): post-tabs ssr

### LINKED ISSUE #7093: [web component ssr] Tabs
### 🧾 Findings

With JavaScript disabled, no tab is selected on initial load:

Image

### 🧪 How to test

**Run the Next.js SSR page**

After updating the component, ensure it is compiled correctly as a React component by running the following commands:

```bash
pnpm components:build
pnpm components-react:build
```

Then, to focus only on the component being tested, isolate it in the Next.js application.
Comment out all components that are not under test and their imports in the following files:

- `packages/nextjs-integration/src/app/ssr/layout.tsx`
- `packages/nextjs-integration/src/app/ssr/page.tsx`

Finally, serve the application with:

```bash
pnpm nextjs:start
```

**Test SSR in the browser**

- Open the application in two different tabs
- In one tab, disable JavaScript and refresh the page
- Compare both pages, they must look the same

### 📃 Tasks

- [x] Ensure Shadow DOM is enabled (explicit `shadow: true`) for the component and potential sub-components
- [x] Verify no hydration or console errors are thrown when running the Next.js SSR page
- [x] Confirm the page looks the same on initial load with JavaScript enabled and disabled across all devices and supported configurations
- [x] Ensure ARIA attribute cross-referencing (e.g. `aria-labelledby`, `aria-describedby`) works as expected

myrta2302:
## 📄 Description

This PR updates the `post-tabs` and its sub-components to achieve SSR compatibility.
 
To avoid breaking changes, active state styling in the no-JS version is driven by CSS custom properties injected at render time. These act as binary switches (0/1) that toggle between default and active styles using `calc()` and `color-mix()`.

_Note:_ High Contrast Mode with an explicit `active-tab` prop produces slightly different styles compared to the CSR version. This (possibly) happens because `color-mix()` values are overridden by the browser in forced colors mode. So the toggle does not work there.

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/styles/src/components/tabs/_tab-title.scss
```diff
@@ -151,17 +158,107 @@ tokens.$default-map: utilities.$post-spacing;
...
+      #{color.$white} calc(var(--active, 0) * 100%),
+      transparent
+    );
+    color: #{color.$black};
+    font-weight: calc(400 + var(--active, 0) * 300);
+
+    &::before {
+      content: '';
+      height: var(--post-core-dimension-4);
+      background-color: #fc0;
```
alionazherdetska: This is hardcoded to `#fc0` while the equivalent non-SSR path and the `post-tabs:not([active-tab])` fallback both use `nav.$nav-tabs-link-active-border-color`. They compile to the same value today, but using the token here too would keep things consistent and make future color updates automatic across all three paths.

```
// before
background-color: #fc0;

// after
background-color: nav.$nav-tabs-link-active-border-color;
```

packages/components/src/components/post-tabs/post-tabs.tsx
```diff
@@ -394,9 +391,14 @@ export class PostTabs {
 
   render() {
     const TabsContainer = this.isNavigationMode ? 'nav' : 'div';
-
+    const activeTabName = this.host.getAttribute('active-tab');
```
alionazherdetska: Since `activeTab` is not reflected, `getAttribute('active-tab')` can return `null` on re-renders even when the prop has a value. Tested this and confirmed: on the second render pass `getAttribute` returns `null` while `this.activeTab` is still `'dua'`. On the server that would mean injecting `--post-tab-panel-null: block` instead of the right variable, so the active tab wouldn't show. Can you use `this.activeTab` here instead?

Image

packages/styles/src/components/tabs/_tab-title.scss
alionazherdetska, line 1: Question / accessibility thought: in navigation mode, tabs are still keyboard-navigable with JS disabled since they're real anchor elements. In panel mode with JS disabled, the tab items render as buttons but have no focus styles and clicking them does nothing, the whole interaction is JS-dependent. For users with disabilities on slow connections, there's no fallback at all in panel mode. I was wondering if we're accepting for now.

https://github.com/user-attachments/assets/ddb782f5-cf9b-4326-9d75-17be38b45439

alionazherdetska (changes requested):

packages/styles/src/components/tabs/_tab-title.scss
myrta2302: Removed hover styles from panel mode in the SSR state to avoid implying interactivity before JS is available.

myrta2302: Maybe the active item should anyway have a `cursor: default` for all variants. (TBD)

alionazherdetska: Maybe when JS is disabled at least the active item should get focus styles in panel mode, similar to what the [W3C tabs example](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/) does . But that's probably out of scope here since this ticket is about making sure there is no hydration errors and the page looks the same on initial load with JavaScript enabled and disabled.
@oliverschuerch What do you think?

alionazherdetska (approved):

oliverschuerch (changes requested):
Since the StencilJs [Build Conditionals](https://stenciljs.com/docs/guides/build-conditionals) are handled in a special way during build time, can we please test this update on the POW repo, before we ship it?

myrtasakellariou:
> Since the StencilJs [Build Conditionals](https://stenciljs.com/docs/guides/build-conditionals) are handled in a special way during build time, can we please test this update on the POW repo, before we ship it?

In the end i tested it on Cargo. More specifically i tested the tab-styles PR (https://github.com/swisspost/design-system/pull/7612) which is based on this. There are no errors coming from the tabs. The functionality with the active styling for SSR is as expected. 
image

packages/styles/src/components/tabs/_tab-title.scss
oliverschuerch: I would say we should not waste to much time to make the panel variant of the tabs work without JS. SSR compatibility has nothing to do with "disabled JS" (even if it may help to disable JS for testing).

If we must ensure the tabs content is consumable for users with disabled JS, we would need to investigate a bit if the `<noscript>` tag, added to the components shadowDom could be a solution (to render conditional HTML/CSS during server-side rendering). Because in this case we would need to show all the panel contents, no matter a tab is active or not.

However, if this is really needed/wanted, we should clearly separate it from this task here.

oliverschuerch (approved):

myrtasakellariou:
To be merged together with https://github.com/swisspost/design-system/pull/7608 and https://github.com/swisspost/design-system/pull/7608.

PR #7501, opened 2026-04-16
fix(components): make breadcrumbs more predictable

### LINKED ISSUE #7460: [bug]: breadcrumb item URL not applied
### 🐞 Describe the bug

Breadcrumbs URLs are not correctly applied in NextJS.

### 🛠️ To Reproduce

1. Start the NextJS app with `pnpm nextjs:start`
2. Inspect the shadow DOM of a breadcrumb item 
3. See the url set on the item is not applied to the `href` on the inner anchor

### ✨ Expected behavior

The behavior should be the same as in the documentation, the `url` property should be use for the `href` property of the anchor rendered by the breadcrumb item.

### 🏷️ Version of the Design System

`10.0.0-next.66`

alizedebray:
## 📄 Description

This PR refactors the breadcrumbs to make them more predictable.
It removes logic that was previously cloning visible elements and parsing properties and use these as they are provided as much as possible.

## 🚀 Demo

https://preview-7501--swisspost-design-system-next.netlify.app/?path=/docs/b7db7391-f893-4b1e-a125-b30c6f0b028b--docs&devModeEnabled=true

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.scss
alionazherdetska, line 1: The focus is clipped for the home-icon and breadcrumb-items:
https://github.com/user-attachments/assets/36d5ed0e-5cbf-402f-a6bd-088f29a0a302

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.tsx
```diff
@@ -14,64 +15,52 @@ export class PostBreadcrumbItem {
...
+    checkEmptyOrOneOf(this, 'variant', VARIANTS);
   }
 
-  connectedCallback() {
-    this.validateUrl();
-  }
+  /**
+   * Defines whether the component renders as a list item or a menu item.
+   */
+  @Prop({ reflect: true }) selected = false;
```
alionazherdetska, line 40: It says "Defines whether the component renders as a list item or a menu item" which is variant's description. Should describe that this marks the item as the current page instead.

packages/components/src/components/post-breadcrumbs/post-breadcrumbs.tsx
alionazherdetska, line 1: After pressing `Enter` or `Space` on the `...` button, focus stays on the button instead of moving to the first `menu-item`:

Image

packages/components/src/components/post-breadcrumbs/post-breadcrumbs.tsx
```diff
@@ -1,235 +1,188 @@
...
+  private updateBreadcrumbItemProps() {
+    const breadcrumbItems = this.host.querySelectorAll('post-breadcrumb-item');
+    breadcrumbItems.forEach((item, index) => {
+      item.setAttribute('variant', this.shouldRenderMenu ? 'menuitem' : 'listitem');
+      item.setAttribute('selected', String(index === breadcrumbItems.length - 1));
+    });
+  }
+
+  private renderMenu() {
+    const menuId = `m${nanoid(6)}`;
```
alionazherdetska: It seems that the `menuId` is regenerated on every render: I added a quick log to confirm and every time the component resizes into the concatenated state the new menuId is getting generated:

Image

packages/components/src/components/post-breadcrumbs/post-breadcrumbs.tsx
```diff
@@ -1,235 +1,188 @@
...
+    Promise.all(breadcrumbItems.map(item => componentOnReady(item))).then(() => {
+      this.shouldRenderMenu = hiddenNav.scrollWidth > hiddenNav.clientWidth;
+      this.updateBreadcrumbItemProps();
+    });
+  }
+
+  private prepareBreadcrumbItemsForClone() {
+    this.shouldRenderMenu = false;
+    this.updateBreadcrumbItemProps();
+  }
```
alionazherdetska, line 146: Resetting `shouldRenderMenu` before cloning breaks the overflow measurement

Calling `updateBreadcrumbItemProps` here sets all items back to [variant="listitem"] before the clone is measured. This forces Stencil to re-render the live items at full width, so `checkOverflow` ends up measuring a wider DOM than the container allows and incorrectly decides the menu is not needed.

The result: when items overflow and the component should switch to menu mode, it stays in list mode and visually overflows instead:

Image

alionazherdetska (changes requested):

packages/components/src/components/post-breadcrumbs/post-breadcrumbs.tsx
alizedebray: Are you sure about that? 
Rendering all the items as `listitem` before cloning is necessary because this is the state we want to measure. If we try and measure when the items are in a menu, we will never get an overflow.

Also you wrote: 
> `checkOverflow` ends up measuring a wider DOM than the container allows and incorrectly decides the menu is not needed.

But it is the opposite: if `checkOverflow` measurs a wider DOM than the container allows, the menu **is** needed.

I've checked both the docs and the components playground but could not reproduce the state shown in your screen shot. Which browser are you on?

packages/internet-header/src/components/post-internet-breadcrumbs/readme.md
alionazherdetska, line 1: Can you please fix he tests for preconfigured `breadcrumb`:

Image

packages/components/src/components/post-breadcrumbs/post-breadcrumbs.tsx
alionazherdetska: https://github.com/user-attachments/assets/062bb34b-f8f9-44fb-9026-632b5835e4c3

I noticed this while testing in Edge by dynamically injecting items via appendChild in the console: the concatenation state flips incorrectly as items are added. I think this is not relevant since dynamically injecting breadcrumb items at runtime isn't a supported use case, so feel free to ignore this comment.

apps/documentation/src/stories/components/breadcrumb/breadcrumbs.stories.ts
```diff
@@ -7,100 +7,52 @@ const meta: MetaComponent = {
...
-  render: (args: Args) => html`
-    <post-breadcrumb-item url=${args.url} label=${args.label} description=${args.description}
-      >${args.content}</post-breadcrumb-item
-    >
-  `,
+type Story = StoryObj;
+
+export const Default: Story = {};
+
+export const Concatenated: Story = {
```
alionazherdetska, line 54: Small thing, feel free to ignore:
There is a horizontal scroll for the concatenated Breadcrumbs story for mobile and tablet viewports:

https://github.com/user-attachments/assets/c6e3173a-a6a8-4e77-aaa2-b1978cddfa44

apps/documentation/src/stories/components/breadcrumb/breadcrumb-item.stories.ts
```diff
@@ -0,0 +1,51 @@
...
+    design: {
+      type: 'figma',
+      url: 'https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-Next-Level?node-id=558-16158&m=dev',
+    },
+  },
+  args: {
+    variant: 'listitem',
+    label: 'Products section',
+    description: 'This section contains all the products you can buy.',
+    selected: false,
```
alionazherdetska, line 25: Toggling `selected` to true in the controls panel doesn't apply the selected styles to the item.

Image

apps/documentation/src/stories/components/breadcrumb/breadcrumbs.stories.ts
alionazherdetska, line 1: Image
Just noticed the Concatenated story has the same control table as Default, was just wondering whether that might be confusing for users to have two control tables for the same component. Not relevant to this PR though since it is not something you added and it was already there, so feel free to ignore.

packages/components/src/components/post-breadcrumbs/post-breadcrumbs.tsx
alionazherdetska, line 1: When navigating the dropdown menu with the keyboard, pressing **Home** moves focus to the first item as expected, but it also scrolls the page to the top at the same time.

https://github.com/user-attachments/assets/41834887-542f-4c81-9514-609477cab6e5

alionazherdetska (changes requested):

packages/components/src/components/post-breadcrumbs/post-breadcrumbs.tsx
alizedebray: This issue is also reproducible in the current version of the breadcrumbs, and it should be handled by the `post-menu` component directly.

Could you open a separate ticket for this?

alionazherdetska: https://github.com/orgs/swisspost/projects/3/views/26?pane=issue&itemId=184641101&issue=swisspost%7Cdesign-system%7C7605

apps/documentation/src/stories/components/breadcrumb/breadcrumb-item.stories.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+function render(args: Args) {
+  return html`
+    <post-breadcrumb-item
+      url=${args.url || nothing}
+      variant=${args.variant !== 'listitem' ? args.variant : nothing}
+      label=${args.label || nothing}
+      description=${args.description || nothing}
+      ?selected=${args.selected}
```
alionazherdetska, line 43: Boolean props should have the value true or false, right now if the user sets `selected=true` it reflects the emty string in the markup:

Image

It should be similar to the `post-tabs` stories with `full-width` prop:

Image

apps/documentation/src/stories/components/breadcrumb/breadcrumb-item.stories.ts
alionazherdetska, line 1: ### Small thing, feel free to ignore:
If selected is set to true and the user then switches to the post-menu variant, the list variant is still shown instead.

Image

alionazherdetska (dismissed):

apps/documentation/src/stories/components/breadcrumb/breadcrumb-item.stories.ts
alizedebray: True and false are invalid values for boolean attributes in HTML: https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML

apps/documentation/src/stories/components/breadcrumb/breadcrumb-item.stories.ts
alizedebray: The selected item does not change depending on the variant, it is always outside the menu.

leagrdv: Should we show this prop in the docs? It's calculated based on the available space, I feel like it might be misleading to show this option to the user

leagrdv: Same goes for the `selected` props, it shouldn't be set by the user so I think we can remove it from the docs.

leagrdv (changes requested):

packages/components/src/components.d.ts
leagrdv, line 1: When building locally, I got some changes in this file, could you update it?

leagrdv (changes requested):

apps/documentation/src/stories/components/breadcrumb/breadcrumb-item.stories.ts
alizedebray: I've left this part of the docs simply because it was already there, but I can remove it.

apps/documentation/src/stories/components/breadcrumb/breadcrumbs.stories.ts
```diff
@@ -7,100 +7,52 @@ const meta: MetaComponent = {
...
-      <post-breadcrumb-item url="/section7">Section 7</post-breadcrumb-item>
-      <post-breadcrumb-item url="/section8">Section 8</post-breadcrumb-item>
-      <post-breadcrumb-item url="/section9">Section 9</post-breadcrumb-item>
-      <post-breadcrumb-item url="/section10">Section 10</post-breadcrumb-item>
-    </post-breadcrumbs>
-  `,
-};
+export default meta;
 
-export const BreadcrumbItem: Story = {
```
leagrdv, line 66: My comment was simply regarding the `selected` and `variant` prop. Having the breadcrumb item props that the user can update on the docs made sense in my opinion (since `label` and `description` are not shown in other examples, without this the user has no way to know those props exist). It's simply those new props that were added with your changes, they only should be coming from the parent `post-breadcrumb` and user should not set them themselves so those are the ones I would have simply excluded from the table.

leagrdv (approved):

PR #7510, opened 2026-04-17
chore(eslint): preserve attribute casing and prevent empty attr serialization

### LINKED ISSUE #7491: [bug]: Running the migration script changes Angular props to lowercase
### 🐞 Describe the bug

When running the automigration scripts on a project, the following errors occur:
- Some Angular props are being changed to lowercase, such as `ngClass` or `routerLink`, which breaks the page.
- Empty props, directives or anything else on a node gets added an `=""` automatically

**Example:** 

Before the migration
```
<nav
    class="col-md-3 col-lg-2 d-md-flex bg-light sidebar p-md-3"
    id="sidebar"
    [ngClass]="{'closed': !openedMenu}"
>
<a routerLink="ui/home" translate>template.menu.home</a>
```

After the migration
```
<nav class="col-md-3 col-lg-2 d-md-flex sidebar p-md-16" id="sidebar" [ngclass]="{'closed': !openedMenu}">
<a routerlink="ui/home" translate="">template.menu.home</a>
```

According to Alizé, issue might potentially come from the `node.toCheerio` function

### 🛠️ To Reproduce

Update a project to v10 and run the migration script as described in the migration guide

### ✨ Expected behavior

Only the classes or eventually the components names should be targeted but everything other than the specific parts that need to be migrated should stay as they are.

myrta2302:
## 📄 Description

When running automigration rules on Angular templates, Cheerio's serializer was causing two issues:
- Attribute names were being lowercased (e.g. `[ngClass]` → `[ngclass]`, `routerLink` → `routerlink`)
- Valueless attributes were getting `=""` appended (e.g. `translate` → `translate=""`)

## Solution
- Added `parse5-case-sensitive` override to preserve attribute casing in Cheerio's parser  [https://github.com/antfu/parse5-case-sensitive](https://github.com/antfu/parse5-case-sensitive)
- Added `removeEmptyAttrs` utility for rules that still use Cheerio serialization, which strips `=""` added by Cheerio unless the original source already had empty attributes. The alternative would be to remove Cheerio from serialization entirely, which would require a more significant refactor.
---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


leagrdv (changes requested):
I tried to run the unit tests locally with some added empty attributes and it's still adding `=""` on them
Image

packages/eslint/test/rules/html/migrations/no-deprecated-btn-rg.spec.ts
```diff
@@ -28,8 +28,8 @@ htmlRuleTester.run(name, rule, {
   ],
   invalid: [
     {
-      code: '<button class="btn btn-rg">Click me</button>',
-      output: '<button class="btn btn-sm">Click me</button>',
+      code: '<button class="btn btn-rg" testattr test-attr1="" test-attr2="">Click me</button> ',
```
leagrdv: Could you just update those tests by modifying on attribute to camel case? (e.g. `testAttr`)

leagrdv (approved):

PR #7546, opened 2026-04-23
feat(dialog): add bottom sheet variant to dialog

JoaoSequeira489:
## 📄 Description

Add Bottom sheet variant to Dialog component

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


gfellerph:
@JoaoSequeira489, can you add a changeset (`pnpm changeset`)? Something like: "Added a bottom-sheet variant for the dialog component." and mark it as feature for the styles and documentation?

apps/documentation/src/stories/components/dialog/dialog.docs.mdx
```diff
@@ -46,6 +46,40 @@ import PackageTag from '@/shared/package-tag';
...
+/>
+
+<Source
+  code={`<div class="container-fluid">
+  <dialog class="post-dialog" data-variant="bottom-sheet">
+    ...
+  </dialog>
+</div>`}
+  language="html"
+/>
```
oliverschuerch: Please extract multiline code snippets into specific `*.sample.html` files, where you ensure correct indentation.
Have a look at the [back-to-top](https://github.com/swisspost/design-system/tree/main/apps/documentation/src/stories/components/back-to-top) component, if you want to see an example.

packages/styles/src/components/dialog.scss
```diff
@@ -15,6 +17,7 @@ tokens.$default-map: utilities.$post-spacing;
 
 $dialog-types: 'info', 'error', 'success', 'warning';
 $border-radius: 0.5rem;
+$dialog-bottom-sheet-selector: "[data-variant='bottom-sheet']";
```
oliverschuerch, line 20: The variable is as long as the real selector. Therefore it is not really helpful, imo. And also it makes it harder to read the SCSS below. I would prefer to use the selector directly below and get rid of this useless variable.

apps/documentation/src/stories/components/dialog/dialog.snapshot.stories.ts
```diff
@@ -41,3 +41,29 @@ export const Dialog: Story = {
...
+            width: '',
+            content:
+              'Contentus momentus vero siteos et accusam iretea et justo. Contentus momentus vero siteos et accusam iretea et justo.',
+          },
+          context,
+        )}
+      `,
+    );
+  },
+};
```
oliverschuerch, line 74: ## Problem

The snapshot story is not working as expected:
https://preview-7546--swisspost-design-system-next.netlify.app/iframe.html?id=snapshots--dialog-bottom-sheet&viewMode=story

The bottom-sheet dialog is positioned fixed and sticks to the bottom of the page. The light and dark variants are overlaying each other.

The goal should be to have them rendered in their corresponding `Color Scheme` container:

Image

## Proposals

**Easy but somehow incorrect**

Set `position: relative` to the `dialog` element.

**More advanced**

Render the story in an iframe by adding the following parameters to the `DialogBottomSheet` story:

```ts
export const DialogBottomSheet: Story = {
  parameters: {
    docs: {
      story: {
        inline: false,
      },
    },
  },
  render: ...
}
```

Related [Docs](https://storybook.js.org/docs/api/doc-blocks/doc-block-story#inline)

It might be necessary to set a fixed `height` for the story.
One way to do this, is by using a [decorator](https://storybook.js.org/docs/writing-stories/decorators) for this specific story.

apps/documentation/src/stories/components/dialog/dialog.stories.ts
```diff
@@ -15,9 +15,11 @@ const meta: Meta = {
     title: 'Dialog',
     content: 'This is a dialog',
     type: 'default',
+    variant: 'default',
```
oliverschuerch, line 28: I would move this one line up, so it shows up before the `type` control on the docs page.

Image

apps/documentation/src/stories/components/dialog/dialog.stories.ts
```diff
@@ -63,11 +76,19 @@ const meta: Meta = {
...
       options: ['pop-in', 'slide-in', 'none'],
       table: { category: 'Variant' },
     },
+    width: {
+      name: 'Width override',
+      description:
+        'Optional `--post-dialog-width` override. This replaces the bottom-sheet variant’s responsive default width.',
+      control: 'text',
+      table: { category: 'Variant' },
+    },
```
oliverschuerch, line 103: This control should show up conditionally. Only show it when `variant` is set to `bottom-sheet`.
See [Docs](https://storybook.js.org/docs/api/arg-types#if) to learn how to do it.

apps/documentation/src/stories/components/dialog/dialog.stories.ts
```diff
@@ -157,20 +178,27 @@ const getControls = () => {
...
       <dialog
         class="post-dialog"
         data-type="${args.type !== 'default' ? args.type : nothing}"
+        data-variant="${isBottomSheet ? args.variant : nothing}"
         data-size="${args.size}"
         data-position="${args.position}"
-        data-animation="${args.animation}"
+        data-animation="${animation}"
         open="${args.open || nothing}"
+        style="${style}"
```
oliverschuerch: I'd prefer to have that implemented a bit more dynamic, so we can add more variants, without the need to touch this logic again.

Note: This is just a suggestion and completely untested!

```ts
import { Args, Meta, StoryObj } from '@storybook/web-components-vite';
import { html, nothing } from 'lit-html';

const DEFAULT_ANIMATION = 'pop-in';
const VARIANT_ANIMATION_MAP = {
  'bottom-sheet': 'slide-in'
};
const WIDTH_OVERRIDE_ALLOWED_VARIANTS = [
  'bottom-sheet'
];

const meta: Meta = {
  ...
  args: {
    ...
    animation: DEFAULT_ANIMATION,
    ...
  },
  ...
};

...

const Template = {
  render: (args: Args) => {
    const postDialogCloseButton = args.closeButton ? getCloseButton() : nothing;
    // every variant not defined in the MAP is using the default animation
    const defaultVariantAnimation = VARIANT_ANIMATION_MAP[args.variant] ?? DEFAULT_ANIMATION;
    // if the currently configured animation matches the default animation for the specific variant, don't add it as a parameter
    const animation = args.animation === defaultVariantAnimation ? nothing : args.animation;
    // ensure only variants, with the possibility to override the width contain the inline style
    // this might not be necessary, if the `width` argument is conditionally
    const style = WIDTH_OVERRIDE_ALLOWED_VARIANTS.includes(args.variant) && args.width ? `--post-dialog-width: ${args.width}` : nothing;

    // Don't declare default values or show empty containers
 // if (args.animation === 'pop-in') args.animation = nothing; remove this line! 
    if (args.position === 'center') args.position = nothing;
    if (args.size === 'medium') args.size = nothing;

    return html`
      <dialog
        class="post-dialog"
        data-type="${args.type !== 'default' ? args.type : nothing}"
        data-variant="${isBottomSheet ? args.variant : nothing}"
        data-size="${args.size}"
        data-position="${args.position}"
   <!-- data-animation="${args.animation}" remove this line! -->
        data-animation="${animation}"
        open="${args.open || nothing}"
        style="${style}"
        aria-labelledby="dialog-title"
        aria-describedby="dialog-description"
      >
      ...
```

apps/documentation/src/stories/components/dialog/dialog.stories.ts
```diff
@@ -248,6 +276,47 @@ const CustomContentTemplate = {
...
+  render: (args: Args) => {
+    const postDialogCloseButton = args.closeButton ? getCloseButton() : nothing;
+    const style = args.width ? `--post-dialog-width: ${args.width}` : nothing;
+
+    return html`
+      <dialog
+        class="post-dialog"
+        data-variant="bottom-sheet"
+        data-type="${args.type !== 'default' ? args.type : nothing}"
+        data-animation="${args.animation === 'slide-in' ? nothing : args.animation}"
```
oliverschuerch: **Optional**
There is no control table to change these options for this story, so you could simply use the `args` passed to the function, without any condition.

---

For the future we should reduce the amount of Templates to one, which can handle everything and allows to pass custom content to the dialog. 🎉

packages/styles/src/components/dialog.scss
```diff
@@ -83,6 +86,50 @@ dialog {
     bottom: 2rem;
   }
 
+  &#{$dialog-bottom-sheet-selector} {
+    --_dialog-sheet-default-width: 100%;
```
oliverschuerch, line 90: It's currently not written anywhere, but we do prefix every external facing piece of code (e.g. css variables, etc.) with `--post-`.

We also don't use the BEM or any similar naming format, therefore, please only use letters, numbers and the basic dash `-` for custom-property names.

```suggestion
    --post-dialog-default-width: 100%;
```

packages/styles/src/components/dialog.scss
```diff
@@ -83,6 +86,50 @@ dialog {
...
+    > .dialog-grid {
+      max-height: 80dvh;
+    }
+
+    > .dialog-grid > post-closebutton {
+      inset-block-start: tokens.get('utility-gap-16');
+      inset-inline-end: tokens.get('utility-gap-16');
+      transform: none;
+      z-index: 2;
+    }
```
oliverschuerch, line 130: Our components currently add the close button in the top-right corner, half outside, half inside of the component:

Image

I know this is a problem for the mobile variant of the dialog bottom-sheet implementation (100% width), so I would like to ask you to clearify this topic with your UX designer.

Let me know if there is none, then I'll ask our designers how to deal with that.

packages/styles/src/components/dialog.scss
```diff
@@ -255,10 +317,20 @@ dialog:not([data-animation='none']) {
...
+      }
+    }
+  }
+
+  &#{$dialog-bottom-sheet-selector} {
+    @media (prefers-reduced-motion: reduce) {
+      transition-duration: 0s;
+
+      &::backdrop {
+        transition-duration: 0s;
```
oliverschuerch:
```suggestion
      transition: none;

      &::backdrop {
        transition: none;
```

packages/styles/src/components/dialog.scss
oliverschuerch, line 1: Make sure ALL the changes in the styles ONLY affect the `bottom-sheet` variant!

packages/styles/src/components/dialog.scss
```diff
@@ -184,8 +241,10 @@ dialog {
 // deprecated
 dialog:not([data-animation='none']) {
   --_dialog-transition-duration: var(--dialog-transition-duration, 0.5s);
+  --_dialog-closed-scale: 0.8;
+  --_dialog-closed-translate-y: 0;
```
oliverschuerch, line 245: If you did not plan to make it easy for DEVs to override these properties, you should not use custom-properties here.

**Option 1**
Switch to SCSS variables.

**Option 2**
Inline the values (e.g. `transform: translateY(0) scale(0.8)`).

packages/styles/src/components/dialog.scss
```diff
@@ -243,8 +302,11 @@ dialog:not([data-animation='none']) {
     }
   }
 
-  &[data-animation='slide-in'] {
-    transform: translateY(calc(var(--_dialog-slide-in-offset)));
+  &[data-animation='slide-in'],
+  &#{$dialog-bottom-sheet-selector}:not([data-animation]) {
+    --_dialog-slide-in-offset: 3rem;
+    --_dialog-closed-scale: 1;
+    --_dialog-closed-translate-y: var(--_dialog-slide-in-offset);
```
oliverschuerch, line 309: Same here.

oliverschuerch (changes requested):

apps/documentation/src/stories/components/dialog/dialog.docs.mdx
JoaoSequeira489: Thanks for pointing to an example 😄  Done let me know if it's ok now

packages/styles/src/components/dialog.scss
JoaoSequeira489: It is not meant to shorten the selector. It centralizes the bottom-sheet variant, which is reused in multiple composed selectors. I’d keep it to avoid having a situation where we inline it everywhere. 

E.g. A future rename from data-variant='bottom-sheet' to another attribute/class would require touching several places and makes it easier to miss one.

apps/documentation/src/stories/components/dialog/dialog.snapshot.stories.ts
JoaoSequeira489: Thanks for this 😄 I went with the easy solution but mostly too keep it consistent with what was already there for the Default snapshot story. Let me know if it's ok like this

image

apps/documentation/src/stories/components/dialog/dialog.stories.ts
JoaoSequeira489: Done

apps/documentation/src/stories/components/dialog/dialog.stories.ts
JoaoSequeira489: Thanks for the reference 😄 Done !

apps/documentation/src/stories/components/dialog/dialog.stories.ts
JoaoSequeira489: I went with a smaller simpler approach to avoid doing too many refactors that could touch the other stories too much. Have a look and let me know your thoughts

apps/documentation/src/stories/components/dialog/dialog.stories.ts
JoaoSequeira489: Done.

I agree and the main story is currently able to handle everything. I added this one.because there were some extra use-cases in the specification which might be worth documenting in storybook.

packages/styles/src/components/dialog.scss
JoaoSequeira489: In this case it wouldn't make sense to rename because it's not meant to be external. This value is an internal fallback behind the documented public override **--post-dialog-width**. 

The notation used was not meant to be BEM but something that was already used in this file. I used **--_dialog-*** to match the existing internal variables in this file which were already on *main* branch, such as:
-  **--_dialog-transition-duration** 
-  **--_dialog-slide-in-offset**. 

Is this not the notation for internal variables? I'm a bit torn on what the repo convention is here 😅

packages/styles/src/components/dialog.scss
JoaoSequeira489: Yeah we talked about it with Christian Zundel and I believe he already spoke to the DS Designer about this variant and designs. 

Here is the designs:
[Figma](https://www.figma.com/design/6gsh0EC0itp2amxwU5m5K1/Nebula---UI-Kit?node-id=887-20955)

packages/styles/src/components/dialog.scss
JoaoSequeira489: Thanks for that! Done!

packages/styles/src/components/dialog.scss
JoaoSequeira489: This is related to the above notation problem. 
I'd avoid inlining values that should be consistent across several different classes.

Switching to SCSS variables could be a good approach but I'd avoid having both **--_dialog-*** and SCSS variables for internals.

packages/styles/src/components/dialog.scss
JoaoSequeira489: Same as above

packages/styles/src/components/dialog.scss
JoaoSequeira489: I tested both variants before commiting, seems fine 😄 . Did you find anything that could impact the regular version?

apps/documentation/src/stories/components/dialog/dialog.snapshot.stories.ts
oliverschuerch: It still renders in the with `position: absolute` in the preview.
image

apps/documentation/src/stories/components/dialog/dialog.stories.ts
oliverschuerch: Did you push your changes? The preview link is still showing `type` before `variant`.
image

apps/documentation/src/stories/components/dialog/dialog.stories.ts
oliverschuerch: Same here. The control is still visible always.

packages/styles/src/components/dialog.scss
oliverschuerch: I had the feeling that there could be something bleeding to the `default` version. But it's sometimes hard to understand the whole file with all the changes inbetween 😆

You tested it, so I'm fine with it.

packages/styles/src/components/dialog.scss
oliverschuerch: There are no repo conventions about that specific case, since there is no such thing as internal custom properties in pure CSS components.

Anyway, I get your point and I'll accept your solution, since you only followed what was there and ensured consistency (which I like 😆).

oliverschuerch (approved):

PR #7548, opened 2026-04-24
fix(components): fix esm directory import for consuming packages

oliverschuerch:
## 📄 Description

In Angular 22 (~ May 2026) Jasmine/Karma support will be removed. Projects need to migrate to Vitest instead.

We got a support request, that the migration is not working, because of an error caused by `@swisspost/design-system-components-angular`, which tries to import `@swisspost/design-system-components/loader`, without specifying a valid entry file.

image

While this seems to be no issue for code that runs through build tools, vitest does not resolve these paths to the underlaying index (cjs|mjs|js) file, but instead follows strictly the import behavior of node.

To fix this, I have added a proper export definition in the `package.json` file (similar to the one in https://github.com/swisspost/design-system/labels/v10) of our components package. This allows our `@swisspost/design-system-components-angular` package to import `@swisspost/design-system-components/loader`, while getting the desired entry file for its setup.

---

## 📝 Checklist


myrta2302 (approved):

packages/components/package.json
```diff
@@ -23,6 +23,27 @@
...
+    },
+    "./dist/*": {
+      "default": "./dist/*"
+    },
+    "./loader": {
+      "types": "./loader/index.d.ts",
+      "import": "./loader/index.js",
+      "require": "./loader/index.cjs.js"
+    },
+    "./loaders": {
```
myrta2302, line 41:
```suggestion
    "./loaders/*": {
      "default": "./loaders/*",
    "./loaders": {
```

myrta2302 (changes requested):

myrta2302 (approved):

PR #7594, opened 2026-05-05
fix(migration): prevent chain collision in automigration rules by marking conflicting classes as `manual-only`

### LINKED ISSUE #7493: [bug]: Auto-migration issue with `-rg-` breakpoint classes
### 🐞 Describe the bug

When running the automigration scripts, there is an issue with the `-rg-` breakpoint classes, as they are being renamed to `-xs-` instead of `-sm-`. Investigate on why that's happening, as the tests in the eslint package are passing correctly.

### 🛠️ To Reproduce

1. Use an existing project using `v9` of the Design System or create one
2. Add some classes with the `-rg-` breakpoint in it (e.g.: `col-rg-6`)
3. Run the automigration rules as described in the docs on the project

### ✨ Expected behavior

The classes should be renamed as followed for the breakpoint:
- `*-sm-*` -> `*-xs-*` ✅
- `*-rg-*` -> `*-sm-*` ❌ -> currently changes to `*-xs-*`
- `*-xxl-*` -> `*-xl-*` ✅

alionazherdetska:
## 📄 Description

This PR fixes the chain collision bug in the automigration rules by removing the autofix for classes whose renamed output would immediately be picked up by another migration rule and renamed again to the wrong final value.
 
### Affected rules
 
| Rule | Conflicting values |
|---|---|
| `no-deprecated-breakpoints` | `*-rg-*` — chains into `sm → xs` |
| `no-deprecated-spacing-utilities` | `*-1`, `*-4`, `*-hair`, `*-micro` |
| `no-deprecated-gutter-utilities` | `*-1`, `*-4` |
| `no-deprecated-gap-utilities` | `*-1`, `*-4` |
 
### Changes
 
- Extended the mutation tuple type from `[string, string]` to `[string, string, boolean?]`:  when the third element is `true`, the rule reports the violation but provides no autofix
- Added `conflictingOldValues` parameter to `setUpClassesMutations` to mark specific values as manual-only
- Updated `generatedDataTester` to handle manual-only tuples correctly in tests
- Updated docs and migration guide to document which classes require manual migration

## 🚀 Demo

[Migration page
](https://preview-7594--swisspost-design-system-next.netlify.app/?path=/docs/c23b1d0b-76b3-4e38-aa76-b10c29bb873f--docs&devModeEnabled=true)
---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -633,16 +633,28 @@ export class MigrationV910Component extends LitElement {
                         Breakpoints updated
                         <ul>
                           <li>
-                            All classes containing <code>*-rg-*</code> are no longer effective
+                            All classes containing <code>*-sm-*</code> are renamed to
```
myrta2302: Should this also be _manual only_? Probably not very possible, but users could do the manual changes first and the automigration after. It would be safer to completly exclude all classes involved in conflicting pairs.

myrta2302 (approved):

apps/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -633,16 +633,32 @@ export class MigrationV910Component extends LitElement {
                         Breakpoints updated
                         <ul>
                           <li>
-                            All classes containing <code>*-rg-*</code> are no longer effective
+                            All classes containing <code>*-xxl-*</code> are renamed to
```
leagrdv, line 636: If people check the "Hide changes covered by the automatic 🪄 migration rules" checkbox, it will hide everything. I think the ones that need manual migration should still be there and their `migration rules` tag should be switched to a different tag (.e.g. `partial migration rules`)

packages/eslint/docs/rules/html/migrations/no-deprecated-breakpoints.md
```diff
@@ -3,7 +3,18 @@
 Flags all deprecated breakpoint size classes and replaces with existing ones.
 
 - Type: problem
-- 🔧 Supports autofix (--fix)
+- 🔧 Supports autofix (--fix) — **except for `rg` and `sm` breakpoint classes (see note below)**
+
+## ⚠️ Manual migration required for `-rg-` and `-sm-` classes
```
leagrdv, line 8: If we told user to do the automigration first, and then do the manual ones, we could keep the `sm` -> `xs` one as it would not create any problems. Then users would simply have to manually update the `rg` to `sm` ones. What do you think?

packages/eslint/docs/rules/html/migrations/no-deprecated-gap-utilities.md
```diff
@@ -3,7 +3,17 @@
 Flags all deprecated bootstrap gap utility classes and replaces them with the new ones.
 
 - Type: problem
-- 🔧 Supports autofix (--fix)
+- 🔧 Supports autofix (--fix) — **except for classes using values `1` and `4` (see note below)**
+
+## ⚠️ Manual migration required for `*-1` and `*-4` classes
```
leagrdv, line 8: Same thing here, we could do the `4` to `16` in the automigration, and then users would do the `1` to `4` manually

leagrdv (changes requested):

packages/eslint/docs/rules/html/migrations/no-deprecated-gap-utilities.md
leagrdv, line 1: (Ignore this comment, I'm just trying to submit the review)

leagrdv (changes requested):

packages/eslint/docs/rules/html/migrations/no-deprecated-gap-utilities.md
alionazherdetska: Same as for the prev comment:)

packages/eslint/docs/rules/html/migrations/no-deprecated-breakpoints.md
alionazherdetska: I thought about this, but wouldn't it break if the user does the manual rename first and then runs the automigration? rg → sm would immediately get picked up by sm → xs and end up as xs. Keeping both manual feels safer since there's no guaranteed order, what do you think?

leagrdv: Yea I agree .. we're not too sure in which order they're gonna do it. I'm resolving the comment

leagrdv (approved):

PR #7600, opened 2026-05-06
chore(components): turn property validators into decorators

alizedebray:
## 📄 Description

This PR creates decorators to check Stencil component properties. 

They're meant to be used on to of the `@Prop()` decorator, and perform checks on `componentDidLoad` and every time the value changes, without the need to manually watch the property:

```ts
// BEFORE
@Prop({ reflect: true }) readonly headingLevel!: HeadingLevel;

@Watch('headingLevel')
validateHeadingLevel() {
  checkRequiredAndOneOf(this, 'headingLevel', HEADING_LEVELS);
}

componentDidLoad() {
  this.validateHeadingLevel();
}

// AFTER
@Required()
@OneOf(HEADING_LEVELS)
@Prop({ reflect: true }) readonly headingLevel!: HeadingLevel;
```

> [!IMPORTANT]  
> This PR only creates the decorators, a second PR will follow to actually use them in the components and remove the checker functions altogether.

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


oliverschuerch (changes requested):
There is one propertyChecker missing we have in the current implementation (`empty-or`), is this intentional?

alizedebray:
> There is one propertyChecker missing we have in the current implementation (`empty-or`), is this intentional?

I've added the "empty or" check to all the validator as default behavior so: 
- if you allow empty value you can use the decorator alone (ex: `@Type('string')` allows strings but also null or undefined)
- if you don't want to allow empty value you can combine decorators (ex: `@Required() @Type('string')` only allows non-empty strings)

packages/components/src/utils/property-decorators/create-validator-decorator.ts
```diff
@@ -0,0 +1,153 @@
...
+/** Returns or creates a value in a WeakMap for a given key. */
+function getOrCreate<K extends object, T>(map: WeakMap<K, T>, key: K, factory: () => T): T {
+  let value = map.get(key);
+  if (!value) {
+    value = factory();
+    map.set(key, value);
+  }
+
+  return value;
+}
```
oliverschuerch: Since this is the hearth of your feature, I'd like to ask you to add extended JSDoc comments per function, so it becomes easier for others to understand the logic.

**Example**

```suggestion
/**
 * Returns or creates a value in a WeakMap for a given key.
 *
 * @param map - The WeakMap to look up.
 * @param key - The object key to retrieve or initialize.
 * @param factory - A factory function called to create the value if it doesn't exist.
 * @returns The existing or newly created value.
 */
function getOrCreate<K extends object, T>(map: WeakMap<K, T>, key: K, factory: () => T): T {
  let value = map.get(key);
  if (!value) {
    value = factory();
    map.set(key, value);
  }

  return value;
}
```

oliverschuerch:
> > There is one propertyChecker missing we have in the current implementation (`empty-or`), is this intentional?
> 
> I've added the "empty or" check to all the validator as default behavior so:
> 
> * if you allow empty value you can use the decorator alone (ex: `@Type('string')` allows strings but also null or undefined)
> * if you don't want to allow empty value you can combine decorators (ex: `@Required() @Type('string')` only allows non-empty strings)

Can we document that somewhere in a specific code file or so?

oliverschuerch (changes requested):

packages/components/src/utils/property-decorators/create-validator-decorator.ts
alizedebray: Done. By the way, it would be good if you could add JSDocs on the code you wrote for the icons package for the same reasons.

oliverschuerch (approved):

packages/components/src/utils/property-decorators/create-validator-decorator.ts
oliverschuerch: The code written there is 1/3 from @gfellerph, 1/3 from me and 1/3 from @alionazherdetska. But you're right, we should add in-code documentation there as well.

I've created a ticket: #7658

PR #7608, opened 2026-05-08
feat(components): implement tabs scroll buttons functionality

### LINKED ISSUE #6768: [web component styles] content tabs
### 📃 Tasks

- [ ] Complete documentation in Storybook and make it accessible to all users
- [ ] Design review

myrta2302:
## 📄 Description

The former component design handled overflow with `overflow-x: auto`. The new tabs design integrates **scroll buttons** for scroll navigation. Scroll button visibility is managed reactively based on scroll position, container size changes and DOM mutations.

This PR also includes the renaming of panel/navigation mode to Content / Page Tabs in code and documentation.

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -9,6 +9,15 @@
...
 
+.tabs {
+  // Hide scrollbar across browsers while keeping scroll functionality
+  scrollbar-width: none; // Firefox
+
+  &::-webkit-scrollbar {
+    display: none; // Chrome, Safari, Edge
+  }
+}
+
```
oliverschuerch: Should we nest that similar to the structure below? What do you think?
```scss
:host {
  .tabs-wrapper {
    .tabs {
      ...
    }
  }
}
```

packages/components/src/components/post-tabs/post-tabs.tsx
```diff
@@ -15,11 +15,19 @@ import { version } from '@root/package.json';
...
 import { componentOnReady, checkRequiredAndType } from '@/utils';
 
+// Extends the HTMLButtonElement interface to include ariaControlsElements, which is part of the
+// ARIA reflection API but not yet present in TypeScript's built-in DOM type definitions.
+declare global {
+  interface HTMLButtonElement {
+    ariaControlsElements: Element[];
+  }
+}
+
```
oliverschuerch: Please move this to a specific file (e.g. `packages/components/types/dom-aria.d.ts`), so we don't have to add it again for any other element in other components.

Also it would be better to extend the Element interface if the attribute can be used on other elements too.

```ts
declare global {
  interface Element {
    // optional, since not all environments implement it yet
    ariaControlsElements?: Element[];
  }
}
```

Add a comment that states something like: `This can be removed as soon as the installed typescript version adds the declaration by default`.

packages/components/src/components/post-tabs/post-tabs.tsx
```diff
@@ -120,41 +145,53 @@ export class PostTabs {
...
-      attributeFilter: ['data-navigation-mode', 'aria-current'],
+      attributeFilter: ['data-pages-variant', 'aria-current'],
     };
 
     this.contentObserver = new MutationObserver(this.handleContentChange.bind(this));
     this.contentObserver.observe(this.host, config);
   }
 
+  private setupResizeObserver() {
+    this.resizeObserver = new ResizeObserver(this.updateScrollButtons);
```
oliverschuerch, line 183: Without any guard, this will probably create a hydration error.

```ts
this.resizeObserver = globalThis.ResizeObserver ? new ResizeObserver(...) : null;
this.resizeObserver?.observe(this.tabsContainer);
```

Also you would need to add the question mark to every usage of `this.resizeObserver` in the component (as shown on the second line above.

If you checked it already and it is not throwing, it's fine.

packages/components/src/components/post-tabs/post-tabs.tsx
```diff
@@ -412,8 +483,24 @@ export class PostTabs {
               }}
             />
           </TabsContainer>
+          <button
+            ref={el => (this.rightScrollButton = el!)}
+            class="scroll-btn scroll-btn-right"
+            type="button"
+            aria-label="Scroll button right"
```
oliverschuerch: Don't we need to get the text for this label through a text-property on the `post-tabs` component?
Same for the other buttons label.

packages/components/src/components/post-tabs/post-tabs.tsx
```diff
@@ -412,8 +483,24 @@ export class PostTabs {
...
+            tabindex={this.showRightScrollButton ? 0 : -1}
+            hidden={!this.showRightScrollButton}
+            onClick={() =>
+              this.tabsContainer.scrollBy({
+                left: this.tabsContainer.clientWidth,
+                behavior: 'smooth',
+              })
```
oliverschuerch: Please move this inline clickHandler to a private function in the component class.
Same for the other buttons clickHandler.

packages/components/src/components/post-tabs/post-tabs.tsx
oliverschuerch, line 1: The buttons are visible and causing layout shifts. Is this wanted?

Image
 
Image

Tested in Chrome 148 on Mac OS Tahoe 26.4.1

oliverschuerch (changes requested):

oliverschuerch (commented):
Add a changeset.

packages/components/src/components/post-tabs/post-tabs.tsx
myrta2302: They are completely unstyled. Styles are handled by a different ticket.

packages/components/src/components/post-tabs/post-tabs.tsx
myrta2302: I tested again and i get no hydration errors. Probably because its called at componentDidLoad().

packages/components/src/components/post-tabs/post-tabs.tsx
oliverschuerch: In general, that's a bad workflow, because it means, we could possibly release something in a broken state. But ok...

oliverschuerch (approved):

oliverschuerch (approved):

myrtasakellariou:
There were some new e2e errors, and i did some changes. So i had to re-request review. Thank you!

packages/components/src/components/post-tabs/post-tabs.tsx
```diff
@@ -29,12 +38,22 @@ import {
 })
 export class PostTabs {
   private currentActiveTab: HTMLPostTabItemElement;
-  private showing: Animation;
-  private hiding: Animation;
+  private showing: Animation | null;
+  private hiding: Animation | null;
```
oliverschuerch: This makes sense, if you immediately initiallize the class members with `null`.
Your change would work properly, if you...

- ...either initialize the class members in the constructor.
  e.g.
  ```ts
  constructor() {
    this.showing = null;
    this.hiding = null;
  }
  ```
- ...or you add type `undefined`. Because that's what both class members are if you don't initialize them.
  ```ts
  private showing: Animation | undefined;
  private hiding: Animation | undefined;
  ```

Also I don't know which IDE you're using, but my VS-Code (even without any extension), tells me there is an typescript issue with this implementation.

Image

---

The same is actually true for all class members (e.g. `currentActiveTab`, etc.).
For now I would focus on fixing the once you have changed/added with this pr.

packages/components/src/components/post-tabs/post-tabs.tsx
```diff
@@ -29,12 +38,22 @@ import {
...
   private contentObserver: MutationObserver;
+  private resizeObserver: ResizeObserver;
+  private tabsContainer: HTMLElement;
+
+  @State() isPagesVariant: boolean = false;
+
+  @Element() host: HTMLPostTabsElement;
+
+  @State() private showLeftScrollButton = false;
+  @State() private showRightScrollButton = false;
```
oliverschuerch: It would be better readable (imo), if you would group similar things (e.g. private class members, states, etc.) together.

oliverschuerch (changes requested):

packages/components/src/components/post-tabs/post-tabs.tsx
oliverschuerch: What about the first two sentences?

1. Moving the declaration to a file under `components/types/*`, allows typescript to use it for all our components.
2. Switching from `HTMLButtonElement` to the `Element` interface, seems to be technically correct, since the first sentence on the [Element: ariaControlsElements property](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaControlsElements) docs says the following:
   
   > The ariaControlsElements property of the [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) interface is an array containing the elements that are controlled by the element it is applied to.

   It means, the `aria-controls` attribute can be used on any HTML element, not only specifically on a `<button>`.

oliverschuerch (changes requested):

packages/components/src/components/post-tabs/post-tabs.tsx
myrta2302: I actually did it already, but had forgot to remove it from here.

oliverschuerch (approved):

PR #7612, opened 2026-05-08
chore(components): web component styles tabs

### LINKED ISSUE #6768: [web component styles] content tabs

myrta2302:
## 📄 Description

Implement Tabs styles according to the new [Design](https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-Next-Level?node-id=19294-91721&m=dev).

Demo

https://preview-7612--swisspost-design-system-next.netlify.app/?devModeEnabled=true

Variant Snapshots
-------------
Content Tabs
https://preview-7612--swisspost-design-system-next.netlify.app/iframe.html?id=snapshots--content-tabs&viewMode=story

Page Tabs
https://preview-7612--swisspost-design-system-next.netlify.app/iframe.html?id=snapshots--page-tabs&viewMode=story

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


myrtasakellariou:
On Hold until Design is updated.

apps/documentation/src/stories/components/tabs/tabs.stories.ts
```diff
@@ -175,22 +175,31 @@ function renderPagesVariant(
...
+    <post-tabs
+      full-width="${fullWidth ? true : nothing}"
+      label="${ifDefined(label)}"
+      size="${ifDefined(size)}"
+    >
       <post-tab-item name="first">
         <a href="/first" aria-current="page">First page</a>
       </post-tab-item>
-      <post-tab-item name="second">
+      <post-tab-item name="second" disabled>
```
alionazherdetska: The default story now has a disabled tab item but there's no docs or explanation around it anywhere: what it does, how to use it, and especially the accessibility concerns. Would be nice to have something about it, similar to what the button docs do:

Image

apps/documentation/src/stories/components/tabs/tabs.stories.ts
alionazherdetska, line 1: The controls table has two `Props` sections: **size**, **text-next-tab-items** and **text-prev-tab-items** ended up in a separate group:

Image

alionazherdetska, line 1: Just a suggestion: it would be nice to document `text-next-tab-items` and `text-prev-tab-items` the same way as the accessible label props in other components:

Image

Image

Image

As you can see `default value` is shown in the `control input` rather than in the `default` column.

apps/documentation/src/stories/components/tabs/tabs.stories.ts
alionazherdetska: The disabled tab is reachable via keyboard which is correct: according to the [W3C APG keyboard interface guide](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/), tab elements should remain focusable when disabled. But when you land on it, the screen reader announces it as a regular tab with no indication it's disabled. Worth checking if aria-disabled="true" should be set on it so the state is communicated properly:

Image

https://github.com/user-attachments/assets/9f6dfca8-d9b4-4645-a73d-b9a37267ace9

packages/styles/src/components/tabs/_tab-title.scss
```diff
@@ -1,267 +1,438 @@
...
-    color: LinkText;
-
-    &:hover,
-    &:focus,
-    &:focus-within,
-    &:focus-visible {
-      outline: tokens.get('utility-gap-2') solid Highlight;
-      outline-offset: calc(tokens.get('utility-gap-4') * -1);
+  // hovered
+  &:hover:not([disabled]) {
```
alionazherdetska, line 71: When hovering over the Page Tabs, the cursor stays as the default arrow instead of changing to a pointer:

Image

packages/styles/src/components/tabs/_tab-title.scss
```diff
@@ -1,267 +1,438 @@
...
+        in srgb,
+        tokens.get('post-tabs-content-tab-selected-border-color') calc(var(--active, 0) * 100%),
+        transparent
+      );
+    }
+    font-weight: calc(400 + var(--active, 0) * 300); // TOKEN HERE ALSO
     z-index: calc(var(--active, 0) * 1);
-    border-right-color: color-mix(
+
+    @include utilities-mx.high-contrast-mode() {
```
alionazherdetska, line 349: In HCM  the active tab item text in the Content Tabs variant doesn't have enough contrast:

Image

packages/styles/src/components/tabs/_tab-title.scss
```diff
@@ -1,267 +1,438 @@
...
+  // Content Tabs: Style active post-tab-item for SSR if there is a [active-tab] set
   post-tab-item.tab-title.ssr {
+    &::after {
+      background-color: color-mix(
+        in srgb,
+        tokens.get('post-tabs-content-tab-selected-border-color') calc(var(--active, 0) * 100%),
+        transparent
+      );
+    }
+    font-weight: calc(400 + var(--active, 0) * 300); // TOKEN HERE ALSO
```
alionazherdetska: Looks like a leftover comment, was this meant to be a token?

packages/styles/src/components/tabs/_tab-title.scss
```diff
@@ -1,267 +1,438 @@
...
+      }
+
+      // focused
+      &:focus,
+      &:focus-visible,
+      &:focus-within {
+        padding-block: tokens.get('post-tabs-page-small-tab-focused-padding-block');
+        height: tokens.get('post-tabs-page-small-tab-focused-height');
+      }
+    }
```
alionazherdetska, line 323: Same focus styles appear in both large and small variants, maybe worth a mixin?

packages/styles/src/components/tabs/_tab-title.scss
```diff
@@ -1,267 +1,438 @@
...
+    padding-block: calc(
+      tokens.get('post-tabs-page-large-tab-selected-padding-block') + var(--active, 0) *
+        (
+          tokens.get('post-tabs-page-large-tab-enabled-padding-block') - tokens.get(
+              'post-tabs-page-large-tab-selected-padding-block'
+            )
+        )
     );
-    color: #{color.$black};
     font-weight: calc(400 + var(--active, 0) * 300);
```
alionazherdetska: Same calculation as above, could this be a shared variable or token?

packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -1,33 +1,228 @@
...
+// Content Tabs variant
+:host(.content-tabs) {
+  .tabs-wrapper {
+    button.scroll-btn {
+      flex-shrink: 0;
+      color: tokens.get('post-tabs-content-scroll-button-enabled-fg');
+
+      // hovered
+      &:hover {
+        border-color: tokens.get('post-tabs-page-scroll-button-hovered-border-color') !important;
```
alionazherdetska, line 179: This is inside :host(.content-tabs) but uses a page tabs token.

packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -1,33 +1,228 @@
...
-    //Internal padding to align the tabs content to the container
     .tabs {
       padding-inline: var(--post-container-offset);
     }
   }
 }
+
+:host(.page-tabs[full-width]) {
+  .tabs-wrapper {
+    margin-inline: calc(-1 * (var(--post-container-offset) + 40px));
```
alionazherdetska: Magic number 40px, should this be a token or variable?

packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -1,33 +1,228 @@
...
+  }
+}
+
+:host(.page-tabs[size='large']) button.scroll-btn {
+  top: tokens.get('post-tabs-large-scroll-button-padding-block');
+}
+
+:host(.page-tabs[size='small']) button.scroll-btn {
+  top: tokens.get('post-tabs-small-scroll-button-padding-block');
+}
```
alionazherdetska: These two size rules sit outside the main `:host(.page-tabs)` block while all other page tabs styles are inside it, maybe you cna move it inside.

alionazherdetska (changes requested):
Focus indicator is missing on the aria-current="page" tab for the `dark-color-scheme` all other tabs show an outline when focused, but the active one doesn't:

https://github.com/user-attachments/assets/fc8748e8-08e7-43e0-a2d4-1bf757c31492

packages/components/src/components/post-tabs/post-tabs.scss
myrta2302: Not sure how these could be moved inside, because the sizes live on the host and must go directly into the `:host()` parenthesis.

packages/styles/src/components/tabs/_tab-title.scss
myrta2302: removed

apps/documentation/src/stories/components/tabs/tabs.stories.ts
myrta2302: Created a ticket to add it later.

myrtasakellariou:
> Focus indicator is missing on the aria-current="page" tab for the `dark-color-scheme` all other tabs show an outline when focused, but the active one doesn't:
> 
>  Screen.Recording.2026-06-22.164724.mp4

The correct color was set. But there is a visual issue (of the border mixing with the focus) to be discussed further with UX. Possible solution would be to create a token for the offset that depends on the palettes. 
image

packages/styles/src/components/tabs/_tab-title.scss
myrta2302: I see this: 
image
Should be highlightText

packages/components/src/components/post-tabs/post-tabs.scss
myrta2302: This need a new token creation. Possibly along with the focus token.

apps/documentation/src/stories/components/tabs/tabs.docs.mdx
alionazherdetska, line 13: Image 
The Figma link still points to the old design page

apps/documentation/src/stories/components/tabs/tabs.stories.ts
```diff
@@ -42,47 +42,15 @@ const meta: MetaComponent<
...
-        'The name of the tab that is initially active. If not specified, it defaults to the first tab.\n\n**Changing this value after initialization has no effect.**',
-      control: 'select',
-      options: ['first', 'second', 'third'],
-      if: { arg: 'variant', eq: 'Content Tabs' },
-      table: {
-        category: 'Props',
-        type: { summary: 'string' },
-      },
-    },
     'activeTab': {
```
alionazherdetska, line 44: activeTab prop shows now for both `Content` and `navigation` variants, although it is only relevant for the `Content` variant:

Image

apps/documentation/src/stories/components/tabs/tabs.stories.ts
alionazherdetska, line 1: The red asterisk indicating a required prop is missing from all required props:

Image

packages/styles/src/components/tabs/_tab-title.scss
alionazherdetska: From what I can see tin the following examples:
1. https://carbondesignsystem.com/components/tabs/code/
2. https://codepen.io/pen

The cursor is still a `hand-pointer` even when hovering over the `post-tab-item` which is open/active at the moment.

packages/styles/src/components/tabs/_tab-title.scss
alionazherdetska: I use the Contrast Themes for Windows:
image
---
### Aquatic theme 

image
image


---
### Desert theme
image
image

---
### Dusk theme
image
image

packages/styles/src/components/tabs/_tab-title.scss
alionazherdetska: This one is not addressed yet.☺️

packages/components/src/components/post-tabs/post-tabs.scss
alionazherdetska: This one is not addressed yet.☺️

packages/components/src/components/post-tabs/post-tabs.scss
alionazherdetska: Got it, makes sense.

packages/styles/src/components/tabs/_tab-title.scss
```diff
@@ -1,267 +1,450 @@
...
+  cursor: default;
+}
 
-  @include utilities-mx.high-contrast-mode() {
-    border-left-color: ButtonText;
-    border-right-color: ButtonText;
-    color: Highlight;
+// Large variant (default)
+post-tabs post-tab-item,
+post-tabs[size='large'] post-tab-item {
```
alionazherdetska: `post-tabs post-tab-item` matches all tab items regardless of size, so the large styles also apply to small tabs and get overridden instead of never being set. Maybe you should scope it to `post-tabs[size='large'] post-tab-item` only, like the small variant does.

alionazherdetska (changes requested):

apps/documentation/src/stories/components/tabs/tabs.docs.mdx
myrta2302: It seems to still be in the Work in Progress.

packages/styles/src/components/tabs/_tab-title.scss
myrta2302: ok. i m removing the specific style.

packages/components/src/components/post-tabs/post-tabs.scss
myrta2302: Weird, i had already updated with this: ` margin-inline: calc(-1 * var(--post-container-offset));` Maybe i was missing the last push..

packages/styles/src/components/tabs/_tab-title.scss
myrta2302: I felt this is a bit of overkill because its just two lines of code. But if you disagree i can sure implement a mixin.

packages/styles/src/components/tabs/_tab-title.scss
myrta2302: Correct, i updated to `post-tabs:not([size='small']) post-tab-item`.

apps/documentation/src/stories/components/tabs/tabs.docs.mdx
alionazherdetska: Maybe then it is better to confirm with Sandra, because it feels wrong the both ways: leaving the old link since it looks completely different or linking them to the page `In Progress`.

packages/styles/src/components/tabs/_tab-title.scss
alionazherdetska: Good for me. Just thought you did not see it.

apps/documentation/src/stories/components/tabs/tabs.stories.ts
alionazherdetska, line 1: Size is also a required prop, but there is no asterisk:

Image

Image

apps/documentation/src/stories/components/tabs/tabs.stories.ts
```diff
@@ -94,9 +62,9 @@ const meta: MetaComponent<
       },
     },
     'post-tabs-content': {
-      name: 'post-tabs-content ', // trailing space is intentional to avoid conflict with auto-generated part
-      description: 'The container element that displays the content of the currently active tab.',
+      name: 'post-tabs-content ',
```
alionazherdetska: Can you add the comment back so it is clear for the other developers, who may work on this file later on: 

Image

apps/documentation/src/stories/components/tabs/tabs.stories.ts
```diff
@@ -94,9 +62,9 @@ const meta: MetaComponent<
...
       control: false,
+      description: 'The container element that displays the content of the currently active tab.',
```
alionazherdetska, line 67: same here

packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -1,15 +1,130 @@
...
 :host {
   display: block;
 
   .tabs-wrapper {
+    position: relative;
+    display: flex;
+    align-items: center;
+
+    // Scroll buttons
+    button.scroll-btn {
```
alionazherdetska, line 27: Can you please also change the color for the scroll button in the HCM (dark scheme), since the contrast is not high enough (you can only see it after you hovered once over the scroll button, inititally the color is correct):

https://github.com/user-attachments/assets/e487c806-e6c8-4cee-8566-c89a80468050

packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -1,15 +1,130 @@
...
+      }
+
+      &[hidden] {
+        display: none;
+      }
+
+      @include utilities-mx.focus-style() {
+        outline-offset: $tabs-outline-offset !important;
+
+        @include utilities-mx.high-contrast-mode() {
```
alionazherdetska, line 59: The scroll button is missing the focus indicator in dark mode when navigating with the keyboard (only for the Page Tabs):

https://github.com/user-attachments/assets/c47621b5-f8f0-424d-9673-8babb74f8835

apps/documentation/src/stories/components/tabs/tabs.stories.ts
alionazherdetska, line 1: Maybe we can add the separate story entries or the toggle in the controls for the tabs with only icons, sicne now we have such variation in Figma (just an idea):

Image

apps/documentation/src/stories/components/tabs/tabs.snapshot.stories.ts
alionazherdetska, line 1: The 5 rows per palette look identical except for subtle width differences. Maybe we could reduce this to just 2 combinations (default width and full-width for each size) to keep the snapshot readable. What do you think:

Image

alionazherdetska (changes requested):

apps/documentation/src/stories/components/tabs/tabs.stories.ts
myrta2302: It should not be required as it defaults to large, i will remove this.

apps/documentation/src/stories/components/tabs/tabs.snapshot.stories.ts
myrta2302: There was a typo that prevented the `active-tab` from showing up in the variants which is now fixed. And also the full-width makes no visual change for the Content Tabs, so i removed it. But for the Page Tabs i made no cahnges because it still iterrates for the active-tab nd the full width for container and container-fluid.

apps/documentation/src/stories/components/tabs/tabs.stories.ts
myrta2302: Sure, But i will add it in separate ticket for Next.

sandra-post:
@myrta2302 

### Design Review Content Tabs

- Tabs large
  - Icon size
    - IS: 20 / 16 / 14
    - SHOULD: 24 / 20 / 16
- In the storybook, there are no controls/toggles to show/hide icons and labels (so that 3 versions are possible: icon only, label only, icon + label)

Rest looks good, thank you.

sandra-post:
@myrta2302 

### Design Review Page Tabs

- In storybook, the controls/toggle for show/hide icon and label (so that there can be 3 versions in the end: icon only, label only, icon + label) --> #7961 
- In storybook, there is the control "Full width" true/false . The tabs should always be full width --> the space between first tab and left edge (navTabs padding left) equals the page layout margins (Desktop LG: 40 px, Tablet SM: 16 px, etc.)
- Tabs LARGE
  - Icon scroll button
    - IS: 20 / 16 / 14
    - SHOULD: 24 / 20 / 16
  - Height (whole tabs compo)
    - IS: 70 / 62 / 54
    - SHOULD: 72 / 64 / 56 (tabs "stacked" on border bottom, so e.g. for desktop: tab height (70) + border bottom (2) = 72)
- Tabs SMALL
  - Height (whole tabs compo)
    - IS: 54 / 54 / 46
    - SHOULD: 56 / 56 / 48 (tabs "stacked" on border bottom, so e.g. for desktop: tab height (54) + border bottom (2) = 56)

packages/components/src/components/post-tabs/post-tabs.tsx
alionazherdetska: The size prop still has the `Required` type decorator

packages/components/src/index.html
alionazherdetska, line 1: Do not forget to clean this file

alionazherdetska (changes requested):

alionazherdetska (approved):

packages/components/src/components/post-tabs/post-tabs.scss
myrta2302: Will be probably fixed by adding a palette focus token.

PR #7664, opened 2026-05-20
chore(components): use property decorators

alizedebray:
## 📄 Description

This PR updates all components to use property decorators instead of property checkers.

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/components/post-stepper/post-stepper.tsx
```diff
@@ -25,43 +25,43 @@ export class PostStepper {
...
    * Use `{number}` as a placeholder — it will be replaced with the current step number at runtime.
    */
-  @Prop({ reflect: true }) textStepNumber!: string;
+  @Prop({ reflect: true })
+  @Required()
+  @Pattern(/\{number\}/)
+  textStepNumber!: string;
 
   @Watch('textStepNumber')
   validateTextStepNumber() {
```
alionazherdetska: Since validation moved to the `@Pattern` decorator, this method only handles the side effect now and no longer does any validation. How about renaming it to `handleTextStepNumberChange` to reflect that? 
`post-language-menu` does the same with `handleVariantChange`.

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -155,75 +148,76 @@ export class PostDatePicker {
...
 
   /**
    * Label for the toggle button that opens the calendar.
    * It is only needed when the calendar is connected to the input.
    */
-  @Prop() textToggleCalendar?: string;
+  @Prop()
+  @Type('string')
+  textToggleCalendar?: string;
   @Watch('textToggleCalendar')
```
alionazherdetska: Since this is the only `text*` prop that kept a manual `@Watch`, maybe we shouldd add a short comment explaining why? Would help other devs understand why it's an exception.

alionazherdetska (changes requested):
## Notes:
**Copilot review:** I requested the Copilot review: the few points it flagged are valid, so I didn't flag them again.

---

**A little out of context for this PR:**

While testing, I noticed that passing a non-string value like `123` to `@IsoDate` produces a slightly misleading error:

```
Received: "123"  ← stringified version instead of the actual value
```
image


What do you think about guarding `typeof value === 'string'` first, like `@Pattern` does? That way `Received` would show the actual value and the message could be clearer too (e.g. `"must be a string in ISO format"`).

See: https://github.com/swisspost/design-system/blob/main/packages/components/src/utils/property-decorators/iso-date.ts

alionazherdetska (approved):

PR #7674, opened 2026-05-22
feat(components): post-closebutton ssr

### LINKED ISSUE #7088: [web component ssr] Close button
### 🧾 Findings

`post-closebutton` does not have Shadow DOM enabled

### 🧪 How to test

**Run the Next.js SSR page**

After updating the component, ensure it is compiled correctly as a React component by running the following commands:

```bash
pnpm components:build
pnpm components-react:build
```

Then, to focus only on the component being tested, isolate it in the Next.js application.
Comment out all components that are not under test and their imports in the following files:

- `packages/nextjs-integration/src/app/ssr/layout.tsx`
- `packages/nextjs-integration/src/app/ssr/page.tsx`

Finally, serve the application with:

```bash
pnpm nextjs:start
```

**Test SSR in the browser**

- Open the application in two different tabs
- In one tab, disable JavaScript and refresh the page
- Compare both pages, they must look the same

### 📃 Tasks

- [ ] Ensure Shadow DOM is enabled (explicit `shadow: true`) for the component and potential sub-components
- [ ] Verify no hydration or console errors are thrown when running the Next.js SSR page
- [ ] Confirm the page looks the same on initial load with JavaScript enabled and disabled across all devices and supported configurations
- [ ] Ensure ARIA attribute cross-referencing (e.g. `aria-labelledby`, `aria-describedby`) works as expected

myrta2302:
## 📄 Description

This PR update the post-closebutton:
1. Enables shadow:true
2. Updates the styles for shadowdom
3. Adds form association so the button can automatically close a parent dialog

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/cypress/support/commands.ts
oliverschuerch, line 1: The whole logic (specifying specific focusable selectors) seems to be very shaky in my opinion. It would be much more secure to check every elements focusability, by setting focus to it (`element.focus()`) and then compare it with the active element (`document.activeElement`).

However, I would at least declare the `focusableSelector` and `focusDisablingSelector` globally and reuse them in both commands (`getFocusableElements` and `getDeepFocusableElements`).

packages/components/cypress/support/commands.ts
```diff
@@ -129,3 +129,53 @@ Cypress.Commands.add(
     return cy.wrap(Array.from(focusableElements));
   },
 );
+
+Cypress.Commands.add(
```
oliverschuerch, line 147: Please add JSDocs comments for both focus commands. Describe at least input/output and the main purpose of each of these functions, so one can easily know the difference between them.

```ts
/**
 * @function {name}
 * {purpose}
 * @param {name} {description}
 * @returns {description}
 */
Cypress.Commands.add(...
```

packages/components/src/components/post-closebutton/post-closebutton.tsx
```diff
@@ -69,17 +75,30 @@ export class PostClosebutton {
     this.validatePlacement();
     this.validateSize();
 
-    if (!this.host.querySelector('.visually-hidden').textContent) {
+    const slot = this.visuallyHidden?.querySelector('slot') as HTMLSlotElement;
```
oliverschuerch: Why is the optional chaining operator needed here?

As far as I understand, if `this.visuallyHidden` is not defined (e.g. when `checkContent` is called before hydration on the client), `slot` will also not be defined, which means `hasContent` will become `false`, causing the error to be  thrown.

The solution to this could be to keep the initialization of the `MutationObserver` in the `componentDidLoad` lifecycle hook.

packages/components/src/components/post-closebutton/post-closebutton.tsx
```diff
@@ -69,17 +75,30 @@ export class PostClosebutton {
...
+      ?.assignedNodes({ flatten: true })
+      .some(node => node.textContent?.trim());
+    if (!hasContent) {
       console.error(`The \`${this.host.localName}\` component requires content for accessibility.`);
     }
   }
 
+  private handleClick() {
+    if (this.buttonType === 'submit') this.internals.form?.requestSubmit();
+    else if (this.buttonType === 'reset') this.internals.form?.reset();
```
oliverschuerch: I have some thoughts:

1. Which user would ever expect a close button to submit a form?
   As an alternative, the close button could call `closeButton.closest('dialog')?.close()`.
   Imo, this must be discussed with the whole team.
2. If we'll keep `submit` and `reset` for the close button type, we're missing the default `type="button"`, which still also must close the dialog.

My personal opinion:

Keep form association only for `type="reset"` (maybe even remove `type="submit"`) and use `closeButton.closest('dialog')?.close()` for all other variants.

oliverschuerch (changes requested):

packages/components/src/components/post-closebutton/post-closebutton.tsx
```diff
@@ -40,35 +45,53 @@ export class PostClosebutton {
...
+  private handleClick() {
+    if (this.buttonType === 'reset') this.internals.form?.reset();
+    else if (this.buttonType === 'button') {
+      this.host.closest('dialog')?.close();
+    } else if (this.buttonType === 'submit') {
+      console.warn(
+        'A close button with type="submit" is semantically incorrect. Use a regular `<button type="submit">` inside your form instead.',
+      );
+      this.host.closest('dialog')?.close();
+    }
```
oliverschuerch: After a talk with @gfellerph (correct me if I'm wrong) we can do the following:

1. Remove the `buttonType` `submit` option (defined in the `./types.ts` file).
  Since it was only meant to close the dialog, which is now done with `this.host.closest('dialog')?.close();`, we don't need it anymore.
2. If `this.buttonType === 'reset'`, reset the form as you do, then close the dialog, no matter if `buttonType` is set to `button` or `reset`.

oliverschuerch (changes requested):

packages/components/src/components/post-closebutton/post-closebutton.tsx
```diff
@@ -1,18 +1,23 @@
-import { OneOf, Required } from '@/utils';
+import { Component, Element, h, Host, Prop, AttachInternals } from '@stencil/core';
 import { version } from '@root/package.json';
-import { Component, Element, h, Host, Prop } from '@stencil/core';
 import { BUTTON_TYPES, ButtonType, Placement, PLACEMENT, SIZE, Size } from './types';
+import { Required, OneOf } from '@/utils';
```
oliverschuerch: As long as the team does not agree on a specific order (I saw some changes ordering imports alphabetically), please don't re-arrange the import statements without a technical reason.

I've created a ticket with a general proposal:
- https://github.com/swisspost/design-system/issues/7912

oliverschuerch (changes requested):

oliverschuerch (approved):

oliverschuerch (approved):

PR #7727, opened 2026-06-02
feat(components): support nesting without id references for `<post-collapsible>` component

### LINKED ISSUE #7688: Nested collapsible without the need for ids
### ✨ Is your feature request related to a problem? Please describe it.

The `<post-collapsible>` and `<post-collapsible-trigger>` work together through id refs. For other trigger/main component connection, nesting is possible to reduce the need for artificial ids. For the `<post-sidenavigation>` this behaviour would be very convenient and reduce complexity.

### 💡 Describe the solution you'd like

```HTML
<post-collapsible-trigger>
  <button>Collapse/expand</button>
  <post-collapsible>
    <p>Some content here</p>
  </post-collapsible>
</post-collapsible-trigger>
```

### 🔄 Describe alternatives you've considered

None; alignment with other APIs.

### 🔗 Additional context

https://github.com/swisspost/design-system/issues/6277

alionazherdetska:
## 📄 Description

This PR adds support for controlling a <post-collapsible> via direct nesting inside <post-collapsible-trigger>, removing the hard requirement to link via for="<id>". Also fixes aria-controls deduplication to use token-based comparison instead of substring matching, preventing false positives when multiple collapsibles are controlled by the same trigger.

## 🚀 Demo

[Collapsible docs
](https://preview-7727--swisspost-design-system-next.netlify.app/?path=/docs/6a91848c-16ec-4a23-bc45-51c797b5b2c3--docs&devModeEnabled=true)
---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/src/stories/components/collapsible/collapsible.stories.ts
leagrdv: The decorator here has no impact on the wrapped collapsible, as the trigger and collapsible are no longer two separate elements. Could you remove the decorator from there and add a `mb-16` to the button so that it's not stuck to the collapsible element?

packages/components/src/components/post-collapsible-trigger/post-collapsible-trigger.tsx
leagrdv: We should add a safe-guard here to make sure the button is not within the `post-collapsible` but outside of it, otherwise this edge case fails:

```
    <post-collapsible-trigger>
      <post-collapsible>
        <p class="border rounded p-24">This is collapsible content that can be shown or hidden.</p>
        <button class="btn btn-secondary">Toggle Collapsible</button>
      </post-collapsible>
      <button class="btn btn-secondary">Toggle Collapsible</button>
    </post-collapsible-trigger>
```

In that use case, because we're getting the first button we find, we would get the button that is inside of the `post-collapsible`, rather than the one outside of it.

leagrdv (changes requested):

leagrdv (approved):

PR #7758, opened 2026-06-09
feat(styles, documentation): add application header

### LINKED ISSUE #7734: [web component styles] Application header
### 📝 Description

The application header is a variant of the header that aims to be used by application that are not available on the internet. The term "variant" is a bit stretched on this component since the primary distinctions are made based on what content is slotted or what slots are filled and which are empty. Application header in technical terms refers to the header used in an `appearance: compact` environment as opposed to e.g. a microsite header in an `appearance: default` environment.

### ♿ Accessibility

The application header inherits the accessibility features already implemented in the self-managed header.

### ⚙️ Requirements
The application headers one distinguishing feature is the collapsing of the global header (the yellow part) when all slots except for the logo are empty. This should only happen, if the environent is `appearance: compact`. In default appearance, the global header should not collapse when all slots are empty.

- global header is empty except from the logo
- `appearance: compact` is set

To achieve that, we manually set a variable `--_post-initial-global-header-height` that determines initial global header height in the appearance token files. For compact appearance, that variable is set to `32px`, for default appearance it's set to `72px`. In the header, this variable is used to determine initial global header height.

1. Default is `72px`
2. If all slots except logo are empty, apply `--_post-initial-global-header-height`
3. If the header is scrolled, overwrite global header height with scroll value

### 🧪 Testing

New visual regression tests should be added for this header variant.

### 💻 Proposed implementation

The application header mostly differs from other variant by the size of the global and local headers. 
These sizes are controlled by custom properties: https://github.com/swisspost/design-system/blob/main/packages/styles/src/components/header/_custom-properties.scss.

_See updates in requirements._ A new variant must be added for the application header, the selector to target this variant is checking whether the there are elements in the global header slots: `audience`, `global-nav-primary`, `global-nav-secondary`, `language-menu`, `post-login`.

### 📃 Tasks

- [x] Implement styles according to design
- [x] Complete documentation in Storybook and make it accessible to all users
- [x] Add an example in the docs where the language menu is in the local header
- [x] Add visual regression tests
- [ ] Design review

myrta2302:
## 📄 Description

Added a Header variant for Internal Applications.

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/.storybook/helpers/with-compact-appearance.ts
```diff
@@ -0,0 +1,9 @@
+import { StoryFn } from '@storybook/web-components-vite';
+
+export function withCompactAppearance(story: StoryFn) {
```
leagrdv: I feel it would make more sense to call this (and the file) `forceCompactAppearance` as it's more clear that it overrides the user selected one.

apps/documentation/.storybook/helpers/with-compact-appearance.ts
```diff
@@ -0,0 +1,9 @@
...
+  const link = document.createElement('link');
+  link.rel = 'stylesheet';
+  link.href = '/styles/post-compact.css';
```
leagrdv: The stylesheet used in the iframe is added at the end of the `head` but the `post-default.css` file is also still there. Can you check if you can override it, rather than add it on top?

Image

apps/documentation/src/stories/components/header/header.docs.mdx
leagrdv:
```suggestion
1. **Global navigation** (Yellow Background): provides navigation links and information that are global to the entire organization.
2. **Local navigation** (White Background): displays navigation links and information specific to the current application.
```

Could be other better terms but I think it'd be good to remove this "Application header" here as well as calling the header parts "Header", otherwise it'll get too confusing.

apps/documentation/src/stories/components/header/header.docs.mdx
```diff
@@ -98,6 +98,18 @@ see the [Heading elements documentation](https://developer.mozilla.org/en-US/doc
 
 <Canvas sourceState="shown" of={HeaderStories.OnePagerH1} inline="false" />
 
+### Application
+
+The Application header is a header variant intended for internal applications. It uses the `appearance: compact` configuration and is distinguished by the absence of any content in its global (yellow) part except from the Logo.
```
leagrdv:
```suggestion
The Application header is a header variant intended for internal applications. It uses the `appearance: compact` styles bundle and is distinguished by the absence of any content in its global (yellow) part except from the Logo.
```

Maybe a link to the bundle details would also be useful: https://next.design-system.post.ch/?path=/docs/e53e2de8-0bbf-4f70-babc-074c5466f700--docs#available-bundles

apps/documentation/src/stories/components/header/header.stories.ts
```diff
@@ -154,6 +156,18 @@ const meta: MetaComponent = {
       },
       table: { category: 'state' },
     },
+    languageMenuPosition: {
```
leagrdv: If you go on the header docs, and you switch it to `local` (without modifying anything else in the table), the language menu disappears, because the `Local controls` was not set to `true`. 

There should be a condition for the presence of this argument to avoid this.

packages/styles/src/components/header/_custom-properties.scss
```diff
@@ -120,6 +120,35 @@ post-header:not(:has([slot='main-nav'])) {
...
+:root:has(post-header):not(:has(post-header > [slot='audience'])):not(
+    :has(post-header > [slot='global-nav-primary'])
+  ):not(:has(post-header > [slot='global-nav-secondary'])):not(
+    :has(post-header > [slot='language-menu'])
+  ):not(:has(post-header > [slot='post-login'])) {
+  --post-global-header-expanded-height: var(--_post-initial-global-header-height);
+  --post-global-header-reduced-height: 2rem;
+
+  @include media.only(desktop) {
+    --post-local-header-expanded-min-height: 4rem;
```
leagrdv: Why do we have specific height for the local header here? Should it defer from the rest of the headers?

packages/styles/src/components/header/_custom-properties.scss
```diff
@@ -120,6 +120,35 @@ post-header:not(:has([slot='main-nav'])) {
...
+    --post-logo-expanded-offset: 0.625rem;
+    --post-logo-reduced-offset: 0.625rem;
```
leagrdv, line 137: Is this necessary? When inspecting the header, I never see the `.logo` in the styles with this `margin-inline-start`

leagrdv (changes requested):

packages/styles/src/components/header/_custom-properties.scss
myrta2302: It is applied, but there are actually two elements with class `.logo`. This concerns the `.logo` which is in the `post-header` -> `shadowRoot` > `global header` -> `.logo`. 
image

apps/documentation/.storybook/helpers/with-compact-appearance.ts
leagrdv: Can you also rename the file accordingly?

leagrdv (approved):
Can you just rename the `with-compact-appearance.ts` to `force-compact-appearance.ts` to be consistent with the function inside?

PR #7762, opened 2026-06-09
feat(component): add `post-side-navigation`

### LINKED ISSUE #6277: [web component func] Side Navigation
### 📃 Tasks

- [ ] Add functional tests

alionazherdetska:
## 📄 Description

Ticket: https://github.com/swisspost/design-system-planner/issues/6?issue=swisspost%7Cdesign-system%7C6277

This PR implements the `post-sidenavigation` and `post-sidenavigation-trigger` web components.

## 🚀 Demo

[Side navigation
](https://preview-7762--swisspost-design-system-next.netlify.app/?devModeEnabled=true)---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/src/stories/raw-components/side-navigation/side-navigation.docs.mdx
```diff
@@ -0,0 +1,86 @@
...
+  <Controls of={SidenavigationStories.Default} />
+</div>
+
+<ComponentsPackageInstall component="<post-sidenavigation>" />
+
+<StylesPackageImport components={['sidenavigation', 'list']} />
+
+## Navigation content
+
+The navigation content is authored by you and projected through the default slot. It must be a `<nav>` landmark, with a heading (`<h*>`, or a `<p>` if no heading level fits) linked via `aria-labelledby`, and `<ul>`/`<li>` lists for the items. Nested levels are nested `<ul>`s.
```
myrta2302: I feel the text tone here "authored by you..." is a bit different than in the rest of the documentation. But please ignore this if you think is not the case.

apps/documentation/src/stories/raw-components/side-navigation/side-navigation.stories.ts
myrta2302, line 1: Did not check all the stories for this, but at least the Default story markup has a consequent `<ul> <ul>`. which causes an accessibility error (tested with Siteimprove).

Image

oliverschuerch (changes requested):
Please add a changeset

packages/components/src/components/post-side-navigation/post-side-navigation.tsx
myrta2302, line 1: Should the component names be `post-side-navigation` and `post-side-navigation-trigger`? I think we had resulted to this naming convention, but maybe better confirm it first.

apps/documentation/src/stories/raw-components/sidenavigation/sidenavigation.docs.mdx
```diff
@@ -0,0 +1,86 @@
...
+<div className="docs-title">
+  # Side Navigation
+  <link-design of={JSON.stringify(SidenavigationStories)}></link-design>
+</div>
+
+<PackageTag meta={meta} />
+
+<div className="lead">The side navigation is a container of navigation menu items that let the user navigate through different parts of a web application. It is a hierarchical structure that provides a maximum of 4 nested navigation levels which can be expanded or collapsed.</div>
+
+<Canvas sourceState="shown" of={SidenavigationStories.Default} />
```
oliverschuerch: The story must be implemented similar to the header stories... in an iFrame, with a min-height on the body.

Image

packages/components/src/styles/components/post-side-navigation.scss
myrta2302, line 1: I think that it is in the requirements that the side-navigation for mobile should not be initially displayed before hydration.

alionazherdetska:
> Please add a changeset

This is the Raw component implementation, the component will be published later on after the styling ticket is implemented.
Tasks defined in the ticket:
image

myrta2302 (changes requested):
Not totally relevant to this PR: But do you think we should ask UX to rethink the close-button position? Should we have two options? One inside and one as is right now? Because in the curent state, it cannot be used with the side nav.

apps/documentation/src/stories/raw-components/side-navigation/side-navigation.stories.ts
```diff
@@ -0,0 +1,253 @@
...
+          </li>
+          <li class="local-login">
+            <a href="">
+              <span>Login</span>
+              <post-icon name="login"></post-icon>
+            </a>
+          </li>
+        </ul>
+      </post-header>
+      ${renderSidenavigation(content, args, resolvedId, undefined, false)}
```
oliverschuerch: The story does not show how to add content aside the navigation. And adding content just after the `post-sidenavigation` is not working as expected:

Image

---

**Proposal**

```html
<post-header>...</post-header>
<post-sidenavigation>...</post-sidenavigation>

<main>
  <div class="container">
    <h1>Title</h1>
    <p>Content</p>
  </div>
</main>
```

Image

packages/components/src/components/post-side-navigation-trigger/post-side-navigation-trigger.tsx
oliverschuerch, line 1: Why does the `button` need to be slotted in?
Can't we just internalize it, add a text-property for the label and maybe a slot to override the default icon?

packages/components/src/components/post-side-navigation/post-side-navigation.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  private getDialog(): HTMLDialogElement | null {
+    const dialog = this.host.shadowRoot?.querySelector('dialog');
+
+    if (!dialog && this.device !== 'desktop') {
+      console.warn('No dialog was found.');
+    }
+
+    return dialog ?? null;
+  }
```
oliverschuerch: The name is already telling you, that this is actually a getter.
Thanks to the nullish coalescing operator, `const dialog` will always either be  `HTMLDialogElement` or `null`, but never `undefined.

```suggestion
  get dialog(): HTMLDialogElement | null {
    const dialog = this.host.shadowRoot?.querySelector('dialog') ?? null;

    if (dialog === null && this.device !== 'desktop') {
      console.warn('No dialog was found.');
    }

    return dialog;
  }
```

But also, since you define the `dialog` element in the shadowDOM, what could be the reason it does not exist (the only reason I could think of, is, if you access the getter before the first render cycle, when the shadowDOM has not yet been rendered)?
With this in mind, the final getter could look like this (but the above example is also fine):

```suggestion
  get dialog(): HTMLDialogElement {
    return this.host.shadowRoot?.querySelector('dialog') as HTMLDialogElement;
  }
```

packages/components/src/components/post-side-navigation/post-side-navigation.tsx
```diff
@@ -0,0 +1,175 @@
...
+          }}
+        >
+          <slot />
+          <post-closebutton onClick={() => this.hide()}>
+            {this.textClose}
+          </post-closebutton>
+        </dialog>
+      </Host>
+    );
+  }
```
oliverschuerch, line 156: Move the `host` element to the `render` function.

packages/components/src/components/post-side-navigation/post-side-navigation.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+    const trigger = document.activeElement?.closest<HTMLPostCollapsibleTriggerElement>(
+      'post-collapsible-trigger',
+    );
+
+    if (!trigger) return;
+
+    const collapsible = trigger.querySelector<HTMLPostCollapsibleElement>(
+      ':scope > post-collapsible',
+    );
```
oliverschuerch: You could extract this into getter functions, to simplify the `handlKeyDown` method. But it's ok if you keep it as is.

In general, the more specific and robust you set up the basics in your component, the less problems you'll face later. Smaller and more specific functions/class members are always welcome, easier to understand and better maintainable, imo.

```ts
private get collapsibleTrigger<HTMLPostCollapsibleTrigger>(): HTMLPostCollapsibleTrigger | null {
  return document.activeElement?.closest('post-collapsible-trigger') ?? null;
}

private get collapsible<HTMLPostCollapsible>(): HTMLPostCollapsible | null {
  return this.collapsibleTrigger?.querySelector(':scope > post-collapsible') ?? null;
}
```

apps/documentation/src/stories/raw-components/side-navigation/side-navigation.stories.ts
```diff
@@ -0,0 +1,253 @@
...
+
+    <post-sidenavigation id="${resolvedId}" text-close="${args.textClose}">
+      <nav
+        aria-labelledby="${ariaLabel ? nothing : titleId}"
+        aria-label="${ariaLabel ?? nothing}"
+      >
+        <h2 id="${titleId}" class="post-sidenavigation-heading">Section title</h2>
+        <ul>
+          ${navContent}
+        </ul>
```
oliverschuerch: You are rendering a `<ul>` in a `<ul>`!
Check it in the stories code panel.

packages/components/src/styles/components/post-side-navigation.scss
oliverschuerch, line 1: These are web-component specific global styles, so I would move them into the global styles of the components package.

oliverschuerch, line 1: The indentation is wrong!
First level should be fixed with https://github.com/swisspost/design-system/pull/7762/changes#r3427061408

**Figma**
Image

**Docs**
Image

oliverschuerch (changes requested):

packages/components/src/components/post-side-navigation-trigger/post-side-navigation-trigger.tsx
alionazherdetska: This follows the same pattern as other trigger elements e.g. `post-collapsible-trigger`, `post-popover-trigger`: it doesn't render its own button, just enhances whatever gets slotted in. The spec actually calls this out too:

image

Thus it is just to keep things consistent. Although, if you think this one should work differently though, I can bring up this topic for the Daily.

alionazherdetska:
> Not totally relevant to this PR: But do you think we should ask UX to rethink the close-button position? Should we have two options? One inside and one as is right now? Because in the curent state, it cannot be used with the side nav.

@myrta2302 

The concept is already there in Figma:
https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-Next-Level?node-id=31617-140456&m=dev
image

apps/documentation/src/stories/raw-components/sidenavigation/sidenavigation.docs.mdx
alizedebray: The issue is that iFrames cannot be controlled by the Args table (if you change the `text-close`, the snippet is not updated), that's why the first header story is not an iFrame.

oliverschuerch: I see, but for the moment we only have the `text-close` property which is interactive.

Therefore, I'd prefer to have a proper alignment of the component in the story iFrame and accept the fact, that the `text-close` prop is not interactively updating on change, over the confusing visualization of the component in the preview window, as shown in the screenshot above.

packages/components/src/components/post-side-navigation-trigger/post-side-navigation-trigger.tsx
oliverschuerch: Technically the first slotted element must be an interactive one, accessibility-wise it must be a `button[type="button"]`, since anything else would be semantically wrong or at least worse than that, but somehow we still think it's better to leave it to all of our users, to implement this element on there side.

To be 100% clear, I'm not accusing anyone, but I also don't get the thoughts behind it.

Imo, these components should make our users life easier, instead we tell them:
> That's how we want it, now go and implement it your own.
And by the way, if we missed something, all of you will have to life with it, until we force you to update it during a major version migration.

---

Btw. we did it the other way around in our [post-closebutton](https://github.com/swisspost/design-system/blob/main/packages/components/src/components/post-closebutton/post-closebutton.tsx) 😉

Whatever, let's fasten things up for us this time and keep it as it is for the release.

apps/documentation/src/stories/raw-components/side-navigation/side-navigation.styles.scss
oliverschuerch, line 1: The story is not rendered in an iframe. Is this still needed or can we at least decrease the height a little?

oliverschuerch (changes requested):

apps/documentation/src/stories/raw-components/side-navigation/side-navigation.styles.scss
alionazherdetska: It is intentional: otherwise the user wouldn't be able to interact with the `show-icons toggle` or change the `text-close` prop, which would make the story non-interactive. 

We could potentially address and improve it later. Should I add a ticket for `Next` milestone maybe?

I decreased the height of the story for the moment.

apps/documentation/src/stories/raw-components/side-navigation/side-navigation.stories.ts
```diff
@@ -0,0 +1,118 @@
...
+          <post-logo slot="post-logo" url="/">Homepage</post-logo>
+          <p slot="title">[Application Title]</p>
+          <ul slot="local-nav">
+            <li>
+              <post-side-navigation-trigger for="${navigationId}">
+                <button>
+                  <span>Menu</span>
+                  <post-icon aria-hidden="true" name="burger"></post-icon>
+                </button>
+              </post-side-navigation-trigger>
```
leagrdv, line 65: This component is not documented, could you add a section explaining how it works, how to link it to the `post-side-navigation` itself and where to place it in the header?

leagrdv (changes requested):

oliverschuerch (approved):

myrta2302 (approved):

apps/documentation/src/stories/raw-components/side-navigation/side-navigation.stories.ts
alionazherdetska: Will be addressed in this ticket: https://github.com/orgs/swisspost/projects/3/views/12?pane=issue&itemId=203835346&issue=swisspost%7Cdesign-system%7C7932

leagrdv (approved):

PR #7790, opened 2026-06-10
fix(styles): disabled and button link styles

### LINKED ISSUE #7754: [bug]: Wrong behaviour on disabled buttons
### 🐞 Describe the bug

A few issues with the disabled buttons on the docs:
- [x] Disabled buttons have a `cursor: pointer` on hover
- [x] Disabled button links have an hover effect (text-decoration is hidden on hover)
- [ ] #7788 (will be done in a separate ticket)
- [x] In the docs, user is technically able to set a button tag to `link` with the disabled state, which results in an example which shows this `<a class="btn btn-link" href="#" disabled="true">`. This use case should not be possible, and the `disabled` prop should only be shown for tags `button` and `input`

### 📷 Screenshots

https://github.com/user-attachments/assets/86742a3f-8068-4ca0-9cc0-34fdacd75187

Image

leagrdv:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/styles/src/mixins/_button.scss
```diff
@@ -93,8 +94,11 @@
   }
 }
 
-@mixin button-variant-def($type, $color) {
+@mixin button-variant-def($type, $color, $variant: '') {
```
oliverschuerch, line 98: This would be a good opportunity to refactor the mixin.
However, it's not the content of this task, so if you don't want to mix up let me know and I'll create a ticket.

```scss
// $variant => required
// one of button.$btn-variants defined in styles/src/variables/components/_button.scss

// $state => required
// former $type

// The $variant to $color mapping should be done within the mixin to ensure it's correctly working, no matter where the mixin is used!

@mixin button-variant-def($variant, $state) {
  $color: if($variant == 'default', 'secondary', if($variant == 'link', 'tertiary', $variant));

  // rest of mixin can stay as is
}
```

oliverschuerch (changes requested):
There are other files using the `button-variant-def` mixin. Can you confirm they don't need to be updated?

- components package `post-language-menu-item.scss`
- components package `post-megadropdown.scss`
- styles package `toast.scss`
- styles package `header/mixins.scss`

leagrdv:
> There are other files using the `button-variant-def` mixin. Can you confirm they don't need to be updated?
> 
> * components package `post-language-menu-item.scss`
> * components package `post-megadropdown.scss`
> * styles package `toast.scss`
> * styles package `header/mixins.scss`

Yes, these are not affected. The fix was only for the buttons of type `link` on dark mode, to prevent them from having a grey background there.

packages/styles/src/mixins/_button.scss
leagrdv: Yes, please create a ticket so that we can have a look at this after the release

oliverschuerch: FYI: #7925

oliverschuerch (approved):

PR #7779, opened 2026-06-10
chore(docs): improve the formatting for `code-block` component

alionazherdetska:
## 📄 Description

Please include a summary of the changes made in this PR.

## 🚀 Demo

[Migration docs
](https://preview-7779--swisspost-design-system-next.netlify.app/?path=/docs/c23b1d0b-76b3-4e38-aa76-b10c29bb873f--docs&devModeEnabled=true)---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

- [ ] Visual changes detected and approved _(Check this box if VRT fails and changes are intentional)_

## 📝 Checklist

- ✅ My code follows the style guidelines of this project
- 🛠️ I have performed a self-review of my own code
- 📄 I have made corresponding changes to the documentation
- ⚠️ My changes generate no new warnings or errors
- 🧪 I have added tests that prove my fix is effective or that my feature works
- ✔️ New and existing unit tests pass locally with my changes

PR #7800, opened 2026-06-11
chore(docs): add migration information for `ngbProgressbar`

### LINKED ISSUE #7723: Add migration information for `ngbProgressbar`
Add migration guidance for replacing `NgbProgressbar` with the new `progressbar` component.

### Research Required

Review project usage of:
- `NgbProgressbar`
- Dynamic progress updates

### Tasks
- [ ] Research common implementations.
- [ ] Document migration scenarios.
- [ ] Provide code snippet examples.
- [ ] Add migration guidance to the v9 → v10 migration guide.

alionazherdetska:
## Description
Adds migration documentation for replacing `NgbProgressbar` (v9) with `post-progressbar` (v10) in the migration guide.
 
### Research Validation
 
I analyzed NgbProgressbar usage across **48 files in 10+ projects** to make sure our migration guide actually helps people with real problems they're facing:
 
| Pattern | Projects | Main Challenge |
|---------|----------|-----------|
| **Stepper forms** | GMSP, MindWrite, Simba, DXS, Vevol-Zapa | `[value]` binding → `[attr.value]` |
| **File uploads** | EHR, EVD, BBS | `[showValue]="true"` → need separate element |
| **Job progress** | Sortpla, LAD | Dynamic value calculations preserved |
| **Grid cells** | DOPLA | Standalone component compatible |
| **Global config** | test-evoting | `striped`/`animated` → CSS alternatives |
 
### How We Built the Guide
 
The migration documentation is based on what we actually found in the codebase:
 
- **Research showed:** Developers can't use `[showValue]` anymore and need a workaround for percentage displays
- **Guide includes:** Practical example with wrapper + separate `<span>` pattern
- **Result:** Documentation covers the exact problem developers will hit
 
---

## 📄 Note:

The code block formatting in the migration guide will be improved in a following PR which implements the fix the indentation of the opening comment line in template examples:
image
https://github.com/swisspost/design-system/pull/7779

## 🚀 Demo

[Migration docs
](https://preview-7800--swisspost-design-system-next.netlify.app/?path=/docs/c23b1d0b-76b3-4e38-aa76-b10c29bb873f--docs&devModeEnabled=true)
---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -327,10 +328,95 @@ export class MigrationV910Component extends LitElement {
...
+                                  <p><strong>Key differences:</strong></p>
+                                  <ul>
+                                    <li>
+                                      <strong>Attributes instead of properties:</strong> Use standard HTML
+                                      attributes (<code>min</code>, <code>max</code>, <code>value</code>)
+                                      instead of ng-bootstrap directives.
+                                    </li>
+                                    <li>
+                                      <strong>No visual variants:</strong> The new component does not support
+                                      <code>type</code> (color variants), <code>striped</code>, or
```
leagrdv: It's not exactly correct for the type, as the previous "success", "danger" and "warning" still exist. Also, saying that style customization can be done through CSS might not be the best here, as I don't think we want to encourage users to do whatever they want with the components.

apps/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -327,10 +328,95 @@ export class MigrationV910Component extends LitElement {
...
+                                      <code>animated</code> properties. Style customization is done through
+                                      CSS custom properties and the component's public API.
+                                    </li>
+                                    <li>
+                                      <strong>Semantic HTML:</strong> Uses the native HTML role and ARIA
+                                      attributes for better accessibility (<code>role="progressbar"</code>,
+                                      <code>aria-valuemin</code>, <code>aria-valuemax</code>,
+                                      <code>aria-valuenow</code>).
```
leagrdv: This was handled "under the hood" in the ngbprogressbar, and that is still the case now for the `post-progressbar`, so that makes no difference for the user. I don't think it's needed to be mentioned

apps/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -327,10 +328,95 @@ export class MigrationV910Component extends LitElement {
...
+                                  <p>
+                                    To update the progress value dynamically in Angular, bind directly to the
+                                    <code>value</code> attribute:
+                                  </p>
+                                  <code-block
+                                    code=${`<!-- template.html -->\n<post-progressbar\n  [attr.value]="progressValue"\n  min="0"\n  max="100"\n></post-progressbar>`}
+                                  ></code-block>
+                                  <code-block
+                                    code=${`// component.ts\nexport class MyComponent implements OnInit {\n  progressValue = 0;\n\n  ngOnInit() {\n    const interval = setInterval(() => {\n      if (this.progressValue < 100) {\n        this.progressValue += 10;\n      } else {\n        clearInterval(interval);\n      }\n    }, 500);\n  }\n}`}
+                                  ></code-block>
```
leagrdv: The code block above makes sense, but this one just shows how to increase a variable's value. Since `ngbProgressbar` was not handling it either, I don't think it needs to be shown here.

apps/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -327,10 +328,95 @@ export class MigrationV910Component extends LitElement {
...
+                                  <p><strong>Showing percentage values:</strong></p>
+                                  <p>
+                                    Since <code>[showValue]</code> is no longer supported, create a separate element to display percentages:
+                                  </p>
+                                  <code-block
+                                    code=${`<!-- template.html -->\n<div class="progress-with-label">\n  <post-progressbar\n    [attr.value]="progressValue"\n    [attr.max]="100"\n  ></post-progressbar>\n  <span>{{ progressValue }}%</span>\n</div>`}
+                                  ></code-block>
```
leagrdv: This is incorrect:
- The class `.progress-with-label` does not exist.
- There is no way to show the value on the progressbar itself, but as described in the docs, it can be added above the progressbar with the class `.progress-value`

apps/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -327,10 +328,95 @@ export class MigrationV910Component extends LitElement {
...
+                                  <p><strong>Styling with labels and messages:</strong></p>
+                                  <p>
+                                    The new component supports wrapper elements for additional labels and
+                                    status information. See the
+                                    <a href="/?path=/docs/a1b2c3d4-e5f6-7890-abcd-ef1234567890--docs"
+                                      >post-progressbar documentation</a
+                                    >
+                                    for complete examples including labels, helper messages, and state
+                                    indicators.
+                                  </p>
```
leagrdv: Those wrapper elements with additional labels and status information were not available on the `ngbProgressbar` component, this is all new feature so I don't think it's necessary to mention it on the migration guide.

leagrdv (changes requested):

apps/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
leagrdv: Also, quick note: the progress-value part can only be added if a progress-label is also present.
But as stated before, it's not exactly what "showValue" was doing (meaning having the percentage on the bar itself), this needs to be more precise that's it's not a 1:1 equivalent

apps/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
```diff
@@ -323,7 +323,86 @@ export class MigrationV910Component extends LitElement {
...
+                                accessible and framework-agnostic solution.
+                              </p>
+                              <p><strong>Before (v9 — NgbProgressbar)</strong></p>
+                              <p>
+                                The NgbProgressbar component was used with two-way binding and
+                                properties like <code>type</code>, <code>striped</code>, and
+                                <code>animated</code>:
+                              </p>
+                              <code-block
+                                code=${`<!-- template.html -->\n<ngb-progressbar\n  type="primary"\n  [value]="65"\n  [striped]="true"\n  [animated]="true"\n>\n</ngb-progressbar>`}
```
leagrdv, line 567: Can you add back the part about the `type` but explain that now we have the states instead so that 4 previous types now have state (and mention we don't have the colored ones, it would default to neutral for like primary, secondary and the others that are not those 4 states)?

leagrdv (changes requested):

apps/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
alionazherdetska: I added the type info back! I used `CSS classes` instead of `state` since what the user actually needs to do is replace the type prop with a modifier class on the wrapper, calling it "state" felt confusing when there's no actual state prop. Let me know if you have a different opinion.

apps/documentation/src/stories/misc/migration-guide/migrationv9-10.component.ts
leagrdv, line 21:
```suggestion
      typeahead: false,
      progressbar: false,
```

leagrdv (approved):
Can you just add the `progressbar: false` to the default state on the migration guide?
Otherwise, it's all good now! Thanks for the updates

PR #7841, opened 2026-06-15
chore(docs): removed all status tags

gfellerph:
Removed deprecated status tags "experimental" and "in progress". Components will no longer be tagged for status because it confused more than it helped.

### DIFF
--- a/.changeset/petite-vans-cross.md
+++ b/.changeset/petite-vans-cross.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': patch
+---
+
+Removed deprecated status tags "experimental" and "in progress". Components will no longer be tagged for status because it confused more than it helped.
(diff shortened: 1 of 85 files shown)


.changeset/petite-vans-cross.md
leagrdv, line 1: Could you also remove the `status:Deprecated` from the internet-header component as well as the other `status:Stable`?

Otherwise we can still filter them in here and that might be confusing:
Image

And if we're removing those status, I think it'd be best to remove the mechanism for the status icon that's in the `manager.ts` file. We can always come back to this PR if we ever want to add back status.

leagrdv (changes requested):

.changeset/petite-vans-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-documentation': patch
+---
+
+Removed deprecated status tags "experimental" and "in progress". Components will no longer be tagged for status because it confused more than it helped.
```
leagrdv:
```suggestion
Removed all status tags. Components will no longer be tagged for status because it confused more than it helped.
```

leagrdv (approved):

PR #7855, opened 2026-06-15
feat(components): decouple visual value from model in date-picker

### LINKED ISSUE #7412: [web component func] Date Picker - Decouple visual value from model
Currently we use a slotted `input` element with `type="text"` to bind the Air Date Picker to.

Instead we could let the user slot an `input` element with `type="date"`, keep it invisible and use an `input` with `type="text"` internally (and also bind the Air Date Picker to that), just to show the formatted value. While the internal input can show any pre formatted string or mask, the slotted one has a strict behavior about its `value` format.

By the way, user would actually only interact with the slotted `input` element.

The advantage of this would be, that we can show whatever we want to the user, but keep the value of the `input` in the format that it would have on a native `input[type="date"]` element.

I've created a small POC for the localization topic, which works like so already.
Have a look: https://codepen.io/oliverschuerch/pen/qEaPYqJ

---

**Acceptance Criterias**

- [ ] For range Date Picker, DEVs need to slot two `input[type="date"]` elements!
- [ ] The `input[type="date"]` elements value must either be a valid `YYYY-MM-DD' or an empty string (e.g. ''). Make sure the component does not try to apply anything else to the input, or it will end up as an error in the console.
- [ ] Make sure `input[type="date"]` elements can be used to define an initial value (e.g. `<input type="date" value="2026-01-31" />` and the value is always in sync with the date-picker. DEVs may want to use the inputs `value` property in their model, instead of rely on custom update events or manually adding event listener to the input. The visual date-picker component is just a helper, to set the value of the `input[type="date"]` elements.

oliverschuerch:
## 📄 Description

Implements the most simple solution to decouple the date picker mask from the value.

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


.changeset/grumpy-shoes-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': minor
+---
+
+Decoupled the visual value from the model in the `post-date-picker` component. The `selectedStartDate` and `selectedEndDate` properties are no longer mutated by the component itself; user interactions and manual input are now reflected through the `postUpdateDates` event instead. A new `value()` method was added to read the date picker's current value (`string | string[]`).
```
alizedebray, line 5: Can you rewrite this changeset to clearly explain the changes?
The visual value is still coupled to the model, and the rest of the description is difficult to understand.

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -92,27 +92,50 @@ export class PostDatePicker {
...
+  handleSelectedEndDateChange() {
+    const dpDates = this.inputDates || [];
+    const startDate = dpDates[0] || null;
+
+    if (this.range && this.datesChanged(dpDates)) {
+      this.dpInstance.clear({ silent: true });
+      this.dpInstance.selectDate(dpDates);
+      this.dpInstance.setViewDate(startDate);
+      this.updateMask();
+    }
```
alizedebray, line 137: Please remove this code duplication.

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -220,6 +243,14 @@ export class PostDatePicker {
    */
   @Event() postUpdateDates: EventEmitter<string | string[]>;
 
+  /**
+   * Returns the date packer current value (Date| Date[] | '').
```
alizedebray, line 247: Please remove the type from the JSDoc.
The method returns ISO strings, not `Date` objects.

alizedebray, line 247:
```suggestion
   * Returns the date picker's current value (Date| Date[] | '').
```

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -220,6 +243,14 @@ export class PostDatePicker {
...
+   */
+  @Method()
+  async value() {
```
alizedebray, line 250: This is a method, not a getter, so its name should start with a verb (for example, `getSelectedDates()`).

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -560,45 +633,34 @@ export class PostDatePicker {
     );
   }
 
-  private skipOnSelectCount = 0;
+  // private skipOnSelectCount = 0;
```
alizedebray, line 636: Is this comment necessary?

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -560,45 +633,34 @@ export class PostDatePicker {
...
-        this.dpInstance.setViewDate(date);
-        this.postUpdateDates.emit(this.dateToIso(date));
-      } else {
-        this.resetSelection();
+  private updateInstanceDates([start, end]: Date[]) {
+    if (this.isValidDate(start)) {
+      if (this.datesChanged([start, end])) {
+        this.dpInstance.clear({ silent: true });
+        this.dpInstance.selectDate(this.range ? [start, end] : start);
+        this.dpInstance.setViewDate(start);
```
alizedebray, line 658: This logic is duplicated here as well.

packages/components/src/components/post-date-picker/readme.md
```diff
@@ -56,6 +56,16 @@ Type: `Promise<void>`
 
 
 
+### `value() => Promise<string | string[]>`
```
alizedebray, line 59: It is unclear why this method is needed. There are already two ways to retrieve the value: reading it from the input itself or listening to the `postUpdateDates` event. This method introduces a third way to access the same information only with a different format.

In practice, consumers will probably rarely use a method to retrieve the value. A method does not provide any indication when the value changes, so it would still need to be combined with a change or input event listener. At that point, the input value is directly available.

alizedebray (changes requested):
This PR does not appear to address the requirements described in the ticket.

The ticket states:

> For range Date Picker, DEVs need to slot two input[type="date"] elements!

However, this is still the current behavior.

The other requirement:

> Make sure the component does not try to apply anything else to the input, or it will end up as an error in the console.

is not fulfilled either, as the input value still contains the mask.

Based on the current implementation, none of the ticket requirements appear to be addressed by this PR.

gfellerph:
Check event return value when input is empty. It should not return the mask string. Also check return value when date is  cleared by user.

1. Idea: removing the mask could be a workaround for returning an empty string. But does not restrict random user input. Optionally, putting readonly could be used to force picker selection instead of manual entry. 

2. Hidden mask, that is active but not showing as a string in the input value (`lazy` setting in iMask)

3. Document how to interact with the event for reactive forms for now and ship as is.

alizedebray:
Closing in favor of #7955

PR #7866, opened 2026-06-16
chore(components): fixed missing header border for hcm

### LINKED ISSUE #7811: [bug]: Header should have bottom line in windows HCM
### 🐞 Describe the bug

Image

The border bottom should be visible in HCM to distinguish header area from content area, especially because the header is sticky.

### 🛠️ To Reproduce

1. Enable HCM or forced-color mode in browser

myrta2302:
## 📄 Description

Fixed missing header border in High Contrast Mode.

## 🧪 Visual regression tests


packages/components/src/components/post-header/post-header.scss
```diff
@@ -39,6 +39,10 @@ header,
   inset-block-start: var(--post-header-top);
   z-index: commons.$zindex-header;
   box-shadow: elevation.$elevation-300;
+
+  @media (forced-colors: active) {
```
oliverschuerch: Please use the utility mixins for the media query:
https://github.com/swisspost/design-system/blob/main/packages/styles/src/mixins/_utilities.scss

packages/styles/src/components/header/_post-header.scss
```diff
@@ -23,6 +23,11 @@ post-header {
   &:not([data-hydrated]) {
     display: block;
     height: var(--post-header-height);
+
+    @media (forced-colors: active) {
+      box-sizing: content-box;
+      border-bottom: 1px solid;
+    }
```
oliverschuerch: Why is this necessary?

1. The border styles on the `:host` selector should do the trick.
  If the component is server-side rendered, the styles will be present in the components template tag, while when it is client-side rendered, there will be a layout shift anyway, since the components shadowDOM only shows up after client-side hydration.
2. Changing the `box-sizing` property from `border-box` (default by [reset.scss](https://github.com/swisspost/design-system/blob/main/packages/styles/src/elements/reset.scss)) to `content-box` adds the risk for unwanted side-effects.

oliverschuerch (changes requested):

packages/styles/src/components/header/_post-header.scss
myrta2302: When i first tried in the non HCM i was getting a pre and after-hydration mismatch by 1px. So after i moved the styles to the HCM only, this would be to prevent the same issue, when HCM is on. Currently we do not have any E2E tests in HCM to check. In any case, If you think its not necessary i 'll remove it.

oliverschuerch: As mentioned, if the component is server-side rendered, the scoped component styles are present. If on the other side, the component is matching one of the above selectors, it means, it is probably used in a client-side only environment.

---

The hydration issue you encountered, could be related to the fact, that the selector probably use the wrong logic. These fallback styles should only be applied to `post-header` if it has not been server-side rendered and also the client-side hydration did not yet run (which means the scoped styles are not yet present).

At the moment we're using an OR selector:

```scss
// IF
// not defined
post-header:not(:defined)),
// OR
// not server-side rendered
post-header:not([data-hydrated]) {
  // do...
}
```

Before client-side hydration kicks in, `post-header:not(:defined)` is always `true`, no matter if the component has been server-side rendered (which means it contains the `data-hydration` attribute) or not.

Instead it should be:

```scss
// IF not defined AND not server-side rendered
post-header:not(:defined):not([data-hydrated]) {
  // do...
}
```

---

1. For now, please remove the HCM styles from this file.
2. If you run into a hydration issue, try to update the selector as showed above.
3. If this really does not work, we can still apply your workaround.

oliverschuerch (changes requested):

packages/styles/src/components/header/_post-header.scss
myrta2302: I have reverted the file.

oliverschuerch (approved):

PR #7877, opened 2026-06-17
fix(components): increase edge gap for `post-popover`

### LINKED ISSUE #7774: [bug]: `post-closebutton` outside of its container
### 🐞 Describe the bug

Now that the close button has been changed to be outside of its container, it is cutoff when the elements "touch" the right or top edge of the screen.
To avoid this, a safe zone (margin around the containers) should be set on elements containing the `post-closebutton`.

### 📷 Screenshots

Image

### Tasks
- [ ] Implement the safe zone around the popover

leagrdv:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [X] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/components/post-popover/post-popover.tsx
```diff
@@ -86,6 +86,7 @@ export class PostPopover {
         <post-popovercontainer
           arrow={this.arrow}
           placement={this.placement}
+          edgeGap={17}
```
alionazherdetska: Small suggestion: `17` reads as a magic number here. It seems to be half the close button's diameter, but the actual size varies by breakpoint (28px on mobile/tablet, 34px on desktop), so we're adding a bit more gap than needed on smaller screens:

Image

What do you think about deriving it dynamically? That way we won't need to remember to update this if the close button size ever changes 🙂

alionazherdetska (changes requested):

packages/components/src/components/post-popover/post-popover.tsx
leagrdv: Good idea, I've set it dynamically based on the `--post-close-small-size` token :)

packages/components/src/components/post-popover/post-popover.tsx
```diff
@@ -69,6 +72,26 @@ export class PostPopover {
     if (isOpen) this.focusFirstEl();
   }
 
+  componentDidLoad() {
+    this.updateEdgeGap();
+    this.resizeObserver = new ResizeObserver(() => this.updateEdgeGap());
```
alionazherdetska: The **ResizeObserver** fires on every single pixel of window resize.

We already have a breakpoint utility that emits postBreakpoint:device only when the 
device actually changes (mobile/tablet/desktop) and post-footer and post-header use this pattern. 

Do you think it will make sense to use that here instead? 🙂

packages/components/src/components/post-popover/post-popover.tsx
```diff
@@ -69,6 +72,26 @@ export class PostPopover {
...
+    this.resizeObserver.observe(document.documentElement);
+  }
+
+  disconnectedCallback() {
+    this.resizeObserver?.disconnect();
+  }
+
+  // Dynamically update edge gap based on the post-closebutton icon's width
+  private updateEdgeGap() {
+    const rawValue = getComputedStyle(this.host).getPropertyValue('--post-close-small-size').trim();
```
alionazherdetska: I looked into this and the token `--post-close-small-size` doesn't actually reflect 
the real rendered size of the close button. From DevTools you can see the button size 
comes from `.btn-icon.btn-sm`. So `--post-close-small-size` returns 24px but the 
actual button is 34px on desktop and 28px on tablet/mobile 🙂

Image
Image

alionazherdetska (changes requested):

packages/components/src/components/post-popover/post-popover.tsx
alionazherdetska, line 54:
```suggestion
@Method()
async show(target: HTMLElement) {
  await this.popoverRef.show(target);
  this.updateEdgeGap();
  this.focusFirstEl();
}
```

packages/components/src/components/post-popover/post-popover.tsx
```diff
@@ -68,6 +70,35 @@ export class PostPopover {
     if (isOpen) this.focusFirstEl();
   }
```
alionazherdetska, line 75:
```suggestion
@Method()
async toggle(target: HTMLElement, force?: boolean) {
const isOpen = await this.popoverRef.toggle(target, force);
  if (isOpen) {
    this.updateEdgeGap();
    this.focusFirstEl();
  }
}
```

packages/components/src/components/post-popover/post-popover.tsx
```diff
@@ -68,6 +70,35 @@ export class PostPopover {
...
+    requestAnimationFrame(() => this.updateEdgeGap());
+  };
+
+  connectedCallback() {
+    globalThis.addEventListener('postBreakpoint:device', this.breakpointChange);
+  }
+
+  componentDidLoad() {
+    requestAnimationFrame(() => this.updateEdgeGap());
+  }
```
alionazherdetska: This can be removed since the popover is hidden on load and getBoundingClientRect 
returns 0 at that point, so updateEdgeGap always exits early without setting anything:

Image

🙂

alionazherdetska (changes requested):

alionazherdetska (approved):

PR #7879, opened 2026-06-17
fix(styles): fix anchor button styling in HCM

### LINKED ISSUE #7813: [bug]: Link buttons have no border in HCM
### 🐞 Describe the bug

Image

`<a>` tags as primary and secondary buttons should have a visible border in windows HCM or forced colors mode.

Tertiary button links however have a border, but it's too close to the text:

Image

### Tasks

- [ ] Provide a border for primary, secondary and tertiary with padding
- [ ] Provide a `text-decoration: underline` for tertiary without padding

alionazherdetska:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/styles/src/elements/anchor.scss
```diff
@@ -28,5 +28,16 @@ a {
...
     }
+
+    &.btn {
+      border: 1px solid ButtonBorder !important;
+      padding: 0.25rem 0.75rem;
+    }
+
+    &.btn.btn-tertiary.px-0 {
+      border: 0 !important;
+      padding: 0 !important;
```
leagrdv: I don't think you need to force this as `px-0` already sets `padding: 0`

packages/styles/src/elements/anchor.scss
```diff
@@ -28,5 +28,16 @@ a {
     &:visited {
       color: VisitedText;
     }
+
+    &.btn {
+      border: 1px solid ButtonBorder !important;
+      padding: 0.25rem 0.75rem;
```
leagrdv: There should not be padding-specific code on HCM, as it would change the layout on the page and HCM should only change color contrasts there. Also this currently adds padding to all padding, regardless if they're standard or `.btn-link`

leagrdv (changes requested):

packages/styles/src/elements/anchor.scss
```diff
@@ -28,5 +28,14 @@ a {
     &:visited {
       color: VisitedText;
     }
+
+    &.btn {
+      border: 1px solid ButtonBorder !important;
```
leagrdv: Forcing the `solid` style here removed the dashed border on disabled elements. Same as the `1px`: our buttons have a wider border than this. I would simply force the `border-color` here, while making sure for disabled elements, it's the same color as the disabled text.

packages/styles/src/elements/anchor.scss
leagrdv, line 1: `.btn-link` elements have a border as well as the text-decoration, I think it could simply have the text-decoration (as it does outside of HCM)
Image

leagrdv (changes requested):

packages/styles/src/elements/anchor.scss
alionazherdetska: Updated to use border-color only. And about the disabled state here: since this is anchor.scss, the rule only affects `<a>` tags, and those don't support disabled at all. You can actually see this in our own Storybook: if you switch the tag to `<a>` on the Button page, the disabled control just disappears.

packages/styles/src/elements/anchor.scss
leagrdv: Some button links elements still have both the border and text decoration underline
Capture d&#39;écran 2026-06-29 162135

packages/styles/src/elements/anchor.scss
leagrdv: It's probably in the button file directly then, but some buttons still have the lined border when disabled, while some other don't. It would make sense for it to be consistent with the "normal mode", meaning dashed border when it has a border
Capture d&#39;écran 2026-06-29 162217

leagrdv (changes requested):

packages/styles/src/elements/anchor.scss
alionazherdetska: I have created a ticket which will address it: https://statics.teams.cdn.office.net/evergreen-assets/safelinks/2/atp-safelinks.html

packages/styles/src/elements/anchor.scss
alionazherdetska: I have created a ticket which will address the button styles in HCM: https://statics.teams.cdn.office.net/evergreen-assets/safelinks/2/atp-safelinks.html

leagrdv (approved):

PR #7882, opened 2026-06-17
feat(header): update the post login in the configurable header

### LINKED ISSUE #7786: KLP Login Widget updates
After PI planning, we've decided to provide a version of the Login Widget for self-managed headers that is truly self-managed. This means that the Design System will **not** make API calls or manage content, it will only provide the styles and functionality of the menu.

The pre-configured approach remains unchanged. However, the markup currently in the `packages/components` package needs to move into the `packages/internet-header` package as an **internal** component.

### Tasks

- [x] Remove the `<post-login-widget>` documentation from `apps/documentation`
- [x] Delete the `<post-login-widget>` component from `packages/components`
- [x] Update the login API in the internet header to match the [defined contract](https://github.com/swisspost/design-system/discussions/5422#discussioncomment-15541785)
- [x] Create an internal login widget component in `packages/internet-header/src/components/internal`, reusing the removed `<post-login-widget>` markup, and import it directly in `<swisspost-internet-header>`
- [x] Improve the self-managed user menu documentation on the header's documentation page, including guidance on how consumers can retrieve user information from the account API
- [ ] In the self-managed user menu, add a slot for the company name: https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-Next-Level?node-id=32040-38606&t=5Ir2T0KBXcuDFcrJ-4
- [x] Fill in the company name in the pre-configured login widget (if available from the `/subscribe`, `company` key when `userType` is `B2B`)
- [x] #7898

alizedebray:
## 📄 Description

This PR integrate the post login mechanism into the `swisspost-internet-header` component.

## 🚀 Demo

https://preview-7882--swisspost-design-system-next.netlify.app/?path=/docs/ebb11274-091b-4cb7-9a3f-3e0451c9a865--docs&devModeEnabled=true

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
```diff
@@ -111,16 +112,8 @@ export class PostInternetHeader {
...
-      state.environment = this.environment.toLocaleLowerCase() as Environment;
-      if (this.language !== undefined) state.currentLanguage = this.language;
-
-      await this.updateConfig();
-    } catch (error) {
-      console.error(error);
-    }
+    await this.fetchHeaderConfig();
+    await this.fetchUserData();
   }
```
alionazherdetska, line 118: **Small perf thing: the two fetches are sequential, which delays the first render**

Right now `componentWillLoad` awaits them one after the other:

```typescript
await this.fetchHeaderConfig();
await this.fetchUserData();
```

Since nothing renders until `componentWillLoad` fully resolves, the header is invisible while we wait for the session API, even though the config might already be ready. I added lifecycle logs to the component and measured this in the playground across two network conditions.

The first two screenshots show the Network waterfall for the sequential approach. You can see `subscribe` only starts after `test-configuration.json` finishes (queued at 744ms vs 734ms), 
so its full ~557ms gets added on top:
[

Image
Image

](url)
The next two show `Promise.all` where both requests fire at the same time (queued at ~330ms each), so we only wait for whichever takes longer:

Image
Image

The last two show the lifecycle logs on Slow 4G: with the sequential approach `componentDidRender` fires at +5283ms, and with `Promise.all` at +4917ms. The gap only grows on worse connections:

Image
Image

Switching to `Promise.all` so both fire at the same time would be a pretty straightforward fix:

```typescript
await Promise.all([this.fetchHeaderConfig(), this.fetchUserData()]);
```

There is also a more aggressive option which is `void fetchUserData()`. That would render the header as soon as the config arrives and let the session resolve in the background. The catch is that logged-in users would briefly see the Login button before it switches to the UserMenu once the session comes back, so it probably needs a placeholder or some UX thought before going there. For now `Promise.all` feels like the easy win.

Anyway, just something I noticed while digging into it, totally up to you whether it's worth addressing or there are constraints agains it.

packages/internet-header/src/components/internal/UserMenu.tsx
```diff
@@ -1,35 +1,40 @@
...
+  const userMenuId = createIdFrom(userFullname);
   return (
     <div slot={slot}>
       <post-menu-trigger for={userMenuId}>
         <button class="btn btn-link" type="button">
-          <Avatar user={config.user} description={textCurrentUser} />
-          <span class="visually-hidden">{textUserLinks}</span>
+          <Avatar
+            user={config.user}
+            description={textCurrentUser.replace('{user}', `${userFullname}`)}
```
alionazherdetska, line 29: The `{user}` placeholder in `textCurrentUser` isn't documented anywhere. Tested with a mock logged-in user (Jane Smith) and the markup from `index.html`:

Image
Image

NVDA announces `"Current user is John Doe."` because the prop description gives no hint that a placeholder is needed:

https://github.com/user-attachments/assets/40e2abb4-4940-4f12-86cf-a32d4d22823e

**Side note:** `textCurrentLanguage` uses `#name` for the same concept, maybe we should align on one convention.

packages/internet-header/src/index.html
```diff
@@ -21,7 +21,7 @@
       text-current-language="The currently selected language is #name."
       text-current-user="Current user is John Doe."
```
alionazherdetska:
```suggestion
      text-current-user="Current user is {user}."
```

packages/internet-header/cypress/fixtures/pages/ppm-config.html
```diff
@@ -21,7 +21,19 @@
...
   <body>
-    <swisspost-internet-header project="ekp" environment="int01"></swisspost-internet-header>
+    <swisspost-internet-header
+      project="ekp"
+      environment="int01"
+      text-main="Main"
+      text-menu="Menu"
+      text-change-language="Change the language"
+      text-current-language="The currently selected language is #name."
+      text-current-user="Current user is John Doe."
```
alionazherdetska:
```suggestion
      text-current-user="Current user is {user}."
```

alionazherdetska (changes requested):

alionazherdetska (approved):

PR #7885, opened 2026-06-17
chore(docs): better document self-managed user menu

alizedebray:
## 📄 Description

This PR updated the documentation for the header user menu.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/src/stories/components/header/header.docs.mdx
```diff
@@ -100,15 +101,29 @@ see the [Heading elements documentation](https://developer.mozilla.org/en-US/doc
...
+The authentication mechanism and conditional display are handled by each project individually.
+The Design System provides the UI components, while the application manages the authentication logic and decides which content to render in the `post-login` slot.
+
+### Fetching user data
+
+Projects can retrieve the current Post user session data via the following endpoint:
+
+<Source code={SampleSessionSubscription} language="typescript" />
+
+If the request fails or returns no `data`, the user is not authenticated and the login link can be displayed instead.
```
oliverschuerch: The required base URL varies by runtime environment (INT, TEST, PROD, etc.). Displaying only the PROD base URL can mislead developers, while exposing all environment-specific base URLs can introduce security risks.

Projects can also validate user authentication through other, more robust approaches (for example, server-to-server authentication flows).

@philipp suggested removing this information entirely, or at minimum exposing only the URL slug.

```suggestion
One method for the authentication mechanism is, to use the following KLP endpoint:

<Source code="{EnvSpecificBaseURL}/v1/session/subscribe" language="bash" />

The user is logged in, if the request returns with a server status code of `200` and the response body contains a valid `userdata` object (e.g. containing `name`, `surname`, `userType`, etc. properties).

In all other cases, the login link must be displayed!
```

oliverschuerch (changes requested):

apps/documentation/src/stories/components/header/header.docs.mdx
```diff
@@ -100,15 +101,30 @@ see the [Heading elements documentation](https://developer.mozilla.org/en-US/doc
...
+The Design System provides the UI components, while the application manages the authentication logic and decides which content to render in the `post-login` slot.
+
+### Fetching user data
+
+One method to retrieve the current Post user session data is via the session subscription endpoint.
+The URL shown below is for the **production** environment, adapt it to match your target environment (e.g. int, test).
+
+<Source code={SampleSessionSubscription} language="typescript" />
+
+If the request fails or returns no `data`, the user is not authenticated and the login link can be displayed instead.
```
oliverschuerch: Looking at the endpoint docs, I still don't fully agree about this sentence.
Because in the following cases, the endpoint responds with status code `200` and a response body of `{ "type": "ukn" }`:

- No Cookie
- Empty Session ID
- Cookie Present But Session Expired
- Not found

```suggestion
If the request fails or returns no valid `userdata`, the user is not authenticated and the login link can be displayed instead.
```

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #7895, opened 2026-06-18
feat(styles)!: remove unused entry files

### LINKED ISSUE #7894: Remove unused CSS files
> **Resolution**:
> 1. Remove src/elements.scss
> 2. No, remove if CDN Stats show that it's unused
> 3. Yes, create a ticket to document custom import setup @oliverschuerch  

 _Originally posted by @gfellerph in [#7524](https://github.com/swisspost/design-system/issues/7524#issuecomment-4477749188)_

### Tasks

- [x] Remove `src/elements.scss`
- [x] Remove `src/fonts.scss`
- [x] Create migration guide entry

gfellerph:
Removed unused CSS entry files `fonts.css` and `elements.css`. Use `/components/fonts.css` and direct imports from `/elements/...` instead.


packages/styles/src/elements.scss
alizedebray, line 1: This file is used in other entry files: https://github.com/swisspost/design-system/blob/main/packages/styles/src/post-default.scss#L10

alizedebray (changes requested):

packages/styles/src/elements.scss
gfellerph: Is this not referring to the folder and the `_index.scss` file within that folder as with the utilities for example? https://sass-lang.com/documentation/at-rules/use/#index-files

gfellerph:
@alizedebray I had to fix the tests because importing post-default.css also loaded Swiss Post Sans which was not present before and that changed the width of the text in a way that 19 items always overflowed the containerised header. Setting the header to full-width and the viewport to 1920px solved this for the arrow visibility test.

alizedebray (approved):

PR #7899, opened 2026-06-19
chore(styles): fix timepicker time alignement on ios

### LINKED ISSUE #7825: [bug]: Timepicker on iOS misaligned
### 🐞 Describe the bug

Image

Image

Timepicker value is center aligned on iOS, it should be left aligned. Also, floating label is floated initially even if value is empty.

Tested on iPhone 15 Pro Max, iOS 17.3.

myrta2302:
## 📄 Description

Fixed `time` input text alignment in Safari.

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


alizedebray (changes requested):
The issue with the floating label is not solved

myrtasakellariou:
> The issue with the floating label is not solved

The floating label is initially in the floated state because the `:placeholder-shown` is always `false` for `input[type="time"]` and so the applied style `&:not(:placeholder-shown)` is always `true` and makes the label float. This happens because the `placeholder` attribute is not applicable to the type="time" (https://html.spec.whatwg.org/#local-date-and-time-state-(type=datetime-local))
image
Additionally, the UI of the input `type="time"` is browser and OS dependent so in mobile devices it shows empty. 

Possible solution would be to exclude the input `type="time"` from the above rule so that the label does not initially float, apply `opacity:0` to  the `--:--` and show it on focus and after a value is entered.
But i could not find a selector that can differentiate whether a type="time" input is empty or filled.

So i think there is currently no solution for this.

alizedebray (approved):

PR #7923, opened 2026-06-23
chore(setup): prepare changelog-github for exiting pre mode

gfellerph:
## 📄 Description

This updates the old changelog-github fork for writing our changelogs in a way that requests to the github api are staggered in order to respect github api rate limits while exiting pre mode with more than 900 changesets. It also fixes a long standing bug that did not mention co-contributors in the changelog.


oliverschuerch (approved):
This package is going to be released on [npm](https://www.npmjs.com/package/@swisspost/design-system-changelog-github?activeTab=versions) and has 30 weekly downloads (probably form us).

However, I'll leave it to you if you add a changeset or not.

PR #7930, opened 2026-06-23
fix(tooltip): remove conflicting long-press event

### LINKED ISSUE #7851: [bug]: Tooltip behaves strange on touch devices
### 🐞 Describe the bug

During a long-press event, the tooltip shows up, but disapears after release. On click, the tooltip shows up briefly, but disappears immediately.

Tested on Android/Chrome 149 and on iOS/Safari by @leagrdv.

**The tooltip should be remove for these devices until a better solution is found.**

gfellerph:
Removed conflicting long press event from tooltips. As of now, tooltips are only supported for pointers and focused elements. Alternatively, use an information button (i) that triggers a popover, which is accessible for all users. The tooltip concept will be improved in a future release and remains partially inaccessible for touch devices.


.changeset/empty-books-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Removed conflicting long press event from tooltips. As of now, tooltips are only supported for pointers and focused elements. Alternatively, use an information button (i) that triggers a popover, which is accessible for all users. The tooltip concept will be improved in a future release and remains partially inaccessible for touch devices.
```
leagrdv, line 5: I would say this is a breaking change if projects have to manually change tooltips to popovers, no?

apps/documentation/src/stories/components/tooltip/tooltip.docs.mdx
```diff
@@ -17,6 +18,15 @@ import ComponentsPackageInstall from '@/shared/components-package-install.mdx';
 <div className="lead">For short, informative messages or explanations.</div>
 
 <Canvas sourceState="shown" of={TooltipStories.Default} />
+
+<post-banner type="warning">
```
leagrdv: Could you move this up above the canvas (or under the table)? So that the args table is always right after the canvas.

apps/documentation/src/stories/components/tooltip/tooltip.docs.mdx
```diff
@@ -17,6 +18,15 @@ import ComponentsPackageInstall from '@/shared/components-package-install.mdx';
...
+  **Limited accessibility on touch devices:** Tooltips currently have limited accessibility on
+  touch devices, because the long-press event used to trigger them conflicts with native gestures
+  in mobile browsers (such as text selection or the context menu). Touch users may therefore be
+  unable to reveal tooltip content reliably. If the information is essential, use the [accessible
+  alternative](#accessible-alternative-for-essential-content) below instead.
```
leagrdv: Could you change the link to the page ID + the `#accessible-alternative-for-essential-content`? Not sure if it's because of this, but clicking on that link opens the story itself in full page with no navigation or toolbar anymore

apps/documentation/src/stories/components/tooltip/tooltip.docs.mdx
```diff
@@ -28,6 +38,31 @@ The `<post-tooltip-trigger>` component enables you to show tooltips with customi
...
+    <post-icon aria-hidden="true" name="info"></post-icon>
+    <span class="visually-hidden">See more information</span>
+  </button>
+  <post-popover class="palette palette-accent" placement="top" text-close="Close" arrow>
+    <p class="mb-0">
+      Follow your letter's journey with automatic updates at key delivery milestones.
+    </p>
+  </post-popover>
+</post-popover-trigger>
+```
```
leagrdv: Can you add `sourceState="shown"` to the canvas to avoid duplicating the code here?

apps/documentation/src/stories/components/tooltip/tooltip.docs.mdx
```diff
@@ -28,6 +38,31 @@ The `<post-tooltip-trigger>` component enables you to show tooltips with customi
 <Canvas sourceState="shown" of={TooltipTriggerStories.Default} />
 <Controls of={TooltipTriggerStories.Default} />
 
+
+
+## Accessible alternative for essential content
+
+When the additional information is essential and must be available to every user, including on
```
leagrdv, line 46: Should we maybe mention that the limitation with going with the popover approach instead of the tooltip is that it's only gonna work with elements that don't trigger actions. Meaning if we have a button that triggers an event which has a tooltip on hover, then we cannot switch to the popover (or the (i) would have to be outside of the button which might look a bit off)

leagrdv (changes requested):

.changeset/empty-books-pull.md
gfellerph: I've went for patch because it's removing a broken feature. The popover is more of a recommendation because I think for most desktop use-cases the tooltip still works fine.

leagrdv (approved):

PR #7948, opened 2026-06-24
fix(select): floating label covering options with multiple set

### LINKED ISSUE #7768: [bug]: Multiple select with floating label
### 🐞 Describe the bug

The `select` HTML element with the `multiple` prop + floating label does not appear correctly.
When the first option is selected by default, it makes for a weird-looking UI. On focus, an arrow appears even though there is no dropdown.

### ✨ Expected behavior

- The label should take the whole width of the select, in order to avoid a selected option behind to be visible
- When an option is selected (also by default), the label should shrink and move up (currently it stays where it is)
- The select multiple should have a padding-top set for the options so that when scrolling, content does not go behind the label (mostly important for disabled state as the bg is transparent so content should not pass under)
- Since the multiple select does not have a dropdown, it should never show an arrow (_should also be fixed on the non-floating version_)

### 📷 Screenshots

Firefox (left: option selected/right: untouched)
Image

Edge (left: option selected/right: untouched)
Image

Edge: disabled (same thing on FF)
Image

gfellerph:


packages/styles/src/components/form-select.scss
alionazherdetska, line 1: The fix resolves the empty option case, but looking at the snapshots for Multiple - Floating Label (https://preview-7948--swisspost-design-system-next.netlify.app/iframe.html?id=snapshots--select&viewMode=story), the label still overlaps the first option in all variants except the `is-valid`, `is-invalid`, and `:disabled` ones:

Image

alionazherdetska (changes requested):

packages/styles/src/components/form-select.scss
gfellerph: Updated the padding of multiple selects to take offset into account.

alionazherdetska (approved):

PR #7949, opened 2026-06-24
fix(components): rework id logic on accordion items

### LINKED ISSUE #7937: [bug]: post-accordion with Angular
- [ ] setting an `id` attribute on the `post-accordion-item` does not override the internal id, therefore the `toggle` method cannot be used
- [ ] the `postToggle` event is not read as a `CustomEvent`, accessing the `detail` property required manual type assignation with `(event as CustomEvent<boolean>).detail`
- [ ] update documentation of the `postToggle` to inform about the target being the accordion-item that has just been toggled

alizedebray:
## 📄 Description

This PR removes the auto-generated id on the `post-accordion-item` that couldn't be overiden in an Angular application.
It also cleans up the accordion item component and the accordion docs.

## 🚀 Demo

https://preview-7949--swisspost-design-system-next.netlify.app/?path=/docs/4d1b4185-e04d-494a-ab38-2b56c1778b0b--docs&devModeEnabled=true

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/src/stories/components/accordion/accordion.stories.ts
```diff
@@ -23,17 +23,8 @@ const meta: MetaComponent<HTMLPostAccordionElement & HTMLPostCollapsibleElementE
     logoSrc: '',
   },
   argTypes: {
-    postToggle: {
```
alizedebray, line 26: This was incorrectly documented: the `post-accordion` has no `postToggle` event.

alionazherdetska (changes requested):
1. Since this PR removes the auto-generated ids, users now need to set the `id` on `<post-accordion-item>` themselves for `toggle()` to work. Could we update `accordion-custom-trigger.sample.ts` to make this clear? Something like:

```html
<post-accordion-item id="my-item">...</post-accordion-item>
```

```typescript
const accordion = document.querySelector('post-accordion') as HTMLPostAccordionElement;
accordion.toggle('my-item'); // id must match the one set on the item
```

Otherwise users might not realize they need to set the id manually and wonder why `toggle()` isn't working 😊

---

2. maybe we can also add a test to cover the main fix: something that verifies `accordion.toggle('my-item')` works when `id="my-item"` is set on the item? Feel free to ignore this one though since I know this is a blocker PR 😊

gfellerph (approved):

alionazherdetska (approved):

PR #7952, opened 2026-06-24
chore(components): update popovercontainer

### LINKED ISSUE #7940: [bug]: post-menu with Angular
- [ ] **`toggleMenu` is misleading, does not read as an event, but as a method should be `postToggle`**
- [ ] the buttons should be as wide as the menu, not smaller

### LINKED ISSUE #7951: post-popovercontainer update
- [x] the `show` and `toggle` methods should probably not open the menu if no target is provided otherwise it is shown at a random place in the view
- [x] docs for the `show` and `toggle` methods must be improved to explain what the `target` is, maybe just updating the signature for `triggerElement` instead of `target` would be enough

alizedebray:
## 📄 Description

This PR improve the documentation of the `show()`and `toggle()` methods on popup components. 
It also prevent the methods being called without a target from actually triggering the popovercontainer.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/components/post-popover/post-popover.tsx
```diff
@@ -59,9 +59,8 @@ export class PostPopover {
   }
 
   /**
-   * Toggle popover display
-   * @param target A focusable element inside the <post-popover-trigger> component that controls the popover
-   * @param force Pass true to always show or false to always hide
```
myrta2302, line 64: Was this maybe removed by mistake? `@param force Pass true to always show or false to always hide`

myrta2302 (approved):

packages/components/src/components/post-popover/post-popover.tsx
alizedebray: No, I removed it because it is not shown in the docs.

PR #7953, opened 2026-06-24
fix: release workflow

oliverschuerch:
## 📄 Description

This should fix the release action workflow, which failed for the last two weeks (unnoticed), because we added the explicit pnpm-package-manager definition in the `devEngines` field in our `package.json`.

Next.73 and Next.74 packages were not released to npm at all.
The docs on the other side have been updated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,8 +107,8 @@
     "tokens:lint": "pnpm --filter design-system-tokens lint",
     "tokens:lint:fix": "pnpm --filter design-system-tokens lint:fix",
     "tokens:usage": "node ./utilities/token-checker.cjs",
-    "changeset:publish": "pnpm changeset publish",
-    "changeset:version": "pnpm changeset version && pnpm install --lockfile-only",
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -58,6 +58,7 @@ jobs:
         uses: changesets/action@e0145edc7d9d8679003495b11f87bd8ef63c0cba
         env:
           GITHUB_TOKEN: ${{ secrets.SWISSPOSTDEVS_ACCESS_TOKEN }}
+          NPM_CONFIG_FORCE: 'true'
```
oliverschuerch: https://docs.npmjs.com/cli/v9/using-npm/config#force

gfellerph: I don't think we should do that. Why is there not an alternative?

gfellerph (changes requested):

alionazherdetska (approved):

gfellerph (approved):

PR #7955, opened 2026-06-24
feat(components): datepicker improvements

### LINKED ISSUE #7945: [bug]: post-date-picker improvements (behavior)
- [x] **form value includes the mask**
- [x] setting an initial value on the input is not showing as expected with the mask 

Image

- [ ] the form control is not marked as invalid if the value is invalid
- [x] mask is not optimal: 
  - you cannot change only the day or the month you have to erase from the year
  - you need to erase the separators, even though you did not add them
- [ ] the `lang` attribute is not taken into account to set the locale
- [x] an empty `locale` attribute is shown in the code snippet but it thrown an error when you copy/path the code 
- [x] the state of the date picker is not reset when hidden
   - if your in year/month view then close and re-open you are still in year/month view
   - if you close the date picker mid range, then reopen it, the range styles are still applied

alizedebray:
## 📄 Description

This PR updates the date picker to always return valid ISO dates as a value.

## 🚀 Demo

https://preview-7955--swisspost-design-system-next.netlify.app/?path=/docs/eb77cd02-48b2-42e1-a3e4-cd8a973d431e--docs&devModeEnabled=true

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/src/stories/components/datepicker/date-picker.stories.ts
myrta2302, line 1: In the Default Story:
1. Toggling to Inline in default example does not show the flyout, only the input. Also after toggling to Inline and Back the icon does not open the flyout anymore.
2.  Entering Min/Max dates does not disable any periods
3. Not sure if this is an issue, probably its normal. but i m just mentioning it. This happens if you set min/max in the story -after a value is already set that is not in the valid period- (This also happens in playground when you enter a date manually, and try to add a year that is out of the range,)

Image
Image

packages/components/src/components/post-date-picker/post-date-picker.tsx
myrta2302, line 1: 1. When input is not included in the Inline markup then this error is also thrown along with the dedicated error: 

Image

2. When the typed value is not completed (e.g. trying to set a date outside of the valid period) then the restriction is affecting the flyout header info. Not sure what should happen here.

Image

3. General question: Can the input really be validated depending on the date? Since the mask is already restricting the period entered.

myrta2302 (changes requested):
_Comments for Next milestone:_
1. Should all other cell attributes be immutable, so only `disabled` and `classes` can be customized via `cellConfig`?
2. General thought: should it be possible to clear the value, especially for inline?
3. Arabic does not have correct margins around the date grid
4. If you erase the range-start the whole mask is erased, it could keep the range-end value in case user wants to type or select only a new range-start date.
5. Range should be working for a period of one day

alizedebray:
> Should all other cell attributes be immutable, so only `disabled` and `classes` can be customized via `cellConfig`?

If I am not mistaken, this is already what I've implemented in the PR. Or do you mean you would like a more flexible customization on the contrary? 

> If you erase the range-start the whole mask is erased, it could keep the range-end value in case user wants to type or select only a new range-start date.
> Range should be working for a period of one day

That's Air Datepicker's behavior, not sure we can change it: https://air-datepicker.com/examples

packages/components/src/components/post-date-picker/post-date-picker.tsx
alizedebray: > When input is not included in the Inline markup then this error is also thrown along with the dedicated error

If the component is used incorrectly, errors are expected, so I don't think it's necessary to spend time handling those scenarios.

> Can the input really be validated depending on the date?

Please check the comment I added in the ticket yesterday, I guess it addresses this point: https://github.com/swisspost/design-system/issues/7945#issuecomment-4797250948

myrtasakellariou:
> > Should all other cell attributes be immutable, so only `disabled` and `classes` can be customized via `cellConfig`?
> 
> If I am not mistaken, this is already what I've implemented in the PR. Or do you mean you would like a more flexible customization on the contrary?

It is currently possible to manipulate e.g. the aria-labels of the grid dates from the console.
> 
> > If you erase the range-start the whole mask is erased, it could keep the range-end value in case user wants to type or select only a new range-start date.
> > Range should be working for a period of one day
> 
> That's Air Datepicker's behavior, not sure we can change it: https://air-datepicker.com/examples

"Range should be working for a period of one day"
There is a possibility this could work if toggleSelected is false. But at any case this would be for later. https://air-datepicker.com/docs?scrollTo=toggleSelected

apps/documentation/src/stories/components/datepicker/date-picker.docs.mdx
myrta2302: This should maybe be rephrased now that the input is attached for inline  version too.

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -825,38 +734,133 @@ export class PostDatePicker {
...
+      get: this.getInputIsoValue.bind(this),
+      set: this.setInputIsoValue.bind(this),
+      configurable: true,
+      enumerable: true,
+    });
+  }
+
+  private getIsoDates(): string[] {
+    const nativeVal = getNativeValue(this.dpInput);
+    const DIGITS_PER_DATE = 8; // dd(2) + mm(2) + yyyy(4)
```
oliverschuerch, line 764: This may conflict with the `baseMaskOptions`, which allow the year to be set between `101` and `9999` (`1` to `9999` was not working, because air-datepicker broke if the year was smaller than `101`).

Anyway, updating the `baseMaskOptions` to allow years from `1000` to `9999` would probably solve the conflict.

myrta2302 (approved):

PR #7958, opened 2026-06-24
feat(components): add `side-nav` slot to `post-header`

### LINKED ISSUE #7932: Define the concept for `sidenav trigger` placement in Application header
## Context

The **Application Header** design defined in Figma implies that the `post-side-navigation-trigger` lives to the left of the Application title. Currently there is no such slot in `post-header` in code.

### **Expected (Figma):**
Image

---

### **Current trigger placement:**
Image

---

## Problem

We need to define and implement a concept for how to resolve the `post-sidenavigation-trigger` placement in code.

alionazherdetska:
## Notes

1. **The trigger is guarded behind a title check** in the stories (`args.sideNav && args.title !== ''`). Without a title there is no local header on mobile and tablet, so the trigger would have nowhere to sit, on desktop the side nav is always visible anyway so toggling the control without a title would be misleading.

2. Not sure if we need a changeset since the side navigation still lives in raw components and this is where we refer the user.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/src/stories/components/header/header.docs.mdx
```diff
@@ -100,7 +100,7 @@ see the [Heading elements documentation](https://developer.mozilla.org/en-US/doc
 
 ### Application
 
-The Application header is a header variant intended for internal applications. It uses the `appearance: compact` styles bundle and is distinguished by the absence of any content in its global (yellow) part except from the Logo.
+The Application header is a header variant intended for internal applications. It uses the `appearance: compact` styles bundle and is distinguished by the absence of any content in its global (yellow) part except from the Logo. For applications that require in-app navigation, the Application header supports [Side Navigation](/?path=/docs/9f26d86e-7edb-5804-ac96-92g22f91c9d9--docs).
```
oliverschuerch:
```suggestion
The Application header is a header variant intended for internal applications. It uses the [compact styles bundle](/?path=/docs/e53e2de8-0bbf-4f70-babc-074c5466f700--docs&devModeEnabled=true#available-bundles) and is distinguished by the absence of any content in its global (yellow) part except from the Logo. For applications that require in-app navigation, the Application header supports [Side Navigation](/?path=/docs/9f26d86e-7edb-5804-ac96-92g22f91c9d9--docs).
```

Btw. why are the Side Navigation docs still in "raw components"? Is it planned to move it into "components" when styles are merged?

apps/documentation/src/stories/components/header/header.stories.ts
oliverschuerch, line 1: There is no way to show the side-nav in any of the included stories here (or I couldn't find one). And since you add it inside of a decorator, even if a user could configure the default story to show the side-nav, its markup would not show up in the code panel.

Will we keep the standalone side-nav docs page or integrate the story from there in this docs page?

oliverschuerch (changes requested):
Please link the related issue to the pr, if there is one.

And please add a changeset. It's you who add the functionality and this should be reflected in the changeset too, not only the change itself.

apps/documentation/src/stories/components/header/header.docs.mdx
alionazherdetska: Yes, this is the PR, which adds the styles and also moves the side-nav doc page from Raw components to the Components:
https://github.com/swisspost/design-system/pull/7972/changes

apps/documentation/src/stories/components/header/header.stories.ts
alionazherdetska: **This was the resolution of the daily yesterday:** to keep just the reference to the side nav story in the Header docs page in the Application header example for now and improve the concept after the release (regarding all the header example stories since there are already quite a few).

oliverschuerch (approved):

apps/documentation/src/stories/components/header/header.stories.ts
myrta2302, line 1: Should the Header + Side Nav Trigger **markup** be also mentioned - e.g. in a sample - in the Side Navigation page (and maybe also on the Application Header specific example?

myrta2302 (approved):

leagrdv (approved):

PR #7983, opened 2026-06-27
feat(header): implement account switch in the `user-menu`

### LINKED ISSUE #7898: Implement account switch in the user-menu
- [ ] Add an `accountSwitch: IconLinkConfig;` in the `postLogin` object of the header config
- [ ] Insert the account switch link as the first link in the global header user menu, before the other `userLinks`
- [ ] Add account switching logic based on v9 KLP login widget

alionazherdetska:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/internet-header/src/components/internal/UserMenu.tsx
```diff
@@ -18,8 +18,14 @@ export const UserMenu: FunctionalComponent<{ config: UserMenuConfig } & UserMenu
   textUserLinks,
   textAccessUserLinks,
 }) => {
+  const canSeeAccountSwitch = 
+  config.user.changeUserAndProfile === 'userAndProfile' ||
+  config.user.changeUserAndProfile === 'profile' ||
+  (config.user.userType === 'B2B' && config.user.canChangeCompany === true);
```
myrta2302: Should this also maybe check for  ` && config.user.changeUserAndProfile == null` ?

alionazherdetska: Good catch! Thanks for digging into the original logic

So the `changeUserAndProfile == null` check in `isOldChangeCompany()` was there to stop the B2B branch and the `'profile'` branch from overlapping. But either way both branches end up returning `true`, so the actual outcome never changes regardless of which one fires

I went with a flat OR here, happy to add the `== null` check back if you'd rather keep it closer to the `v9 version`.

oliverschuerch: Regarding the KLP Widget API Docs, it should not be necessary to handle `changeUserAndProfile === "profile"` anymore, because every user must have migrated to Swiss ID until summer 2023 (new users where forced to use Swiss ID since April 2022).

Image

---

Also how do we know `userType === "B2B" && canChangeCompany === true` without `changeUserAndProfile === "userAndProfile"` is enough, to show the AccountSwitch?

And I would rather check if `canChangeCompany === true && company !== ""`, instead of checking if it's a B2B client or not.

If you need more insights on the API docs, let me know 😉

packages/internet-header/src/models/user.model.ts
```diff
@@ -3,4 +3,7 @@ export interface UserConfig {
   surname: string;
   email: string;
   company?: string;
+  userType?: string;
```
oliverschuerch: I think the `userType` should be required.

packages/internet-header/src/models/user.model.ts
```diff
@@ -3,4 +3,7 @@ export interface UserConfig {
...
+  canChangeCompany?: boolean;
+  changeUserAndProfile?: 'userAndProfile' | 'profile';
```
oliverschuerch: Same for `changeUserAndProfile`. It should be require.

oliverschuerch (changes requested):
Can we see these changes somewhere in action?

packages/internet-header/src/components/internal/UserMenu.tsx
alionazherdetska: Hey Oli, thanks for the API docs, really helpful.

After going through widget-api.html I can see that userType and changeUserAndProfile are indeed required in both SessionData models, so you were right.

**Updated based on your feedback:**
- Made userType and changeUserAndProfile required in UserConfig
- Dropped changeUserAndProfile === 'profile'
- Replaced the B2B check with canChangeCompany === true && company !== '' as you suggested

packages/internet-header/src/components/internal/UserMenu.tsx
```diff
@@ -18,8 +18,13 @@ export const UserMenu: FunctionalComponent<{ config: UserMenuConfig } & UserMenu
   textUserLinks,
   textAccessUserLinks,
 }) => {
+  const canSeeAccountSwitch =
+    config.user.changeUserAndProfile === 'userAndProfile' ||
+    (config.user.canChangeCompany === true && config.user.company !== '');
```
oliverschuerch: Please add a comment per line so we still now, if the check is related to `B2C` or `B2B` or alternatively do it in code, for example like so:


```ts
  const isB2C = config.user.userType === 'B2C';
  const isB2B = config.user.userType === 'B2B';

  const canChangeUserAndProfile = (isB2C || isB2B) && config.user.changeUserAndProfile === 'userAndProfile';
  const canChangeCompany = isB2B && config.user.canChangeCompany && config.user.company;

  const canSeeAccountSwitch = canChangeUserAndProfile || canChangeCompany;
```

packages/internet-header/src/models/user.model.ts
```diff
@@ -3,4 +3,7 @@ export interface UserConfig {
   surname: string;
   email: string;
   company?: string;
+  userType: string;
+  canChangeCompany?: boolean;
+  changeUserAndProfile: 'notAvailable' | 'profile' | 'userAndProfile';
```
oliverschuerch:
```suggestion
  userType: 'B2C' | 'B2B';
  canChangeCompany?: boolean;
  changeUserAndProfile: 'notAvailable' | 'userAndProfile';
```

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #8054, opened 2026-07-02
chore(components): bug post date picker styles improvements

### LINKED ISSUE #7956: [bug]: post-date-picker improvements (styles)
- [x] icon has a blue focus ring
- [x] the calendar icon is misaligned when used with non-floating label
- [x] the validation icons overflow the calendar icon
      
Image

- [x] show a floating label datepicker in the main story since this is the recommended usage
- [x] if possible, add controls for the input (for example, the same as we currently have for the number input)

### LINKED ISSUE #8112: [bug]: accessible label for `post-date-picker`'s slotted input
## Problem

`post-date-picker` expects consumers to slot in their own `<input>`, but the component does nothing to ensure that input has an accessible name. If a consumer does not add a `<label>`, `aria-label`, or `aria-labelledby`, the input gets flagged as unlabeled.

Image

## Flagged by

- **axe DevTools** 
- **ARC Toolkit** 
- **Siteimprove**
- **WAVE**

## Question:

**1. Do we fix this in the component, or just document that consumers must label their own input**?

**2. If we do fix it in the component** should we add an optional `textInputLabel` prop now and then make it `@Required()` in `v11`?

myrtasakellariou:
## 📄 Description

This PR:
- Fixes the icon blue focus ring (issue on firefox)
- Updates the component, styles and markup to allow the option of a floating label
- Fixes the validation icon position for all cases
- Shows by default a floating label datepicker in the default story
- Adds controls for the input (as per number-input)
- Updates the form-input styles to use logical css properties

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/styles/src/components/form-input.scss
```diff
@@ -350,9 +351,15 @@ input.form-control {
...
+    inset-block-start: var(
+      --post-form-floating-validation-icon-block-start,
+      calc(#{tokens.get('input-sizing-height')} / 2)
+    );
     transform: translateY(-50%);
-    right: tokens.get('input-validation-icon-position-inline-end');
+    inset-inline-end: var(
+      --post-form-floating-validation-icon-inline-end,
+      #{tokens.get('input-validation-icon-position-inline-end')}
+    );
```
leagrdv, line 365: Why did you make changes here, when you're redefining it on the `post-date-picker.scss` file?

packages/components/src/components/post-date-picker/post-date-picker.scss
```diff
@@ -711,16 +712,24 @@ $current-day-inset: calc(var(--adp-cell-border-width) * -1);
...
     &[disabled] {
       color: tokens.get('post-datepicker-grid-day-cell-element-color-disabled-fg');
     }
+
+    &:focus-visible {
+      outline-offset: 4px !important;
+      outline-style: solid !important;
+      outline-width: tokens.get('focus-outline-width', helpers.$post-focus) !important;
+      outline-color: tokens.get('focus-outline-color', helpers.$post-focus) !important;
+    }
```
leagrdv, line 738: Can you use the `@mixin focus-style` instead here, with just an override on the outline-offset?

packages/components/src/components/post-date-picker/post-date-picker.scss
```diff
@@ -762,3 +775,33 @@ $current-day-inset: calc(var(--adp-cell-border-width) * -1);
...
+    inset-block-start: var(--post-form-floating-validation-icon-block-start);
+    transform: translateY(-50%);
+    inset-inline-end: var(--post-form-floating-validation-icon-inline-end);
+    z-index: 1;
+    pointer-events: none;
+    background-image: var(--post-signal-icon-error);
+    width: #{tokens.get('input-sizing-icon', components.$post-text-input)};
+    height: #{tokens.get('input-sizing-icon', components.$post-text-input)};
+  }
+}
```
leagrdv: Most of this has been duplicated from `form-input.scss`, can we just override the things that differ? Or have a mixin somewhere with the common code to avoid duplicated code

apps/documentation/src/stories/components/datepicker/date-picker.stories.ts
```diff
@@ -98,18 +104,125 @@ const meta: MetaComponent = {
...
+
+  const label = html`<label class="form-label" for="${args.id}-input"
+    >${getLabelText(args)}</label
+  >`;
+
   return keyed(
-    `${args.id}-${args.inline}`,
-    html`
+    `${args.id}-${args.inline}-${args.floatingLabel}-${args.validation}`,
+    html`<div dir=${dir}>
```
leagrdv: Can you remove the div? The `dir` prop could be set on the `post-date-picker` directly

packages/styles/src/components/form-input.scss
```diff
@@ -350,9 +351,15 @@ input.form-control {
   &:has(> input.form-control:-webkit-autofill.is-valid)::after {
     content: '';
     position: absolute;
-    top: calc(#{tokens.get('input-sizing-height')} / 2);
+    inset-block-start: var(
```
leagrdv, line 357: Are those changes necessary? They're being written in the `post-date-picker.scss` file too

leagrdv (changes requested):

packages/components/src/components/post-date-picker/post-date-picker.scss
myrta2302: I 've tried using the @mixin focus-style and the @mixin focus-style-custom but i did not manage to make it work. (The issue with the blue outline is reproducible only on Firefox)

packages/styles/src/components/form-input.scss
myrta2302: i removed the duplicate styles from datepicker.scss

packages/styles/src/components/form-input.scss
myrta2302: i removed the duplicate styles from datepicker.scss

packages/components/src/components/post-date-picker/post-date-picker.scss
myrta2302: This was removed.

apps/documentation/src/stories/components/datepicker/date-picker.stories.ts
```diff
@@ -98,18 +104,122 @@ const meta: MetaComponent = {
         category: 'Input',
       },
     },
+    floatingLabel: {
```
leagrdv, line 107: Can you hide all label/form related arguments when the datepicker is set to `inline`? Otherwise we can randomly add a label to it even though it's not an input

apps/documentation/src/stories/components/datepicker/date-picker.stories.ts
```diff
@@ -123,20 +233,25 @@ function render(args: Args) {
...
 }
 
 type Story = StoryObj;
 
-export const Default: Story = {};
+export const Default: Story = {
+  decorators: [
+    (story: StoryFn, context: StoryContext) => {
+      const locale = typeof context.args.locale === 'string' ? context.args.locale : '';
+      const dir = locale ? getLocaleDir(locale) : nothing;
```
leagrdv: You're adding it outside of the content the user can see, it would make sense to have this conditional div within the story itself

leagrdv (changes requested):

apps/documentation/src/stories/components/datepicker/date-picker.stories.ts
myrta2302: Ok. i updated to render it conditionally in the story - only for rtl.

apps/documentation/src/stories/components/datepicker/date-picker.stories.ts
```diff
@@ -123,17 +230,45 @@ function render(args: Args) {
...
+    id="${args.id}-input"
+    type=${args.inline ? 'hidden' : 'text'}
+    class=${args.inline ? nothing : `form-control${validation}`}
+    value=${args.value ?? nothing}
+    aria-describedby=${args.inline ? nothing : ariaDescribedBy}
+    ?aria-invalid="${VALIDATION_STATE_MAP[args.validation]}"
+    ?required="${args.requiredOptional === 'required'}"
+  />`;
+
+  const label = html`<label class="form-label" for="${args.id}-input"
```
leagrdv, line 262: Even when setting the datepicker to `inline`, the label is still showing in the code:

Image

leagrdv (changes requested):

leagrdv (approved):

PR #8067, opened 2026-07-06
feat: improve versions.json and add npm shields per package page

### LINKED ISSUE #6109: Add dist-tag specific package badges per documentation version
**User Story**

As a user of the Swiss Post Design System documentation, I want the package badges on each documentation version (v8, v9, next/v10) to display and link to the correct package version for that documentation, so that I always see the relevant version information and can easily access the corresponding package on npm.

Image

**Acceptance Criteria**

- [ ] The versions.json file includes a dist-tag for each package, mapping documentation versions (v8, v9, next/v10) to the correct npm dist-tag (e.g., version-8, latest, next).
- [ ] Each documentation fetches the versions.json from https://design-system.post.ch/assets/versions.json to receive the relevant data.
- [ ] Package badges in documentation pages use a service which allows to specify dist-tags for a badge (e.g. https://shields.io/badges/npm-version-with-dist-tag).
- [ ] Each badge displays the correct version for the documentation being viewed.
- [ ] Each badge links to the corresponding version on https://npmjs.com.

oliverschuerch:
## 📄 Description

This PR fixes the limitation of the the old version.json, which only contained the peer dependencies of the @swisspost/design-system-styles package. Since each of our packages has different peers, it makes sense to show them per package on the corresponding docs pages.

I've fixed the NPM dist-tags for some of our packages and added a step in our `release.yaml`, which updates every former major version in the `versions.json`. The current version is still updated by as before.
Note: This change is backwards compatible.

As stated in a [comment](https://github.com/swisspost/design-system/pull/8067#discussion_r3530088192) below, the preview is dynamically loading the `versions.json` file from prod (https://design-system.post.ch/assets/versions.json), which means, the preview URL is not showing the final output. It will only show up in the `chore(changesets): 🦋📦 publish packages (main)` PR, as soon as this one here is merged or if you check out the [branch](https://github.com/swisspost/design-system/tree/feat/improve-docs-package-version-shields) and visit the docs locally.

## Demo

https://preview-8067--swisspost-design-system.netlify.app/?path=/docs/e53e2de8-0bbf-4f70-babc-074c5466f700--docs&devModeEnabled=true

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/public/assets/versions.json
oliverschuerch, line 1: The peerDependencies are read from the `monorepo.json`. It can be created for example like so `pnpm m ls --depth=0 --json > monorepo.json`, which is done during the release workflow run. Within this file you get all the needed information per package, so you can summarize things dynamically (you don't need to load specific package.json files, etc.).

Doing something similar in the docs without this file would be way less strait forward and more of a manual task to always import/map package.json files together.

Since the preview of this pr uses the versions.json from prod, not the local one, it will show some wrong entries:

| Package Docs Page | Package History (opens in dialog) |
| :- | :- |
| image | image |

As soon as versions.json is updated correctly and deployed to prod, it will show up like so:

| Package Docs Page | Package History (opens in dialog) |
| :- | :- |
| image | image |
| image | image |

oliverschuerch:
@myrta2302 I know I once created a ticket for this and now I found it.
We already estimated priority and effort on September 1st 2025.
https://github.com/swisspost/design-system/issues/6109

So I guess the PR can be reviewed/merged as regularly.

apps/documentation/.storybook/addons/version-switcher/VersionSwitcher.tsx
```diff
@@ -66,12 +53,10 @@ function VersionSwitcher() {
...
                   className={['addon-dropdown__item', isActive].filter(c => c).join(' ')}
-                  key={version.title}
-                  href={version.url}
+                  key={v.title}
+                  href={v.url}
                 >
-                  <span className="item__title">
-                    v{isActive ? CURRENT_VERSION : version.version}
-                  </span>
+                  <span className="item__title">v{getVersion(v.version, 'major')}</span>
```
alionazherdetska, line 59: I noticed the dropdown now shows just the major (v10, v9), before it showed the full version like v10.2.0. Was this intentional to keep it cleaner?

apps/documentation/.storybook/addons/version-switcher/VersionSwitcher.tsx
alionazherdetska, line 1: Small thing: entries without dependency badges (v10, v5) are shorter than the rest, which makes the rows a bit uneven visually. Maybe a min-height can help?

Image

apps/documentation/src/shared/package-shields.tsx
```diff
@@ -0,0 +1,197 @@
...
+      <React.Fragment key={version?.title || ''}>
+        <div className={`d-sm-flex position-relative${renderTitle ? '' : ' mt-8 pt-8 border-top'}`}>
+          {renderVersionShields(version, renderTitle)}
+          {renderVersionPeerShields(version, renderTitle)}
+        </div>
+        {isCurrent && (
+          <div className="text-end mt-8">
+            <button
+              command="show-modal"
+              commandfor="PackageShieldsDialog"
```
alionazherdetska: `commandfor` doesn't seem to work in Firefox yet,  tested the "Show Package History" button and nothing happens. Should we add an onClick fallback so it works everywhere?

.github/workflows/release.yaml
```diff
@@ -303,23 +329,154 @@ jobs:
...
+              }
+            }
+
+            // Enumerate the workspace package names that exist on a release/v* branch.
+            const packageNames = worktree => {
+              const monorepo = JSON.parse(
+                execSync('pnpm m ls --depth=0 --json', { cwd: worktree, encoding: 'utf8' }),
+              )
+              return monorepo.map(pkg => pkg.name).filter(Boolean)
+            }
```
alionazherdetska, line 417: `npmView` and `branchExists` both have try/catch, but this one doesn't. If `pnpm m ls` fails on an old branch it would take down the whole step. Maybe wrap it too for consistency?

apps/documentation/src/shared/package-shields.tsx
```diff
@@ -0,0 +1,197 @@
...
+        data-position="top"
+        data-animation="slide-in"
+        style={{ width: '90vw', minWidth: 'min(1000px, 90vw)' }}
+      >
+        <div className="dialog-grid">
+          <h3 className="dialog-header" id="PackageShieldsDialogLabel">
+            Package History
+          </h3>
+          <div className="dialog-body">
+            {versions.filter(v => getVersionPackage(v)).map((v, i) => renderVersion(v, i === 0))}
```
alionazherdetska: The dialog body renders all past-version badges on page load even though the dialog is closed. The browser sees the `` tags in the DOM and fetches them immediately.

For `components-angular` with peer dependency badges it's 30+. These happen every time someone opens any package page, even if they never click "Show Package History".

Could we defer rendering the dialog content until it's actually opened? Something like a `dialogOpen` state that conditionally renders the body.

Image

apps/documentation/.storybook/helpers/get-versions-json.ts
```diff
@@ -0,0 +1,143 @@
...
+        title: '',
+        version: '',
+        description: '',
+        url: '',
+        dependencies: {},
+        peerDependencies: {},
+      }
+    );
+  });
+}
```
alionazherdetska: This fallback object is truthy, so if (currentVersion) checks in consumers will always pass even when the version wasn't actually found it will silently render empty badges instead of showing nothing or handling the error. Would returning null here be safer?

apps/documentation/.storybook/helpers/get-versions-json.ts
```diff
@@ -0,0 +1,143 @@
...
+
+export function getCurrentDistTag(): Promise<string> {
+  return getCurrentVersion().then(currentVersion => getDistTag(currentVersion));
+}
+
+export function getDistTag(version: Version): Promise<string> {
```
alionazherdetska, line 159: the dist-tag logic depends on the array order in versions.json, if someone reorders the entries this would silently break and badges would show `not found`. maybe worth a comment noting that newest-first order is requried?

apps/documentation/src/shared/package-shields.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+  function renderVersion(
+    version: VersionWithDistTag | null,
+    renderTitle: boolean = true,
+  ): React.ReactNode {
+    const isCurrent = version === currentVersion;
+
+    return (
+      <React.Fragment key={version?.title || ''}>
+        <div className={`d-sm-flex position-relative${renderTitle ? '' : ' mt-8 pt-8 border-top'}`}>
```
alionazherdetska: on some screen widths the package badges and peer dependency badges kind of blend together, maybe add a bit more gap between them?

Image

alionazherdetska (changes requested):
Way better than the manual tables! Left just a few small comments.

apps/documentation/.storybook/helpers/get-versions-json.ts
oliverschuerch: Good catch. Instead of commenting it, I made it sort order independent.

apps/documentation/.storybook/addons/version-switcher/VersionSwitcher.tsx
oliverschuerch: Philipp once told me the complete version doesn't matter here, since our docs always document the latest version of the current major. Therefore I removed it.
But if you feel different, we can ask the whole team.

alionazherdetska (approved):

PR #8088, opened 2026-07-08
chore(docs): improve side navigation and header docs

### LINKED ISSUE #7989: [bug]: Outside an iframe, the sidebar takes the whole screen
### 🐞 Describe the bug

Because of its fixed position, the sidebar takes over the whole left of the screen when it's not contained within an iframe. 
In the sidebar story, I've put it within an iframe to avoid the issue but now that the sidebar has been added to the header, when the argument is set to `true`, it looks buggy because it's going out of the header view.

_Quick fix solution_: Only have the sidebar option on the Application header story

### 📷 Screenshots

Image

- [ ] Disable the side navigation control int the controls table and link to the side navigation docs
- [ ] Add a second example for the application header with a side navigation

leagrdv:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/src/stories/components/side-navigation/side-navigation.docs.mdx
oliverschuerch, line 1: Should we add a message/banner to tell people this component is restricted to the application header?
A the moment it shows up as a regular component, everyone can use in any project.

I know it's document in the header docs, but if a dev visits this components docs page directly the information is missing.

oliverschuerch (changes requested):

apps/documentation/src/stories/components/side-navigation/side-navigation.docs.mdx
leagrdv: I've added the following: _The <post-side-navigation> component has been designed to be used alongside the application header._ in the trigger documentation part, so that people know what to use it with.

I would not state that it's "restricted" to the application header, as it's just a slot like any other in the headers, so technically if users want to use it with other types of header, they could.

apps/documentation/src/stories/components/header/header.stories.ts
```diff
@@ -144,12 +144,10 @@ const meta: MetaComponent = {
       },
     },
     sideNav: {
```
alizedebray, line 146: Actually I wonder if we shouldn't simply remove the control since it is not a header property. What do you think?

apps/documentation/src/stories/components/side-navigation/side-navigation.docs.mdx
gfellerph: Sites like the post shop are using a side navigation as well. "Designed to be used with" fits pretty well IMO.

oliverschuerch (approved):

apps/documentation/src/stories/components/header/header.stories.ts
leagrdv: Yes, makes sense! I've simply hidden it from the controls table, so that we can still use the argument to display the side navigation on the application header example

alizedebray (approved):

PR #8111, opened 2026-07-13
fix(components): fix accessibility violations for `<post-date-picker>` component

### LINKED ISSUE #7620: [bug]: fix accessibility violations fo `post-date-picker`
## 🐞 Describe the bug

Accessibility audit of the `post-date-picker` component revealed 3 violations. One is confirmed by all three tools (Axe, ARC Toolkit and Siteimprove browser extension) and must be fixed. Two additional violations are reported only by the Siteimprove browser extension and should be investigated for applicability.

---

```
Input has no accessible name
Description: The form control lacks an accessible name.
WCAG 2.2 Success Criterion:A [4.1.2 Name, Role, Value External link](https://www.w3.org/TR/WCAG22/#name-role-value)
<input type="text" class="form-control"
```

---

```
The accessible name of any interactive element should contain its visible text label.
Occurrences of this issue:
```

Image

---

```
This page element has a role that only makes sense in the context of another role.
```

Image

---

## 🛠️ To Reproduce

1. Open the `post-date-picker` story in Storybook **in the iframe** to avoid checks for Storybook-related markup
2. Run an accessibility audit using Axe, ARC Toolkit, or the Siteimprove browser extension
3. Observe the reported violations on the date input and calendar grid

alionazherdetska:
## 📄 Description

Fixes two accessibility violations in `<post-date-picker>`, flagged independently by axe DevTools, ARC Toolkit, and Siteimprove:

1. **Grid cells missing required `row` context** : AirDatepicker renders calendar cells as a flat list, so `role="gridcell"` had no `role="row"` ancestor between it and the `role="grid"` container. Scanners correctly flagged this as invalid ARIA structure.
2. **Label in Name mismatch (WCAG 2.5.3)**:  the nav title button's `aria-label` fully replaced its accessible name (e.g. visible "July 2026" vs. announced "Switch to year view"), so voice-control users saying the visible label couldn't activate the button.
3. A third violation was also flagged on this component ("Form field missing a label," on the slotted `<input>`), that one is out of scope here and is being addressed separately in #8054.

## How

- Added `wrapCellsInRows()`, which groups the flat cell list into `role="row"` wrapper divs, keeping the required `grid > row > gridcell` hierarchy intact. Uses CSS `subgrid`
- `gridObserver` disconnects before `wrapCellsInRows()` mutates the DOM and reconnects after, so the observer doesn't react to its own change
- Replaced the nav title's `aria-label` with a `.visually-hidden` span appended after the visible text, so the accessible name now contains the visible label plus the extra context, instead of replacing it.

## Testing

- Verified with ARC Toolkit / axe DevTools that both violations are resolved on days/months/years views.
- Manually confirmed no visual regression (subgrid inherits the parent's column tracks, so cell layout is unchanged).
- Manually confirmed keyboard navigation (arrows, month/view switching) still works correctly with the new row wrappers in place.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-date-picker/post-date-picker.tsx
+++ b/packages/components/src/components/post-date-picker/post-date-picker.tsx
@@ -229,6 +229,13 @@ export class PostDatePicker {
 
   private static readonly FLYOUT_OFFSET: number = 4;
 
+  // Cells per row for each view
+  private static readonly COLUMNS_BY_VIEW: Record<AirDatepickerViews, number> = {
+    days: 7,
+    months: 4,
+    years: 4,
+  };
+
   private currentViewMonth: number;
   private currentViewYear: number;
   private currentViewType: AirDatepickerViews = 'days';
@@ -370,7 +377,7 @@ export class PostDatePicker {
         break;
     }
 
-    return Array.from(this.dpContainer.querySelectorAll(selector));
+    return Array.from(this.dpContainer.querySelectorAll<HTMLElement>(selector));
   }
 
   private setActiveCell(date: Date, focusOnDate: boolean = true) {
@@ -647,6 +654,34 @@ export class PostDatePicker {
     }
   };
 
+  // ARIA requires role="grid" to own role="row", and role="row" to own role="gridcell"
+  // AirDatepicker renders cells flat, so we wrap them in row divs to satisfy that
+  private wrapCellsInRows(body: Element) {
+    this.gridObserver?.disconnect();
+
+    body.querySelectorAll(':scope > [role="row"]').forEach(row => {
+      while (row.firstChild) body.insertBefore(row.firstChild, row);
+      row.remove();
+    });
+
+    const cells = Array.from(body.children).filter(el =>
+      el.classList.contains('air-datepicker-cell'),
+    );
+    const columns = PostDatePicker.COLUMNS_BY_VIEW[this.currentViewType];
+
+    for (let i = 0; i < cells.length; i += columns) {
+      const row = document.createElement('div');
+      row.setAttribute('role', 'row');
+      row.classList.add('a11y-row');
+      body.insertBefore(row, cells[i]);
+      for (let j = i; j < i + columns && j < cells.length; j++) {
+        row.appendChild(cells[j]);
+      }
+    }
+
+    this.gridObserver.observe(body, { childList: true, subtree: false });
+  }
+
   private enhanceAccessibility(focusOnDate: boolean = true) {
     let body = this.dpContainer.querySelector('.air-datepicker-body--cells');
 
@@ -658,6 +693,7 @@ export class PostDatePicker {
     if (!body) return;
 
     body.setAttribute('role', 'grid');
+    this.wrapCellsInRows(body);
 
     this.getCells().forEach(c => {
       c.setAttribute('aria-selected', c.classList.contains('-selected-') ? 'true' : 'false');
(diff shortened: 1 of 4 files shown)


packages/components/src/components/post-date-picker/post-date-picker.tsx
leagrdv, line 1: I'm seeing a regression on your PR when in the year view and navigating between the years with the arrows:
If I'm on the next decade, and I try to navigate back to our decade (2020-2029), when our current decade appears, the focus is set to the current year (2026), not to the last element on the list.
Note that on main, there is another bug with the first/last element of the list which I've created a bug for: https://github.com/swisspost/design-system/issues/8121

packages/components/src/components/post-date-picker/post-date-picker.tsx
```diff
@@ -863,8 +899,8 @@ export class PostDatePicker {
     if (this.dpContainer) {
       const options: AirDatepickerOptions<HTMLDivElement> = {
         navTitles: {
-          days: `<button aria-label="${this.textSwitchYear}"><strong>MMMM yyyy</strong><post-icon name="chevrondown"></post-icon></button>`,
-          months: `<button aria-label="${this.textSwitchYear}"><strong>yyyy</strong><post-icon name="chevrondown"></post-icon></button>`,
+          days: `<button><strong>MMMM yyyy</strong><span class="visually-hidden">, ${this.textSwitchYear}</span><post-icon name="chevrondown"></post-icon></button>`,
```
leagrdv, line 915: This would announce "July 2026, _Switch to year view_" (depending on what user sets for `textSwitchYear`). I feel like it could be better if we'd update this to have the `textSwitchYear` first in order for user to be able to set something more intuitive such as "_Switch to year view, current month is:_ July 2026". We could document it better to go in that direction maybe. What do you think?

leagrdv (changes requested):

packages/components/src/components/post-date-picker/post-date-picker.tsx
alionazherdetska: For the years view this one actually was introduced by my PR. For the days view, though this bug is already on `next`. 

## Steps to reproduce it on Next (in the Days view":
- Start on **August 1**, press **ArrowLeft** → lands on **July 15** (today's date) instead of July 31.
- Start on **July 1**, press **ArrowLeft** → lands on **June 1** instead of May 31.

https://github.com/user-attachments/assets/f774fcfa-d125-49e4-aa29-4fb146326144


The bug is only when moving backward not forward both on next for the days and in my PR for the years.

Same root cause in both cases: crossing a month boundary triggers a grid redraw, which reactively re-runs `enhanceAccessibility()` via `gridObserver`. That second call has no way of knowing a keyboard navigation is in flight, so it defaults to focusing today's date and steals focus back from the correct cell.

Pushed a fix (`pendingFocusDate`) that covers both views, so the keyboard handler and the reactive redraw agree on the same target cell instead of fighting over it.

packages/components/src/components/post-date-picker/post-date-picker.tsx
alionazherdetska: Good catch, and I actually looked into this before replying, turns out reordering it would break WCAG 2.5.3 😅 The official W3C guidance specifically calls out "words before the visible label" as one of the known failure patterns, and gives "accessible name starts with the visible label" as the correct example instead.
Screenshot 2026-07-15 162259

https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html

The reason makes sense once you dig into it: voice control tools like  Voice Control match starting from the beginning of the accessible name. So if we put "Switch to year view" before "July 2026", someone who sees "July 2026" and says exactly that might not be able to activate the button anymore.

leagrdv (approved):

PR #8135, opened 2026-07-16
chore(components): Fix tabs scroll btns position

myrta2302:
## 📄 Description

Fixes the tabs scroll button position.

### DIFF
--- a/.changeset/thin-cats-wave.md
+++ b/.changeset/thin-cats-wave.md
@@ -2,4 +2,4 @@
 '@swisspost/design-system-components': patch
 ---
 
-Fixed tabs scroll buttons position.
+Fixed the position of the scroll buttons in the `post-tabs` component.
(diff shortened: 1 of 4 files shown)


.changeset/thin-cats-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': patch
+---
+
+Fixed tabs scroll buttons position.
```
alizedebray:
```suggestion
Fixed the position of the scroll buttons in the `post-tabs` component.
```

packages/components/src/components/post-tabs/post-tabs.scss
alizedebray, line 1: Please also remove the variable on line 16 as it is no longer used.

alizedebray (approved):

PR #8160, opened 2026-07-22
fix(components): make `post-listbox` blank-slate slot content reactive to DOM changes for async data sources

### LINKED ISSUE #8217: [bug]: `blank-slate` slot stays stuck when options are added after an initial empty mount (backend-driven data) for `post-autocomplete` component
### 🐞 Describe the bug

`post-listbox` decides whether to show the `blank-slate` slot based on `this.visibleOptions.length`, a plain class field, not a Stencil `@State()`/`@Prop()`. Mutating a plain field doesn't trigger a re-render in Stencil, so once the listbox mounts with zero options (the normal case for backend/async-driven autocompletes, since data hasn't arrived yet), it renders `blank-slate`. Even after real `<post-listbox-option>` elements are added later, the component never re-renders, since nothing reactive actually changed, so `blank-slate` stays stuck visible indefinitely, even though matching options genuinely exist in the DOM.

alionazherdetska:
## Background
 
We got a request in the support channel about `post-autocomplete` suggestions not showing up when the options come from a backend search.
 
## Problem
  
If `post-listbox` mounts with zero options, it gets stuck showing the blank-slate message forever - even once real options get added later. That's exactly what happens with async filtering: listbox mounts empty, backend responds with real streets, dropdown opens but still says "nothing to see here."

 
**Fix:** moved the blank-slate visibility logic out of JS and into CSS, using `:has()` to derive it live from the DOM instead of a JS-tracked state that Stencil wasn't reactively picking up on:
 
```scss
post-listbox [slot='blank-slate'] {
  display: none;
}
 
post-listbox:not(:has(post-listbox-option:not([hidden]))) [slot='blank-slate'] {
  display: block;
}
```
 
Since visibility is now derived straight from the actual DOM, it updates correctly whenever options are added, removed, or hidden, no matter when that happens, including after an initial empty mount.

## Additional info
In addition, this PR adds a focus handler to `post-autocomplete`. When `filterThreshold` is `0`, focusing the input (while empty) now emits `postFilteringEvent` with an empty string, the same way typing does. This lets consuming projects with an async/backend-driven data source respond to that event and render an initial list of all available options as soon as the user focuses the input, instead of only reacting to typed input.

## Note:
The changeset is missing because `post-autocomplete` is not yet published.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/styles/components/post-listbox.scss
oliverschuerch, line 1: I'd prefer to move the other styles applied to the slotted element as well (e.g. `padding`, `color`, etc.), so we can get rid of the slotted selector in the component entierly.

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #8163, opened 2026-07-23
fix(sonar): resolve accessibility, redundant-cast, and deprecated-API code smells

alionazherdetska:
## 📄 Description

This PR resolves a batch of open SonarCloud findings across multiple packages.

## Changes
- **Accessibility (WCAG)**
  - `live-region-atomic.sample.html`, `live-region-bad-angular.sample.html`, `live-region-good-angular.sample.html`: replaced `role="status"` with native `<output>` element
  - `index.html` (styles): fixed mismatched `label[for]`/`select[id]` pair
- **Confusing code / `this` usage**
  - `ngx-toastr.sample.ts`, `ngx-toastr-dismissible.sample.ts`: wrapped bare `this.toastr` calls in a proper class context
- **Redundant labels / declarations**
  - `theme-ag-grid-usage.sample.js`: fixed missing `const` causing label + undeclared-variable errors
- **Redundant type casts**
  - `create-class-update-rule.ts`: removed redundant `?` on already-optional `RuleType`
  - `report.ts`: removed unnecessary `as MergedReport` / `as IconSetStats` casts
- **Deprecated APIs**
  - `eslint.config.mjs` (components-angular, internet-header): replaced deprecated `ts.config(...)` wrapper with plain flat-config array
- **Missing test assertions**
  - `components.spec.ts`: added explicit `expect()` assertion to console-error test

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components-angular/eslint.config.mjs
```diff
@@ -6,7 +6,7 @@ import globals from 'globals';
 import ts from 'typescript-eslint';
 
 /** @type {import('eslint').Linter.Config[]} */
-export default ts.config(
+export default [
```
myrta2302, line 9: Could this flattening unintentionally broaden rule scope, so TypeScript and Angular TypeScript rules apply to non-TypeScript files as well? (same for the internet-header eslint.config.mjs)

alionazherdetska: I tested this with eslint `--inspect-config` against a sample `.js` file on both main and current PR branch. The only difference between the two: on this branch, the inspector shows the @typescript-eslint/* rules listed explicitly under a "disabled" section (61 rules, all marked off) for .js files, whereas on main they don't show up in that list at all. But in both cases, none of those rules are actually active for .js files, so it seems there is no scope leak, just a difference in how the inspector displays already-disabled rules:
image

myrta2302 (approved):

oliverschuerch:
I can't review the output, since the preview is not rendering correctly. Please check/fix and let me know, so I can have a look at it.

oliverschuerch (approved):
The following files would need a changeset, since these are patches of the docs, consumable by our users and imo, we should inform them about the updates.

- Toast samples (two files)
- Theme AG Grid sample (one file)

PR #8165, opened 2026-07-23
fix(sonar): resolve sonar redundant nesting and deprecated `word break` issues

alionazherdetska:
## 📄 Description
Fixes two SonarCloud findings:

1. **Redundant nested style rule (`css:S8769`)** in `_notification.scss`: removed a no-op `& { }` wrapper around `notification-body`'s declarations.
2. **Deprecated `word-break: break-word` (`css:S6810`)** in `post-breadcrumb-item.scss`, `post-stepper-item.scss`, `teaser.scss` :  replaced with `overflow-wrap: anywhere; word-break: normal; line-break: strict;`. Verified with long-text Storybook stories (Stepper, Card Teaser).

**Reference:** [[word-break: break-word is Deprecated — Why You Should Switch to overflow-wrap: anywhere](https://dev-yakuza.github.io/en/posts/word-break-break-word/)](https://dev-yakuza.github.io/en/posts/word-break-break-word/):  explains the deprecation and the min-content sizing bug that motivated this fix.

## 🚀 Demo

---
## 🔮 Design review
- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/styles/src/mixins/_notification.scss
```diff
@@ -56,13 +56,10 @@
...
-  & {
-    display: flex;
-    flex-direction: column;
-    justify-content: center;
-    gap: notification.$notification-body-gap;
-  }
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  gap: notification.$notification-body-gap;
```
oliverschuerch: The problem with mixins, adding css properties to the element where they were used on is, that they often produce a [mixed-decls](https://sass-lang.com/d/mixed-decls) warning! If this is the case usually depends on the CSS structure of the element styles and where the mixin is used.

That's why we had to add the `& {}` solution in the first place.

oliverschuerch (changes requested):
1. Please update your PR, so I can check the preview (which is not working currently).
2. Add a changeset.
3. What do you think, should we document your `work-break` solution in an ADR or somewhere else and/or create a mixin for it?

oliverschuerch (approved):
❤️

Issue #8039, opened 2026-07-01
[tokens] Tab tokens renaming proposals


sandra-post:
(split up from #7979)

| Token name  IS                                      | Token name renaming proposal                  | 
| :-------------------------------------------------- | :-------------------------------------------- | 
| `post.tabs.page.scroll.button.enabled-bg`           | `post.tabs.page.scroll-button.enabled-bg`     |
| `post.tabs.page.scroll.button.enabled-fg`           | `post.tabs.page.scroll-button.enabled-fg`     |
| `post.tabs.page.scroll.button.enabled-border-color` | `post.tabs.page.scroll-button.enabled-border` |
| `post.tabs.page.scroll.button.hovered-bg`           | `post.tabs.page.scroll-button.hovered-bg`     |
| `post.tabs.page.scroll.button.hovered-fg`           | `post.tabs.page.scroll-button.hovered-fg`     |
| `post.tabs.page.scroll.button.hovered-border-color` | `post.tabs.page.scroll-button.hovered-border` |
| `post.tabs.page.tab.selected-border-color`          | `post.tabs.page.tab.selected-border`          |
| `post.tabs.page.tab.enabled-border-color`           | `post.tabs.page.tab.enabled-border`           |
| `post.tabs.page.tab.hovered-border-color`           | `post.tabs.page.tab.hovered-border`           |
| `post.tabs.page.tab.disabled-border-color`          | `post.tabs.page.tab.disabled-border`          |
| `post.tabs.page.tab.selected-hovered-border-color`  | `post.tabs.page.tab.selected.hovered-border`  |
| `post.tabs.page.border-color`                       | `post.tabs.page.border`                       |
| `post.tabs.page.scroll-button.border.radius`        | `post.tabs.page.scroll-button.border-radius`  |
| `post.tabs.content.scroll.button.enabled-fg`        | `post.tabs.content.scroll-button.enabled-fg`  |
| `post.tabs.content.scroll.button.hovered-fg`        | `post.tabs.content.scroll-button.hovered-fg`  |
| `post.tabs.content.tab.hovered-border-color`        | `post.tabs.content.tab.hovered-border`        |
| `post.tabs.content.tab.selected-border-color`       | `post.tabs.content.tab.selected-border`       |
| `post.tabs.small.font.size`                         | `post.tabs.small.font-size`                   |

- [x] Rename tokens as above
- [x] Check only the update variable option

sandra-post:
[...]

Issue #8037, opened 2026-07-01
[bug]: [tokens] Minor Tab color corrections


sandra-post:
(split up from #7979)

## Bug 1

<img width="743" height="121" alt="Image" src="https://github.com/user-attachments/assets/2154b96a-b9f5-4f94-b199-303e98a500e6" />

```yml

post.tabs.page.tab.enabled-bg: 
  Theme: Cargo
    Palette: Accent
      ↳ IS: '#CDCCC8 / #CDCCC8' 
      ↳ SHOULD: '#050400 / #050400' 

```

- [x] Adjust palette color

## Bug 2

<img width="375" height="100" alt="Image" src="https://github.com/user-attachments/assets/3d6f4b81-41dc-48fa-928a-b4f37b277a9c" />

```yml
post.tabs.page.border-color:
  Theme: Post
    Palette: Default + Alternate
      ↳ IS: '#050400 / #CDCCC8' 
      ↳ SHOULD: '#050400 / #FAFAFA' 
    Palette: Accent
      ↳ IS: '#CDCCC8 / #050400' 
      ↳ SHOULD: '#FAFAFA / #050400' 
  Theme: Cargo
    Palette: Default + Alternate
      ↳ IS: '#050400 / #CDCCC8' 
      ↳ SHOULD: '#050400 / #FAFAFA' 

```

- [x] Check and adjust palette colors for the bottom line

## Bug 3

<img width="427" height="100" alt="Image" src="https://github.com/user-attachments/assets/f0dc72fa-81c8-4d98-8d72-9a8e6e5f5f8d" />

```yml

post.tabs.page.border-color:
  Theme: Cargo
    Palette: Brand
      ↳ IS: '#696864 / #696864'
      ↳ SHOULD: '#FAFAFA / #FAFAFA' 
```

## Bug 4

```yml
# MISSING/NEW TOKENS
post.tabs.page.tab.selected.hovered-bg: Same raw values as post.tabs.page.tab.selected-bg

```

- [x] Create semantic and component tokens for the post.tabs.page.tab.selected.hovered-bg

## Bug 5
Fix / add token for scroll buttons focus color 

~~Create palette token for scroll button focus color~~

_* This was probably added because of the focus offset issue which is being handled in the #8191_ 

## 📚 References

[...]

PR #8191, opened 2026-07-28
chore(components, styles): fix focus ring offset and update focus mixins

### LINKED ISSUE #8041: Fix focus ring offset
(split up from #7979)

### Problem

Since in dark mode, the tab border has the same color as the tab background, the focus ring has visually larger gap than in light mode.

Image

_Focus ring with offset -2_

Image

_Focus ring with offset -4_

### Possible solution
- ~~A token to be created for the outline offset via the palettes? @myrta2302 to try out~~
- **Simple approach:** Use an outline with a `-4px` `outline-offset` in both light and dark color schemes.

- **Complex approach:** Remove the focus outline and instead use two inset box shadows:
  
  ```css
  inset 0 0 0 2px light-dark(#f0efed, #000000),
  inset 0 0 0 4px light-dark(#000000, #ffffff)
  ```

  - The outer shadow is **2px** wide.
  - The inner shadow is **4px** wide.
  - Each shadow can be assigned its own color through the palette token layer.

  **Note:** This approach requires a transparent bottom border. As a result, the bottom padding should be reduced by the border width to keep the overall component height unchanged.

____________
Following Roundtable 21/07/26 the task is to :

- [x] Update the focus ring to leave a -2px space for all cases. For this a tabs specific outline-offset token: -6px should be created.
- [x] Update `focus-style` mixin to accept an optional `outline-offset` parameter so we don't have to override with `!important` after using the mixin

myrtasakellariou:
## 📄 Description

- Updated Tabs focus `outline-offset `with a new (Helper) token of -6px so that inner focus leaves at least 2px spacing from the visible edge of the element (border or edge).
- Updated the `focus-style` and `focus-style-custom` mixins to accept  `outline-offset` and `outline-color` as parameters.

**Snapshots**
https://preview-8191--swisspost-design-system.netlify.app/iframe.html?id=snapshots--page-tabs&viewMode=story

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


sandra-post:
@myrta2302 Design review

- #8041 : Now fixed, all good.
- #8037 : Now fixed, all good.

packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -99,7 +95,7 @@ $tabs-outline-offset: calc(
 
       // hovered
       &:hover {
-        border-color: tokens.get('post-tabs-page-scroll-button-hovered-border-color') !important;
+        border-color: tokens.get('post-tabs-page-scroll-button-hovered-border') !important;
```
oliverschuerch, line 98: Why did you rename this token?
The `...-button-hovered-border` token name should be reserved for the `border` shorthand property imo, while the more specific `...-button-hovered-border-color` token name should be used for the `border-color` property.

packages/components/src/components/post-tabs/post-tabs.scss
```diff
@@ -135,7 +131,7 @@ $tabs-outline-offset: calc(
       inset-inline-start: 0;
       width: 100%;
       height: tokens.get('post-tabs-border-width');
-      background-color: tokens.get('post-tabs-page-border-color');
+      background-color: tokens.get('post-tabs-page-border');
```
oliverschuerch, line 134: Same here.

packages/styles/src/components/tabs/_tab-title.scss
oliverschuerch, line 1: Again, is there a good reason, why you renamed those border-color tokens?

packages/styles/src/mixins/_utilities.scss
```diff
@@ -115,16 +118,35 @@
...
+  $outline-color: tokens.get('focus-outline-color', helpers.$post-focus)
+) {
+  &#{$additional-selector} {
+    outline-offset: $outline-offset !important;
+    outline: $outline-color none tokens.get('focus-outline-width', helpers.$post-focus) !important;
+  }
+
   // :has(:focus-visible) mimic a focus-visible-within pseudo-class
   &:is(:focus-visible, :has(:focus-visible), .pretend-focus)#{$additional-selector} {
+    outline-style: tokens.get('focus-border-style', helpers.$post-focus) !important;
```
oliverschuerch, line 133: We use token names `focus-outline-(offset|color|width)`. Why didn't we name this one here `focus-outline-style`?

If you didn't create this token in the current PR, you can ignore this CR.

packages/tokens/tokensstudio-generated/tokens.json
oliverschuerch, line 1: If you did not already do this, please ask @sandra-post for verification of this update.

I'm not sure if all the Figma variables related updates could be a problem for her!

I mean updates like these:
Image

oliverschuerch (changes requested):
Is it possible you have a different configuration of prettier

packages/components/src/components/post-tabs/post-tabs.scss
myrta2302: All these renamings have been suggested by Sandra to be consistent with the other components border tokens.  https://github.com/swisspost/design-system/issues/8039

packages/tokens/tokensstudio-generated/tokens.json
myrta2302: Sandra has checked everything. I thing this is always happening because of variables reindexing.

packages/styles/src/mixins/_utilities.scss
myrta2302: I did not create this token. It has been already used in several other components.

myrtasakellariou:
> Is it possible you have a different configuration of prettier

I don't think so. I did not find any not-commited configs that would affect this.

packages/styles/src/mixins/_utilities.scss
oliverschuerch: Sad, but ok...

packages/components/src/components/post-tabs/post-tabs.scss
oliverschuerch: Ok.
However, as a DEV I know this is not future proof and I want to avoid such cases.

**Explanation**
As soon as we need to specify another `border` related property to the same selector (e.g. `style`, `width`), we'll end up with a naming conflict (This is true for all shorthand properties).
You would probably prefer `border-color: #000;` over `border: #000;` in CSS and so we should align the naming in the token system.

I'll talk with @sandra-post too and come back to you with a resolution.

oliverschuerch (approved):

PR #8206, opened 2026-07-29
fix(component): skip layout space for unused slots for `post-footer`

### LINKED ISSUE #8148: [bug]: `swisspost-internet-footer` reserves layout space for empty grid columns even when no content is slotted in
### 🐞 Describe the bug

When a footer config omits `socialLinks`, `appStoreLinks`, and/or `companyLinks`, `swisspost-internet-footer` correctly skips rendering those slots, but the underlying `post-footer` component still renders its own wrapper `<div>`s for those slots and still reserves visual space for them in the grid layout, leaving a large empty gap in the footer.

### 🛠️ To Reproduce

1. Open the internet-header playground
2. Inspect the `<div class="footer-column">` element inside the `swisspost-intenet-footer`
3. Observe that it still takes space although the slots inside are empty

### 📷 Screenshots

Image

alionazherdetska:
## 📄 Description

This PR fixes `post-footer` reserving layout space for empty sections even when no content is slotted in.

Only the four `grid-{1|2|3|4}` sections tracked slot visibility and collapsed when empty. The other sections  (`socialmedia`, `app`, `businesssectors`, `meta`, `copyright`) were rendered unconditionally, so their wrapper elements stayed in the layout even with no content assigned to their slots. 

This PR extends the same slot-visibility tracking already used for the grid slots to cover all footer sections, so each one collapses individually when empty, and replaces the fragile CSS-based grid collapse with a value computed directly in the component for consistent behavior across breakpoints.

While testing on mobile, also found and fixed a related issue where the grid section has no content but the margin-bottom was still taking space:
Screenshot 2026-07-30 174202

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


oliverschuerch:
Please ask @sandra-post or @rouvenpost for a detailed design review!

packages/components/src/components/post-footer/post-footer.scss
```diff
@@ -136,20 +136,6 @@ footer {
...
-    margin-block-end: 0;
-    padding-block-end: 0;
-    row-gap: 0;
-    column-gap: 0;
-  }
-
-  // mobile: hide the accordion wrapper when all items are empty
-  post-accordion:has(> .d-none + .d-none + .d-none + .d-none) {
-    display: none;
-  }
```
oliverschuerch, line 151: Thanks for removing this abomination 😄

packages/components/src/components/post-footer/post-footer.tsx
```diff
@@ -57,17 +59,23 @@ export class PostFooter {
...
-      this.gridSlotDisplayed = { ...this.gridSlotDisplayed, [slotName]: hasContent };
+  private handleSlotChange(e: Event) {
+    if (e.target instanceof HTMLSlotElement) {
+      this.updateSlotDisplay(e.target.name, e.target.assignedElements().length > 0);
+    }
+  }
+
+  private updateSlotDisplay(slotName: string, hasContent: boolean) {
+    if (this.slotDisplayed[slotName] !== hasContent) {
+      this.slotDisplayed = { ...this.slotDisplayed, [slotName]: hasContent };
```
oliverschuerch, line 80: Why do we need to create a whole new object?
Can't we just update the corresponding slot property?

```suggestion
      this.slotDisplayed[slotName] = hasContent;
```

If this is some kind of limitation of the `@State` decorator, please add a comment, before this line, so no one accidentally breaks it.

packages/components/src/components/post-footer/post-footer.tsx
```diff
@@ -102,6 +110,10 @@ export class PostFooter {
   }
 
   render() {
+    const socialmediaDisplayed = this.slotDisplayed['socialmedia'];
+    const appDisplayed = this.slotDisplayed['app'];
+    const allGridSlotsEmpty = GRID_SLOTS.every(slotName => !this.slotDisplayed[slotName]);
+
```
oliverschuerch, line 121: Please add a `displayed` flag for the remaining slots as well (for consistency)?
e.g. `businesssectors`, `meta`, `copyright`.

And personally I would prefer a naming like `renderSocialmedia`, `renderApp`, etc.
1. Since we're in the render function, they are not yet displayed.
2. Grouping them with a prefix is usually easier to spot during maintenance, than postfixing them.
3. I would always name flags in the present tense, rather than in the past tense.

But I'm ok if you keep the naming as you prefer, as long as it is consistent 😉

oliverschuerch (changes requested):

packages/components/src/components/post-footer/post-footer.tsx
alionazherdetska: I actually tried it out. When I switched to 
```
this.slotDisplayed[slotName] = hasContent,
``` 
removing content from a slot no longer hid it in the DOM, the wrapper div stayed visible and kept its full height, even though the value was technically updated in memory:
Screenshot 2026-08-04 153804
Screenshot 2026-08-04 153813

 Turns out @State only triggers a re-render when this.slotDisplayed itself gets reassigned, not when you just change one of its properties in place.. That is why I added a comment above it explaining why, so it's clear for next time.

packages/components/src/components/post-footer/post-footer.tsx
oliverschuerch: feels good

oliverschuerch (approved):

PR #8222, opened 2026-07-31
fix(component): fix `postBreakpoint:device` listener leak caused by repeated bind calls for `post-megadropdown` component

alionazherdetska:
## 📄 Description

`breakpointChange` was bound via `.bind(this)` separately in `connectedCallback` and `disconnectedCallback`. Since `Function.prototype.bind()` returns a new function reference every time it's called, `removeEventListener` never matched the listener that was actually added on mount, it silently failed to detach it. The listener stayed attached to `globalThis` for the lifetime of the page, even after the component was removed from the DOM.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


oliverschuerch (approved):
This PR needs a changesets, imo, since `post-megadropdown` is a public component and the PR fixes production code.

PR #8223, opened 2026-07-31
fix(internet-header): collapse login, primary nav, and local nav link to icon-only on mobile

### LINKED ISSUE #8149: [bug]: Login button shows both icon and label on mobile in the `preconfigured header`
### 🐞 Describe the bug

In the preconfigured header, the login button displays both the icon and the text label on mobile viewports.

### 🛠️ To Reproduce

1. Open the internet-header playground
2. Resize the viewport to the `mobile`
3. Observe the `Login` button

### ✨ Expected behavior

It should show the icon only on mobile. 

### 📷 Screenshots

Image

alionazherdetska:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


myrta2302 (approved):

packages/internet-header/src/components/internal/Link.tsx
```diff
@@ -24,7 +24,7 @@ export const Link: FunctionalComponent<{ config: LinkConfig } & LinkProps> = ({
       aria-description={config.description}
       aria-current={isActive ? ariaCurrentWhenActive : undefined}
     >
-      {hiddenText ? <span class="visually-hidden">{config.text}</span> : config.text}
+      <span class={{ 'visually-hidden': !!hiddenText }}>{config.text}</span>
```
gfellerph, line 27: I don't see how switching from rendering a span to switching a class should fix this issue. The core logic stays the same and the result is almost the same. I suspect the underlying issue is elsewhere and that this change is a band-aid on top of the bug. There should be styles responsible for hiding the text that probably don't work anymore or not in every situation.

gfellerph (changes requested):

packages/internet-header/src/components/internal/Link.tsx
alionazherdetska: @gfellerph 

Thank you for the feedback!

In the self-managed header markup the `<span>` is present and the icon-only slots behave as expected. Before my fix, the markup of the pre-configured header didn't match the markup of the self-managed header.

If you compare the `post-login` or `search` slot markup between `pre-configured` and `self-managed header` in the current production docs, inspecting both shows the markup differs:

**1. Self-managed header:**
image

**2. Pre-configured header:**
image

All this PR does is align the two implementations, so consumers of the pre-configured header and consumers of the self-managed header get the same markup. This isn't a new problem either, POW hit the similar issue when integrating the self-managed header into their project and skipped `<span>` around Login:
image

However, if you think the styles should be improved to handle both cases: text wrapped in a `<span>` and text without one , I can do that. Just wanted to give you a bit more context first.

gfellerph: I see, so there are styles hiding a `<span>` on mobile and `hiddenText` is not updated based on viewport size as I assumed, but is a static parameter coming from the config. Then the fix is valid.

gfellerph (approved):

PR #8254, opened 2026-08-05
chore(deps, styles, components): update styles package to use icons and tokens package directly

### LINKED ISSUE #7263: Update Styles package to use icons package directly
Currently, icons and tokens are copied from their own packages into the styles package (via a temporary folder) during the build, due to previous pnpm restrictions.

We want to remove this temporary copy step and let styles reference icons directly from the icons package.

Have a look at pnpm hoisting: https://pnpm.io/settings#dependency-hoisting-settings

----------

### Tasks

- [x] Update Styles package to import the icons directly from the icons package
- [x] Remove temporarily-copy-icon-files task
- [x] Do the same for the tokens workaround

myrtasakellariou:
## 📄 Description

Removed old workaround where tokens and icons where copied within styles in /temp folders.
Added hoisting for tokens/icons in .npmrc.
Updated imports and gulp file.

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


alionazherdetska (changes requested):
**- Leftover references:**

1. The `src/tokens/temp/**` output is still listed in `packages/styles/turbo.json` line 6:

```json
"outputs": ["dist/**", "src/tokens/temp/**"]
```

2. `.gitignore` lines 17-18 still reference the temporary folders:
```
# Temporary files
src/tokens/temp/
src/icons/temp/
```

**- Transient Build Warning During Docs Start**

When running `pnpm docs:start`, there's a brief warning (~3-5 seconds) about stylesheet imports before the build settles. This appears to resolve on its own later on but I think should still be investigated:

Image

apps/documentation/.storybook/main.ts
alionazherdetska, line 1: Oli suggested a little bit different approach here, he tested and it worked:  instead of the custom sassPackageImporter (findFileUrl, walking node_modules by hand, checking .css/.scss extensions), we can just use dart-sass's built-in loadPaths:

```
preprocessorOptions: {
  scss: {
    loadPaths: [docsNodeModules],
  },
},
```

tested it and it resolves both the static @use cases and the dynamic meta.load-css() calls in the icons mixin, same as before, just with a lot less code. also lines up better with what we already do in stencil.config.ts for components and internet-header, which both just use `includePaths: ['node_modules']` so this keeps things consistent instead of storybook having its own one-off pattern.

if we go this way we can drop findFileUrl/existsSync/pathToFileURL completely.

alionazherdetska (changes requested):
It works even for the legacy projects, like Paprica😊

alionazherdetska (dismissed):

oliverschuerch (dismissed):

alionazherdetska (dismissed):

oliverschuerch (approved):

PR #8263, opened 2026-08-06
fix(components): fix collapsed header not expanding on `post-language-menu-trigger` focus while scrolled

alionazherdetska:
## 📄 Description

While working on improving test coverage for `post-header`, I came across a bug: focusing `post-language-menu`'s trigger while the collapsed (scrolled) header is shown does not expand the header back to full height. Every other focusable element in the header already triggered the expansion correctly.

https://github.com/user-attachments/assets/1ffc7b37-24a7-4b50-b1db-892ca31e436b

 
## Root cause
 
`onFocusChange()` decides whether to expand the header by checking `document.activeElement?.matches(':focus-visible')`. This only works when the focused element sits directly in the light DOM. `post-language-menu` renders its trigger button inside its own shadow root, so when that button is focused, `document.activeElement` resolves to `<post-language-menu>` itself (the shadow host), not the button. The host was never the literal focus target, so `:focus-visible` never matched and `data-expanded` was never set.
 
## Fix
 
Added a `getDeepActiveElement()` helper that walks nested shadow roots (`activeElement.shadowRoot.activeElement`) to find the actually focused element before checking `:focus-visible`. This also covers any other slotted component that's itself a shadow-DOM element with an internal focusable, not just `post-language-menu`.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


oliverschuerch (approved):
@alionazherdetska 
Not part of this ticket, but somehow related:
Changing the focus to one of the interactive elements in the global header changes the pages scroll location.

This seems to be technically unnecessary, unwanted from perspective of user experience and may be a native browser behavior (tested on Mac, Chrome 151).

Can you please create a ticket and add it to the Roundtable, so we can discuss with UX if we should prevent the page from scrolling, in this cases?

https://github.com/user-attachments/assets/c48a221a-e4b3-483c-b9e0-750d25af8618

PR #8276, opened 2026-08-10
fix(components): always show, but dynamically disable pagination controls

### LINKED ISSUE #8030: [bug]: post-pagination
With the current design of the pagination, arrow buttons are moving preventing the user from clicking continuously. 

If you are on page 1, there is no "Previous" button.
If you click the "Next " button you reach page 2. 
There the "Previous" button appears making all other buttons shift to the right.
You then have to move your mouth to be able to click the "Next" button again. 

___

> Options:
> - disable the buttons that can't be clicked so they take up space in the UI but communicate their inability to act
> - let users click the buttons without any action happening (probably worse for accessibility) 

 _Originally posted by @gfellerph in [#7941](https://github.com/swisspost/design-system/issues/7941#issuecomment-4851836872)_

yomonom:
## 📄 Description

Updated the `post-pagination` component to always show, but dynamically disable the navigation control buttons.

## 🚀 Demo

https://github.com/user-attachments/assets/bcb1993f-32e4-4967-b99d-6c1f2bad7655

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-pagination/post-pagination.tsx
+++ b/packages/components/src/components/post-pagination/post-pagination.tsx
@@ -21,12 +21,37 @@ const MAX_ELLIPSIS_COUNT = 2; // Maximum ellipsis elements
 const GAP_THRESHOLD_FOR_PAGE = 2; // Show page number instead of ellipsis when gap is this value
 const MIDDLE_RANGE_START = 2; // Middle range starts from page 2 (page 1 is always shown separately)
 
+const enum PaginationItemType {
+  Page,
+  Control,
+  Ellipsis,
+}
+
 /**
  * Type-safe pagination item definition using discriminated union.
  */
-type PaginationItem = { type: 'page'; page: number } | { type: 'ellipsis' };
+type PaginationItem =
+  | { type: PaginationItemType.Page; page: number }
+  | { type: PaginationItemType.Control; action: ControlAction; disabled: boolean }
+  | { type: PaginationItemType.Ellipsis };
 
-type SectionType = 'none' | 'page' | 'ellipsis';
+const enum ControlAction {
+  Previous,
+  Next,
+}
+
+interface ControlAttributes {
+  icon: string;
+  label: string;
+  className: string;
+  onClick: () => void;
+}
+
+const enum SectionType {
+  None,
+  Page,
+  Ellipsis,
+}
 
 @Component({
   tag: 'post-pagination',
@@ -297,7 +322,7 @@ export class PostPagination {
   private updatePagesWithValidation() {
     const totalPages = this.getTotalPages();
     this.page = this.clampPageToValidRange(totalPages);
-    this.generatePages(totalPages);
+    this.generateItems(totalPages);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -21,12 +21,38 @@ const MAX_ELLIPSIS_COUNT = 2; // Maximum ellipsis elements
 const GAP_THRESHOLD_FOR_PAGE = 2; // Show page number instead of ellipsis when gap is this value
 const MIDDLE_RANGE_START = 2; // Middle range starts from page 2 (page 1 is always shown separately)
 
+const enum PaginationItemType {
```
alionazherdetska: Just checked const enum isn't used anywhere else in the repo, we stick to regular enum (like in post-avatar) or string unions (like the original code here). 

Could we keep the string unions? It'd stay consistent with the rest of the codebase, and it'd also shrink the diff quite a bit, right now the enum swap touches a lot of lines which makes it harder to focus on the actual behavior change. String unions are also the [generally recommended](https://www.typescriptlang.org/docs/handbook/enums.html#const-enum-pitfalls) pattern, same type safety, just simpler.

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -710,50 +790,66 @@ export class PostPagination {
...
           type="button"
-          class={`pagination-control-button btn btn-icon btn-secondary ${isPrev ? 'prev-button' : 'next-button'}`}
-          aria-label={label}
-          onClick={onClick}
-          onKeyDown={e => this.handleKeyDown(e, onClick)}
+          class={`pagination-control-button btn btn-icon btn-secondary ${attributes.className}`}
+          aria-label={attributes.label}
+          onClick={attributes.onClick}
+          onKeyDown={e => this.handleKeyDown(e, attributes.onClick)}
           tabIndex={0}
```
alionazherdetska: The button has disabled={disabled} but still sets tabIndex={0} unconditionally. You could gate it with `disabled ? -1 : 0`.

packages/components/src/components/post-pagination/post-pagination.tsx
```diff
@@ -710,50 +790,66 @@ export class PostPagination {
...
   /**
-   * Renders a pagination item.
+   * Renders a control button.
    */
-  private renderItem(item: PaginationItem, index: number) {
-    return item.type === 'ellipsis'
-      ? this.renderEllipsis(`ellipsis-${index}`)
-      : this.renderPageButton(item.page);
-  }
+  private renderControlButton(action: ControlAction, disabled: boolean, dummy: boolean) {
```
alionazherdetska: Took me a moment to figure out what dummy does here if I open this code fresh I'd need to read through both branches to get it. When its false you get a full `<li>` with events and labels, when its true its a bare `<button>` with different classes and no wrapper. 

These are basically two different things squeezed into one method. I know its tempting to reduce duplication but sometimes keeping things separate is just eaiser to follow. Maybe keep the hidden markup inlined in renderHiddenItems like before or pull it into its own small method: either way its clear what you're looking at without chasing a boolean around

alionazherdetska (changes requested):
Nice fix!
Just a few suggestions.

packages/components/src/components/post-pagination/post-pagination.tsx
yomonom: True. I extracted the "dummy" button into its own helper method.

packages/components/src/components/post-pagination/post-pagination.tsx
yomonom: Good call; I went a bit overboard with the enums. I reverted back to string unions.

alionazherdetska (approved):

PR #8290, opened 2026-08-11
fix(components): ensure post-autocomplete only executes query when required

### LINKED ISSUE #8199: [bug]: `post-autocomplete`: `filterThreshold` isn't respected once a consumer takes over filtering, empty query still fires below threshold
## Background

Ran into this wiring up `post-autocomplete` inside the Angular app to a real backend search (async filtering via `postFilteringEvent` + `preventDefault()`). With `filterThreshold` set to 4, typing anything shorter than 4 characters still fires `postFilteringEvent` with an empty string: so if you're doing async filtering, you end up firing a full, unfiltered fetch on every single keystroke below the threshold, instead of just not firing anything yet.

Here's what it actually looked like backspacing "bahn" down one character at a time - every keystroke below the threshold fired its own request with an empty query:

```
GET /api/streets?query=
GET /api/streets?query=
GET /api/streets?query=
GET /api/streets?query=bahn
```

If you want to reproduce this against a real backend rather than mocking one, I've got a small test API deployed here: [street-finder-api.onrender.com/api/streets].

## Root cause

In `handleInput` (`post-autocomplete.tsx`):

```typescript
const query = value.length >= this.filterThreshold ? value : '';
const { defaultPrevented } = this.postFilteringEvent.emit(
  query && query.length >= this.filterThreshold ? query : '',
);
```

Anything below `filterThreshold` just collapses to `query = ''`, and that gets emitted every time, no way to tell "not enough characters yet" apart from "input is genuinely empty." For the component's own built-in filtering that's probably fine (empty query there just means "show everything"), but once you've taken over filtering yourself, both cases look identical and both trigger a fetch.

yomonom:
## 📄 Description

- Prevents `post-autocomplete` from firing `postFilterEvent` with an empty query more than once before `filter-threshold` is exceeded.
- Ensures that the count described in the `output` element is in sync with the number of options rendered through custom filtering (see #8291).
- Mitigates the issue of the same filtering results being displayed when the input is re-entered after an option has been selected (see #8291).

## 🚀 Demo

https://github.com/user-attachments/assets/f36b7af8-d467-4862-9bb2-e6f4aa16b727

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/components/post-autocomplete/post-autocomplete.tsx
```diff
@@ -136,31 +135,67 @@ export class PostAutocomplete {
...
+  private async runQuery(query: string, alwaysExpand: boolean = false) {
+    // Prevent running the same query twice.
+    if (this.lastQuery !== (this.lastQuery = query)) {
+      await this.filterListBox(query);
+    }
+
+    // Always update the count if the listbox is expanded.
+    if (this.expanded) this.announceCount();
+    // Automatically show the listbox if the query is non-empty.
+    else if (alwaysExpand || query) await this.showListBox();
```
oliverschuerch: Please use the curly braces for this if-else statement for readability.
I personally like this braceless style, but I prefer consistency ;)

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
oliverschuerch, line 1: There is an issue with the component (not sure if you added it or it was there before):

1. Select a value in the autocomplete
2. Move on (click aside the input)
3. Click on the input again
  => listbox opens
4. Click on the input a second time (or perform a doubleclick)
  => listbox closes, filtering not working
  => only klicking `esc` or leaving the input and focus it again resets the component now, so I see the listbox and can choose a new value

https://github.com/user-attachments/assets/31d0f6ba-247c-471a-8773-2962b542e8b3

oliverschuerch (changes requested):

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
yomonom: This is a bug on main as well. I don't think it's worth addressing right now as the autocomplete will probably be reworked in the future anyway.

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
yomonom: Done.

oliverschuerch (approved):

packages/components/src/components/post-autocomplete/post-autocomplete.tsx
oliverschuerch: If there is already a ticket for reworking the component, please link this CR to it, so the bug can be addressed there.

PR #8326, opened 2026-08-18
fix(components): automatically hide popovers.

yomonom:
## 📄 Description

Automatically hide popovers when their target element moves outside of view, taking 'safe areas' (such as `post-header`) into account.

> [!Note]
> This change introduces a small wrapper around the Floating UI DOM-Platform to ensure that edge cases are handled correctly. For this reason, it introduces `@floating-ui/utils` as a dependency, which was already implicitly included by `@floating-ui/dom`.

https://github.com/user-attachments/assets/26da005c-edbd-484e-9b2f-95539fce2efa

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


yomonom:
Note: We could introduce a `clip` attribute to the `post-popovercontainer` component to specify whether the clipping behaviour should be enabled. This would allow us to opt in to automatic hiding for popovers only, not for tooltips or any other elements that use the `post-popovercontainer` component, thereby reducing the potential impact on other components.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -355,25 +356,21 @@ export class PostPopovercontainer {
...
-   */
-  private getHeaderHeight(): number {
-    const header = document.querySelector('post-header');
-    return header ? Number.parseFloat(getComputedStyle(header).height) : 0;
-  }
+  private async updatePosition() {
+    const { x, y, middlewareData, placement } = await this.calculatePosition();
+
+    // Hide the popover if the target is outside the viewport
+    if (middlewareData.hide?.referenceHidden) {
```
alionazherdetska, line 374: think we're double closing here. `hidePopover()` fires `beforetoggle` which already routes through `handleToggle()` -> `close()`, then we call `this.close()` again right after. tested it with some quick logging and yeah, get two calls for one scroll-hide:

Image

not a leak (autoUpdate gets cleared fine on the first call, second is a no-op there) but `postBeforeToggle`/`postToggle`/`postHide` all fire twice, which post-tooltip's `handleBeforeToggle` picks up too. can we just drop the explicit `this.close()` here and let `hidePopover()` do it like everywhere else in this file?

```ts
if (middlewareData.hide?.referenceHidden) {
  this.host.hidePopover();
}
```

packages/components/src/utils/floating-ui.ts
alionazherdetska, line 1: We have unit tests for most of the utils (breakpoints, clamp, debounce, etc) but floating-ui.ts doesn't have a spec yet. given how much this PR is riding on the safe-area logic (getContainingBlock check, header-detection, the cache) do you think it's worth adding a floating-ui.spec.ts for it, or are you thinking cypres covers this well enough on its own?

Image

alionazherdetska (changes requested):
maybe worths a cypress test for this? not blocking though

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
yomonom: Yea. Not sure how the additional `this.close` got here.

yomonom:
> maybe worths a cypress test for this? not blocking though

Good idea. Perhaps this should be done as part of a separate ticket? This one is time-boxed to one day, and together with the first attempt, I have already spent around one and a half days on this...

packages/components/src/utils/floating-ui.ts
yomonom: I'm not sure if a unit test would be beneficial here, as it would require a lot of mocking. Floating UI mostly uses Playwright tests themselves, so I think proper end-to-end testing should suffice.

alionazherdetska:
> > maybe worths a cypress test for this? not blocking though
> 
> Good idea. Perhaps this should be done as part of a separate ticket? This one is time-boxed to one day, and together with the first attempt, I have already spent around one and a half days on this...


@yomonom 
We usually keep tests in the same ticket. If we split it off it might land in a different milestone or get deprioritized, and then it ships without any guards.

Going over the estimate a bit is fine, these are hard to size right at triage anyway, especially with how many components use popovercontainer.

If you'd rather keep this PR smaller though, you could do the test in a follow-up PR instead, but then it becomes part of a PR train (this one has to merge first, then the next depends on it, etc.), up to you.

packages/components/src/utils/floating-ui.ts
oliverschuerch: Generally, we don't want to unit test external dependencies, as they should (🤞🏼) implement their own unit tests.
Adding e2e tests to check if we're using the lib correctly is always needed/wanted on the other side.

I understand it's difficult to set up a unit test for this deps wrapper utility. I would appreciate, if you could at least try it using copilot, since you add a lot of custom functionality here.

oliverschuerch:
> > maybe worths a cypress test for this? not blocking though
> 
> Good idea. Perhaps this should be done as part of a separate ticket? This one is time-boxed to one day, and together with the first attempt, I have already spent around one and a half days on this...

One day to find a solution!
If none can be found, come back to the team and ask for advice...

Since you found a working solution, it's fine in my opinion to add the e2e test in the same ticket, since it will finally proof, the solution is working as expected.
No need for an additional ticket.

packages/components/package.json
```diff
@@ -105,7 +105,8 @@
     "copy:styles:components-angular": "copyfiles -f www/build/post-components.css ../components-angular/src/stencil-generated/"
   },
   "dependencies": {
-    "@floating-ui/dom": "1.7.6",
+    "@floating-ui/dom": "1.8.0",
+    "@floating-ui/utils": "0.2.12",
```
oliverschuerch, line 109: Just wanted to ask if this is a stable version.
Seems to be pretty stable... 😅

Image

packages/components/src/utils/floating-ui.ts
oliverschuerch, line 1: Please add jsdoc comments per function, to describe input/output.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
oliverschuerch, line 1: The changes here also affect other components than just the popover!
It is used in:

- `post-date-picker`
- `post-listbox`
- `post-menu`
  - `post-breadcrumbs`
  - `post-language-menu`
- `post-popover`
- `post-tooltip`

If this is ok (ask the team), then we should consider to reflect that fact in the changeset (mention the post-language-switch, post-tooltip, etc.).

https://github.com/user-attachments/assets/b6ef169b-150e-4f46-bbf7-432214f3b659

oliverschuerch (changes requested):

packages/components/package.json
yomonom: Yea, it's what's used by `@floating-ui/dom` (I've aligned the versions).

packages/components/src/utils/floating-ui.ts
yomonom: The caching logic is adapted from `@floating-ui/dom`. The only 'real' custom logic is the `applySafeArea` function, which needs to be tested through e2e, imho. I've tried to extend the new e2e tests to cover all possible branches.

I'm happy to try to add some unit tests if you want me to, though.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
yomonom: Should I move this issue to the dev roundtable, then?

If these changes are unwanted in other components, we could also restrict the 'clipping' behaviour through an extra attribute on `postpopover-container`, as this is not a public-facing component anyway. Then we could just change out `computePositionWithSafeArea` with the regular `computePosition` when `clip` (or something the like) is set to `false`, restricting the behaviour to just the popovers themselves.

yomonom:
I've added some new utility functions for testing. These could probably also be used in other tests that interact with popovers. I didn't touch them though, as I feel like this PR is already getting big enough.

It's probably worth to add that to another ticket though. Maybe together with the broken popover tests that still contain the `Promise.all`?

packages/components/cypress/e2e/tooltip.cy.ts
yomonom, line 1: Prettier went wild here (the formatting is 'correct' now though, so I left it in). The only real change is the import on line 2.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
yomonom: Btw, the behaviour seen in the video is specified by `post-language-menu` itself (see `hideMenuOnScroll`).

I have actually included an e2e test that asserts this behaviour does not happen for the `popover-container` by default. (see 'inside header' → it '_should stay open while scrolling, as the header does not affect its own content_').

alionazherdetska: @oliverschuerch The behaviour for the language menu was alreay like this. The PR of Yanis did not alter it, closing the lang menu on scroll was implemented as a part of this PR:
https://github.com/swisspost/design-system/pull/7445

Thus, the users are alreday aware and no need for the new changeset to be added (at least for the lang-menu). 

Probably worths a changeset at least for the `user-menu` since initially it stayed open on scroll and after this PR will be merged it wil collapse on scroll:

https://github.com/user-attachments/assets/b6c0f7ce-775c-471e-8828-123cd3bf5cea

oliverschuerch: > Should I move this issue to the dev roundtable, then?
> 
> If these changes are unwanted in other components, we could also restrict the 'clipping' behaviour through an extra attribute on `postpopover-container`, as this is not a public-facing component anyway. Then we could just change out `computePositionWithSafeArea` with the regular `computePosition` when `clip` (or something the like) is set to `false`, restricting the behaviour to just the popovers themselves.

Yes please!

oliverschuerch: > Btw, the behaviour seen in the video is specified by `post-language-menu` itself (see `hideMenuOnScroll`).
> 
> I have actually included an e2e test that asserts this behaviour does not happen for the `popover-container` by default. (see 'inside header' → it '_should stay open while scrolling, as the header does not affect its own content_').

Just add comments in the affected files, to explain to every dev where/how these lines of code tested.

yomonom: > > Should I move this issue to the dev roundtable, then?
> > If these changes are unwanted in other components, we could also restrict the 'clipping' behaviour through an extra attribute on `postpopover-container`, as this is not a public-facing component anyway. Then we could just change out `computePositionWithSafeArea` with the regular `computePosition` when `clip` (or something the like) is set to `false`, restricting the behaviour to just the popovers themselves.
> 
> Yes please!

I've called the new property `autoHide`, since the flipping behaviour is still determined by the height of the header (already present on main).

alionazherdetska:
> I've added some new utility functions for testing. These could probably also be used in other tests that interact with popovers. I didn't touch them though, as I feel like this PR is already getting big enough.
> 
> It's probably worth to add that to another ticket though. Maybe together with the broken popover tests that still contain the `Promise.all`?

here's the ticket addressing the `Promise.all` issue: https://github.com/orgs/swisspost/projects/3/views/12?pane=issue&itemId=237949730&issue=swisspost%7Cdesign-system%7C8410

Feel free to add it as a task/follow-up on this one.

alionazherdetska (dismissed):

packages/components/cypress/e2e/tooltip.cy.ts
oliverschuerch: It's good to keep format changes, as long as your IDE is setup correctly regarding prettier, etc.

If there are changes in almost every file, you'll need to check your setup, because it could be wrong and the reformatting would become obsolete.

oliverschuerch (approved):

alionazherdetska (approved):

PR #8330, opened 2026-08-18
fix(components, styles): inset the focus outline for stacked elements

### LINKED ISSUE #7455: [bug]: Focus outline crossing borders of list items
### 🐞 Describe the bug

Image

https://career.post.ch/de/lernende

Focus outline should be inset in link lists so that the focus outline does not interfere with the borders of the items.

yomonom:
## 📄 Description

The `.list-interactive` class has been updated to display the focus outline inside, so that it no longer clashes with neighbouring outlines and borders.

The same fix has also been applied to the 'post-accordion-item' component to address the same issue and provide a more consistent experience.

## 🚀 Demo

https://github.com/user-attachments/assets/00500d42-fd31-4e54-8008-d043fc7bd906

https://github.com/user-attachments/assets/3c0ca13a-3afd-45c9-a645-3622cbff8a32

https://github.com/user-attachments/assets/298673f8-f0c4-459a-a240-e3228ab28b42

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist

### DIFF
--- a/packages/components/src/components/post-accordion-item/post-accordion-item.scss
+++ b/packages/components/src/components/post-accordion-item/post-accordion-item.scss
@@ -49,16 +49,6 @@ tokens.$default-map: components.$post-accordion;
   border-block-end: tokens.get('accordion-border-bottom-width')
     tokens.get('accordion-border-bottom-style') transparent;
   cursor: pointer;
-  @include utility-mx.focus-style-none;
-
-  @include utility-mx.focus-style('::before') {
-    display: block;
-    content: '';
-    position: absolute;
-    inset: calc(tokens.get('accordion-group-border-top-width') * -1) 0
-      calc(tokens.get('accordion-border-bottom-width') * -1);
-    border-radius: tokens.get('focus-border-radius', helpers.$post-focus);
-  }
 
   slot::slotted(span[slot='header']) {
     flex-grow: 1;
@@ -88,6 +78,8 @@ tokens.$default-map: components.$post-accordion;
   > ::slotted(.text-truncate) {
     display: block;
   }
+
+  @include utility-mx.focus-inline(3px, 3px);
 
   @include utility-mx.high-contrast-mode() {
     &:hover,
--- a/packages/styles/src/mixins/_utilities.scss
+++ b/packages/styles/src/mixins/_utilities.scss
@@ -115,6 +115,29 @@
   }
 }
 
+// Draws the focus outline on a pseudo-element instead of on the element itself.
+//
+// Use it on full-width stacked elements where the default outline would clash with neighbouring outlines and borders.
+@mixin focus-inline($top: 0, $bottom: 0, $selector: '::before') {
+  $focus-outline-inverse: calc(
+    tokens.get('focus-outline-width', helpers.$post-focus) +
+      tokens.get('focus-outline-offset', helpers.$post-focus)
+  );
+
+  @include focus-style {
+    @include focus-style-none;
+  }
+
+  @include focus-style($selector) {
+    display: block;
+    content: '';
+    position: absolute;
+    // Inset the pseudo-element instead of adjusting the outline offset to ensure that the outline retains its default appearance.
+    inset: calc(#{$focus-outline-inverse} + #{$top}) 0 calc(#{$focus-outline-inverse} + #{$bottom});
+    border-radius: tokens.get('focus-border-radius', helpers.$post-focus);
+  }
+}
+
 @mixin focus-style-custom($additional-selector: '') {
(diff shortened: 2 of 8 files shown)


packages/styles/src/mixins/_utilities.scss
```diff
@@ -115,6 +115,29 @@
   }
 }
 
+// Draws the focus outline on a pseudo-element instead of on the element itself.
+//
+// Use it on full-width stacked elements where the default outline would clash with neighbouring outlines and borders.
+@mixin focus-inline($top: 0, $bottom: 0, $selector: '::before') {
```
alizedebray: `focus-inset` would be more meaningful here

packages/styles/src/mixins/_utilities.scss
```diff
@@ -115,6 +115,29 @@
...
+  @include focus-style {
+    @include focus-style-none;
+  }
+
+  @include focus-style($selector) {
+    display: block;
+    content: '';
+    position: absolute;
+    // Inset the pseudo-element instead of adjusting the outline offset to ensure that the outline retains its default appearance.
+    inset: calc(#{$focus-outline-inverse} + #{$top}) 0 calc(#{$focus-outline-inverse} + #{$bottom});
```
alizedebray: The horizontal inset should probably also take the focus width and focus offset into account, otherwise it looks a little bit unbalance.

```suggestion
    inset: calc(#{$focus-outline-inverse} + #{$top}) $focus-outline-inverse calc(#{$focus-outline-inverse} + #{$bottom});
```

packages/components/src/components/post-accordion-item/post-accordion-item.scss
```diff
@@ -89,6 +79,8 @@ tokens.$default-map: components.$post-accordion;
     display: block;
   }
 
+  @include utility-mx.focus-inline(3px, 3px);
```
alizedebray: Where are these `3px` coming from? It would be good to have them as default values or, even better, have them tokenized.

alizedebray, line 82: Where does the `3px` value come from?
It would be good to have this as a default value in the mixin and, ideally, have it tokenized.

packages/styles/src/mixins/_utilities.scss
alizedebray, line 121: `focus-inset` might be a more meaningful name here, since the mixin is defining an inset focus outline.

packages/styles/src/mixins/_utilities.scss
```diff
@@ -115,6 +115,29 @@
...
 }
 
+// Draws the focus outline on a pseudo-element instead of on the element itself.
+//
+// Use it on full-width stacked elements where the default outline would clash with neighbouring outlines and borders.
+@mixin focus-inline($top: 0, $bottom: 0, $selector: '::before') {
+  $focus-outline-inverse: calc(
+    tokens.get('focus-outline-width', helpers.$post-focus) +
+      tokens.get('focus-outline-offset', helpers.$post-focus)
+  );
```
alizedebray, line 125: Since this calculation is simply adding two values rather than inverting anything, would `$default-offset` be a more meaningful name here?
This could then also include the tokenized `3px` value.

packages/styles/src/mixins/_utilities.scss
alizedebray, line 136: Could we also take the focus width and focus offset into account for the horizontal inset?

As it stands, the outline is inset vertically but outset horizontally, which makes the focus treatment look slightly unbalanced. But this could also be checked with the designers.


```suggestion
    inset: calc(#{$focus-outline-inverse} + #{$top}) $focus-outline-inverse calc(#{$focus-outline-inverse} + #{$bottom});
```

packages/styles/src/mixins/_utilities.scss
alizedebray, line 121: I'd also rename `$top` and `$bottom` to `$top-adjustment` and `$bottom-adjustment`, since these values are added to the default inset rather than being used directly as the top and bottom inset.

alizedebray (changes requested):
In general, it would be good to have a design for the inset focus outline before finalizing the implementation. This would help us define the right tokens and keep the implementation aligned with Figma.

For example, the focus outline in the [List Interactive documentation](https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-V2?node-id=1639-46092&t=UFBVkrD5TbZIMgK1-0) has changed and is now different from what we have in the code.

yomonom:
Didn't see that this was a UX ticket 🫣. Closing for now (until design is finalised).

PR #8340, opened 2026-08-19
fix(fonts): add italic font definitions

gfellerph:
Added definitions for italic font styles for Swiss Post Sans. Italicised text will now use the real font instead of slanting the regular one. This PR also updates the CDN version of the font files to v1-1.

Old italic text (compare letter `a` and `e`):
image

New italic text:
image


alizedebray (dismissed):

oliverschuerch (commented):
Should we document ExtraBlack and/or italic fonts explicitely?
https://design-system.post.ch/?path=/docs/677cfabf-dbf0-4de2-ad07-6d5bfb9e2375--docs

oliverschuerch (approved):

PR #8347, opened 2026-08-20
fix(component): fix `<back-to-top>` button overlaying other page content when it is hidden

### LINKED ISSUE #8319: [bug]: Back to top button must be `display: none` when not displayed
### 🐞 Describe the bug

Image

A bug report from a project came in that shows the back to top button overlaying the close button for a banner (discovered in tests). The back to top button seems to be misplaced. In any case, the back to top button should be completely hidden with `display: none` if it's not displayed. The current `tabindex="-1"` and `aria-hidden="true"` are not enough.

- [x] Check the position against the design in Figma

alionazherdetska:
## 📄 Description

Checked Figma: the position of the back-to-top button matches the design intent.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/components/post-back-to-top/post-back-to-top.tsx
```diff
@@ -34,11 +35,30 @@ export class PostBackToTop {
...
+        translate: this.translateY,
+        fill: 'forwards',
+      });
+      return;
+    }
+
+    this.currentAnimation = fadeSlide(this.host, 'out', {
+      translate: this.translateY,
+      fill: 'forwards',
+    });
```
oliverschuerch: Can you try to refactor that to make it simpler?

```js
// No matter we animate 'in' or 'out', the host must be visible before we can start any animation!
// After the animation, the hosts hidden state is recalculated, dependent on the `belowFold` variable. This state can persist until the next animation is started.
this.host.hidden = false;

this.currentAnimation = fadeSlide(
  this.host,
  newValue ? 'in' : 'out',
  {
    translate: this.translateY,
    fill: 'forwards',
  }
);
```

oliverschuerch (changes requested):
This seems to be the best solution for the current situation. Generally I would prefer a CSS solution.

packages/components/src/components/post-back-to-top/post-back-to-top.tsx
oliverschuerch: Any feedback about that?

oliverschuerch (changes requested):

packages/components/src/components/post-back-to-top/post-back-to-top.tsx
alionazherdetska: I have already implemented what you suggested.

oliverschuerch: Sorry, I didn't see your changes.

oliverschuerch (approved):

PR #8354, opened 2026-08-21
fix(styles): Button can break into two rows if text is too long

### LINKED ISSUE #8077: [bug]: Button text can be cut if it's too long
### 🐞 Describe the bug

Image

### ✨ Expected behavior

We need a better coping strategy when button content is too long. Not displaying text is the worst thing to do. I see two options:
- Make it scrollable
- Make it multi-line

Either will break the UI, but will keep text readable and accessible which should be prioritised over not breaking the UI. Fixing the layout break is then a matter of content or pattern choice.

### Requirements
- Long single words must wrap
- Multiline text must be readable -> adjust border radius
- Icon size must remain stable

### Tasks

- [ ] Allow buttons to be multiline if text is just too long

AndrinAAF:
## 📄 Description

Button can break into two rows if text is too long

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/styles/src/mixins/_button.scss
```diff
@@ -33,7 +33,7 @@
   display: inline-flex;
   position: relative;
   align-items: center;
-  justify-content: center;
+  justify-content: flex-start;
```
alionazherdetska: Hey! 👋 Nice progress on the wrapping fix, it works well for regular buttons. I found one issue though while testing icon-only buttons:

Changing `justify-content` from `center` to `flex-start` in `button-base` makes the icon in `.btn-icon` buttons shift to the left instead of staying centered (see screenshot, the search icon looks off-center now):

Image

https://preview-8354--swisspost-design-system.netlify.app/iframe.html?id=[snapshots--button](https://preview-8354--swisspost-design-system.netlify.app/iframe.html?id=snapshots--button&viewMode=story)&viewMode=story

This happens because icon buttons have a fixed round box that's bigger than the icon itself, so it relies on `justify-content: center` to keep the icon in the middle. Normal text buttons don't show this problem since they just shrink to fit their text, but icon buttons do.

Good news: you don't actually need `justify-content: flex-start` for the wrapping to work. The other changes you made (`white-space: normal`, `overflow-wrap: anywhere`, `text-align: center`) already do the job on their own. So you can just change it back:

```scss
justify-content: center;
```

Could you try that and check if the icon buttons look right again? Also worth a quick look at the "Full-width" button example in Storybook after the change, just to be safe. Thanks! 🙌

.changeset/better-ghosts-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-styles': patch
+---
+
+Fixed the Problem with long text in the button field. The text will now wrap into multiple lines if it is too long.
```
alionazherdetska: maybe we can improve the changeset a bit, something like:

```suggestion
Fixed button text wrapping into multiple lines instead of being cut off when content is too long.
```

alionazherdetska (changes requested):
Great work!
Just a few things to address.

Can you please also add a snapshot example for the wrap case here too? like a button with a long label in a narrow container, so we catch the wrapped look visually:
https://github.com/swisspost/design-system/blob/main/apps/documentation/src/stories/components/button/button.snapshot.stories.ts

Then we wil be able to see them rendered here: [snapshots--button](https://preview-8354--swisspost-design-system.netlify.app/iframe.html?id=snapshots--button&viewMode=story)


You can access it locally using: http://localhost:9000/iframe.html?id=snapshots--button&viewMode=story

alionazherdetska (changes requested):
small suggestion: the existing snapshot matrix in `button.stories.snapshots.ts` won't actually catch long text wrapping, since none of the combinations constrain the button's width. Could be good to add a dedicated example for that, something like this (this file isn't touched in the PR yet, pasting the whole thing so it's clear where it'd go):

```ts
import type { Args, StoryContext, StoryObj } from '@storybook/web-components-vite';
import meta, { Default } from './button.stories';
import { html } from 'lit';
import { schemes } from '@/shared/snapshots/schemes';
import { bombArgs } from '@/utils';

const { id, ...metaWithoutId } = meta;

export default {
  ...metaWithoutId,
  title: 'Snapshots',
};

type Story = StoryObj;

export const Button: Story = {
    return schemes(
      () => html`
        <div class="d-flex flex-wrap align-items-start gap-16">
          ${bombArgs({
            variant: context.argTypes.variant.options,
            size: context.argTypes.size.options,
            tag: context.argTypes.tag.options,
            disabled: [false, true],
            icon: ['null', '2069'],
            secondIcon: ['2069'],
            iconPosition: context.argTypes.iconPosition.options,
          })
            .filter(args => !(args.icon === 'null' && args.iconPosition !== 'textOnly'))
            .filter(args => !(args.tag === 'a' && args.disabled))
            .filter(args => !(args.icon !== 'null' && args.tag === 'input'))
            .map((args: Args) => (args.tag === 'input' ? { ...args, type: 'button' } : args))
            .map((args: Args) =>
              Default.render?.({ ...context.args, ...args, animated: false }, context),
            )}
        </div>

        <div class="d-flex flex-wrap align-items-start gap-16 mt-24" style="max-width: 300px;">
          ${Default.render?.(
            {
              ...context.args,
              text: 'This is a very long button label that should wrap onto multiple lines',
              animated: false,
            },
            context,
          )}
        </div>
      `,
    );
  },
};
```

This puts a button inside a narrower container so the label has no choice but to wrap, and the snapshot will actually show us what that looks like, so when the next developer changes something in the file the snapshot will clearly shows that something wen wrong.

AndrinAAF:
image

Button gets wrapped when the text is to long

PR #8357, opened 2026-08-24
fix(components): resolve desktop layout stuck on mobile viewport on initial load for `post-header`

alionazherdetska:
## 📄 Description

Tested the fix both on the `career` and `cargo` pages. The fix is working.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/utils/breakpoints.ts
```diff
@@ -34,27 +32,26 @@ class Breakpoint {
...
-    (options: { emitEvents: boolean } = { emitEvents: true }) => {
-      const previousBreakpoint = this.currentBreakpoint;
-      const newBreakpoint = this.breakpoints.find(breakpoint => {
-        return breakpoint.minWidth <= innerWidth;
-      });
+  private updateCurrentBreakpoint = (options: { emitEvents: boolean } = { emitEvents: true }) => {
+    const previousBreakpoint = this.currentBreakpoint;
+    const newBreakpoint = this.breakpoints.find(breakpoint => {
+      return globalThis.matchMedia
+        ? globalThis.matchMedia(`(min-width: ${breakpoint.minWidth}px)`).matches
```
oliverschuerch: Instead of generating the string over and over again, we could add an additional property to a `BreakpointDefinition`.

e.g. `{ key: 'xl', device: 'desktop', minWidth: 1280, mediaQuery: '(min-width: 1280px)' }`

Then use it here like `globalThis.matchMedia
        ? globalThis.matchMedia(breakpoint.mediaQuery).matches : ...`

Performance-wise, this would be much better.

oliverschuerch (changes requested):
Thanks for fixing this 💛

packages/components/src/utils/breakpoints.ts
alionazherdetska: Applied

oliverschuerch (approved):

PR #8384, opened 2026-08-26
fix(styles): distinguish completed stepper items

### LINKED ISSUE #8383: [bug]: Completed and active steps are not correctly indicated (v9)
> **Resolution:**
> - keep current implementation as it is
> - add the requested functionality as a new class in v9, it must be _optional_
> - `[aria-current]` is used for the currently visited step, the new class is used for the last completed step
> - in v10 we can add a new property, but we will investigate other libraries to have more information on how this is usually handled.  

 _Originally posted by @alizedebray in [#7012](https://github.com/swisspost/design-system/issues/7012#issuecomment-3784200443)_

Implement the resolution for v9 steps that are both active and completed (when a user navigates back to check on previous steps).

### Requirements
- Completed steps need the new `stepper-item-completed` class. Completed items with `aria-current` show the checkmark instead of the step number

### Tasks
- [ ] Implement the new class and styles
- [ ] Document the new feature in v9 storybook

WilliamK112:
## Summary

- add the optional `.stepper-item-completed` state while preserving the existing `aria-current`-based behavior
- render a checkmark for steps that are both completed and current, including forced-colors styling
- document the independent completed/current states in Storybook and add focused snapshot assertions

## Validation

- `pnpm styles:test`
- `pnpm styles:lint`
- `pnpm --filter design-system-styles... build`
- `pnpm --filter design-system-documentation... build`
- Prettier and ESLint checks for the changed files

Fixes #8383

### DIFF
--- a/.changeset/tidy-steppers-finish.md
+++ b/.changeset/tidy-steppers-finish.md
@@ -1,0 +1,5 @@
+---
+'@swisspost/design-system-styles': minor
+---
+
+Added an optional `.stepper-item-completed` class so completed steps remain visible when the current step changes.
--- a/packages/documentation/cypress/snapshots/components/stepper.snapshot.ts
+++ b/packages/documentation/cypress/snapshots/components/stepper.snapshot.ts
@@ -2,6 +2,14 @@ describe('Stepper', () => {
   it('default', () => {
     cy.visit('/iframe.html?id=snapshots--stepper');
     cy.get('.stepper', { timeout: 30000 }).should('be.visible');
+    cy.get(".stepper-item-completed[aria-current='step'] > .stepper-link").should($links => {
+      expect($links.length).to.be.greaterThan(0);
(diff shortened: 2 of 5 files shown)


WilliamK112:
I checked the failing SonarCloud job: the fork `pull_request` run has an empty `SONAR_TOKEN`, and the scanner stops with “Not authorized or project not found” before analysis. The e2e job passed. This appears to need maintainer-side SonarCloud/CI configuration rather than a change to the stepper implementation; I have left the workflow and secret handling unchanged.

leagrdv:
Closing this PR as it's not working correctly:

When user sets a completed step after the current step, UI breaks:
Screenshot 2026-08-28 at 10 00 17

When user sets completed step the same as the current step, it also breaks:
Screenshot 2026-08-28 at 10 00 42

PR #8394, opened 2026-08-28
fix(components): fix glitches in `post-side-navigation` in ssr environment

### LINKED ISSUE #8213: [bug]: visual glitches in `<post-side-navigation>` component on initial render in the `nextjs-integration` app
### 🐞 Describe the bug

On initial load of the side navigation, two separate visual glitches occur right after the panel becomes visible:

1. `.post-side-navigation-item` links without a leading post-icon briefly render unindented, then shift to align with icon items shortly after.
2. The chevron icon on the collapsible trigger briefly points downward, then flips to pointing upward.

### 🛠️ To Reproduce

1. Run the `integration-next` app
2. Observe the nehaviour of the sidenav

### 📷 Screenshots

https://github.com/user-attachments/assets/d549edef-d780-46cb-a929-456865666926

alionazherdetska:
## 📄 Description

This ticket originally reported two separate visual bugs on initial render in the Next.js SSR integration app:

1. The chevron icon on collapsible triggers briefly showing the wrong rotation before snapping to the correct expanded/collapsed state.
2. An item's indentation/alignment being wrong for a moment when it sits next to a sibling that has an icon, then correcting itself.

Regarding the visual bug number 2 it turned out to not actually be a bug in the component. Dug into it a fair bit and traced it back to the integration app's own markup - the sidenav example had one level-1 item without an icon sitting next to a sibling that had one. But per our own docs, icons on `post-side-navigation` are level-1 only and it's all-or-none - you either put an icon on every level-1 item or none of them, never mixed. Once you follow that rule there's no such thing as "an icon-less item next to an icon item" at level 1 in the first place, so there's nothing for the layout to flash between:
image

So instead of trying to patch the CSS to handle a markup shape we don't actually support, I just fixed the example markup in both the Next.js and Angular integration apps to follow the all-or-none rule (every level-1 item now has an icon, consistent with the docs), and it's gone.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


.changeset/vast-hornets-dress.md
oliverschuerch, line 1: Nice you found a solution to fix this, but...

The fix will not be implemented automatically, just by updating to the released package version containing this fix.
Technically it's not a breaking change, but it still needs manual adjustment on project-side.

I think it would be a good idea to flag such changeset files with a specific tag/keyword (e.g. `manual-fix-required`, etc.), or alternatively explain to the projects, they have to implement the fix on there side.

Can you bring that up in the next DEV RT please?

oliverschuerch (changes requested):

.changeset/vast-hornets-dress.md
oliverschuerch: The documentation included the icons already before, so the actual fix only happens in CSS, while the PostIcon markup was only missing on our integration apps markup.

Conclusion: The fix will work for projects, without any further changes, only by updating to the version containing this update.

oliverschuerch (dismissed):

oliverschuerch (approved):

PR #8405, opened 2026-08-31
feat(styles): error page templates

### LINKED ISSUE #5847: Add additional error page templates
### ✨ Is your feature request related to a problem? Please describe it.

There's a template for 404 not found pages with a nice graphic and example markup for how to create this in a project: [Storybook v9.1 / Patterns / Error page](https://design-system.post.ch/?path=/docs/a536a61d-cac2-4f39-adbf-092bdd445ce5--docs) .

### 💡 Describe the solution you'd like

There are other types of errors. It would be super nice to have respective graphics and maybe some text ideas and a template to copy&paste for easy implementation.

Error pages should use [palettes](https://swisspost-design-system-next.netlify.app/?path=/docs/43481535-5b39-40b5-a273-478b07dc3b31--docs) to set their background, [containers](https://swisspost-design-system-next.netlify.app/?path=/docs/a4ca9660-bb4a-4cc7-adfd-84767382ac03--docs) and [grid](https://swisspost-design-system-next.netlify.app/?path=/docs/7240f2ef-216a-490e-9bd8-c0cef19f7b31--docs) to align their content, like so:
```html
<div class="palette palette-brand">
  <div class="container">
    <div class="row row-cols-1 row-cols-lg-2">
      <div class="col">
        <!-- Content -->
      </div>
      <div class="col">
        <!-- Image -->
      </div>
    </div>
</div>
```

The error image should be a CSS component and include the SVG images as a `background-image`:
```html
<div aria-hidden="true" class="error error-401"></div>
```

### 📖 References
- Storybook v9.1: [Patterns / Error page](https://design-system.post.ch/?path=/docs/a536a61d-cac2-4f39-adbf-092bdd445ce5--docs) (mentioned above)
- Figma: [Layout Examples Next Level / Error pages](https://www.figma.com/design/1U0tACfFegz2s1ThIr4M06/Layout-Examples-Next-Level?node-id=5244-141545&t=9IOserQ0nHvJks3N-0) (V1) 

### 🔄 Describe alternatives you've considered

None, the idea is very good.

### LINKED ISSUE #6407: Error page templates
### 📝 Description

Document the different error page template alternatives:
- Error 400
- Error 401
- Error 403
- Error 404
- Error 451
- Error 500
- Error 503
- Error 504

Additionally, each error image should be a downloadable asset in the documentation.

Error pages should be able to use the grid alongside existing classes to be built without having to create custom CSS for it.
To have the error page image outside of the container, the `section` can be used.

🎨 Design link: [Figma](https://www.figma.com/design/1U0tACfFegz2s1ThIr4M06/Layout-Examples-Next-Level?node-id=4978-5160&p=f&t=hjWTwm6Lb2urRP7P-0)

### 📃 Tasks
- [ ] Update documentation to have the multiple error page options
- [ ] Add images as assets to be downloaded within the documentation page
- [ ] Implement error pages according to design
- [ ] Design review

leagrdv:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/src/stories/templates/error-page/error-page.stories.ts
```diff
@@ -1,47 +1,149 @@
...
+                  </a>
+                `
+              : html`
+                  <a class="btn btn-primary" href="/" @click="${(e: Event) => e.preventDefault()}">
+                    <post-icon name="arrowleft" aria-hidden="true"></post-icon>
+                    Go to homepage
+                  </a>
+                  <a
+                    class="btn btn-secondary"
+                    href="https://www.post.ch/en/pages/suche#t=AllTab"
```
alionazherdetska, line 143: the 500 error page doesn't have search but the button still shows the href for it. secondaryButtonLabel is "Contact the website operator" but the href is still hardcoded to `/pages/suche`. button says one thing, goes somewhere else. maybe it needs its own secondaryButtonHref in contentByType, or a real contact url

.changeset/salty-rabbits-bow.md
```diff
@@ -0,0 +1,8 @@
+---
+'@swisspost/design-system-documentation': minor
+'@swisspost/design-system-styles': minor
+---
+
+Added different pages templates for the following errors: 400, 401, 403, 404, 451, 500, 503 and 504.
```
alionazherdetska:
```suggestion
Added different page templates for the following errors: 400, 401, 403, 404, 451, 500, 503 and 504.
```

apps/documentation/src/stories/templates/error-page/error-page.stories.ts
```diff
@@ -1,47 +1,149 @@
...
+          '504': '504: Gateway time-out',
+        },
+      },
+      options: ['400', '401', '403', '404', '451', '500', '503', '504'],
+      table: {
+        category: 'General',
+      },
+    },
+  },
+  render: ({ type }) => render(type),
```
alionazherdetska: Small suggestion: render calls render, works because of hoisting but kind of risky to touch later. Maybe rename the top-level function to renderErrorPage so it's less confusing?

packages/styles/src/components/error-container.scss
```diff
@@ -1,3 +1,398 @@
...
+      max-width: calc(50% - (var(--post-grid-gutter-x) / 2));
+    }
+
+    @include media.min(xl) {
+      padding-block: 3.75rem;
+    }
+  }
+}
+
+.error-400 {
```
alionazherdetska, line 123: Just as an idea:
The `.error-4xx` blocks all repeat the same structure (background-size/image/position, padding-bottom per breakpoint), just with different values. Could pull the values into one map and loop over it instead, so the numbers stay exactly as they are now but we're not copy-pasting the same selectors 8 times. easier to scan and less risk of a typo hiding in there somewhere. not blocking, just a thought if this file keeps growing

for example, something like:

```scss
$error-types: (
  '400': (mobile-size: 150px, mobile-pos: bottom 2rem left 50%, desktop-size: 320px),
  '401': (mobile-size: 190px, mobile-pos: bottom 2rem left 50%, desktop-size: 210px),
  // ...rest of the types
);

@each $type, $config in $error-types {
  .error-#{$type} {
    &::before {
      background-size: map.get($config, mobile-size);
      background-position: map.get($config, mobile-pos);
      background-image: url('.../#{$type}.svg');

      @include media.min(lg) {
        background-size: map.get($config, desktop-size);
      }
    }
  }
}
```

same output, just one place to add/tweak a value instead of hunting through the whole file

alionazherdetska (changes requested):
Error pages look cool!

apps/documentation/src/stories/templates/error-page/error-page.stories.ts
leagrdv: Changed the link here to `https://www.post.ch/en/help-and-contact`, good catch!

packages/styles/src/components/error-container.scss
leagrdv: Pretty good idea! I did do it for the padding-bottom of the containers, but I tried for the other properties and it made it harder to read 😄 Cause they all have their specificities (e.g. error 400 has two different background images, some have `background-position`, other have `background-position-x`, some change size and some don't...) .

.changeset/salty-rabbits-bow.md
alionazherdetska, line 1: Should we also mark it as deprecated in `v9` docs then:
https://swisspost-design-system-version-9.netlify.app/?path=/docs/a536a61d-cac2-4f39-adbf-092bdd445ce5--docs

alionazherdetska:
@rouvenpost can you please do the design review😊
Here are the preview snapshots: https://preview-8405--swisspost-design-system.netlify.app/iframe.html?id=[snapshots--error-page](https://preview-8405--swisspost-design-system.netlify.app/iframe.html?id=snapshots--error-page&viewMode=story)&viewMode=story

packages/styles/src/components/error-container.scss
```diff
@@ -1,3 +1,366 @@
...
+}
+
+$error-types: (
+  '400': (
+    sm-padding-block-end: 12rem,
+    md-padding-block-end: 16rem,
+    lg-padding-block-end: 20rem,
+    bg-position: bottom 2rem left 0,
+    bg-position-lg: top 0 left 80%,
+    bg-position-xl: top 0 left 75%,
```
alionazherdetska: small suggestion: the `400` entry in `$error-types` has `bg-position`, `bg-position-lg` and `bg-position-xl` keys but the `@each` loop only reads the padding-block-end ones. looks like the start of moving background positioning into the map too, but it's not wired up anywhere so right now they're just dead. worth either finishing that or dropping them so it doesn't look like leftover state

packages/styles/src/components/error-container.scss
```diff
@@ -1,3 +1,366 @@
...
+    position: absolute;
+    background-repeat: no-repeat;
+    pointer-events: none;
+    width: 100%;
+    height: 100%;
+  }
+
+  .container::before {
+    pointer-events: none;
+  }
```
alionazherdetska: `.container::before { pointer-events: none; }` in the base `.error` block is dead code for `400`/`401`/`403`: those types put their image on `.error-XXX::before` directly, not on `.container::before`, so there's no `content` set and the pseudo-element never renders for them. only the `404/451/500/503/504` group actually uses `.container::before`. probably makes more sense to move that rule into their shared block instead of the universal `.error` one, so it's clear which types actually use it

alionazherdetska (changes requested):

.changeset/salty-rabbits-bow.md
leagrdv: Because I did not remove it, but simply deprecated it in `v10`, I don't think we need to deprecate it also in `v9`.

packages/styles/src/components/error-container.scss
leagrdv: Yes, I indeed forgot to remove my tests!

alionazherdetska (dismissed):
Approving with a small suggestion: the 8 error codes are typed out by hand in four places now - the `ErrorType` union, `contentByType`'s keys, and `argTypes.type.options`/`labels`. ts checks `contentByType` against `ErrorType`, but `options` and `labels` are just plain strings with nothing linking them back, so nothing stops them drifting apart. like if `502` gets added to `contentByType` later but someone forgets `options`, storybook just quietly won't offer it as a choice, no error or anything. or the other way round, adding it to `options` but not `contentByType` would throw at render time since `contentByType['502']` is undefined

could just derive `options` from `contentByType`'s keys instead, `options: Object.keys(contentByType)`, so it always matches whatever's actually there and nothing needs sycing by hand.

rouvenpost:
Design is Approved

myrtasakellariou (approved):

Issue #7668, opened 2026-05-20
Context aware close button


gfellerph:
### 📝 Description

The `<post-closebutton>` component should automatically detect the closest closable element or component and perform the appropriate close action on it, without requiring any explicit wiring (event listeners, `onClick` handlers, etc.) from the consumer.

#### Current Behavior

Each component that uses `<post-closebutton>` handles its own close logic:

- `<post-banner>` listens for `click` events via `@Listen('click')` + `@EventFrom('post-closebutton', { allowDescendants: true })` and calls `this.dismiss()`
- `<post-popover>` binds `onClick={() => this.hide()}` directly on the close button rendered in its shadow DOM

#### Desired Behavior

When clicked, `<post-closebutton>` walks up the DOM tree (crossing shadow DOM boundaries) to find the nearest supported closable element/component and calls its close method. No explicit event binding needed from consumers.

#### Supported Closable Targets

Any component that can hide/close should be supported.

| Target | Close method |
|--------|--------------|
| Element with `[popover]` attribute | `togglePopover(false)` |
| `<dialog>` element | `close()` |
| `<post-collapsible>` | `toggle(false)` |
| `<post-accordion-item>` | `toggle(false)` |
| `<post-popover>` | `hide()` |
| `<post-popovercontainer>` | `hide()` |
| `<post-banner>` | `dismiss()` |

#### Shadow DOM Considerations

Standard `Element.closest()` does **not** cross shadow DOM boundaries. The close button can live in three different DOM contexts, and the ancestor-search logic must handle all of them:

##### 1. Light DOM (simplest)

The close button sits directly inside a closable element (e.g. `<dialog>`). Regular `closest()` works.

##### 2. Slotted into another component

The close button is in the light DOM but projected into a component's shadow DOM via `<slot>`. Example: `<post-closebutton slot="close-button">` inside `<post-banner>`. Since `closest()` follows the light DOM tree, this also works with regular `closest()`.

##### 3. Rendered inside another component's shadow DOM

The close button is part of another component's internal template (not slotted by the consumer). Example: `<post-popover>` renders `<post-closebutton>` in its own shadow DOM. Here, `closest()` stops at the shadow root boundary — you must jump to the shadow host via `getRootNode()` and continue walking up.

#### Existing Utilities to Reuse

[...]

PR #8406, opened 2026-08-31
chore(components): fix closebutton e2e error

myrtasakellariou:
## 📄 Description

Issue seems to occur because Cypress tests referenced missing stories. This change updates the test to point to the fixtures (and creates missing fixtures).
Also added `<div class="mock-element">` to the snapshot story so that closebutton is placed correctly in the snapshots.

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/cypress/fixtures/post-accordion-item.test.html
```diff
@@ -0,0 +1,23 @@
...
+    <style>
+      post-accordion-item {
+        display: block;
+        position: relative;
+        width: 20rem;
+      }
+    </style>
+  </head>
+  <body>
+    <post-accordion-item>
```
leagrdv, line 17: Why are we testing an accordion closable with the `close button`? There are no examples of this in the docs, I'm not sure it's something users should do no? So I don't know why we're testing this 🤔

leagrdv (approved):

packages/components/cypress/fixtures/post-accordion-item.test.html
myrta2302: This was the original ticket: https://github.com/swisspost/design-system/issues/7668

PR #8411, opened 2026-08-31
feat(internet-header): add close link support

### LINKED ISSUE #8353: Add close link to pre-configured header
As requested, online services can display a Close button in the header.

This use case is already documented in Figma and supported for the Self-managed Header.
See [#7700](https://github.com/swisspost/design-system/issues/7700) for more information.

The Pre-Configured Header must support the same use case.

**Tasks:**

- [ ] Propose an API change to allow the configuration to include a Close button as the last element of the local navigation.
  - The Close button differs from regular local navigation items, so it cannot be included directly in localHeader.navigation.items.
  - The Close button must be optional.
- [ ] Inform the POW team about the API change and request the corresponding backend implementation.
- [ ] Update swisspost-internet-header to read the new parameter and display the Close button when it is provided.

alionazherdetska:
## 📄 Description

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
```diff
@@ -342,7 +351,7 @@ export class PostInternetHeader {
 
           {localHeader.title && <p slot="title">{localHeader.title}</p>}
 
-          {localHeader.navigation && this.renderNavigation('local-nav', localHeader.navigation)}
+          {localHeader.navigation && this.renderLocalNavigation(localHeader)}
```
myrtasakellariou: Should this method be skipped when there are no navigation items or should the close button be rendered at any case?

packages/internet-header/cypress/e2e/header.cy.ts
```diff
@@ -251,7 +251,7 @@ describe('header', () => {
       cy.get('[slot="local-nav"] > li').as('nav-items');
 
       cy.get('@nav-items')
-        .should('have.length', headerConfig.localHeader.navigation.length)
+        .should('have.length', headerConfig.localHeader.navigation.length + 1)
```
myrtasakellariou, line 255: A comment of what the `+1 ` is would be nice.

myrtasakellariou (changes requested):
Should we add the close button as a control option on the pre-configured header story?

packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
alionazherdetska: Thanks for the question, it actually got me digging in, and I realized my guard wasn't working correctly.

The config comes from the backend, and if it sends a `closeLink`, we want to render it regardless of whether `navigation` items are also present.

alionazherdetska:
> Should we add the close button as a control option on the pre-configured header story?

@myrtasakellariou 
I don't think we need a control for this one though: the close link is fully driven by the backend config (text, url, icon), just like the rest of the navigation items. It's not something the client renders or customizes on its own, so adding a toggle in Storybook could give a slightly misleading impression of how it actually works in a real project. Happy to reconsider if you think it'd still help for testing/demo purposes though.🙂

myrtasakellariou:
> > Should we add the close button as a control option on the pre-configured header story?
> 
> @myrtasakellariou I don't think we need a control for this one though: the close link is fully driven by the backend config (text, url, icon), just like the rest of the navigation items. It's not something the client renders or customizes on its own, so adding a toggle in Storybook could give a slightly misleading impression of how it actually works in a real project. Happy to reconsider if you think it'd still help for testing/demo purposes though.🙂

Ok. It makes sense. Its only that there is no published description of what this is destined for. (e.g. Optional close button, rendered as the last element of the local navigation (e.g. to lead users of an online service back to post.ch).).

myrtasakellariou (approved):

gfellerph (changes requested):
Needs redesign from a new feature request

PR #8427, opened 2026-09-05
feat(component): enable the client side navigation for the `post-breadcrumbs` component

### LINKED ISSUE #8072: [bug]: Breadcrumb links reload the page in a next.js application
### 🐞 Describe the bug

Clicking a link in the breadcrumbs reloads the entire page instead of using next.js routing.

- [ ] Try and find a solution to keep the link in the shadow DOM (timeboxed to 4h)
- [ ] Offer the possibility to slot the link as per our architectural decision 
- [ ] Open ticket for other components that involve routing without allowing slotting an anchor tag to implement the same solution (i.e. post-logo, post-language-menu-item)

alionazherdetska:
## 📄 Description

This PR fixes breadcrumb links reloading the page in client-side routing apps like Next.js.

Previously, post-breadcrumb-item and post-breadcrumbs' home link always rendered their own <a>, based on the url/home-url props. Now users can slot their own <a> (or a framework's Link, e.g. Next.js Link) instead: into post-breadcrumb-item's default slot, or into post-breadcrumbs' home slot. The components detect the slotted anchor and skip rendering their own, so the user's router still handles the click and navigation stays client-side.

**Also includes:**

1. Updated styles so a slotted anchor gets the same colors, hover states, and text-wrapping/overflow-menu behavior as the internal anchor.
2. Documentation update explaining how to use client-side routing with a slotted link, plus a new Storybook story.
3. New Cypress tests covering internal vs. slotted anchor rendering, aria-current behavior, and reactivity to slot changes.
4. A new Playwright test in the integration-next app confirming a slotted next/link navigates without a full reload.

### Note: 
I have also aligned the breadcrumbs Storybook docs with the post-accordion/post-accordion-item pattern, to avoid repeating the same render code across canvases.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/src/stories/components/breadcrumb/breadcrumbs.mdx
myrtasakellariou, line 1: The `breadcrumb-item` code example probably was removed by mistake.

apps/documentation/src/stories/components/breadcrumb/breadcrumbs.stories.ts
myrtasakellariou, line 1: Maybe good to mention that the slotted`<a>` should be a direct child and not be contained in a wrapper element.

packages/components/src/components/post-breadcrumbs/post-breadcrumbs.tsx
```diff
@@ -173,13 +187,19 @@ export class PostBreadcrumbs {
...
+      if (assignedElements.length > 0) {
+        // Insert the assigned elements where the slot used to be, then drop the now empty slot.
+        assignedElements.forEach(element => {
+          target.insertAdjacentElement('beforebegin', element.cloneNode(true) as Element);
+        });
+        target.remove();
+      } else {
+        // Nothing assigned: keep the slot's own fallback content (e.g. the default home link) by
+        // unwrapping the <slot> in place, instead of removing it along with its children.
+        target.replaceWith(...Array.from(target.childNodes));
```
myrtasakellariou: This throws an error in the nextjs app 

Image

myrtasakellariou (changes requested):

apps/documentation/src/stories/components/breadcrumb/breadcrumbs.mdx
alionazherdetska: I aligned the breadcrumb stories with the accordion story, since they're very similar, both have a parent component and item components and only one render. Adding the client story seemed to introduce another render, and things started getting overwhelming, so I aligned it with the accordion story instead. Let me know if you think this is redundant and I'll revert it.

myrtasakellariou (approved):

leagrdv (changes requested):
It seems to not be working on the angular project, I've tested adding a few breadcrumb elements, and they're not being collapsed into the menu as the default breadcrumb:

Image

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.scss
```diff
@@ -17,6 +17,10 @@ tokens.$default-map: components.$post-breadcrumb;
 :host([variant='listitem']) {
   @include mixins.breadcrumb-item;
 
+  ::slotted(a) {
```
leagrdv, line 20: This should only be implemented for the "in-between" links. Adding `white-space: nowrap` to all links makes the last breadcrumb (current page) not wrap, so if the breadcrumb text is long, it adds an overflow because it doesn't wrap into multiple lines.

leagrdv (changes requested):

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.scss
myrtasakellariou: This is handled by the train PR #8459 .

apps/documentation/src/stories/components/breadcrumb/breadcrumbs.mdx
```diff
@@ -24,17 +24,40 @@ The `<post-breadcrumbs>` is a container for `<post-breadcrumb-item>` components,
...
-<Controls of={BreadcrumbItemStories.BreadcrumbItem} />
+### Routing with Client-Side Frameworks
+
+To enable client-side routing (e.g. Next.js, Angular Router), slot your own anchor instead of using the `url`/`home-url` properties: use the default slot for `<post-breadcrumb-item>`, or the `home` slot for the root item of `<post-breadcrumbs>`. Both components detect the slotted `<a>` and stop rendering their own, leaving your link's click handling untouched.
+
+<post-banner type="info">
+  <p>
+    The slotted `<a>` must be a direct child of `<post-breadcrumb-item>` (or of `<post-breadcrumbs>`
+    for the `home` slot), not wrapped in another element, otherwise it won't be detected and the
+    component will fall back to rendering its own internal link.
```
oliverschuerch: Why?
I think, we would do our self a favor, if the RouterLink (e.g. a Angular related `<a>` tag, a nextjs `Link` or any other framework-related router-link, we don't know yet) can be wrapped in other tags (e.g. `div`, `span`, etc.) if necessary.

=> Let the slotted content always win over internal logic, because the users want it that way.

apps/documentation/src/stories/components/breadcrumb/breadcrumbs.mdx
```diff
@@ -24,17 +24,40 @@ The `<post-breadcrumbs>` is a container for `<post-breadcrumb-item>` components,
...
+  </p>
+  <p>
+    The `url`/`home-url` properties are ignored once you slot your own `<a>`. Since the label,
+    description, and `aria-current` are no longer applied automatically, give the slotted `<a>`
+    an accessible name (via its text content or `aria-label`), and add `aria-current="page"`
+    yourself on the item that represents the current page.
+  </p>
```
oliverschuerch: This describes multiple different things in one sentence and I found it hard to anderstand.

e.g.

If you slot in a custom home link, `home-url` prop on `post-accordion` is no longer needed.

{same for post-accordion-item}

Apply an accessible name (via text content or `aria-label`) to the slotted element, and keep in mind to add `aria-current="page"` to the slotted anchor, that represents the current page.

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.tsx
```diff
@@ -58,22 +63,40 @@ export class PostBreadcrumbItem {
...
-  render() {
+  componentWillLoad() {
+    this.checkSlottedAnchor();
+  }
+
+  private checkSlottedAnchor() {
+    this.hasSlottedAnchor = Array.from(this.host.children).some(child => child.tagName === 'A');
+  }
+
+  private renderContent() {
```
oliverschuerch, line 74: Following best practice, I would appreciate it, if both constants would be defined first (href is only a pointer, so no performance loss) and then add an if-else-statement instead of having multiple independent if-statements.

```ts
const slot = ...;
const href = ...;

if (this.hasSlottedAnchor) {
  return ...;
} else if (!href) {
  return ...;
} else {
  return (
    ...
  );
}
```

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #8430, opened 2026-09-07
chore(workflow): add step in the release workflow to check if all packages exist on npm

### LINKED ISSUE #6581: Abort release workflow, when new package does not yet exist on npmjs.com
**User Story**
As a developer of the design system, I'd expect the release workflow will fail, if there is a new package which should be released through trusted publishing, but does not yet exist in the npm registry.

**Explanation**
In our automated github release workflow, we switched to trusted publishing, which means... Instead of using an NPM_TOKEN as a secret, we configured each and every of our packages on npmjs.org to accept new package versions coming from our github repo release workflow, which uses an OIDC (Open ID Connect) object to verify it's ientity.
However, since trusted publishing can only be configred on existing npmjs.org packages (at the moment), we need to ensure, our release workflow fails, whenever we're about to release a new package, which does not exist on npmjs.org.

**Suggestion**
1. Add another step after the dry-run check in our release.yaml
2. Generate the changesets.json file (`pnpm changeset status --output=changesets.json`) and read out the `releases` array. To detect which packages should be released. Don't forget to clean up afterwards.
3. Check if the packages exist in the official registry (https://registry.npmjs.org).
  If it exists, everything is fine. If not, the workflow should fail.
  Maybe `pnpm view <pkg-name>` can be used to do this check.
4. If it fails, we could maybe add an annotation to explain what we need to do to create the package manualy and configure it for trusted publishing.

**Acceptance Criterias**
- [ ] The release workflow must fail, if we're about to release a new, not yet on npmjs.org existing package.

myrtasakellariou:
## 📄 Description

Added a step  in the release workflow (after dry-run) that creates the `changesets.json` and runs a script that checks if each included package already exists on npm. Tested locally with a fake `changesets.json`.

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -54,6 +54,12 @@ jobs:
       - name: Test publish step
         run: pnpm -r publish --dry-run
 
+      - name: Check release packages exist on npm
+        run: |
+          trap 'rm -f changesets.json' EXIT
+          pnpm changeset status --output=changesets.json
+          node .github/workflows/scripts/check-packages-on-npm.mjs
+
       # The changeset action will behave differently based on whether there are
       # new changesets on the github.ref branch:
       #
--- a/.github/workflows/scripts/check-packages-on-npm.mjs
+++ b/.github/workflows/scripts/check-packages-on-npm.mjs
@@ -1,0 +1,34 @@
+// This script checks if the packages about to be released are available on npm
+
+import { execFileSync } from 'node:child_process';
+import fs from 'node:fs';
+
+// Read the changesets.json file and parse it into an object
+const changesets = JSON.parse(fs.readFileSync('changesets.json', 'utf8'));
+
+// Get the list of releases from the changesets object
+const releases = changesets.releases ?? [];
+const missingPackages = [];
+
+// Check if each package exists on npm and log an error for the missing ones
+for (const release of releases) {
+  try {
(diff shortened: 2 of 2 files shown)


leagrdv (approved):

.github/workflows/release.yaml
```diff
@@ -54,6 +54,12 @@ jobs:
       - name: Test publish step
         run: pnpm -r publish --dry-run
 
+      - name: Check release packages exist on npm
```
oliverschuerch, line 57:
```suggestion
      # Trusted publishing on the public npm registry requires the package
      # to already exist before we release it with the "trusted publishing" method.
      # New packages must first be published manually with a placeholder
      # version (e.g. 0.0.0), then configured for trusted publishing on npmjs.com.
      # This check fails fast if a queued package hasn't been bootstrapped yet.
      - name: Verify packages to release exist on the public npm registry
```

Alternatively, you could add the information about the placeholder package with version 0.0.0 in the `check-packages-on-npm.mjs` script.
But I think it would be helpful to know, it can be a dummy package initially, just to configure trusted publishing, before the "real" package will be released through the trusted publishing mechanism by our release workflow.

oliverschuerch (changes requested):

Issue #4487, opened 2025-01-22
Consolidate eslint, stylelint and prettier config


gfellerph:
It's possible to outsource and share stylelint config into it's own package/repo. Doing this would have the benefit that we could share that config between our packages and also other repos like the blueprint.

Sharing prettier config: https://prettier.io/docs/en/sharing-configurationshttps://prettier.io/docs/en/sharing-configurations
Stylelint config: https://stylelint.io/user-guide/get-started

```[tasklist]
### Tasks
- [ ] Delete prettier configs from packages, keep only the one in the root
- [ ] Create a shared config for stylelint and use it in the styles and other packages where we have sass/css files
- [ ] Ensure eslint compatibility with prettier (format a file and then lint it, the file should have no linting errors)
```

PR #8433, opened 2026-09-07
feat(tooling): typescript config package

### LINKED ISSUE #4500: Create a global tsconfig
Most of our packages are using typescript but they all have a separate configuration. 
To improve make sure all packages are aligned, we could implement a global configuration.

Using references and building our packages with the `tsc -b` command could also improve the build process.

Project references documentation: https://www.typescriptlang.org/docs/handbook/project-references.html
Implementation example: https://github.com/typescript-eslint/typescript-eslint

```[tasklist]
### Tasks
- [ ] Put a base tsconfig at the root and extend from all packages
```

yomonom:
## 📄 Description

- Add a new package that provides re-usable typescript configurations
- Ensure type checks are run on every build in existing packages and apps

## 🚀 Demo

n.A.

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


yomonom:
Code smells were pre-existing. Don't know whether they should be addressed in this PR.

yomonom:
Created a new `/tooling` directory, since at least two more packages will follow #4487.

This also aligns with [PostOne](https://github.com/postch/postone-web/tree/main/tooling).

apps/documentation/package.json
alionazherdetska, line 1: All other packages pass `pnpm -r typecheck` cleanly, but apps/documentation is the only failure, with ~150 errors across .storybook/, src/stories/, and src/utils/.

Since this is a new script and root now runs pnpm -r typecheck, this currently breaks the repo-wide command. 

Given the scope, could we either scope exclude to leave src/stories/ out for now (with a tracking issue to fix it properly), or tackle it in a follow-up?

Image

apps/documentation/tsconfig.json
```diff
@@ -1,25 +1,22 @@
...
-    "experimentalDecorators": true
+
+    // Pre-existing debt. Drop once sources are clean.
+    "noImplicitOverride": false,
+    "noImplicitReturns": false,
+    "noUnusedLocals": false,
+    "noUnusedParameters": false
   },
-  "include": [".storybook/*", "src", "types", "../internet-header/src/components.d.ts"],
+  "include": [".storybook/*", "src", "types"],
```
alionazherdetska, line 20: This PR drops `../internet-header/src/components.d.ts` from include in `tsconfig.json`, but it's still there in `tsconfig.eslint.json`. Looks like a leftover, the JSX typings it provides should already come from the @swisspost/internet-header workspace dependency's built types. Can we remove it from tsconfig.eslint.json too, or is there a reason it still needs it there specifically?

alionazherdetska (changes requested):
quite a neat PR, nice to see everything consolidated into a shared config like this.

One thing I noticed is that none of the workflows in `.github/workflows` actually call typecheck yet so it's something we can run locally but it doesn't gate anything in CI. That means a PR could still merge with type errors even though we now have all this infrastructure in place to catch them. Should we maybe add it as a step in lint.yaml alongside the existing lint step, once the documentation typecheck is sorted out?

apps/documentation/tsconfig.json
yomonom: Nope, will remove.

yomonom:
> quite a neat PR, nice to see everything consolidated into a shared config like this.
> 
> One thing I noticed is that none of the workflows in `.github/workflows` actually call typecheck yet so it's something we can run locally but it doesn't gate anything in CI. That means a PR could still merge with type errors even though we now have all this infrastructure in place to catch them. Should we maybe add it as a step in lint.yaml alongside the existing lint step, once the documentation typecheck is sorted out?

I would very much like that, and I have already discussed it with Oli. However, the same is also true of formatting: currently, the CI does not guard against incorrectly formatted files. Therefore, I think it's best to address adjusting the CI in a separate issue.

However, I made sure that `typecheck` is run as part of every `build` script (except for the docs package), so it is actually run as part of the CI (just not that clean).

apps/documentation/package.json
yomonom: Excluding `src/stories` still leaves 16 hard errors (which means basically the whole source directory would have to be excluded), so I think it's best to tackle it in a follow-up.

I'm already planning to create separate issues for all the individual packages where I had to set overrides (see the _"Pre-existing debt. Drop once sources are clean."_ comments ).

alionazherdetska (approved):
Small one, but worth fixing before this gets published:

json
"repository": {
  "type": "git",
  "url": "https://github.com/swisspost/design-system.git",
  "directory": "packages/tsconfig"
}

This still points at packages/tsconfig, but the package now lives at `tooling/typescript/package.json`, the leftover from the earlier moves

yomonom:
> Small one, but worth fixing before this gets published:
> 
> json "repository": { "type": "git", "url": "https://github.com/swisspost/design-system.git", "directory": "packages/tsconfig" }
> 
> This still points at packages/tsconfig, but the package now lives at `tooling/typescript/package.json`, the leftover from the earlier moves

This will be addressed in #8443. See [this diff](https://github.com/swisspost/design-system/pull/8443/changes#diff-620b0c69237d522dcf9927999727613b07a1ab4520d4f96985f75b021a0f0ab8).

tooling/typescript/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "url": "https://github.com/swisspost/design-system.git",
+    "directory": "packages/tsconfig"
+  },
+  "homepage": "https://design-system.post.ch",
+  "bugs": {
+    "url": "https://github.com/swisspost/design-system/issues"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
```
oliverschuerch: This is only needed when the package is about to be released. As soon as this is the case we can add it back. Maybe along with other options like `linkDirectory: true`, etc.

tooling/typescript/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "name": "@swisspost/design-system-tsconfig",
+  "version": "10.4.0",
+  "description": "Shared TypeScript configurations for the Swiss Post Design System.",
+  "author": "Swiss Post <design-system@post.ch>",
+  "license": "Apache-2.0",
+  "private": true,
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/swisspost/design-system.git",
+    "directory": "packages/tsconfig"
```
oliverschuerch: We should create a ticket to add this property also in all of our other packages.

tooling/typescript/README.md
```diff
@@ -0,0 +1,33 @@
...
+![Swiss Post Design System splash screen](https://github.com/swisspost/design-system/assets/1659006/e84f1fea-e666-4853-8c85-726a6bf22e6c)
+
+Shared TypeScript configurations used across the Swiss Post Design System.
+
+## Available Configurations
+
+| Configuration           | Use for                                            |
+| ----------------------- | -------------------------------------------------- |
+| `base.tsconfig.json`    | Options shared by all environments                 |
+| `dom.tsconfig.json`     | Bundled browser code                               |
```
oliverschuerch:
```suggestion
| `browser.tsconfig.json`     | Bundled browser code                               |
```

tooling/typescript/README.md
```diff
@@ -0,0 +1,33 @@
...
+| `node.tsconfig.json`    | Node.js code such as build scripts and CLI tooling |
+| `stencil.tsconfig.json` | Stencil packages                                   |
+| `angular.tsconfig.json` | Angular libraries and applications                 |
+
```
oliverschuerch, line 24: Explain to the users, which of the configs fits best for them (or how to choose the right config).
Suggestion: "Go bottom-up and choose the first config, which fits best to your project setup".

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #8443, opened 2026-09-08
feat(tooling): consolidate eslint stylelint and prettier config

### LINKED ISSUE #4487: Consolidate eslint, stylelint and prettier config

yomonom:
## 📄 Description

- Add new packages that provide re-usable eslint, stylelint, and prettier configurations (structure aligned with POW)

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


tooling/eslint/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "@swisspost/design-system-eslint-config",
+  "version": "10.5.0",
+  "description": "Shared ESLint configurations for the Swiss Post Design System.",
+  "author": "Swiss Post <design-system@post.ch>",
+  "license": "Apache-2.0",
+  "private": true,
```
leagrdv, line 7: The 3 added tooling packages are set to `private`, if we want to be able to use them outside of this repo (e.g. the blueprint or just to let people use them), they should be published.

apps/documentation/package.json
```diff
@@ -7,7 +7,8 @@
   "private": true,
```
leagrdv, line 7: In the `apps/documentation`, there is a `prettier-options.ts` file with some conflicting options (e.g. `singleQuote` set to `false` instead of `true` like it's set in the prettier package). This file might need some Storybook-specific overrides but the base should import and extend the shared prettier config.

tooling/eslint/package.json
leagrdv: Or if we don't want to publish them now:
- Update the `README.md` files to remove the installation guidelines 
- Create a ticket for the publishing of the packages later

leagrdv (changes requested):

tooling/eslint/package.json
yomonom: They are intended to be made public. However, I discussed it with Oli and he said this should be done separately from this ticket, so I've left them `private` for the time being. I have already created a separate issue (#8503) to make them public.

Oli also wanted the installation and usage guidelines to be included in the README for `tooling/typescript`, so I have left them in for the other packages as well. Happy to take them out if you want, though.

apps/documentation/package.json
yomonom: Done. I've also removed the JavaScript-specific options (`trailingComma` and `arrowParens`), as `parser` is set to `HTML` and `embeddedLanguageFormatting` is off, so they don't have an effect anyway.

packages/changelog-github/package.json
```diff
@@ -7,7 +7,8 @@
   "private": false,
   "repository": {
     "type": "git",
-    "url": "https://github.com/swisspost/design-system.git"
+    "url": "https://github.com/swisspost/design-system.git",
+    "directory": "packages/changelog-github"
```
leagrdv, line 11: Is there a reason you added the directory on all packages?

tooling/prettier/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@swisspost/design-system-prettier-config",
```
leagrdv, line 2: I'm wondering if those four tooling packages should not be merged into one, so that if a user wants to follow our styling/linting, they don't have to install 4 different dependencies. What do you think?

packages/changelog-github/package.json
yomonom: Oli requested this in https://github.com/swisspost/design-system/pull/8433#discussion_r4014898691. He suggested creating a separate ticket for it, but I felt that, since it's such a small change, it could be done as part of this ticket.

tooling/prettier/package.json
yomonom: I'd rather keep them separate.

Each package comes with its own set of dependencies and peer dependencies. ESLint, in particular, has many transitive dependencies. Someone who only wants the Prettier config would inherit version
conflicts and security warnings for plugins they never opted into.

The current structure also aligns with POW, so I'd keep it as is.

yomonom: There's also going to be an additional Cypress tooling package, see #8526.

leagrdv (approved):

PR #8459, opened 2026-09-10
feat(components): implement segment specific breadcrumbs

### LINKED ISSUE #8416: [web component func] Segment specific breadcrumbs
### 📝 Description

Breadcrumbs receive a new feature for segment specific home links. Implement according to Design as a feature (non-breaking) and document accordingly.

### ⚙️ Requirements

The first breadcrumb segment needs to be customisable with text, replacing the home icon.

### 🧪 Testing

Add tests with long texts in both the first and the last segment. Text should always be displayed and readable, never truncated. Middle segments should be truncated if there is not enough space available either due to the last or the first segment being too long.

### 🔗 References

Design: https://www.figma.com/design/JIT5AdGYqv6bDRpfBPV8XR/Foundations---Components-V2?node-id=34316-112180&t=j4KpSg0hAdPY7Bsp-4

### 📃 Tasks

- [ ] Add functional tests

alionazherdetska:
## 📄 Description
This PR adds a `show-home-text` prop to `<post-breadcrumbs>`. When true, it shows `text-home` as visible text instead of the home icon, so the root segment can link to a segment's own home page instead of the generic site root (e.g. "Private customers").
 
### Degrade behavior
 
Home and the last (selected) segment never truncate. As per design, when space runs out:
 
1. Middle items collapse into the overflow menu first
2. Home collapses into its own menu next
3. Only as a last resort does the last item wrap onto multiple lines

This PR also adds a `text-expand-home` prop. It's required only once home actually collapses into its own menu, and it's used as the accessible label for its trigger.
 
### Docs
 
This PR improves the breadcrumbs docs to make them more user friendly and easier to find. It introduces a layout with tabs for the examples section (similar concept that is used in the Self managed header page).

## Tests

There were no e2e tests for post-breadcrumbs before this PR. Since the concatenation logic was recently improved, I also added test coverage for it to make sure no one breaks it later on, alongside tests for the new segment specific breadcrumbs behavior.

## 🚀 Demo

image

---

## 🔮 Design review

- [ ] Design review done
- [ ] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


apps/documentation/src/stories/components/breadcrumb/breadcrumbs.mdx
```diff
@@ -36,13 +36,28 @@ The `<post-breadcrumbs>` is a container for `<post-breadcrumb-item>` components,
 
 ## Examples
```
myrtasakellariou, line 38: There is an extra :`:before` line added from docs styles, that does not belong to the tabs design.

packages/components/src/components/post-breadcrumbs/post-breadcrumbs.tsx
```diff
@@ -50,13 +56,23 @@ export class PostBreadcrumbs {
...
+   * The label of the root (home) breadcrumb item. Displayed visibly when `show-home-text` is
+   * `true`, otherwise used as an accessible label alongside the home icon.
    */
   @Prop({ reflect: true })
   @Required()
   @Type('string')
   textHome!: string;
 
+  /**
+   * Whether `text-home` is displayed visibly instead of the home icon, enabling segment specific
```
myrtasakellariou, line 68:
```suggestion
   * Whether `text-home` content is displayed instead of the home icon, enabling segment specific
```

packages/components/src/components/post-breadcrumbs/post-breadcrumbs.scss
myrtasakellariou, line 1: When collapsed, the home "menu" shows a third chevron before the ellipsis, that does not exist in design.

Code
Image

Figma
Image

myrtasakellariou (changes requested):

.changeset/funny-train-ride.md
```diff
@@ -0,0 +1,5 @@
+---
+'@swisspost/design-system-components': minor
+---
+
+Added a `show-home-text` property to `<post-breadcrumbs>` for segment specific breadcrumbs, displaying `text-home` as visible text instead of the home icon. The home item's own overflow menu is labelled via the new `text-expand-home` property.
```
leagrdv, line 5: Only issue I found was the extra chevron but regarding behaviour in general, I know it corresponds to the design, but I don't really get why the first element collapses into its own menu:
- It's confusing: the user needs to click on both "..." to see where whatever link they want is.
- It adds an extra label property to the component, so we now have "text-expand-home" and "text-more-items" which is a bit unclear which ones does which. Feels like if we keep this, I'd go with "text-home-item" or something similar that complements the existing one.
- Collapsing the first element makes it that the user needs to click twice to open a link (one for the menu and one for the link inside), in the case that the breadcrumb is for instance `A very long home page name > A very long subpage name > My page` and the screen can't fit both, you're left with `... > ... > My page` with two menus that both contain just one link.


I think it'd make more sense to have either:
- All in the same menu: The middle ones would collapse first, and if space doesn't allow for the first and last to show, the first gets into the middle menu
- First link in multiple lines: If the first link is too long and can't fit together with the last link, then have the first link also go into multiple lines.

What do you think? If you agree with this, I'd say maybe we could add it to the pattern discussion.

leagrdv (changes requested):

.changeset/funny-train-ride.md
yomonom: The first solution (all in the same menu) would also resolve the issue regarding the extra label property and could be implemented very easily:

Instead of having a `show-home-text` property, there could be a `show-home` property which hides the home item altogether if set. This would mean that you could just use the first breadcrumb item as the 'Home' item.

myrtasakellariou: I also agree that the two menus are very confusing. And using a menu component for a single item does not seem very elegant. I would vote for either: 1. all in one menu or 2. hide home icon and use the next one as the custom home one. 3. show an icon on mobile?

gfellerph: - If there's a home icon, the first item should never collapse -> behaviour should stay as is today `🏠 > ... > Some absurdly long last breadcrumb item`
- If there's a very long first item text, no home icon and it does not fit, only show one menu, like `... > Some absurdly long last breadcrumb item`

Having the first item break on multiple lines as well would be problematic in my opinion. The home icon should never be in the overflow menu.

Length calculation logic might be a bit hard to define. I'd set the reference element to `white-space: nowrap` and check if it's longer than the available width. If it is, collapse the middle section and test width again. If it's still wider than the available width, also collapse the first item. Then render, allowing the last item to break on multiple lines as needed. Recalculate on viewport resize and on slotchange.

myrtasakellariou:
image

.changeset/funny-train-ride.md
myrtasakellariou: Update after discussion on Daily 16/9/26: Remove extra ellipsis/checvron and have only one collapsed menu.

yomonom: I forgot to ask during the discussion: should we keep the `text-home` property? I still think that using just `show-home` and setting it to `false` (adding the 'home text' as a regular breadcrumb, but keeping the behaviour suggested by Philipp) would be much easier for the user and reduce the number of properties on the component.

oliverschuerch: Late to the party, but I see it similar as @yomonom.
Instead, I would keep `text-home` only, because I would not render the `home` item as a regular breadcrumb, since we would need to implement the logic in the `post-breadcrumb-item`, producing unnecessary overhead for every other item.

**How to**

1. When the property `text-home` is not set (defaults to `''`), the component renders the home icon.
2. When it is set to something else than `''`, the component shows the text instead of the home icon.

apps/documentation/src/stories/components/breadcrumb/breadcrumbs.stories.ts
```diff
@@ -57,6 +59,36 @@ export const Concatenated: Story = {
...
+      <post-breadcrumb-item url="/section1">Send letters</post-breadcrumb-item>
+      <post-breadcrumb-item url="/section2">Letters abroad</post-breadcrumb-item>
+      <post-breadcrumb-item url="/section3">Europe</post-breadcrumb-item>
+      <post-breadcrumb-item url="/section4">Neighbouring countries</post-breadcrumb-item>
+      <post-breadcrumb-item url="/section5">Letters to Switzerland</post-breadcrumb-item>
+    </post-breadcrumbs>
+  `,
+};
+
 export const ClientSideRouting: Story = {
```
oliverschuerch, line 71: This story still renders the unnecessary `text-home` property on the `post-breadcrumbs` host.
Can we make it optional and only render it if set to something else than `''`?

Image

apps/documentation/.storybook/styles/components/index.scss
oliverschuerch, line 1: Why are these changes part of that PR?
Can we extract it to an own `chore` PR?

apps/documentation/.storybook/styles/components/tabs.scss
oliverschuerch, line 1: Why are these changes part of that PR?
Can we extract it to an own `chore` PR?

oliverschuerch (changes requested):

.changeset/funny-train-ride.md
oliverschuerch: @myrtasakellariou After our discussion were you explained that the `text-home` property is used as an `aria-label` if `show-home-text` is `false` , I think I would close this CR, because the logic works as intended.

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.scss
```diff
@@ -17,20 +17,29 @@ tokens.$default-map: components.$post-breadcrumb;
...
+    // Last resort, once home has also collapsed. `overflow`/`text-overflow` undo
+    // button-base's default ellipsis truncation.
+    flex-shrink: 1;
+
     :is(a, span),
     ::slotted(a) {
       white-space: normal;
+      overflow: visible;
+      text-overflow: unset;
       @include utilities.word-break;
```
oliverschuerch, line 41: Do we need both (custom CSS and the mixin)?

oliverschuerch (changes requested):

packages/components/src/components/post-breadcrumb-item/post-breadcrumb-item.scss
myrtasakellariou: Only the  `white-space: normal;` is needed from the  extra css rules.

apps/documentation/src/stories/components/breadcrumb/breadcrumbs.stories.ts
oliverschuerch: Sadly we can't just remove the `text-home` property in the story but keep it required in code. This throws an error in the console.
image

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #8478, opened 2026-09-10
chore: configure pnpm dependency build permissions

oliverschuerch:
## 📄 Description

Configure explicit pnpm build permissions, allowing required native and binary builds while blocking unnecessary dependency install scripts.

**Verification**
`pnpm install` passes successfully without ignored-build errors.

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -69,3 +69,6 @@ allowBuilds:
   msgpackr-extract: true
   sharp: true
   unrs-resolver: true
+minimumReleaseAgeExclude:
+  # Renovate security update: svgo@4.1.0
+  - svgo@4.1.0
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -60,11 +60,11 @@ overrides:
 allowBuilds:
   '@parcel/watcher': true
   '@percy/core': true
-  core-js-pure: true
+  core-js-pure: false
   cypress: true
-  es5-ext: true
+  es5-ext: false
   esbuild: true
-  iframe-resizer: true
+  iframe-resizer: false
   lmdb: true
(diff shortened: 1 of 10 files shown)


pnpm-workspace.yaml
```diff
@@ -60,11 +60,11 @@ overrides:
...
   '@parcel/watcher': true
   '@percy/core': true
-  core-js-pure: true
+  core-js-pure: false
   cypress: true
-  es5-ext: true
+  es5-ext: false
   esbuild: true
-  iframe-resizer: true
+  iframe-resizer: false
```
leagrdv: The `iframe-resizer` dependency doesn't seem to be used anywhere anymore, so this line (as well as the dependency) can be removed completely.

leagrdv (changes requested):

pnpm-workspace.yaml
oliverschuerch: Only for traceability: This dependency could have been removed with commit https://github.com/swisspost/design-system/commit/1357f7b02dfed9f8160b0e95bff5869b27415fff

leagrdv: I don't think this comment adds anything useful for traceability, since that's already available on git. It mostly serves to highlight something that was missed previously, which can come across as a bit negative.

leagrdv (approved):

pnpm-workspace.yaml
oliverschuerch: The dependency was once used by the breadcrumbs component and got removed earlier.
Since no other component, as well as no npm script uses it anymore and it's also not a peer dependency of any other installed module, it can be removed safely.

PR #8483, opened 2026-09-11
fix(klp): use env specific URLs for subscribe call

### LINKED ISSUE #8452: [bug]: KLP Login URL on int does not go to int
### 🐞 Describe the bug

When a project runs on integration, KLP rejects calls to `/subscribe` with a CORS policy if they go to KLP prod.

### ✨ Expected behavior

Use the following URL table for the specified environments as the base URL for `/subscribe` calls.

Environment | Base URL
-- | --
PROD | https://n.account.post.ch:443
INT2 | https://n.accountint2.post.ch:443
INT1 | https://n.accountint1.post.ch:443

gfellerph:
## 📄 Description

Updated URL mapping for KLP subscribe calls to take environment into account.

---

## 🔮 Design review

- [ ] Design review done
- [X] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist

### DIFF
--- a/apps/documentation/src/stories/components/header/self-managed/session-subscription.sample.ts
+++ b/apps/documentation/src/stories/components/header/self-managed/session-subscription.sample.ts
@@ -1,5 +1,5 @@
 // Import the KLP URL configuration from the internet-header package
-import { KLP_BASE_URLS, getSessionUrl } from '@swisspost/internet-header/config/klp-urls';
+import { getSessionUrl } from '@swisspost/internet-header/config/klp-urls';
 
 // Get the appropriate URL based on your environment
 const environment = 'prod'; // Change to your environment (int01, int02, etc.)
--- a/apps/documentation/src/stories/components/header/self-managed/session-subscription.sample.ts
+++ b/apps/documentation/src/stories/components/header/self-managed/session-subscription.sample.ts
@@ -1,4 +1,19 @@
-const response = await fetch('https://n.account.post.ch/v1/session/subscribe', {
+// Map environments to their corresponding KLP base URLs
+const KLP_BASE_URLS = {
+  dev01: 'https://n.accountint1.post.ch',
+  dev02: 'https://n.accountint1.post.ch',
+  devs1: 'https://n.accountint1.post.ch',
+  test: 'https://n.accountint1.post.ch',
+  int01: 'https://n.accountint1.post.ch',
+  int02: 'https://n.accountint2.post.ch',
+  prod: 'https://n.account.post.ch',
+};
+
+// Get the appropriate URL based on your environment
+const environment = 'prod'; // Change to your environment (int01, int02, etc.)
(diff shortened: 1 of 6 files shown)


packages/internet-header/src/components/post-internet-header/post-internet-header.tsx
```diff
@@ -19,7 +19,17 @@ import {
...
+
+const KLP_BASE_URLS: Record<Environment, string> = {
+  dev01: 'https://n.accountint1.post.ch',
+  dev02: 'https://n.accountint1.post.ch',
+  devs1: 'https://n.accountint1.post.ch',
+  test: 'https://n.accountint1.post.ch',
+  int01: 'https://n.accountint1.post.ch',
+  int02: 'https://n.accountint2.post.ch',
+  prod: 'https://n.account.post.ch',
+};
```
oliverschuerch: If you would use them, exported from a `.ts` file, they could probably being reused in the testing.

oliverschuerch (dismissed):

packages/internet-header/src/config/klp-urls.ts
oliverschuerch, line 1: Very nice!

apps/documentation/src/stories/components/header/self-managed/session-subscription.sample.ts
```diff
@@ -1,4 +1,11 @@
-const response = await fetch('https://n.account.post.ch/v1/session/subscribe', {
+// Import the KLP URL configuration from the internet-header package
+import { getSessionUrl } from '@swisspost/internet-header/config/klp-urls';
```
oliverschuerch, line 2: I'm relatively sure this won't work 😆
You don't even export the file standalone from the package. It only exists under `dist/esm-es5/swisspost-internet-header.entry.js`.

oliverschuerch (changes requested):

apps/documentation/src/stories/components/header/self-managed/session-subscription.sample.ts
gfellerph: Right, bad LLM.

apps/documentation/src/stories/components/header/self-managed/session-subscription.sample.ts
oliverschuerch, line 1: It feels wrong to expect every project to implement these constants and the helper function themselves. And every necessary adjustment on our part has the potential to become a support issue. If only there were a way to implement these things in our own code to keep things simple for the projects 🤔

We're lucky, because there is one 😃

Image

What do you think, should we give that approach a try?
I have only one concern: I'm not sure if this ends up in a project, who only imports standalone components or how they would need to import/use this entry file without importing the whole library.

oliverschuerch (changes requested):

oliverschuerch (approved):

PR #8486, opened 2026-09-11
feat(appearance): add experimental condensed layout

gfellerph:
This is an experimental PR for a condensed appearance. It works the same way as compact: it just renders mobile device tokens only with the effect, that every breakpoint just uses mobile tokens. Simple but effective.

The basic plan would be to release these entry files and label them as beta in the documentation. Applications already on v10 or ones migrating could use or switch to this and we'd get early feedback if we're on the right track here or if more specific styles are necessary. If tweaks can be done through the appearance layer without any other change, even better because we'd have a good basis and use the appearance layer only for the things that really matter.

### DIFF
--- a/apps/documentation/.storybook/addons/styles-switcher/StylesSwitcher.tsx
+++ b/apps/documentation/.storybook/addons/styles-switcher/StylesSwitcher.tsx
@@ -2,7 +2,7 @@ import React, { useEffect, useState } from 'react';
 import { IconButton, WithTooltip } from 'storybook/internal/components';
 
 const THEMES = ['Post', 'Cargo'] as const;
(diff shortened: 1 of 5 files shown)


gfellerph:
Roundtable: is this an approach that could speed up application appearance development? What would happen if we'd add specific appearance tokens on top of it? Would it just work?

gfellerph:
image

Some cases with custom media queries are misaligned. Search for other instances of this and check if components are compatible.

gfellerph:
**Resolution**
- Accepted proposal, go forward with implementation. Workaround already communicated to UX

gfellerph:
Closing draft but keeping it as a reference for#8399

PR #8490, opened 2026-09-14
chore: add action to label stale PRs

### LINKED ISSUE #6101: [discussion]: Should we mark PRs as stale after X days?
At the time of writing, there are 38 open PRs, 15 of which are older than a month. The longer a PR is open, the smaller the chance to be able to merge it without conflicts. Should we add an actions workflow that marks PRs as stale after a certain amount of time? The workflow could also add the label `needs: 🏓 dev roundtable` to the PR and add it to the project board for better visibility.

gfellerph:
## 📄 Description

This action runs once a week on Sunday evening and adds stale PRs (older than one month) to the dev roundtable.


.github/workflows/label-stale-review-prs.yaml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - name: Label pull requests awaiting review for 30 days
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
+        with:
+          script: |
+            const label = 'needs: 🏓 dev roundtable';
+            const cutoff = new Date();
+            cutoff.setDate(cutoff.getDate() - 30);
+            const date = cutoff.toISOString().slice(0, 10);
+            const query = `repo:${context.repo.owner}/${context.repo.repo} is:pr is:open draft:false review:required updated:<${date} -label:"${label}"`;
```
leagrdv: Should we filter with `review:required`? There are other states that could be there which should still result in a stale PR (e.g. changes requested). Maybe we should remove this filter altogether. An approved PR that hasn't been touched in a month should probably still be discussed.

leagrdv (changes requested):

.github/workflows/label-stale-review-prs.yaml
gfellerph: True. Also removed the draft filter so we can clean those up as well.

leagrdv (approved):

PR #8494, opened 2026-09-14
chore: add ADR for web component styling

### LINKED ISSUE #8078: [adr] Using `:host` and `::slotted` selectors
## Context

A web component host (`<post-tooltip>`) element can be styled with the `:host` selector and slotted elements can be styled via the `::slotted` pseudo-selector. The cascade order of these selectors however, is very low because the styles are applied from within the shadow DOM. Light dom styles are applied with higher specificity (see [codepen example](https://codepen.io/tuelsch/pen/pvRWvyj)), except when defined with `!important`. This has led to conflicts between project styles and Design System styles. Future unintentional overrides are very likely.

## Decision

Host styles should be:
- Reduced to the minimum
- The default display for host elements should be `contents`
- Host styles on existing components should be moved to an element within the component
- The `:host` selector can be used to react to component properties, e.g. `:host([variant="big"]) button { font-size: 2rem; }`

Slotted elements should be:
- Web Components defined in the components package
- HTML/CSS components defined in the styles package
- HTML elements without specific class name styled in the [components global styles](https://github.com/swisspost/design-system/tree/main/packages/components/src/styles) file

## Consequences

`:host` and `::slotted` selector usage is strictly limited. Exceptions can be made where necessary by using `!important` on `::slotted` elements. These exceptions need to be justified, e.g. by reducing overhead, and need to be documented.

## Customizing slotted element styles

Chances are that slotted elements need a little tweak here and there. The preferred method to pass contextual style overrides to a slotted component (needs to be a design system component) is via CSS custom properties:

```CSS
/* Host element styles (post-menu) */
:host {
  --post-menu-item-active-background-color: pink;
}

/* Slotted element styles (post-menu-item) */
.list-item {
  background-color: var(--post-menu-item-active-background-color, --token-value, black);
}
```

_[CodePen example for passing styles as props to child components](https://codepen.io/tuelsch/pen/vEgwpzN?editors=1000)_

The `<post-menu-item>` accepts a custom property that is able to override the default value like a prop on a web component. This custom property can also be documented to make it available to other users if it is necessary. This allows Design System components to override specific values contextually with a namespaced value, reducing the risk of unintentional overrides.

gfellerph:


.contributing/ADR/styling-web-components.md
```diff
@@ -0,0 +1,52 @@
...
+A web component host (`<post-tooltip>`) element can be styled with the `:host` selector and slotted elements can be styled via the `::slotted` pseudo-selector. The cascade order of these selectors however, is very low because the styles are applied from within the shadow DOM. Light dom styles are applied with higher specificity (see [codepen example](https://codepen.io/tuelsch/pen/pvRWvyj)), except when defined with `!important`. This has led to conflicts between project styles and Design System styles. Future unintentional overrides are very likely.
+
+## Decision
+
+<!-- What is the change that we're proposing and/or doing? -->
+
+Host styles should be:
+
+- Reduced to the minimum
+- The default display for host elements should be `contents`
```
oliverschuerch: If we do that, the :host element will not be highlighted when selected with the dev tools and applied margins, etc. won't work anymore. This could become specifically an issue if a component is part of the text flow, where it needs to arrange itself between two paragraphs, etc. And there might be other drawbacks, I would be careful using this property as a default style on any element!

.contributing/ADR/styling-web-components.md
```diff
@@ -0,0 +1,52 @@
+# Styling web components
+
+## Context
+
+<!-- What is the issue that we're seeing that is motivating this decision or change? -->
+
+Reference: <https://github.com/swisspost/design-system/issues/8078>
+
+A web component host (`<post-tooltip>`) element can be styled with the `:host` selector and slotted elements can be styled via the `::slotted` pseudo-selector. The cascade order of these selectors however, is very low because the styles are applied from within the shadow DOM. Light dom styles are applied with higher specificity (see [codepen example](https://codepen.io/tuelsch/pen/pvRWvyj)), except when defined with `!important`. This has led to conflicts between project styles and Design System styles. Future unintentional overrides are very likely.
```
oliverschuerch:
```suggestion
A web component host (`<post-tooltip>`) element can be styled with the `:host` selector and slotted elements can be styled via the `::slotted` pseudo-selector. The cascade layer of these selectors however, is very low because the styles are applied from within the shadow DOM. Light DOM styles are applied on a higher cascade layers, so they need minimal specificity (see [codepen example](https://codepen.io/tuelsch/pen/pvRWvyj)) to override the styles from layers below, except when those are defined with `!important`. This has led to conflicts between project styles and Design System styles. Future unintentional overrides are very likely.
```

Maybe it would be worth to add a graphic or some links.
What do you mean?

https://mdn.github.io/shared-assets/images/diagrams/css/at-rules/layer-cascade.svg
[Origin](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@layer)
[Licencing](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Attrib_copyright_license)

.contributing/ADR/styling-web-components.md
```diff
@@ -0,0 +1,52 @@
...
+- Reduced to the minimum
+- The default display for host elements should be `contents`
+- Host styles on existing components should be moved to an element within the component
+- The `:host` selector can be used to react to component properties, e.g. `:host([variant="big"]) button { font-size: 2rem; }`
+
+Slotted elements should be:
+
+- Web Components defined in the components package
+- HTML/CSS components defined in the styles package
+- HTML elements without specific class name styled in the [components global styles](https://github.com/swisspost/design-system/tree/main/packages/components/src/styles) file
```
oliverschuerch:
```suggestion
- HTML elements without specific class name styled in the [components global styles](https://github.com/swisspost/design-system/tree/main/packages/components/src/styles)
```
It's not one file anymore.

.contributing/ADR/styling-web-components.md
```diff
@@ -0,0 +1,52 @@
...
+
+/* Slotted element styles (post-menu-item) */
+.list-item {
+  background-color: var(--post-menu-item-active-background-color, --token-value, black);
+}
+```
+
+_[CodePen example for passing styles as props to child components](https://codepen.io/tuelsch/pen/vEgwpzN?editors=1000)_
+
+The `<post-menu-item>` accepts a custom property that is able to override the default value like a prop on a web component. This custom property can also be documented to make it available to other users if it is necessary. This allows Design System components to override specific values contextually with a namespaced value, reducing the risk of unintentional overrides.
```
oliverschuerch:
```suggestion
The `<post-menu-item>` accepts a custom property that is able to override the default value like a prop on a web component. This custom property should also be documented to make it available to other users if it is necessary. This allows Design System components to override specific values contextually with a namespaced value, reducing the risk of unintentional overrides.
```

oliverschuerch (changes requested):

.contributing/ADR/styling-web-components.md
gfellerph: Changed to `inline`, which is the browser default.

.contributing/ADR/styling-web-components.md
gfellerph: Added the graphic, but the edits don't reflect what I wanted to express.

oliverschuerch: Ok, but there are still some typos...

> Light dom styles are applied with higher specificity (...), except when defined with `!important`.

=> You probably want to say something like:
expect, if the shadow DOM styles are defined with `!important`

---

Write `Light DOM ...` instead of `Light dom ...`

---

> The cascade order of these selectors however, is very low ... Light dom styles are applied with higher specificity ...

=> This still somehow feels, as if you compare cascade layer order with specificity.
I understand what you mean, but it feels wrong on the first read.
However, I'm fine to keep it as is, if you think all our DEVs will understand it correctly.

oliverschuerch (changes requested):

.contributing/ADR/styling-web-components.md
gfellerph: Updated the wording for more clarity.

oliverschuerch (approved):

PR #8495, opened 2026-09-15
feat(components): prevent `post-popovercontainer` from overflowing the viewport

### LINKED ISSUE #8415: [bug]: navigating long menus is difficult
Currently, when `post-menu` contains more items than can fit the viewport, moving focus with the arrow keys does not scroll the focused item into view. Focus moves invisibly: the item is active and can be triggered with `Space`/`Enter`, but the user can't see it unless they scroll manually with `Page Up`/`Page Down`.

https://github.com/user-attachments/assets/f9e219bc-4161-4a4d-93c9-612d2cfdacad

Items that overflow are also difficult for users to discover, as there is no visual indication that there are more items below. This is particularly apparent on mobile devices, where overflow occurs at much lower item counts.

Image

Therefore, it may be worth constraining the menu's size (similar to how `post-popover` is handled) and making the menu itself scrollable, rather than letting it grow past the viewport.

yomonom:
## 📄 Description

- Restrict the size of the `post-popover-container` component to the available space.
- Ensure the `post-menu` component is scrollable when necessary.
- Adjust the focus outline for the `post-menu-item` component.

> [!Note]
> Scroll-locking the scrollable ancestors while a menu is open was considered but ultimately decided against. The POC can be found [here](https://github.com/swisspost/design-system/tree/poc/scroll-lock-popovercontainer).

## 🚀 Demo

https://github.com/user-attachments/assets/203a1440-c60f-44a1-a4d8-93a315fcc8d0

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


packages/components/src/components/post-menu-item/post-menu-item.scss
```diff
@@ -47,14 +52,28 @@ post-menu-item {
...
     }
 
+    @include utilities-mx.focus-style-custom('::after', $outline-offset);
+
     &:hover {
       background-color: #504f4b;
       color: #fff;
+
+      @include utilities-mx.focus-visible('::after') {
+        outline-color: tokens.get('focus-outline-color-inverted', helpers.$post-focus) !important;
```
leagrdv, line 75: Because the component itself is not tokenized yet, this value for the color should be hardcoded as well, otherwise the focus color changes to white on dark scheme, but the actual menu stays white, making the focus not visible at all.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
leagrdv, line 1: A few bugs I found while testing the PR:
Some of them might have been present before your changes and might not come from them, but because they're part of the topic, I think it should be addressed here even if that's not the case.

A really long popover gets its content cut by the window, and not all text can be readable in that case. Also, if it's long, it sometimes open on top of the actual trigger.
Image

The close button gets cut on small screens (done in #7774), seems to be a regression maybe
Image

Tooltip with a really long text overflows the container
Image

leagrdv, line 1: Testing it, it seems that the available space calculated omits the header. Because this component is quite big and takes a lot of screen, I think it could make sense to have the popover/menus go over it, just to use as much free space as possible.

leagrdv, line 1: Menu closes as expected when the trigger leaves the screen, but that's not the case for the `post-datepicker` which also uses the `post-popovercontainer`

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
```diff
@@ -186,366 +149,243 @@ export class PostPopovercontainer {
...
-      this.clearAutoUpdate();
-    }
+  private async handleOpen() {
+    this.postBeforeToggle.emit({ willOpen: true });
+    this.postBeforeShow.emit({ first: !this.hasBeenOpened });
 
-    if (this.safeSpace) {
-      globalThis.removeEventListener('mousemove', this.boundMouseTrackingHandler);
-    }
+    await this.updatePosition(true);
```
leagrdv, line 240: This forces `autoSize` to `true` even if the prop has been set to `false`

leagrdv (changes requested):

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
yomonom: This is intentional. Every popover is sized (`max-width` and `max-height` set) initially on open, so that the `edgeGap` property can correctly constrain its dimensions and the placement can be computed against the constrained size. The same is currently done on the main. `autoSize` just means that the size can be re-computed once the popover is open, for example when the anchor is scrolled.

I've renamed the property to `autoResize` and updated its description to make this clearer.

packages/components/src/components/post-popovercontainer/post-popovercontainer.tsx
yomonom: This is intentional. The `post-datepicker` is not scrollable, so it could theoretically overflow the available viewport, meaning the user would need to scroll the page to see or interact with it fully. Automatically closing the date picker when the anchor leaves the viewport could therefore prevent the user from interacting with certain parts of it. We could make the date picker scrollable, but I find that a bit confusing.

What are your thoughts on this?

PR #8515, opened 2026-09-18
feat(components): compose the signed-in state as a real user menu

oliverschuerch:
## 📄 Description

The signed-in widget was a flat list of anchors. It now uses the design system's own menu: a post-menu-trigger button carrying the avatar and the user's name, opening a post-menu whose header repeats the identity and the company the user is acting for, followed by post-menu-item entries.

Both the trigger and the menu live in the widget's shadow root, so the menu id is scoped and a constant cannot collide between instances. The named slots keep working: post-menu resolves a slot assigned into a slot, so a consumer's own markup still wins over the link properties.

KlpLink.description is finally rendered. It goes to aria-describedby pointing at a visually hidden sibling rather than aria-description, which is not baseline available, and sits outside the anchor so it describes the link without joining its accessible name.

The avatar is built from the name alone. Passing the email would send a hash of it to a third-party gravatar host on every page view, which is not something a login widget should do on the user's behalf. It is also marked aria-hidden, because the name it depicts is already next to it.

nav-item-base is included per selector rather than for a group: the focus mixin emits a malformed `:is(:focus-visible{` for grouped selectors, which silently dropped the focus outline.

src/index.html gains the widget in the header's post-login slot, with the platform faked in the page so all three states -- anonymous, account switch, company switch -- can be seen without a backend.

## 🚀 Demo

---

## 🔮 Design review

- [ ] Design review done
- [x] No design review needed

## 🧪 Visual regression tests

## 📝 Checklist


oliverschuerch:
I had to revert that, because of a lot new relevant findings.

